PR #1, opened 2019-01-14
initial commit

josojo:
Very first intital commit.

it allows us to load mock data into the mongodb database

it allows us to fetch the data again out of it.


src/main.rs
```diff
@@ -0,0 +1,67 @@
+extern crate serde_json;
+extern crate serde;
+
+
+const ACCOUNTS: i32 = 8;
```
fleupold: Would it be possible to import these constants from the same file?

src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+const TOKENS: i32 = 4;
+const SIZE_BALANCE: usize = (ACCOUNTS * TOKENS) as usize;
+
+#[macro_use]
+extern crate serde_derive;
+
+#[derive(Serialize, Deserialize)]
+struct State {
```
fleupold: same with state.

src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+            _ => panic!("Expected title to be a string!"),
+        },
+        Some(Err(_)) => panic!("Failed to get next from server!"),
+        None => panic!("Server returned no results!"),
+    }
+
+    let coll = client.db("dfusion").collection("State");
+
+
+    let cursor = match coll.find(Some(doc! { "curState"	: "0x0000000000000000000000000000000000000000", },) , None) {
```
fleupold: should this use the `cur_state` variable instead of hardcoded curstate?

src/bin/database_setup.rs
```diff
@@ -0,0 +1,78 @@
...
+    // Insert document into 'dfusion.CurrentState' collection
+    coll.insert_one(doc.clone(), None)
+        .ok().expect("Failed to insert CurrentState.");
+
+    let coll = client.db("dfusion").collection("State");
+
+
+	let state = State {
+	    curState: "0000000000000000000000000000000000000000".to_owned(),
+    	prevState: "0000000000000000000000000000000000000000".to_owned(),
```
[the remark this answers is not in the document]
bh2smith: Dang, you got to it first!

fleupold:
Would it make sense to move the rust code in a subdirectory (something like driver?). I was hoping that in this repository we could have a few tools (django app, rust driver) and all connected with one docker-compose file.

src/bin/database_setup.rs
```diff
@@ -0,0 +1,78 @@
...
+    coll.insert_one(temp.clone(), None)
+        .ok().expect("Failed to insert CurrentState.");
+
+    let coll = client.db("dfusion").collection("Deposits");
+
+    let doc2 = doc! {
+        "depositHash": "0000000000000000000000000000000000000000",
+        "depositIndex": "0000000000000000000000000000000000000000",
+        "slot": 1,
+        "addressIndex": 0,
```
bh2smith: Please call this accountId or accountIndex. 

Also account IDs start at 1!

src/bin/database_setup.rs
```diff
@@ -0,0 +1,78 @@
...
+        "tokenIndex": 1,
```
bh2smith: Could we call this and accounts by ID?

src/bin/database_setup.rs
```diff
@@ -0,0 +1,78 @@
...
+extern crate serde;
+
+#[macro_use]
+extern crate serde_derive;
+
+#[derive(Serialize, Deserialize)]
+struct State {
+  	curState: String,
+   	prevState: String,
+  	nextStates: String,
```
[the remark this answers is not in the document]
josojo: made it singular, with the newest non-forking decision

src/bin/database_setup.rs
```diff
@@ -0,0 +1,78 @@
...
+   	slot: i32,
+   	balances: [i64; SIZE_BALANCE]
```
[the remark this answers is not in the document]
josojo: changed it to PrimeField

src/bin/database_setup.rs
```diff
@@ -0,0 +1,78 @@
...
+    	balances: [0; SIZE_BALANCE]
+	};
+
+    let document = serde_json::to_string(&state).ok().expect("Failed to convert first State");
+    
+	println!("{}", document);
+
+    //    let document: String = String::from(r#"{"curState":"0000000000000000000000000000000000000000","prevState":"0000000000000000000000000000000000000000","nextStates":"0000000000000000000000000000000000000000","slot":0,"balances":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}"#);
+    //	let temp = doc!( r#document#);
+    let temp = doc! {"curState":"0000000000000000000000000000000000000000","prevState":"0000000000000000000000000000000000000000","nextStates":"0000000000000000000000000000000000000000","slot":0,"balances":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]};
```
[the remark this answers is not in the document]
josojo: Still searching for a solution

src/main.rs
josojo: Yeah will do it.

src/main.rs
josojo: Yes, if I know how to make a mongodb::bson::document out of this, I will do it.

src/bin/database_setup.rs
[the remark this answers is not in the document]
josojo: With the traits PrimeFields I am not able to the deserializer, even with strings its not working as I would get 

balances: ["0".to_owned(); global_helpers::SIZE_BALANCE]                                                                                                                                       
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `std::string::String`

josojo:
Currently, I am hanging on 3 points:

- how to deserialize strings,
- how to make a mongodb::bson::document from a string or another variable
- how to deserialize arrays of different size: ie. [i32, 32] is implemented, but [i32, 64] is not

driver/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "driver"
+version = "0.1.0"
+authors = ["josojo <josojo@hotmail.de>"]
```
fleupold: Maybe use your gnosis email here

driver/src/bin/global_helpers.rs
```diff
@@ -0,0 +1,14 @@
+pub const ACCOUNTS: i32 = 8;
```
fleupold: nit: I would call this file `model.rs` or `defines.rs` something similar. Right now there are no helper functions in here, and global doesn't sound great either. 
Eventually we might want to read ACCOUNTS/TOKENS etc from environment variables so that we can ensure that all parts are running with the same configuration.

driver/src/bin/global_helpers.rs
```diff
@@ -0,0 +1,14 @@
...
+pub const TOKENS: i32 = 8;
+pub const SIZE_BALANCE: usize = (ACCOUNTS * TOKENS) as usize;
+
+use pairing::{ PrimeField };
```
fleupold: Is this used?

josojo:
I think this very first basic PR is ready to merge. What do you guys think?

driver/src/bin/models.rs
```diff
@@ -14,9 +14,9 @@ pub struct State {
...
 pub struct Deposits {
   	pub depositHash: String,
-    depositIndex: String,
-    slot: i32,
-    addressId: i32,
-    tokenId: i32,
-    amount: i32,
+    pub depositIndex: String,
+    pub slot: i32,
+    pub addressId: i32,
```
bh2smith: This should be `accountId`!

driver/src/bin/models.rs
```diff
@@ -14,9 +14,9 @@ pub struct State {
 #[derive(Serialize, Deserialize)]
 pub struct Deposits {
   	pub depositHash: String,
-    depositIndex: String,
-    slot: i32,
-    addressId: i32,
-    tokenId: i32,
-    amount: i32,
+    pub depositIndex: String,
```
bh2smith: This value equivalent to slot and not used.

josojo: No, we agreed to have a counter per slot to sort them in order, didn't we?

How else would I know which one is the latest deposit with the correct depositHash? 
@bh2smith

driver/src/bin/models.rs
josojo: yes

driver/src/bin/models.rs
bh2smith: Interesting. Yes, this definitely matters. I will make this happen. Whoops.

bh2smith: In that case, I think this depositIndex will have to be an `Integer`

driver/src/bin/models.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Deposits {
+  	pub depositHash: String,
+    pub depositIndex: String,
+    pub slot: i32,
+    pub accountId: i32,
+    pub tokenId: i32,
+    pub amount: i64,
```
fleupold: I think we need a solution to make this support the full width (100 bit)

driver/src/bin/fetch_db.rs
```diff
@@ -0,0 +1,87 @@
...
+    let v: serde_json::Value = serde_json::from_str(&query).expect("Failed to parse query to serde_json::value");
+    let bson = v.into();
+    let mut _temp: bson::ordered::OrderedDocument = mongodb::from_bson(bson).expect("Failed to convert bson to document");
+    
+    let coll = client.db("dfusion").collection("Deposits");
+
+    let cursor = coll.find(Some(_temp) , None)
+        .ok().expect("Failed to execute find.");
+
+    let docs: Vec<_> = cursor.map(|doc| doc.unwrap())
```
fleupold: Nit: you should be able to do all of this in one map block

driver/src/bin/fetch_db.rs
```diff
@@ -0,0 +1,87 @@
...
+    let docs: Vec<_> = cursor.map(|doc| doc.unwrap()).collect();
+
+    let json: String = serde_json::to_string(&docs[0]).expect("Failed to parse json");
+
+    let deserialized: models::State = serde_json::from_str(&json).unwrap();
+    Ok(deserialized)
+}
+
+
+fn get_deposits_of_slot(slot: i32) -> Result<Vec< models::Deposits >, io::Error>{
```
fleupold: This function is returning a result but I can only see it ever returning Ok. Do you want to pass on the errors?

driver/src/bin/fetch_db.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let client = Client::connect("localhost", 27017)
+        .expect("Failed to initialize standalone client.");
+
+    let mut query=String::from(r#" { "slot": "#);
```
fleupold: Creating the query should be a one-liner with format strings: https://doc.rust-lang.org/std/fmt/

driver/src/bin/fetch_db.rs
```diff
@@ -0,0 +1,87 @@
...
+extern crate serde;
+extern crate mongodb;
+use std::io;
+use mongodb::bson;
+use mongodb::{Client, ThreadedClient};
+use mongodb::db::ThreadedDatabase;
+
+fn get_current_balances() -> Result<models::State, io::Error>{
+
+    let client = Client::connect("localhost", 27017)
```
fleupold: You probably want to read the hose/port from some configuration file somewhere

driver/src/bin/fetch_db.rs
```diff
@@ -0,0 +1,87 @@
...
+
+extern crate serde_json;
+extern crate serde;
+extern crate mongodb;
+use std::io;
+use mongodb::bson;
+use mongodb::{Client, ThreadedClient};
+use mongodb::db::ThreadedDatabase;
+
+fn get_current_balances() -> Result<models::State, io::Error>{
```
fleupold: In order to unit test the components that use these methods later, you might have to put these methods into a trait that can be mocked.

fleupold:
Turns out rust has support for uint128 (https://doc.rust-lang.org/book/print.html#integer-types)

driver/src/bin/database_setup.rs
```diff
@@ -0,0 +1,77 @@
...
+
+fn main() {
+
+    let client = Client::connect("localhost", 27017)
+        .expect("Failed to initialize standalone client.");
+
+    let coll = client.db("dfusion").collection("CurrentState");
+
+    let doc = doc! {
+        "CurrentState": "0000000000000000000000000000000000000000",
```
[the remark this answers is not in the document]
josojo: as discussed I will implement sha-hashes in another PR

driver/src/bin/database_setup.rs
```diff
@@ -0,0 +1,77 @@
...
+        "depositHash": "0000000000000000000000000000000000000000",
+        "depositIndex": "0000000000000000000000000000000000000000",
+        "slot": 1,
+        "addressId": 1,
+        "tokenId": 1,
+        "amount": 55465465,
+    };
+
+    // Insert document into 'dfusion.CurrentState' collection
+    coll.insert_one(doc2.clone(), None)
```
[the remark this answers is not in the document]
bh2smith: We are not currently emitting the depositHash with the deposit event... is this going to be a problem?

driver/src/bin/database_setup.rs
```diff
@@ -0,0 +1,77 @@
...
+        .ok().expect("Failed to insert CurrentState.");
+
+    let coll = client.db("dfusion").collection("State");
+
+
+	let state = models::State {
+	    curState: "0000000000000000000000000000000000000000".to_owned(),
+    	prevState: "0000000000000000000000000000000000000000".to_owned(),
+    	nextState: "0000000000000000000000000000000000000000".to_owned(),
+    	slot: 0,
```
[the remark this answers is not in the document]
josojo: It was meant to check for slot increments: From the current state, we could always see the current slot as well. But as discussed it is better to get the current slot from the current blcok

bh2smith:
Is this still an active branch? Or is this equivalent to the apply Deposits PR?

driver/src/bin/fetch_db.rs
josojo: sorry, I could still not figure this one out. The issue is that I have to work with raw... or represent the " somehow else. 

I will come to it back later.

fleupold:
No longer relevant.

PR #4, opened 2019-01-17
Event listener - first working demo

bh2smith:
This is a minimally functional event listener that filters Deposit and StateTransition events and posts them into respective collections (names deposits and transitions respectively)

screenshot 2019-01-17 at 4 57 53 pm

screenshot 2019-01-17 at 4 58 31 pm

This can be tested in it current state as follows;

Run ganache and mongo-db
```
ganache-cli
mongod
```

deploy a SnappBase Contract and paste its address into the SNAPP_CONTRACT_ADDRESS  in `/dex-services/.env`

make some transactions with this contract (I have been using the "successful apply deposit" unit test in the contract project by marking it with it.only and console.logging the contracts address after deployment)

execute the django-test as follows;

```
python manage.py test
```

Similarly, one can run this listener indefinitely via

```
python manage.py run_listener
```


bh2smith:
Great suggestion @josojo. I will do that.

event_listener/dfusion_db/event_receivers.py
```diff
@@ -0,0 +1,80 @@
...
+    @abstractmethod
+    def real_save(self, decoded_event, block_info=None):
+        pass
+
+    @abstractmethod
+    def real_rollback(self, decoded_event, block_info=None):
+        pass
+
+
+class DepositReceiver(GenericEventReceiver):
```
fleupold: In general, I prefer using composition over inheritance unless there is a good reason (generally composition gives more design flexibility and ends up being simpler as it's not trying to shoehorn different things into a common base). Happy to discuss this principle more in detail.

Here, I'd suggest to give GenericEventReceiver access to a mapping of String => AbstractEventReceiver and have an abstract event receiver for each type of event (DepositReceiver, StateTransitionReceiver).

This would avoid creating these "real_save", "real_deposit methods".

The GenericEventReceiver could be called something like DfusionEventDispatcher or something along those lines.

event_listener/dfusion_db/event_pusher.py
```diff
@@ -0,0 +1,43 @@
...
+from django.conf import settings
+
+client = MongoClient(
+    host=settings.MONGO_HOST,
+    port=settings.MONGO_PORT
+)
+db = client.get_database(settings.DB_NAME)
+
+
+def post_deposit(event: dict):
```
fleupold: Why did you chose to expose this as global methods rather than a class? Given the name I was expecting a `EventPusher` class.

event_listener/dfusion_db/event_pusher.py
```diff
@@ -0,0 +1,43 @@
...
+    :param event: dict
+    :return: bson.objectid.ObjectId
+    """
+    # Verify integrity of post data
+    assert event.keys() == {'accountId', 'tokenId', 'amount', 'slot'}, "Unexpected Event Keys"
+    assert all(isinstance(val, int) for val in event.values()), "One or more of deposit values not integer type"
+
+    deposits = db.deposits
+    deposit_id = deposits.insert_one(event).inserted_id
+    return deposit_id
```
fleupold: Maybe it's not supposed to be this layer, but I'd hope we could have one component that abstracts all the db specific code (in this case the fact that we are using mongoDb) away. I'm not sure if that layer should do any validation (e.g. the asserts above).

It would also require to return database agnostic return types (in this example a success boolean vs. deposit_id), in the case where we fetch data the raw json or a data model object.

event_listener/dfusion_db/event_pusher.py
bh2smith: I did this to avoid continually reinstatiating connections to the DB. If I move the connection factory into a class, it would be connecting on each event. I had considered a class, but thought this would be more valuable when there are more different type of posts being made. Anyway, I can start changing it today.

event_listener/dfusion_db/event_receivers.py
bh2smith: I considered using the old dictionary trick to avoid a long ugly if-elif-...-else and had decided against it assuming you wouldn’t like it. There is something that one need to be aware of when using dictionaries like this with classes as values. The python interpreter evaluates all of them when the mapping is declared. This means if we try to stuff a Deposit and StateTransition event into this, we could get failure when the event parameters aren’t correct.

bh2smith: I’m any case I am in favour of the dispatcher idea, but may need to be careful about using classes as values.

bh2smith: I’m any case I am in favour of the dispatcher idea, but may need to be careful about using classes as values.

bh2smith: I’m any case I am in favour of the dispatcher idea, but may need to be careful about using classes as values.

event_listener/config/db_settings.py
```diff
@@ -0,0 +1,3 @@
+MONGO_HOST = 'localhost'
```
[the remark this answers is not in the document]
bh2smith: Some in the .env file I assume you mean?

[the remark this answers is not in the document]
josojo: let's have two files for the settings on the top level.
the current .env, which is ignored by default as it has changing values like Snapp address
and another params.env, which contains non-varying variables like the host.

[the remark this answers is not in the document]
josojo: I would like to have it top level, so that I can use it in my rust code as well

event_listener/dfusion_db/event_receivers.py
bh2smith: We also need to consider what our dependencies allow us to do.

event_listener/README.md
```diff
@@ -0,0 +1,113 @@
+This can be tested as a standalone service as follows;
+
+
+Installation
+============
+
+OS requirements
```
josojo: I would suggest to simply write into the README how to start the docker compose up.
Then we would not require users to install specific python and what not!

bh2smith: Cool with me. Although, in this case the readme needs to go at the top level of the project.

event_listener/dfusion_db/event_pusher.py
bh2smith: Although, I would prefer if we make this part of the database infrastructure a proper issue and handle it in a separate branch.

PR #5, opened 2019-01-18
Setting up docker-compose for event_listener

fleupold:
This change sets up docker compose to work with the event_listener, mongo_db and ganache.

To reproduce, run:
1. `docker-compose up`
2. from the smart contracts repo run: `truffle migrate --network development`*
3. Take the contract address of snapp base and past it into the .env file of the dex-services repo.
4. `docker-compose restart listener`
5. Back in the smart contracts repo run:
 
```
truffle console --network development
> const me = (await web3.eth.getAccounts())[0]
> const instance =  await SnappBase.deployed()
> await instance.openAccount(1)
> const token = await ERC20Mintable.deployed()
> await instance.addToken(token.address)
> await token.mint(me, 10)
> await token.approve(instance.address, 10)
> await instance.deposit(1, 1)
```
6 now run mongoDB and verify you see new entries:
```
mongo
> use test_db
> db.deposits.find()
... should list deposits
```

_* where development is configured to connect to port 8545, also make sure ERC20Mintable is deployed as well*_


bh2smith:
So far all looks good, I am stuck trying to "make sure ERC20Mintable is deployed as well"

fleupold:
> Will any changes to the docker configuration need to be made once I move the DB configuration to the .env file?

Not as long as the adjusted hostname (mongo instead of local host) is copied correctly. 

I will let you merge this once the underlying event listener has been merged.

josojo:
For ubuntu users:
after changing the environment variable, you might have to run:
docker-compose up -d --remove-orphans --force-recreate
before
docker-compose restart listener

josojo:
Nope, was wrong. If I would have had .env from the start, it would not have been a problem at all.

PR #13, opened 2019-01-30
Withdraw Listener

bh2smith:
- Handles WithdrawRequest Event
- Handles StateTransition in the case of applied Withdrawals (i.e. performs balance updates on state transition of withdraw type).


event_listener/dfusion_db/event_pusher.py
```diff
@@ -69,11 +78,34 @@ def update_accounts(event: Dict[str, Union[int, str, str, int]]):
...
+        requested_withdraws = db.withdraws.find({'slot': event['slot']})
+
+        for deposit in requested_withdraws:
+            a_id = deposit['accountId']
+            t_id = deposit['tokenId']
+            amount = deposit['amount']
+
+            index = num_tokens * (a_id - 1) + (t_id - 1)
+            # Balances are stored as [b(a1, t1), b(a1, t2), ... b(a1, T), b(a2, t1), ...]
+            if balances[index] - amount >= 0:
```
josojo: I would prefer to check here also the bitmap from the snark, which will indicate whether a withdrawal was processed.

bh2smith: Right although we are not emitting the bitmap in the transition event. This bitmap is constructed based on the account states at the time of apply withdrawal....

bh2smith: I would argue that this is the correct logic since this is exactly how your driver code will construct the bitmap.

event_listener/dfusion_db/event_pusher.py
```diff
@@ -69,11 +78,34 @@ def update_accounts(event: Dict[str, Union[int, str, str, int]]):
         db.accounts.insert_one(new_account_record)
 
     elif transition_type == 1:  # Withdraw
```
fleupold: I wonder if this code could be made a bit cleaner and shared with deposits (it looks 50% similar). 

Nit: can transition_type be an enum to make the documentation explicit?

event_listener/dfusion_db/event_pusher.py
```diff
@@ -69,11 +78,34 @@ def update_accounts(event: Dict[str, Union[int, str, str, int]]):
...
-        pass
+
+        requested_withdraws = db.withdraws.find({'slot': event['slot']})
+
+        for deposit in requested_withdraws:
```
fleupold: bit: withdraw

event_listener/dfusion_db/event_pusher.py
```diff
@@ -69,11 +78,34 @@ def update_accounts(event: Dict[str, Union[int, str, str, int]]):
...
+            t_id = deposit['tokenId']
+            amount = deposit['amount']
+
+            index = num_tokens * (a_id - 1) + (t_id - 1)
+            # Balances are stored as [b(a1, t1), b(a1, t2), ... b(a1, T), b(a2, t1), ...]
+            if balances[index] - amount >= 0:
+                _log.info("Decreasing balance of account {} - token {} by {}".format(a_id, t_id, amount))
+                balances[index] -= amount
+            else:
+                _log.info("Insufficient balance: account {} - token {} for amount {}".format(a_id, t_id, amount))
```
fleupold: Should we really leave the balance untouched, rather than making it negative?

I wonder if this can be attacked now, by posting a state transition that withdraws an infinite amount from an account, but leaves the root hash the same (which needs to be valid, in case there were only invalid withdraws).

As far as a driver can tell the transition would be valid (since all balances in the db still hash to the same state root) and they would likely not challenge the transition. However, if the balance was negative, they would know to challenge.

event_listener/dfusion_db/event_pusher.py
bh2smith: Are you aware of the "Suggestion" button here? I would be way easier for me to change lines like this directly in the browser.

bh2smith: Ah, yes, but I will need to do it on several lines.

event_listener/dfusion_db/event_pusher.py
bh2smith: I thought the point was that insufficient balance withdraw requests are untouched (as in not included in the bitmap). If we make the balances negative, then people would never be able to withdraw their actual balances in the future.

Suppose someone has a balance of 5 and request a withdraw of 10. We subtract their balance to -5 and don't process their request. Now they can never access their 5 again.

Please tell me if I am missing something. Keeping in mind that untouched request are simply encoded in the bitmap published to the claimable state.

event_listener/dfusion_db/event_pusher.py
bh2smith: Yes I like the Enum idea, but I still think deposit and withdraw logic are different and should be kept separate. This is due to the fact that some withdraw requests are not processed (i.e. due to insufficient balances mentioned in another place).

event_listener/dfusion_db/event_pusher.py
bh2smith: Ahhh, so we are trying to consider a case where none of the withdraw requests were valid and the state transition has an identical hash as the previous. I don't really see how this is a problem or should be challenged. I also don't yet understand the purpose of the "infinite" withdraw request as part of this example. Maybe you could elaborate. 

As of now, I am still unconvinced that the current logic is inconsistent.

event_listener/dfusion_db/event_pusher.py
bh2smith: Fixed!

event_listener/dfusion_db/event_pusher.py
fleupold: I think this discussion is somewhat similar with the bitmap discussion above. After thinking about it more, I think you are right that this should probably just mirror the driver's logic.

event_listener/dfusion_db/event_pusher.py
fleupold: I think this logic should be shared, since it's identical:

```
a_id = deposit['accountId']
t_id = deposit['tokenId']
amount = deposit['amount']
index = num_tokens * (a_id - 1) + (t_id - 1)
```

bh2smith:
@fleupold 

Here is a proposed refactor of the deposit withdraw application for updating balances, but I am not sure this is the best approach.

```python


COLLECTION_MAP = {
    0: 'deposits',
    1: 'withdraws',
}

applied_data = db.get_collection(COLLECTION_MAP[transition_type]).find({'slot': event['slot']})


for datum in applied_data:
    a_id = deposit['accountId']
    t_id = deposit['tokenId']
    amount = deposit['amount']
    index = num_tokens * (a_id - 1) + (t_id - 1)

    
    if transition_type == TRANSITION_TYPES['Deposit']:
        # Balances are stored as [b(a1, t1), b(a1, t2), ... b(a1, T), b(a2, t1), ...]

        _log.info("Incrementing balance of account {} - token {} by {}".format(a_id, t_id, amount))
        balances[num_tokens * (a_id - 1) + (t_id - 1)] += amount

    elif transition_type == TRANSITION_TYPES['Withdraw']:

        if balances[index] - amount >= 0:
            _log.info("Decreasing balance of account {} - token {} by {}".format(a_id, t_id, amount))
            balances[index] -= amount
        else:
            _log.info("Insufficient balance: account {} - token {} for amount {}".format(a_id, t_id, amount))

new_account_record = {
    'stateIndex': state_index,
    'stateHash': state_hash,
    'balances': balances
}

db.accounts.insert_one(new_account_record)
```

fleupold:
Yeah, looks reasonable. I'd use a proper enum instead of a mapping and not insert_one in case of auction or else just yet. 

You can go ahead and merge this one first. I can rebase my refactoring on top of it.

bh2smith:
@fleupold  Alright I'm gonna merge it and can start adapting this to the refactored branch. Just another point, I don't think enums are a thing in python, unless you mean some sort of hand made one.

PR #19, opened 2019-01-31
Dex-contracts as submodule

josojo:
I have only adapted the driver abi's reading.

@bh2smith could you do the adaption for the event_listener. I could not add it quickly. sry


bh2smith:
Hey @josojo I would love to help with this, but wonder if we might want to wait till @fleupold's refactoring of the Event Receiver is merged

josojo:
Just talked to felix, it should not have an influence... you can already go ahead. Of course only once you have time @bh2smith

bh2smith:
@josojo I am sorry to say, but I really don't think you should have `driver/build/contracts/SnappBase.json` copied into this location. Can't you reference `/dex-contracts/build/contracts`?

PR #20, opened 2019-01-31
New linting

josojo:
finally got a rust linter installed


bh2smith:
Can we include a linter in the travis build? We should also use pylint for the python!

I notice that the indentation in rust is very very deep? Is this normal? Configurable?

josojo:
I made a new issue for outstanding tasks. Sure, we want it. But I will do it later!

PR #21, opened 2019-01-31
Refactoring EventListener into dispatcher with separate base class

fleupold:
Implementation of suggestion #4 

The concrete event receiver don't have to confirm to `AbstractEventReceiver`. Instead there is one generic EventDispatcher, that parses the event and forwards it to implementations of `SnappEventReceiver`.

This makes the class hierarchy simpler and doesn't rely on inherited logic to be executed. In order to register a new event, one now has to implement a `SnappEventReceiver` and register its instance in the `RECEIVER_MAPPING` dict.

Currently rollback is not implemented, but it would be added to the `SnappEventReceiver` protocol and dispatched in a similar way as save.


event_listener/dfusion_db/snapp_event_receiver.py
```diff
@@ -1,58 +1,17 @@
...
-    @abstractmethod
-    def real_save(self, decoded_event, block_info=None):
-        pass
-
+class SnappEventListener(ABC):
+    """Abstract SnappEventReceiver class."""
     @abstractmethod
-    def real_rollback(self, decoded_event, block_info=None):
-        pass
+    def save(self, event:Dict[str, Any], block_info): pass
```
bh2smith: Should we also include rollback here?

fleupold: Right now, there is no functionality for rollbacks (it's a TODO to add it). Once we add this functionality, we will have to add a method there, yes.

bh2smith: Good enough for me!

bh2smith:
Yes, all great, the sooner we merge this, the sooner I can migrate the Withdrawal logic.

PR #26, opened 2019-02-01
[Refactor] Introduce Database interface as class that can be mocked

fleupold:
This diff abstracts all read/writes to the database into a layer whose interface is agnostic to the concrete database used (can be mongo, sqlite, postgres). 

All other "logic" is moved into the event receiver, which should eventually be unit-tested.

This redesign is a prerequisite for having a unit-testable architecture, as we can now inject a mocked db interface into the receiver. The mock can assert that the correct states/deposits/constants are inserted.

I will probably make another PR that converts all loosely typed dictionaries into strongly typed value classes before adding the unit tests.


event_listener/dfusion_db/snapp_event_receiver.py
```diff
@@ -59,29 +65,51 @@ def update_accounts(self, event: Dict[str, Union[int, str, str, int]]):
...
+            amount = datum['amount']
+
+            # Balances are stored as [b(a1, t1), b(a1, t2), ... b(a1, T), b(a2, t1), ...]
+            index = num_tokens * (a_id - 1) + (t_id - 1)
+
+            if transition_type == self.TRANSITION_TYPES['Deposit']:
 
-                # Balances are stored as [b(a1, t1), b(a1, t2), ... b(a1, T), b(a2, t1), ...]
+                self.logger.info("Incrementing balance of account {} - token {} by {}".format(a_id, t_id, amount))
                 balances[num_tokens * (a_id - 1) + (t_id - 1)] += amount
```
bh2smith: Could use the index declared above here for consistency. (I forgot when reorganizing)

bh2smith:
It’s going to be tricky making the strongly typed value classes. We should start with some fundamentals like bytes32 and stuff like this.

fleupold:
>It’s going to be tricky making the strongly typed value classes.

I think it already helps a lot passing around classes that explicitly declare what fields they have (and require them in the constructor) rather than passing some magic dictionary with an assumption of keys. I'd imagine `save` to just parse the dictionary into whatever data model is required for the methods below and pass it on to some method `saveParsed` or so which then will be the entry point of the unit tests.

What do you think?

bh2smith:
Yes that seems alright. I was really hoping that we could formally define a state transition type class where we could ensure certain values were bytes 32 and other such EVM primitives (doing this might actually help contribute to a python library for EVM and could be very useful for others.

PR #27, opened 2019-02-12
Introducing value objects (NamedTuples) for all data objects

fleupold:
Also  and enforcing strict type checks in travis


event_listener/dfusion_db/database_interface.py
```diff
@@ -1,64 +1,91 @@
...
-    def get_account_state(self, index: int) -> Dict:
-        return self.db.accounts.find_one({'stateIndex': index})
+    def get_account_state(self, index: int) -> AccountRecord:
+        record = self.db.accounts.find_one({'stateIndex': index})
+        return AccountRecord(record["stateIndex"], record["stateHash"], record["balances"])
 
-    def get_deposits(self, slot: int) -> Dict:
-        return self.db.deposits.find({'slot': slot})
+    def get_deposits(self, slot: int) -> List[Deposit]:
+        return list(map(lambda d: Deposit.fromDictionary(d), self.db.deposits.find({'slot': slot})))
```
bh2smith: python style guide uses `snake_case` for variables and functions

event_listener/dfusion_db/snapp_event_receiver.py
```diff
@@ -1,118 +1,80 @@
...
 
-    def get_data_to_apply(self, transition_type, slot):
-        if transition_type == 0:
-            return self.db.get_deposits(slot)
-        elif transition_type == 1:
-            return self.db.get_withdraws(slot)
+    def __get_data_to_apply(self, transition: StateTransition) -> Union[List[Withdraw], List[Deposit]]:
+        if transition.transition_type == TransitionType.Deposit:
+            return self.db.get_deposits(transition.slot)
+        elif transition.transition_type == TransitionType.Deposit:
```
bh2smith:
```suggestion
        elif transition.transition_type == TransitionType.Withdraw:
```

event_listener/dfusion_db/models.py
```diff
@@ -0,0 +1,68 @@
...
+    Auction = 2
+
+class StateTransition(NamedTuple):
+    transition_type: TransitionType
+    state_index: int
+    state_hash: str
+    slot: int
+
+    @classmethod
+    def fromDictionary(cls, data: Dict[str, Any]) -> StateTransition:
```
bh2smith: Can we actually refer to a Class before it has been defined? My linter says Unresolved Reference.

.travis.yml
```diff
@@ -1,8 +1,16 @@
...
+    - language: rust
+      rust:
+        - 1.31.0
+      cache: cargo 
+      script:
+        - cd driver/
+        - cargo build --verbose --all
+        - cargo test --verbose --all
+    - language: python
+      python: "3.7-dev"
```
bh2smith: We are restricted to using python 3.6 because of our dependency on django-eth-events which only works in 3.6

event_listener/dfusion_db/models.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
```
bh2smith: This annotations library was not found on my end and does not appear to be in use here. What is it for?

event_listener/dfusion_db/snapp_event_receiver.py
```diff
@@ -1,118 +1,80 @@
...
 class SnappEventListener(ABC):
     """Abstract SnappEventReceiver class."""
-    def __init__(self, database_interface=MongoDbInterface()):
+    def __init__(self, database_interface: DatabaseInterface = MongoDbInterface()):
         self.db = database_interface
         self.logger = logging.getLogger(__name__)
 
     @abstractmethod
-    def save(self, event:Dict[str, Any], block_info): pass
+    def save(self, event:Dict[str, Any], block_info: Dict[str, Any]) -> None: pass
```
bh2smith:
```suggestion
    def save(self, event: Dict[str, Any], block_info: Dict[str, Any]) -> None: pass
```

event_listener/dfusion_db/models.py
fleupold: that's what the annotation import below is for

.travis.yml
fleupold: Why would it not work on something newer?

event_listener/dfusion_db/models.py
fleupold: For evaluating class annotations after class definitions. You'll need python 3.7 for it

event_listener/requirements.txt
```diff
@@ -5,3 +5,5 @@ django-model-utils==3.1.2
 ethereum==2.3.2
 pymongo==3.7.2
 web3==4.8.2
```
bh2smith: Are we planning to include pylint inthe requirements?

If so, I believe it would be
```
pylint==2.1.1
```

At least this is what I have locally for my python3.7

PR #29, opened 2019-02-18
[Easy] fix event receiver

fleupold:
My refactoring broke the event receiver (which I discovered by running the e2e tests). This PR should fix the issues introduced.

### DIFF
--- a/event_listener/dfusion_db/models.py
+++ b/event_listener/dfusion_db/models.py
@@ -37,16 +37,15 @@ class Deposit(NamedTuple):
 
     @classmethod
     def from_dictionary(cls, data: Dict[str, Any]) -> "Deposit":
-        assert data.keys() == {'accountId', 'tokenId', 'amount',
-                               'slot', 'slotIndex'}, "Unexpected Event Keys"
-        assert all(isinstance(val, int) for val in data.values()
-                   ), "One or more of event values not integer"
+        assert all(k in data for k in(
+            'accountId', 'tokenId', 'amount', 'slot', 'slotIndex')), \
+            "Unexpected Event Keys"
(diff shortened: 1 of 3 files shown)


bh2smith:
My only question is what do you expect to happen when the values inside this dict are not of integer type.

event_listener/dfusion_db/models.py
```diff
@@ -37,16 +37,15 @@ class Deposit(NamedTuple):
 
     @classmethod
     def from_dictionary(cls, data: Dict[str, Any]) -> "Deposit":
-        assert data.keys() == {'accountId', 'tokenId', 'amount',
-                               'slot', 'slotIndex'}, "Unexpected Event Keys"
-        assert all(isinstance(val, int) for val in data.values()
```
bh2smith: I see this assertion has been removed...

event_listener/dfusion_db/snapp_event_receiver.py
```diff
@@ -102,7 +102,7 @@ def save(self, event: Dict[str, Any], block_info: Dict[str, Any]) -> None:
 
         try:
             self.initialize_accounts(
-                event['maxTokens'], event['numAccounts'], state_hash)
+                event['maxTokens'], event['maxAccounts'], state_hash)
```
bh2smith: This looks like the reason something was breaking.

fleupold:
> My only question is what do you expect to happen when the values inside this dict are not of integer type.

We still check this since `int()` will throw if it's argument cannot be converted to integer. I can add a unit test.

bh2smith:
Seems good to me. Feel free to merge!

PR #30, opened 2019-02-18
db interaction refactoring

josojo:
While it is not yet a real interface, I managed to separate so that we can easily write tests and no longer need to clone the whole client.

db tests are not yet mocked. But tests do work if docker-compose up is running

@fleupold Need your opinion on whether this goes in the right direction.

### DIFF
--- a/driver/src/db_interface.rs
+++ b/driver/src/db_interface.rs
@@ -1,4 +1,3 @@
-
 pub mod db_interface{    
     use mongodb::bson;
     use mongodb::db::ThreadedDatabase;
@@ -6,6 +5,22 @@ pub mod db_interface{
 
     use std::env;
 
+    
+    extern crate models;
+    extern crate rustc_hex;
+    extern crate web3;
+
+    use web3::types::{Address, H256, U256};
+
+    use std::fs;
+    use std::io;
+    use std::io::{Error, ErrorKind};
+    use std::sync::mpsc;
+    use std::thread;
+    use std::time::Duration;
+    use std::process;
+    
+    #[derive(Clone)]
     pub struct DbInterface{
         pub client: Client,
         pub db_host: String,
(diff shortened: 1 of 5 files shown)


driver/src/main.rs
```diff
@@ -97,11 +42,12 @@ fn main() {
 		let received = rx.recv().unwrap();
 		println!(": {}", received);
 
-		// initializing all needed web3 variables
 		let db_host = env::var("DB_HOST").unwrap();
-		let db_port = env::var("DB_PORT").unwrap();
-		let client = Client::connect(&db_host, db_port.parse::<u16>().unwrap())
-			.expect("Failed to initialize standalone client");
+        let db_port = env::var("DB_PORT").unwrap();
```
fleupold: Could you spend some time to fix your indentation settings in your IDE? I have seen this being messed up a lot in recent PRs.

driver/src/db_interface.rs
```diff
@@ -0,0 +1,137 @@
...
+        let bson = json.into();
+        let temp: bson::Document = mongodb::from_bson(bson).expect("Failed to convert bson to document");
+        
+        // Insert document into 'dfusion.CurrentState' collection
+        coll.insert_one(temp.clone(), None)
+        .ok().expect("Failed to insert test state");
+
+
+        let d = String::from(r#" "0xf00000000000000000000000000000000000000000000000000000000000000f""#);
+		let state_root: H256 =
```
fleupold: Indentation

driver/src/main.rs
```diff
@@ -97,11 +42,12 @@ fn main() {
...
+		let db_instance = db_interface::db_interface::DbInterface::new(db_host, db_port).unwrap_or_else(|err| {
+      		println!("Problem creating DbInterface: {}", err);
+			process::exit(1);
```
fleupold: Do you want to return here instead of exiting? If you really want to exit, I think we should `panic!` instead.

driver/src/db_interface.rs
```diff
@@ -0,0 +1,137 @@
...
+        pub fn new(db_host: String, db_port: String) -> Result<DbInterface, &'static str> {
+            let client = Client::connect(&db_host, db_port.parse::<u16>().unwrap())
+                .expect("Failed to initialize standalone client");
+
+            Ok(DbInterface { client, db_host, db_port })
+        }
+
+        pub fn get_current_balances(&self, current_state_root: H256) -> Result<models::State, Error> {
+            let t: String = format!("{:#x}", current_state_root);
+            let mut query = String::from(r#" { "stateHash": ""#);
```
fleupold: nit: maybe construct the query with one `format!`

driver/src/db_interface.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+            let json: String = serde_json::to_string(&docs[0]).expect("Failed to parse json");
+
+            let deserialized: models::State = serde_json::from_str(&json)?;
+            Ok(deserialized)
+        }
+        
+        pub fn get_deposits_of_slot(&self, slot: i32) -> Result<Vec<models::Deposits>, io::Error> {
+            let mut query = String::from(r#" { "slot": "#);
```
fleupold: ditto

driver/src/db_interface.rs
josojo: Yes, this would be nice. however, I was not able to process a raw string with format!

driver/src/db_interface.rs
```diff
@@ -0,0 +1,153 @@
...
+        current_state_root: H256,
+    ) -> Result<models::State, Error>;
+    fn get_deposits_of_slot(
+        &self,
+        slot: i32,
+    ) -> Result<Vec<models::Deposits>, io::Error>;
+}
+
+#[derive(Clone)]
+pub struct DbMongoInstance {
```
fleupold: nit: naming, a struct is not really an instance (an instance of a struct is). Maybe just call this MongoDB?

driver/src/db_interface.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub client: Client,
+}
+impl DbMongoInstance {
+    pub fn new(db_host: String, db_port: String) -> Result<DbMongoInstance, &'static str> {
+        let client = Client::connect(&db_host, db_port.parse::<u16>().unwrap()).expect("wrong");
```
fleupold: nit: no expect

driver/src/db_interface.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        let coll = self.client.db(models::DB_NAME).collection("accounts");
+
+        let cursor = coll.find(Some(_temp), None)?;
+
+        let docs: Vec<bson::ordered::OrderedDocument> = cursor.map(|doc| doc.unwrap()).collect();
+
+        if docs.len() == 0 {
+            Error::new(ErrorKind::Other, "Error, state was not found");
+            println!("here is the problem")
```
fleupold: cleanup

driver/src/db_interface.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        Ok(DbMongoInstance { client })
+    }
+}
+impl DbInterface for DbMongoInstance {
+    fn get_current_balances(
+        &self,
+        current_state_root: H256,
+    ) -> Result<models::State, Error> {
+        let t: String = format!("{:#x}", current_state_root);
```
fleupold: Nit, query concatenation should be possible in one line with a single format?

driver/src/db_interface.rs
```diff
@@ -0,0 +1,153 @@
...
+         let coll = self.client.db(models::DB_NAME).collection("deposits");
+
+        let cursor = coll.find(Some(_temp), None)?;
+
+        let mut docs: Vec<models::Deposits> = cursor
+            .map(|doc| doc.unwrap())
+            .map(|doc| {
+                serde_json::to_string(&doc)
+                    .map(|json| serde_json::from_str(&json).unwrap())
+                    .expect("Failed to parse json")
```
fleupold: I guess we should warn but ignore incorrectly parsed json

driver/src/main.rs
```diff
@@ -97,11 +41,13 @@ fn main() {
...
 
-		// initializing all needed web3 variables
 		let db_host = env::var("DB_HOST").unwrap();
 		let db_port = env::var("DB_PORT").unwrap();
-		let client = Client::connect(&db_host, db_port.parse::<u16>().unwrap())
-			.expect("Failed to initialize standalone client");
+		let db_instance =
+			db_interface::DbMongoInstance::new(db_host, db_port).unwrap_or_else(|err| {
+				println!("Problem creating DbInterface: {}", err);
+				process::exit(1);
```
fleupold: 🙅‍♂️ please no process::exit

driver/src/db_interface.rs
```diff
@@ -0,0 +1,153 @@
...
+            })
+            .collect();
+
+        docs.sort_by(|a, b| b.slot.cmp(&a.slot));
+        Ok(docs)
+    }
+}
+
+// #[cfg(test)]
```
fleupold: please don't check in this code

driver/src/main.rs
josojo: This will be refactored in the other PR...

driver/src/db_interface.rs
josojo: sure

driver/src/main.rs
fleupold: It would then be nice to not have this change at all in this PR. But not a big deal.

driver/src/db_interface.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        let coll = self.client.db(models::DB_NAME).collection("deposits");
+
+        let cursor = coll.find(Some(_temp), None)?;
+
+        let mut docs: Vec<models::Deposits> = cursor
+            .map(|doc| doc.unwrap())
+            .map(|doc| {
+                serde_json::to_string(&doc)
+                    .map(|json| serde_json::from_str(&json).unwrap()).expect("needed here")
```
fleupold: "needed here" seems off. I think it could print a warning and just ignore the deposit

driver/src/db_interface.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn get_current_balances(
+        &self,
+        current_state_root: H256,
+    ) -> Result<models::State, Box<dyn std::error::Error>> {
+        let query: String = format!("{{ \"stateHash\": \"{:x}\" }}", current_state_root);
+
+        let v: serde_json::Value = serde_json::from_str(&query)?;
+        let bson = v.into();
+        let mut _temp: bson::ordered::OrderedDocument =
+            mongodb::from_bson(bson).expect("Failed to convert bson to document");
```
fleupold: Should this expect still be here?

driver/src/db_interface.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        let coll = self.client.db(models::DB_NAME).collection("accounts");
+
+        let cursor = coll.find(Some(_temp), None)?;
+
+        let docs: Vec<bson::ordered::OrderedDocument> = cursor.map(|doc| doc.unwrap()).collect();
+
+        if docs.len() == 0 {
+            Error::new(ErrorKind::Other, "Error, state was not found");
```
fleupold:
```suggestion
            Error::new(ErrorKind::Other, "State was not found");
```

driver/src/db_interface.rs
josojo: removed

josojo:
I think I addressed all comments and all unwraps are removed in db_inteface. Other cleanings will come with the second pr.

driver/src/db_interface.rs
fleupold: still here. I guess it was removed in the other PR?

driver/src/db_interface.rs
```diff
@@ -0,0 +1,105 @@
...
+            mongodb::from_bson(bson)?;
+
+        let coll = self.client.db(models::DB_NAME).collection("deposits");
+
+        let cursor = coll.find(Some(_temp), None)?;
+
+        let mut docs: Vec<models::Deposits> =vec!();
+        for result in cursor {
+            if let Ok(item) = result {
+                if let Ok(deposit_str) = serde_json::to_string(&item){
```
fleupold: You are `to_string` and `from_string` here right after another? What is this actually achieving? Is there not a more concise way to do it?

driver/src/db_interface.rs
```diff
@@ -0,0 +1,105 @@
...
+                    if let Ok(deposit) = serde_json::from_str(&deposit_str){
```
fleupold: instead of having nested `if let Ok(...)` with the same error message, you can use `map` on Result, which applies a method to the Ok part, and does nothing if it's already an error.

PR #31, opened 2019-02-19
refactor into deposit_driver: extracting logic from main and removing threading

josojo:


fleupold:
The problem is that you are spawning a new thread inside `run_deposit_listener()`, but you are not waiting on it. This means the program exits from this function and therefore destroys the local stack state, including the `thread::spawn` object (which happens to be in the middle of trying to read from the smart contract and therefore says it "timed out").

If you put a thread::sleep inside the function (not a valid fix), you can see that it's working again.

fleupold:
Why do you want to have multiple threads in the first place? 
I think a single `run` should happen within the same thread. Introducing threads increases the complexity dramatically and makes the system harder to reason about. Unless there is a strong reason why we have to do the different steps concurrently (fetch data, etc) we should do them in a single thread in serial.

josojo:
thanks for the feedback, it helped a lot!

driver/src/deposit_driver.rs
```diff
@@ -0,0 +1,160 @@
...
+	);
+
+	// if 20 blocks have past since the first deposit and we are not in the newest slot, we apply the deposit.
+	if current_deposit_ind_block + 20 < current_block
+		&& deposit_ind != current_deposit_ind.low_u32() as i32 + 1
+	{
+		println!("Next deposit_slot to be processed is {}", deposit_ind);
+		let deposits = db_interface::get_deposits_of_slot(db_instance.clone(), deposit_ind)?;
+		println!("Amount of deposits to be processed{:?}", deposits.len());
+		//rehash deposits
```
fleupold: consider extracting this logic into a small helper method `rehash_deposits`, than you have documentation that is unlikely to get outdated for free and you top level method is shorter and easier to read.

driver/src/deposit_driver.rs
```diff
@@ -0,0 +1,160 @@
...
+	// get current state
+	let result = contract.query("getCurrentStateRoot", (), None, Options::default(), None);
+	let curr_state_root: H256 = result.wait()?;
+	let mut state = db_interface::get_current_balances(db_instance.clone(), curr_state_root.clone())?;
+	let accounts = web3.eth().accounts().wait()?;
+
+	// check that operator has sufficient ether
+	let balance = web3.eth().balance(accounts[0], None).wait()?;
+	if balance.is_zero() {
+		panic!("Not sufficient balance for posting updates into the chain");
```
fleupold: let's not panic unless necessary

driver/src/models.rs
```diff
@@ -1,24 +1,17 @@
...
 use web3::types::H256;
 
+
 pub const ACCOUNTS: i32 = 100;
-pub const BITS_PER_ACCOUNT: i32 = 4;
 pub const TOKENS: i32 = 30;
-pub const BITS_PER_TOKENS: i32 = 3;
 pub const SIZE_BALANCE: usize = (ACCOUNTS * TOKENS) as usize;
-pub const BITS_PER_BALANCE: i32 = 30;
+// pub const BITS_PER_BALANCE: i32 = 30;
```
fleupold: Is this not needed anymore? If so, can we remove?

josojo:
The search for the address solution is rediculous:

I coded now a custom solution:
```
let mut snapp_address = env::var("SNAPP_CONTRACT_ADDRESS")?;
	let s = models::decode_hex_uint8(&mut snapp_address[2..],20)?;
    let mut h: [u8; 20] = [0; 20];
	for i in 0..19 {
		h[i] = s[i];
	}
	println!("{:?}", s);
	//let address: Address = Address::from(s.);
	let address: Address = Address::from("0xC89Ce4735882C9F0f0FE26686c53074E09B0D550");
	println!("{:}",address);
	let address: Address = Address::from(h);
	println!("{:}",address);
```
with
```
pub fn decode_hex_uint8(s: &mut str, size: i32) -> Result<Vec<u8>, Box<dyn Error>> {
  // add prefix 0, in case s has not even length
  let mut pretail: &str = "";
  if s.len() % 2 == 1 {
    pretail = "0";
  }
  let p: &'static str = pretail.into();
  let s = format!("{}{}", p, s);

  let v: Result<Vec<u8>, ParseIntError> = (0..s.len())
    .step_by(2)
    .map(|i| u8::from_str_radix(&s[i..i + 2], 16))
    .collect();

  let mut v = v?;
  let mut vv = Vec::with_capacity(size as usize);
  for _i in 0..size {
    vv.push(0);
  }
  for i in 1..v.len() + 1 {
    vv[size as usize - i] = v.pop().unwrap();
  }
  Ok(vv.clone())
}
```


and the output of the address is correct:
driver_1       | [200, 156, 228, 115, 88, 130, 201, 240, 240, 254, 38, 104, 108, 83, 7, 78, 9, 176, 213, 80]
driver_1       | 0xc89c…d550
driver_1       | 0xc89c…d500
ganache-cli_1  | eth_call
driver_1       | Deposit_driver error: please ensure the contract and method you're calling exist! failed to decode empty bytes. if you're using jsonrpc this is likely due to jsonrpc returning `0x` in case contract or method don't exist

Still I am getting this stupid error. WHY?
any hints @fleupold

PR #33, opened 2019-02-20
Adding a Mongo UI to docker-compose

fleupold:
This will give us an easy to use a UI to inspect the current state of the database.

Run `docker-compose up`, visit [http://localhost:3000/](http://localhost:3000/), connect to the default database. On the left side bar, under Connections select accounts and execute query - notice initial account state.

Run e2e test from shell. Re-execute query in UI and observe that we now have three state updates.

screen shot 2019-02-20 at 11 49 04


bh2smith:
Could we include the link to the GUI in the reader?

PR #35, opened 2019-02-20
Making sure withdraws are working from the event listener side

fleupold:
This PR adds the RequestWitdrawlHandler to the receive list and adds an e2e test, which we can as we build the missing pieces (claim script, driver support)


test/e2e-tests-withdraw.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/bash
+set -e
+
+cd dex-contracts/
+
+truffle exec scripts/request_withdraw.js 3 3 18
+truffle exec scripts/mine_blocks.js 21
+
+#TODO @josojo: this should happen by the driver
+truffle exec scripts/apply_withdrawals.js 0 0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 0x0000000000000000000000000000000000000000000000000000000000000001 0x0000000000000000000000000000000000000000000000000000000000000001
```
bh2smith: I have already removed, this bitmap from the contracts. This is no longer necessary.

fleupold: Will have to pull the new contracts before we can remove this

PR #37, opened 2019-02-21
[Driver] Interface for SmartContract interaction

fleupold:
So far just a scaffold, feel free to commandeer


driver/src/contract.rs
```diff
@@ -0,0 +1,116 @@
...
+        self.contract.query(
+            "getDepositCreationBlock",
+            slot,
+            None,
+            Options::default(),
+            None,
+        ).wait()
+    }
+
+    fn get_current_block_number(self) -> Result<U256> {
```
josojo: Maybe we can put these functions into a separate trait with generic implementation

PR #40, opened 2019-02-22
[Driver] process withdraws

fleupold:
This is just the beginning. Leaving this out here, if someone wants to talk over while I'm on vacation.


driver/src/deposit_driver.rs
```diff
@@ -86,10 +86,7 @@ pub fn run_deposit_listener() -> Result<(), Box<dyn Error>> {
 			//applyDeposits signature is (slot, _currStateRoot, _newStateRoot, deposit_slotHash)
 			let slot = U256::from(deposit_ind);
 			let _curr_state_root = curr_state_root;
-			let mut d = String::from(r#" "0x"#);
-			d.push_str(&state.hash()?);
-			d.push_str(r#"""#);
-			let _new_state_root: H256 = serde_json::from_str(&d)?;
+			let _new_state_root = H256::from(state.hash()?);
```
josojo: love it...

driver/src/withdraw_driver.rs
```diff
@@ -53,32 +73,34 @@ pub fn run_withdraw_listener<D, C>(db: &D, contract: &C) -> Result<(), Box<dyn E
...
+            let balances = db.get_current_balances(&state_root)?;
 
             let withdraws = db.get_withdraws_of_slot(slot.low_u32())?;
-            // Hash withdraws and compare with contract
-            let mut withdraw_hash: H256 = H256::zero();
-            for pat in &withdraws {
-                withdraw_hash = pat.iter_hash(&mut withdraw_hash)
+            let withdraw_hash = withdraws.iter().fold(H256::zero(), |acc, w| w.iter_hash(&acc));
+            if withdraw_hash != contract_withdraw_hash {
+                return Err(Box::new(DriverError::new(
```
josojo: I thought that a panic is appropriate here,
If I remember correctly, in the guidelines, it was stated that corrupted data ( which we would have in this case) is a reason for panic. 
Imagine we know the error over here and we know that something is wrong with our data, but then some other part of the program keeps on calculating a wrong status update..

josojo:
And as I stated already in the chat, I am a big fan of this more efficient hashing of the claimableBalance_Merkleroot

test/e2e-tests-withdraw.sh
```diff
@@ -1,16 +1,16 @@
...
-truffle exec scripts/apply_withdrawals.js 1 0x0000000000000000000000000000000000000000000000000000000000000002 0x0000000000000000000000000000000000000000000000000000000000000002
 sleep 5
+truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot'
 
-EXPECTED_HASH="0000000000000000000000000000000000000000000000000000000000000002"
+EXPECTED_HASH="77b01abfbad57cb7a1344b12709603ea3b9ad803ef5ea09814ca212748f54733"
 truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot' | grep $EXPECTED_HASH
 mongo dfusion2 --eval "db.accounts.findOne({'stateHash': '$EXPECTED_HASH'}).balances[62]" | grep 0
+
+truffle exec scripts/claim_withdraw.js 1 3 3 | grep "Success! Token balance before 82, after 100"
```
bh2smith: Shouldn't the grep just be for the "Success!" string? Also I find the wording here confusing.

bh2smith: Never mind.. I see now

PR #46, opened 2019-03-12
Refactoring Error to be package specific type for better unit testing

fleupold:
For use in Mock framework the return types have to be cloneable, which a generic error isn't. Also introducing a kind makes it easier to test, we are failing for the "right" reason.

### DIFF
--- a/driver/src/error.rs
+++ b/driver/src/error.rs
@@ -1,14 +1,85 @@
 use std::error::Error;
 use std::fmt;
+use ethabi;
 
-#[derive(Debug)]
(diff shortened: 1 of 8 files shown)


driver/src/error.rs
```diff
@@ -1,14 +1,85 @@
 use std::error::Error;
 use std::fmt;
+use ethabi;
 
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
+pub enum ErrorKind {
```
bh2smith: PeopleKind!

PR #47, opened 2019-03-12
Creating mocks and testing withdraw_driver

fleupold:
This adds basic unit tests for withdraw and deposit driver.

### DIFF
--- a/driver/src/deposit_driver.rs
+++ b/driver/src/deposit_driver.rs
@@ -1,4 +1,5 @@
 use crate::models;
+
 use crate::db_interface::DbInterface;
 use crate::error::{DriverError, ErrorKind};
 use crate::contract::SnappContract;
@@ -15,7 +16,7 @@ pub fn apply_deposits(
     state.clone()
 }
 
-pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(), DriverError> 
+pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(bool), DriverError> 
     where   D: DbInterface,
             C: SnappContract
 {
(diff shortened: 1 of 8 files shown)


driver/src/models.rs
```diff
@@ -77,8 +81,8 @@ impl State {
 #[allow(non_snake_case)]
 #[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
 pub struct PendingFlux {
-  pub slotIndex: i32,
-  pub slot: i32,
+  pub slotIndex: u32,
```
bh2smith: good catch!

driver/src/deposit_driver.rs
```diff
@@ -86,9 +87,167 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(), DriverErro
...
+        contract.deposit_hash_for_slot.given(slot + 1).will_return(Ok(H256::zero()));
+
+        let db = DbInterfaceMock::new();
+        db.get_current_balances.given(state_hash).will_return(Ok(state));
+
+        assert_eq!(run_deposit_listener(&db, &contract), Ok(false));
+    }
+
+    #[test]
+    fn does_not_apply_if_highest_slot_too_close_to_current_block() {
```
bh2smith: This seems alright, but have we kept in mind the fact that this deposit slot becomes "apply-able" in the event that is has already been filled to its capacity?

fleupold: Yeah, this is a good observation. At the moment the driver always waits for the blocks to pass. I'll create an issue. This is something we should probably solve in a separate fix PR, that has a unit test and the fix.

bh2smith:
I notice that one of the travis tests is still running after two days!

fleupold:
> I notice that one of the travis tests is still running after two days!

 the tests were resubmitted when I changed the base to current master 10 minutes ago. The ones from 2 days ago passed after 11 minutes (https://travis-ci.org/gnosis/dex-services/builds/505295888)

Will wait for @josojo to take a look as well

josojo:
@fleupold could you give your reasoning to choose mock-it? It's so that it runs under stable versions, right? 

Is it an issue to add later another mocking library, so that we don't have to write all this boilerplate?

Over the long run, it might consume quite some time to maintain the boilerplate for the testing interfaces.

driver/src/deposit_driver.rs
```diff
@@ -86,9 +87,167 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(), DriverErro
...
+        let state_hash = H256::zero();
+        let deposits = vec![create_flux_for_test(1,1), create_flux_for_test(1,2)];
+        let state = models::State {
+            stateHash: format!("{:x}", state_hash),
+            stateIndex: 1,
+            balances: vec![100; (models::TOKENS * 2) as usize],
+        };
+
+        let contract = SnappContractMock::new();
+        contract.get_current_deposit_slot.given(()).will_return(Ok(slot));
```
josojo: I think the parenthesis are not needed

fleupold:
@josojo created https://github.com/gnosis/dex-services/issues/53 to explain the choice of mocking framework

PR #49, opened 2019-03-14
[Driver] Consolidating model/hashing logic and adding more tests

fleupold:
This PR moves all hashing logic into the model library, consolidates it behind two common traits, adds unit tests for it and drastically simplifies the current hashing logic for calculating rolling hashes of pending fluxs and state balances.

Depends on https://github.com/gnosis/dex-contracts/pull/53 so the encoding can be shared for calculating root hashes and rolling hashes.


driver/src/models.rs
```diff
@@ -55,26 +22,21 @@ pub struct State {
...
-      let mut bs = [0u8; 64];
-      bs.as_mut()
-        .write_u128::<LittleEndian>(*i)?;
-      for i in 0..32 {
-        bs[i + 32] = bs[i];
-        bs[i] = hash[i];
-      }
-      let bytes: Vec<u8> = bs.to_vec();
+      let mut bs = [0u8; 32];
+      bs.as_mut().write_u128::<LittleEndian>(*i).unwrap();
```
josojo: Why do we unwrap instead of propagating the error?

fleupold: I think it's fine to panic in this case, because we know that this should never happen. We put i which is a balance (u128) into a 32 byte buffer, which should always work. 
If not, there is nothing our app can do and it should go up in flames.

PR #50, opened 2019-03-14
[Driver] Renaming State and Flux members to snake case

fleupold:
Rust naming convention suggests snake_case for its struct members. We originally chose to use CamelCase, because the event listener inputs events using CamelCase keys into the database and we couldn't find a simple way to parse the json into native rust objects if the keys didn't exactly match.

Turns out there is a simple macro for it: Using `#[serde(rename_all = "camelCase")]` will make snake case member variables work and parse from and to CamelCase objects.

🎈

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -66,3 +66,4 @@ services:
       - DB_HOST=mongo
       - DB_PORT=27017
       - DB_NAME=dfusion2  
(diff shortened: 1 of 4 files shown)


josojo:
>#[serde(rename_all = "camelCase")]
cool macro

PR #52, opened 2019-03-15
Update README.md

fleupold:
Adding general architecture overview, as well as updated usage scripts, test instruction and start a troubleshooting section

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Architecture
  </p>
 
 The *Event Listener* registers for certain EVM events via the [Gnosis Trading DB](https://github.com/gnosis/pm-trading-db).
-The dex smart contract emits these events on user interaction (deposit, withdraw, order) as well as when the saved state root hash is updated (state transitions).
+The [dex smart contract](https://github.com/gnosis/dex-contracts) emits these events on user interaction (deposit, withdraw, order) as well as when the saved state root hash is updated (state transitions).
 
 Upon receiving a relevant event from the contract, the event listener computes the implied changes to the underlying state. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,10 +2,43 @@ Intro
 =====
 This repository contains the backend logic for the dfusion exchange based on the specification, see [here](github.com/gnosis/dex-research)
 
-
-Instructions
+Architecture
 ============
+<p align="center">
+<img src="documentation/architecture.png" alt="dex-services architecture" width="500">
```
bh2smith: Is this the correct location and filename? I see `dƒusion architecture.png` as the filename in the root directory.

README.md
```diff
@@ -2,10 +2,43 @@ Intro
...
+Upon receiving a relevant event from the contract, the event listener computes the implied changes to the underlying state. 
+E.g. if a *deposit* event is received, the list of pending deposits is updated.
+Similarily, if a *deposit state transition* event is received it updates the account balances based on the pending deposits that were included in the state transition.
+
+The *Driver* watches state updates to the database and reads relevant data from the smart contract to decide when a state transition can be applied.
+There are four types of state transitions:
+
+- apply deposit
+- apply withdraws
+- find solution for optimization problem
```
bh2smith: Is finding a solution also considered a state transition?

README.md
fleupold: It's because the PR is currently rebased on an old version of master. After the merge it will look fine.

README.md
fleupold: I think so. Since price suggestion (find solution) and solution submission (trade execution) are separate interactions with the smart contract, we need separate handlers in the driver.

Issue #59, opened 2019-03-26
Solver Integration


bh2smith:
- Convert orders into JSON file format as solver input data
- Invoke solver as command line operation `solve --inFile orders.json --outFile out.json`
- Read solver results from `out.json`

All of this can be done with the `SCIP` solver configuration which is free to use.

bh2smith:
@twalth3r  could you potentially include a more accurate account of what needs to be called from the command line to invoke the solver?

Issue #58, opened 2019-03-26
Simple Solver


bh2smith:
At minimum, this simple solver should find at least 1 direct match (i.e. any non-trivial solution to the orders provided). 

This will be used for the purpose of testing (with small examples).

one suggested solution here is to loop over each token pair and find two orders `i for j` and `j for i` with prices that agree. Then set the median of these limit prices as the price between `i` and `j` and continue for other pairs... or stop at this point.

fleupold:
A slightly more sophisticated version, could keep track of a price range for previous matches. As it finds another match for the same trading pair, it checks if their supported price range intersects with the one that was previously found.

But let's start simple and see how far we get.

Issue #57, opened 2019-03-26
Solver Abstraction


bh2smith:
- Solver interface: `fn solve(orders, previous_prices) -> Solution`
- Mock implementation

PR #66, opened 2019-03-28
[Driver] Defining an interface for the driver<->solver interaction

fleupold:
This PR defines a rust interface that the order_processor will interact with to invoke the solving engine. The solving engine can be implemented in different ways
1. Mocked (for unit tests)
2. Simple Solver (#58)
3. Tom's sophisticated solve (#59)

Let's discuss the interface and data structures here. I included the mock implementation so that we can start using this interface.

Addresses #57

### DIFF
--- a/driver/src/price_finding/price_finder_interface.rs
+++ b/driver/src/price_finding/price_finder_interface.rs
@@ -8,6 +8,7 @@ pub struct Solution {
     pub surplus: U256,
     pub prices: Vec<u128>,
     pub executed_sell_amounts: Vec<u128>,
+    pub executed_buy_amounts: Vec<u128>,
 }
 
 pub trait PriceFinding {
@@ -40,11 +41,11 @@ pub mod tests {
 
     impl PriceFinding for PriceFindingMock {
         fn find_prices(
-        &self, 
-        orders: Vec<models::Order>, 
-        state: models::State
-    ) -> Result<Solution, DriverError> {
-        self.find_prices.called((orders, state))
-    }
+            &self, 
+            orders: Vec<models::Order>, 
+            state: models::State
+        ) -> Result<Solution, DriverError> {
+            self.find_prices.called((orders, state))
+        }
     }
 }
--- a/driver/src/price_finding/price_finder_interface.rs
+++ b/driver/src/price_finding/price_finder_interface.rs
@@ -1,0 +1,51 @@
+use crate::models;
+
+use crate::error::DriverError;
+use web3::types::U256;
+
+#[derive(Clone)]
+pub struct Solution {
+    pub surplus: U256,
+    pub prices: Vec<u128>,
+    pub executed_sell_amounts: Vec<u128>,
+    pub executed_buy_amounts: Vec<u128>,
+}
+
+pub trait PriceFinding {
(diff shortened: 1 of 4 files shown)


fleupold:
@bh2smith no that's a good comment. I should name it the same way the smart contract does it.

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use crate::error::DriverError;
+use web3::types::U256;
+
+#[derive(Clone)]
+pub struct Solution {
+    pub surplus: U256,
+    pub prices: Vec<u128>,
+    pub executed_sell_amounts: Vec<u128>,
+    pub executed_buy_amounts: Vec<u128>,
```
josojo: WHy do we need both buy and sells here?

fleupold: It's part of Tom's solution, but you are right we don't strictly need it. I can remove it.

bh2smith: We do need buy and sell in our SellOrder collection on the contract. Are these related?

Notice that `maxSellAmount` and `minBuyAmount` are used to interpret the `limitPrice` of a sell order. @twalth3r can back this up if need be.

fleupold: This is just for the solutions. The smart contract will get the prices and one side of the vector. @twalth3r were these sufficient to recover the exact balances, even when we have rounding or do we need to push the epsilons as well?

fleupold:
The open question here is what is the minimal amount of data of the solutions that we have to push on chain, for an external observer can recover the exact balances.

Assuming 30 Tokens, 1k Orders.

bh2smith:
Another good (open) question here is whether or not the solution information even need to be stored in the contract or simply recovered from internal transaction data.... @fleupold

fleupold:
@bh2smith the contract only has to store a commitment (sha hash of all calldata) that will eventually be passed in to the Snark in case of dispute.

josojo:
> The open question here is what is the minimal amount of solution data that we have to push on chain, so that an external observer can recover the exact balances.
> 
> Assuming 30 Tokens, 1k Orders.

Yeah, let's not optimize it for these numbers. Hopefully, 1k orders will change soon. I think for now it is good enough if just publish all traded volumes for all orders. 
Or do you have new insights about this? @twalth3r

fleupold:
> I think for now it is good enough if just publish all traded volumes for all orders.

That means I will put the buy Volume back into the response, ok?

josojo:
> > I think for now it is good enough if just publish all traded volumes for all orders.
> 
> That means I will put the buy Volume back into the response, ok?

Are you talking about this line?     pub executed_buy_amounts: Vec<u128>

I think it is enough if we have only     pub executed_sell_amounts: Vec<u128>, if we know all sell volumes, we can determine the unique solution, right?
Sorry, probably, I am confused here.

twalth3r:
As I see it, there is no 'unique' solution, so I'd prefer having both buy- and sell-amounts as part of the solution (or at least sell-amount and the delta).

Issue #61, opened 2019-03-26
Improve Docker Configuration


bh2smith:
- Better use of _onion principle_ 
- Remove images and run natively on Travis
- Don't run SQL interface when running tests.

Task to be completed as a team pair programming event mid April.

fleupold:
Spent 2 day so far. Build time wen from close to 20 minutes to 12. 2 more minutes can likely be saved by using base images that have all necessart apt-get packages already installed.

bh2smith:
Can we close this?

fleupold:
I think it's worth doing the remaining optimizations (optimizing base images) if we get the chance.

PR #70, opened 2019-03-29
[Easy] Don't launch unnecessary docker images on travis

fleupold:
As title.

Addresses one of the points suggested in #61


fleupold:
> I guess we don't need the database for integration testing?

We do. But dependent images are started automatically.

PR #71, opened 2019-03-29
[Travis] Better caching

fleupold:
This PR is trying to improve our build time by creating a higher cache hit ratio on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,8 @@ matrix:
       script:
         - pytest --cov=event_listener/dfusion_db/ --cov-report term-missing
         - mypy event_listener/dfusion_db/ --ignore-missing-imports --strict
-    - dist: xenial
+    - language: generic
+      dist: xenial
       cache: 
         npm: true
         directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
       script:
         - pytest --cov=event_listener/dfusion_db/ --cov-report term-missing
         - mypy event_listener/dfusion_db/ --ignore-missing-imports --strict
-    - language: node_js
+    - language: generic
       cache: 
         npm: true
         directories:
@@ -27,19 +27,7 @@ matrix:
           - dex-contracts/node_modules/
       sudo: required
       node_js: 10.13.0
-      env:
-        - DOCKER_COMPOSE_VERSION=1.22.0
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
-          packages:
-            - build-essential
       before_install:
-        - sudo rm /usr/local/bin/docker-compose
-        - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
-        - chmod +x docker-compose
-        - sudo mv docker-compose /usr/local/bin
         - npm install -g truffle
       install:
         - git submodule update --init
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ matrix:
         - mypy event_listener/dfusion_db/ --ignore-missing-imports --strict
     - language: generic
       dist: xenial
+      sudo: required
       cache: 
         npm: true
         directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,6 @@ matrix:
       install:
         - git submodule update --init
         - npm install -C dex-contracts
-        - ganache-cli -p 8545 > /dev/null &
         - cd dex-contracts && truffle compile && cd - # driver/listener require compiled contracts
         - docker-compose up -d driver listener truffle | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,8 @@ matrix:
         - npm install -C dex-contracts
         # driver/listener require compiled contracts
         - cd dex-contracts && truffle compile && cd -
-        - docker-compose up -d driver listener truffle | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
+        - docker-compose up -d driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
       script:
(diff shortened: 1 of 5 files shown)


fleupold:
rust unit tests went from >5min to 2:30min

fleupold:
Not much to do on the python build. Went from ~90s to ~60s

fleupold:
Unfortunately the big pain-point (the e2e test) isn't that easy to improve. We are still seeing build times >10 minutes. The issue is that we are creating a bunch of micro-services with `docker-compose` and test their integrations. 

From the [Travis wiki](https://docs.travis-ci.com/user/caching/#things-not-to-cache):
>Docker images are not cached, because we provision a brand new virtual machine for every build.

I tried force caching the images. It looks like `docker-compose` doesn't support `--cache-from` (as we do for normal docker [here](https://github.com/gnosis/dex-zksnarks/blob/master/.travis.yml#L10)).
I found a way to manually store the images `docker–compose` creates in the travis build cache (see https://stackoverflow.com/a/51748544). This leads to a ~1.5GB large cache folder inside travis and downloading this takes almost the same time as building from scratch.

A different approach for getting faster build times would be to run all services in parallel natively on travis (e.g. using the [`generic`](https://docs.travis-ci.com/user/languages/minimal-and-generic/#generic) image). This would allow us to only cache the build artefacts. 
However, we'd lose the guarantee that our local/production `docker-compose` setup should work fine as long as travis is green (since there might be problems with inter-host communications that would be no longer tested).

A third strategy would be to copy the built artefacts from the docker images back into the travis host system, cache them and populate them back as port of `docker-compose build`.

If any of our dev-ops Gurus @denisgranha @giacomolicari @Uxio0 @luarx have an idea what we can do here, I'd much appreciate that.

fleupold:
Here is a cached run:
https://travis-ci.org/gnosis/dex-services/jobs/513129320

And here a non-cached run:
https://travis-ci.org/gnosis/dex-services/jobs/513104296

Uxio0:
Hi @fleupold. I see some points to improve the build:
- Use a different `docker-compose.yml` for travis. For example, you don't need `mongoclient` in travis

Uxio0:
Hi @fleupold. I see some points to improve the build:
- Use a different `docker-compose.yml` for travis. For example, you don't need `mongoclient` in travis.
- As you already have `npm` and install all the dependencies, I don't see any issue in using local `truffle` and `ganache` instead the ones from `docker`.
- I understand that docker-compose is the best way to test everything and have the same environment, but from my point of view is not very convenient. I mean, you are already installing the `python` libraries in travis, you compile the `rust` project and you install `node_js` and the libraries in a clean `docker` image (the base travis instance). I would remove the entire `docker-compose` flow for normal `travis` builds. You may test it just in merges to `master branch` or `pull requests`, for example. If it's critical to test `docker-compose` images in every commit you can just set up a `git pre-hook` to run it locally before any push. We are doing it this way for every backend project :smiley:

bh2smith:
@fleupold - feel free to merge! We can also test our automated kanban board with this. I already saw that it move into the "review approved" column once it was approved.

fleupold:
Thanks @Uxio0 for your suggestions. We have already excluded the mongo-client and truffle image from the travis builds (cf. https://github.com/gnosis/dex-services/blob/master/.travis.yml#L39)

As for running ganache locally, I think this would only work if we ran the driver and listener components also natively (I believe it's not trivial for a container to access ports opened by host system). Moreover the ganache image is pretty small (28MB) and doesn't require any customisation.

I think we prefer keeping the docker-compose setup for now (it's really convenient to know that we can run the demo immediately). You are right, that we are already compiling everything for the unit tests on the host system, but these jobs are completely unrelated and can run in parallel. Moreover, we reduced the travis runs to only run on PRs and master.

I was able to reduce the setup time for the docker images by a few minutes and think we can even get a further reduction if we start basing them off optimized base images that have all the required tools already installed (at the moment we do a bunch of `apk-get` installs, which account for > 2 minutes of build time in total). What would be the best way to publish these images as part of the Gnosis dockerhub organisation?

Running the tests as part of git pre-commit hook is interesting and could and could guarantee that we always have a clean master. However, pre-commit hooks don't run on auto-merges or other changes that might be performed in github rather than locally. Also, it would be quite annoying to always have to synchronously wait ~5 minutes when running git commit (especially when people work with a lot of small commits). Travis runs asynchronously and is thus more convenient. How expensive is it to run these checks on travis?

Also, @bh2smith @josojo what are your thoughts?

Uxio0:
> Thanks @Uxio0 for your suggestions. We have already excluded the mongo-client and truffle image from the travis builds (cf. https://github.com/gnosis/dex-services/blob/master/.travis.yml#L39)
Nice, I didn't realise!
> As for running ganache locally, I think this would only work if we ran the driver and listener components also natively (I believe it's not trivial for a container to access ports opened by host system). Moreover the ganache image is pretty small (28MB) and doesn't require any customisation.
It's not very difficult to expose it (just configure the docker interface ip as the ganache server), but I didn't know if was so small now. All the node derived images used to be very heavy
> I think we prefer keeping the docker-compose setup for now (it's really convenient to know that we can run the demo immediately). You are right, that we are already compiling everything for the unit tests on the host system, but these jobs are completely unrelated and can run in parallel. Moreover, we reduced the travis runs to only run on PRs and master.
👍 
> I was able to reduce the setup time for the docker images by a few minutes and think we can even get a further reduction if we start basing them off optimized base images that have all the required tools already installed (at the moment we do a bunch of `apk-get` installs, which account for > 2 minutes of build time in total). What would be the best way to publish these images as part of the Gnosis dockerhub organisation?
Nice. I think you can just create one repo for every microservice you have, as we do (e.g https://hub.docker.com/r/gnosispm/safe-relay-service)
> Running the tests as part of git pre-commit hook is interesting and could and could guarantee that we always have a clean master. However, pre-commit hooks don't run on auto-merges or other changes that might be performed in github rather than locally. Also, it would be quite annoying to always have to synchronously wait ~5 minutes when running git commit (especially when people work with a lot of small commits). Travis runs asynchronously and is thus more convenient. How expensive is it to run these checks on travis?
I was meaning `pre-push` hook, `pre-commit` is a little overkill. As far as I know they don't have any cost, just time 😄

Congrats on your really good work

Issue #69, opened 2019-03-29
[e2e-test] Order Collection & Auction Settlement


bh2smith:
Should introduce a bash script in `dex-services/test/e2e-tests-auction.sh` That tests a full circle interaction from on-chain order collection, price-finding and auction settlement.

A good example to use could be the 6 orders between 3 tokens with EUR, OIL and WINE.

PR #74, opened 2019-04-02
E2E Test Order Collection Event Listener

bh2smith:
Regarding issue #69

- Update contract code (to include sell-order.js script)
- Include 6 orders from workshop as example data for testing

## TODO

- Wait for state transition (from auction settlement)
- Check and verify updated balances (or feasibility of results)

### DIFF
--- a/test/e2e-tests-auction.sh
+++ b/test/e2e-tests-auction.sh
@@ -3,6 +3,10 @@ set -e
 
 cd dex-contracts/
 
+EXPECTED_AUCTION=1
+# Ensure no orders yet in auction slot 1
+mongo dfusion2 --eval "db.orders.find({'auctionId': ${EXPECTED_AUCTION}}).size()" | grep 0
+
 # Place 6 orders in current Auction (accountId, buyToken, sellToken, minBuy, maxSell)
 truffle exec scripts/sell_order.js 1 2 3 12 12
 truffle exec scripts/sell_order.js 2 3 2 22 20
@@ -11,19 +15,14 @@ truffle exec scripts/sell_order.js 4 1 2 180 15
 truffle exec scripts/sell_order.js 5 2 1 4 52
 truffle exec scripts/sell_order.js 6 3 1 20 280
 
-truffle exec scripts/mine_blocks.js 21
 sleep 5
 
-# Test Listener (order put into Database) - Query mongodb
-# mongo --eval "db.dfusion.orders.find({'slot': 1})"
-# TODO
+# Test Listener: There are now 6 orders in auction slot 1 and sellAmount for accountId = 6 is 280
+mongo dfusion2 --eval "db.orders.find({'auctionId': ${EXPECTED_AUCTION}}).size()" | grep 6
+mongo dfusion2 --eval "db.orders.findOne({'auctionId': ${EXPECTED_AUCTION}, 'accountId': 6}).sellAmount" | grep 280
 
+truffle exec scripts/mine_blocks.js 21
 
-# Wait for state update because of closed auction
+# TODO - Test Driver: Wait for state update because of closed auction
 
-# TODO
-
-# Check that Auction Settlement is something different
-# In this case the simple solver would match the two orders with agreeing limit prices between two tokens.
-
-# TODO
+# TODO - Test Solver: Ensure that balances are differ after state transition (i.e. found non-trivial solution)
--- a/test/e2e-tests-auction.sh
+++ b/test/e2e-tests-auction.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -e
+
+cd dex-contracts/
+
+# Place 6 orders in current Auction (accountId, buyToken, sellToken, minBuy, maxSell)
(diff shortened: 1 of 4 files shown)


test/e2e-tests-auction.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -e
+
+cd dex-contracts/
+
+# Place 6 orders in current Auction (accountId, buyToken, sellToken, minBuy, maxSell)
+truffle exec scripts/sell_order.js 1 2 3 12 12
```
fleupold: Would it make sense to use a different accountId for some of these trades?

test/e2e-tests-auction.sh
```diff
@@ -0,0 +1,25 @@
...
+truffle exec scripts/sell_order.js 1 3 2 22 20
+truffle exec scripts/sell_order.js 1 1 3 15 10
+truffle exec scripts/sell_order.js 1 1 2 180 15
+truffle exec scripts/sell_order.js 1 2 1 4 52
+truffle exec scripts/sell_order.js 1 3 1 20 280
+
```
fleupold: Can you check here that the orders made it into MongoDB?

fleupold:
I think it makes sense to land this test soon and then add steps as we progress.

test/e2e-tests-auction.sh
bh2smith: Do you mean include some bash command that ensures the orders are in the database?

test/e2e-tests-auction.sh
bh2smith: It would make sense especially for checking updated balances

test/e2e-tests-auction.sh
bh2smith: Is it currently the case that deposit and withdraw tests are verifying this DB inclusion?

fleupold: Withdraws check the balance in the end. It would probably make sense to add intermediate checks for withdraw requests there as well. As mentioned, this will make finding the point of failure much simpler.

For this case I think it's particularly important, because we might accidentally break the listener functionality without noticing, since we don't have the full e2e functionality yet. By checking the database, we could make sure it's doing what it's supposed to do on every PR.

bh2smith:
Order Collection is now tested on the Event Listener side, by querying the DB and ensuring 6 orders are there and that the sell amount of a particular one matches.

test/e2e-tests-auction.sh
```diff
@@ -0,0 +1,28 @@
...
+# Place 6 orders in current Auction (accountId, buyToken, sellToken, minBuy, maxSell)
+truffle exec scripts/sell_order.js 1 2 3 12 12
+truffle exec scripts/sell_order.js 2 3 2 22 20
+truffle exec scripts/sell_order.js 3 1 3 15 10
+truffle exec scripts/sell_order.js 4 1 2 180 15
+truffle exec scripts/sell_order.js 5 2 1 4 52
+truffle exec scripts/sell_order.js 6 3 1 20 280
+
+sleep 5
```
fleupold: Would love to get rid of these hard coded waits as they are really bad practice (they can be flaky and usually cause the test to take longer than necessary). 

Would appreciate any ideas on how we can know once the driver is done processing all events in the queue.

test/e2e-tests-withdraw.sh
```diff
@@ -10,7 +10,7 @@ sleep 5
 truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot'
 
 EXPECTED_HASH="77b01abfbad57cb7a1344b12709603ea3b9ad803ef5ea09814ca212748f54733"
-truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot' | grep $EXPECTED_HASH
+truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot' | grep ${EXPECTED_HASH}
```
fleupold: What’s the difference between the two?

bh2smith: One is a lint warning, the other is not.

test/e2e-tests-auction.sh
bh2smith: Done!

test/e2e-tests-auction.sh
bh2smith: Yes I would like that too!

fleupold:
@bh2smith can you add this script to .travis.yml and then merge?

fleupold:
I think this should run in between the deposit and the withdraw, since there are no fund available otherwise. Or we need to do a few more deposits as part of this script.

bh2smith:
@fleupold 

> I think this should run in between the deposit and the withdraw, since there are no fund available otherwise.

At this point (i.e. just to test that Orders are received), I don't think that this matters (since the contract does not know the balances anyway). This is collecting orders perfectly fine for the purpose of this PR. I have, nonetheless, switched the order of `auction.sh` and `withdraw.sh` in the Travis build for the time being. We will open new issues for further development of e2e-auction testing.

PR #75, opened 2019-04-02
[Driver] PriceFinder that uses liner optimization solver

fleupold:
This PR adds an implementation of the price finding interface that calls into the linear optimization solver @twalth3r implemented in Python.

The component is doing the following steps:
1) Convert Orders/Balances into the json format the solver expects
2) Write json to disk
3) Invoke solver passing the json file that was written
4) Reading the result file into json
5) Converting the result file into solution struct (and storing prices in memory for next run)

I added a custom error class for PriceFinding, which should help debugging in the future (since we will be able to distinguish other error from solver related errors). Once used, that error should probably be wrapped into a DriverError.

I'm planning to add unit tests for the serialization/deserialization methods, but would suggest doing that in the following PR, since this one is already quite big.

To test for now, I added the following code inside the main method:

```rust
    let price_finder = LinearOptimisationPriceFinder::new();
    let orders = vec![Order {
        account_id: 0,
        sell_token: 0,
        buy_token: 1,
        sell_amount: 1000000000000000000,
        buy_amount: 1000000000000000000
    }, Order {
        account_id: 0,
        sell_token: 1,
        buy_token: 0,
        sell_amount: 1000000000000000000,
        buy_amount: 1000000000000000000
    }];
    let state = State {
        state_hash: "".to_string(),
        state_index: 0,
        balances: vec![1000000000000000000; 60],
    };
    price_finder.find_prices(&orders, &state).expect("Price Finding Failed");
```

And see that the result is computed correctly

### DIFF
--- a/driver/src/price_finding/linear_optimization_price_finder.rs
+++ b/driver/src/price_finding/linear_optimization_price_finder.rs
@@ -1,0 +1,170 @@
+use crate::models;
+use crate::price_finding::error::{PriceFindingError, ErrorKind};
+use crate::price_finding::price_finder_interface::{PriceFinding, Solution};
+
+use serde_json::json;
+use web3::types::U256;
+use std::collections::HashMap;
+use chrono::{Utc};
+use std::fs::File;
+use std::process::Command;
+use std::io::BufReader;
+
+const RESULT_FOLDER: &str = "./results/tmp/";
+type Prices = HashMap<String, String>;
+
+pub struct LinearOptimisationPriceFinder {
+    previous_prices: Prices
+}
+
+impl<'a> LinearOptimisationPriceFinder {
+    pub fn new() -> Self {
+        // All prices are 1 (10**18)
+        return LinearOptimisationPriceFinder {
+            previous_prices: (0..models::TOKENS)
+                .map(|t| (token_id(t), "1000000000000000000".to_string()))
+                .collect()
+        }
+    }
+}
+
+fn token_id(token: u8) -> String {
+    format!("token{}", token)
+}
+
+fn serialize_order(order: &models::Order, id: &str) -> serde_json::Value {
+    json!({
+        "accountID": order.account_id.to_string(),
+        "sellToken": token_id(order.sell_token),
+        "buyToken": token_id(order.buy_token), 
+        "sellAmount": order.sell_amount.to_string(),
+        "buyAmount": order.buy_amount.to_string(),
+        "ID": id //TODO this should not be needed
+    })
+}
+
+fn serialize_balances(balances: &Vec<u128>) -> serde_json::Value {
+    let mut accounts: HashMap<String, HashMap<String, String>> = HashMap::new();
+    let mut current_account = 0;
+    for balances_for_current_account in balances.chunks(models::TOKENS as usize) {
+        accounts.insert(current_account.to_string(), (0..models::TOKENS)
+            .map(|t| token_id(t))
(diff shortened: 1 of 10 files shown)


driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+fn serialize_order(order: &models::Order, id: &str) -> serde_json::Value {
+    json!({
+        "accountID": order.account_id.to_string(),
+        "sellToken": token_id(order.sell_token),
+        "buyToken": token_id(order.buy_token), 
+        "sellAmount": order.sell_amount.to_string(),
+        "buyAmount": order.buy_amount.to_string(),
+        "ID": id //TODO this should not be needed
```
bh2smith: I believe you can use the `order.slot_index` for the orderID. Furthermore, I am pretty sure @twalth3r's solver expects an orderID of some sort anyway...

fleupold: Technically it should not require an ID at least from a top level (otherwise orders should be Map<ID => Order> and not a List). Since we expose everything as lists on the smart contract level it should probably stay like that. Discussed this with @twalth3r and if the solver requires IDs for internal representation it will generate them internally going forward.

Also, at this point we are not exposing order.slot_index on the driver level.

bh2smith: Cool with me, but I do think that there is some aspect of "trader's surplus" which involves the time at in which the orders were placed.

### Example

Consider two identical orders (placed at different times) that are about to be fractionally fulfilled. Does the solution prefer to fully fill the earlier order first then the later with remaining surplus, does it fractionally fill them both equally or does it fully fill the later first and then use the remainder for the earlier? 

- Which is most fair? 
- Which maximizes the surplus?
- Is is biased to use the order index as part of the solution?

fleupold: As far as our problem is specified atm the algorithm can chose either. If order mattered, this would be another strong argument for keeping a list rather than doing a mapping and thus reducing the need for an ID altogether.

bh2smith: I am not sure where the **list** versus **mapping** keeps creeping in. Are you referring to the way it is stored in the database?

fleupold: No. 

There is no point in having an ID when we pass around a list as elements in a list can be uniquely identified by their index in the list. We’d only need an ID if we stored an unordered data structure.

bh2smith: And where are these being stored as a list? Not in the contract, not the database... I just don't see any lists here.

fleupold: We are not storing orders in the smart contract at all. However we store the rolling hash of the **list** of orders there. 
Also, when posting a solution, we submit a volume vector, which is expected to line up with the list of orders.

fleupold:
> I have not yet understood why LineraOptimisationPriceFinder needs a lifetime parameter.

You are right, it doesn't. Good catch. This is a leftover from when I stored the price map with a reference. 

Basically, you need struct lifetimes whenever there is a field on the struct that's a reference (&). Otherwise we wouldn't be guaranteed, that that reference is still "alive" when the struct is being passed around and that field is accessed.

PR #76, opened 2019-04-02
[Easy] Typos, Indentation, Rust warnings etc

bh2smith:
Im just getting comfortable with the rust code so that I can start writing the simple solver and e2e-test for running an auction.

There were a few warnings and errors that my IDE brought up while I was reading through. The lest obvious one was the implementation ordering of the SnappContract trait (which warned should be in the same order as declared). 

Another place where my IDE provides a very strict warning is here is in `driver/src/contract.rs`

https://github.com/gnosis/dex-services/blob/b43e7dae6c48e4f3c9df115ca506257b41c32ccc/driver/src/contract.rs#L165-L166

Where it says "cannot assign field of immutable binding". The IDE suggested to make the opt mutable, so I have... Please let me know if this is alright with ya'll.

Screenshot 2019-04-02 at 2 39 48 PM

### DIFF
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -161,7 +161,7 @@ impl SnappContract for SnappContractImpl {
                 "applyWithdrawals",
                 (slot, merkle_root, prev_state, new_state, withdraw_hash),
                 account,    
-                Options::with(|opt| { // usual gas estimate is not working
+                Options::with(|mut opt| { // usual gas estimate is not working
                     opt.gas_price = Some(25.into());
                     opt.gas = Some(1_000_000.into());
                 }),
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -70,6 +70,13 @@ impl SnappContractImpl {
 }
 
 impl SnappContract for SnappContractImpl {
+    fn get_current_block_number(&self) -> Result<U256> {
+        self.web3.eth()
+            .block_number()
+            .wait()
+            .map_err(|e| DriverError::from(e))
+    }
+
     fn get_current_state_root(&self) -> Result<H256> {
         self.contract.query(
             "getCurrentStateRoot", (), None, Options::default(), None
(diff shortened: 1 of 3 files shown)


driver/src/contract.rs
```diff
@@ -161,10 +161,10 @@ impl SnappContract for SnappContractImpl {
                 "applyWithdrawals",
                 (slot, merkle_root, prev_state, new_state, withdraw_hash),
                 account,    
-                Options::with(|opt| { // usual gas estimate is not working
-            opt.gas_price = Some(25.into());
-            opt.gas = Some(1_000_000.into());
-        }),
+                Options::with(|mut opt| { // usual gas estimate is not working
```
josojo: Hm, not sure what is happening under the hood. Maybe your IDE sees the next line as changes of the object opt. But in the end it is only initialization and this is why the compiler is not asking for it.

I am only guessing here. I am curious on Felixs answer

driver/src/contract.rs
```diff
@@ -70,6 +70,13 @@ impl SnappContractImpl {
 }
 
 impl SnappContract for SnappContractImpl {
+    fn get_current_block_number(&self) -> Result<U256> {
```
fleupold: Do you have any idea on what basis this ordering is happening?

driver/src/contract.rs
fleupold: We are mutating opt, so it should be market as mut. I'm surprised it compiled before. I think technically, it should even be `&mut` (https://tomusdrw.github.io/rust-web3/web3/contract/struct.Options.html#method.with)

driver/src/contract.rs
bh2smith: I am using `intellij-rust`

I am also not able to find this trait ordering easily, but will keep looking.

driver/src/contract.rs
bh2smith: I will change it to &mut

bh2smith:
I found a couple other strange things here and wonder about making some fixes.


Screenshot 2019-04-02 at 3 33 13 PM

Especially a bunch of mismatched types in db_interface

Screenshot 2019-04-02 at 3 33 42 PM

Screenshot 2019-04-02 at 3 40 09 PM

driver/src/contract.rs
bh2smith: @fleupold &mut gave errors! I will not change it.

fleupold:
It looks like your intelliJ is too critical. These are definitely not errors, as we can see in our continuous integration the current version compiles and runs just fine. Maybe you can investigate a bit more what's going on there, and if you might have a too strict setting.

PR #77, opened 2019-04-03
[Driver] LinearOptimisationPriceFinder unit tests

fleupold:
Unit tests for #75

### DIFF
--- a/driver/src/price_finding/linear_optimization_price_finder.rs
+++ b/driver/src/price_finding/linear_optimization_price_finder.rs
@@ -14,16 +14,25 @@ const RESULT_FOLDER: &str = "./results/tmp/";
 type Prices = HashMap<String, String>;
 
 pub struct LinearOptimisationPriceFinder {
-    previous_prices: Prices
+    num_tokens: u8,
+    previous_prices: Prices,
+    // default IO methods can be replaced for unit testing
+    write_input: fn(&str, &serde_json::Value) -> std::io::Result<()>,
+    run_solver: fn(&str) -> Result<(), PriceFindingError>,
+    read_output: fn() -> std::io::Result<serde_json::Value>,
 }
 
 impl LinearOptimisationPriceFinder {
     pub fn new() -> Self {
         // All prices are 1 (10**18)
         return LinearOptimisationPriceFinder {
+            num_tokens: models::TOKENS,
             previous_prices: (0..models::TOKENS)
                 .map(|t| (token_id(t), "1000000000000000000".to_string()))
-                .collect()
+                .collect(),
+            write_input,
+            run_solver,
+            read_output,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -14,16 +14,25 @@ const RESULT_FOLDER: &str = "./results/tmp/";
...
+    previous_prices: Prices,
+    // default IO methods can be replaced for unit testing
+    write_input: fn(&str, &serde_json::Value) -> std::io::Result<()>,
+    run_solver: fn(&str) -> Result<(), PriceFindingError>,
+    read_output: fn() -> std::io::Result<serde_json::Value>,
 }
 
 impl LinearOptimisationPriceFinder {
     pub fn new() -> Self {
         // All prices are 1 (10**18)
```
bh2smith: Does this mean 1 = 10**18?

fleupold:
🏓 @bh2smith @josojo can you review this please?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -168,3 +178,275 @@ fn read_output() -> std::io::Result<serde_json::Value> {
...
+#[cfg(test)]
+pub mod tests {
+
+    use super::*;
+    use std::error::Error;
+
+    #[test]
+    fn test_solver_keeps_prices_from_previous_result() {
+        let state = models::State {
+            state_hash: "hash".to_string(),
```
bh2smith: is this really supposed to be a string with the word hash in it? Or was this a place holder?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -43,11 +52,12 @@ fn serialize_order(order: &models::Order, id: &str) -> serde_json::Value {
     })
 }
 
-fn serialize_balances(balances: &Vec<u128>) -> serde_json::Value {
+fn serialize_balances(balances: &Vec<u128>, num_tokens: u8) -> serde_json::Value {
```
bh2smith: Shouldn't this function also have something to do with `num_accounts`?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -168,3 +178,275 @@ fn read_output() -> std::io::Result<serde_json::Value> {
...
+            "orders": [
+                {
+                    "execSellAmount": "0",
+                    "execBuyAmount": "0",
+                    "execSurplus": "0"
+                },
+                {
+                    "execSellAmount": "318390084925498118944",
+                    "execBuyAmount": "95042777139162480000",
+                    "execSurplus": "15854632034944469292777429010439194350"
```
bh2smith: Do all of these long numbers have any significance? They all appear to be pretty random. Would is be less distracting to use numbers like `100...00`, `200...00`?

driver/src/price_finding/linear_optimization_price_finder.rs
fleupold: It's a unit test, so yeah it's just a dummy string.

driver/src/price_finding/linear_optimization_price_finder.rs
fleupold: We don't actually store num_accounts in our system yet. It's assumed that everything we get from the database is a multiple of `TOKENS`, that multiple being num_accounts.

driver/src/price_finding/linear_optimization_price_finder.rs
fleupold: Yes, good point.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: I believe num_accounts is stored as part of the [SnappInitialization event](https://github.com/gnosis/dex-contracts/blob/4df0570affbe7de86207f790a3a2f8c7f1ccf07b/contracts/SnappBase.sol#L56)

bh2smith:
@josojo 

> Pretty exhausting testing. Great!

This is a quote for the books! I assume you mean "exhaustive", but I agree it was also exhausting!

driver/src/price_finding/linear_optimization_price_finder.rs
josojo: Yeah, currently, we are not using these values from the db at all. Also models::Tokens is also just a constant. I would be in favour of making a seperate issue for this and merging this PR

josojo: #86

bh2smith: Cool beans!

bh2smith:
Who Wants to merge!?

PR #79, opened 2019-04-04
[Driver] some more refactoring needed for the order processor

fleupold:
This PR refactors some things around how we serialize/hash our basic models (to submit it to the blockchain). The goal is to make it easy to reuse the same business logic for PendingFluxs as well as for Orders.

It also factor some utility functions that all driver components need (to get the currently highest slot and check whether it can be processed) into their own module and abstracts the contract specific calls away so they can be called with a small closure (providing the exact method call, e.g. order vs. withdraw vs. deposit)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,8 @@ matrix:
         - npm install -C dex-contracts
         # driver/listener require compiled contracts
         - cd dex-contracts && truffle compile && cd -
-        - docker-compose up -d driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
+        - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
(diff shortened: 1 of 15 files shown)


.travis.yml
```diff
@@ -34,7 +34,8 @@ matrix:
         - npm install -C dex-contracts
         # driver/listener require compiled contracts
         - cd dex-contracts && truffle compile && cd -
-        - docker-compose up -d driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
+        - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
```
bh2smith: Hasn't this change already been pushed to Master?

josojo: Yeah...
also I only made this commit to get travis started again, but travis is a bitch today

Issue #60, opened 2019-03-26
Run Order Driver


bh2smith:
1. Monitor the database and smart contract for the appropriate time to gather orders and invoke solver.
2. Publish Solution (a.k.a. State Transition) to the smart contract.

fleupold:
Done

PR #80, opened 2019-04-04
[Driver] Order Processing

fleupold:
This is a work in progress. It creates the main driver loop for checking if we have pending orders and processing them if appropriate.

There are still a bunch of TODOs and it lacks testing, but the general structure is there.
Task description: Issue #60

### DIFF
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -70,6 +70,13 @@ impl SnappContractImpl {
 }
 
 impl SnappContract for SnappContractImpl {
+    fn get_current_block_number(&self) -> Result<U256> {
+        self.web3.eth()
+            .block_number()
+            .wait()
+            .map_err(|e| DriverError::from(e))
+    }
+
     fn get_current_state_root(&self) -> Result<H256> {
         self.contract.query(
             "getCurrentStateRoot", (), None, Options::default(), None
@@ -82,15 +89,9 @@ impl SnappContract for SnappContractImpl {
         ).wait().map_err(|e| DriverError::from(e))
     }
 
-    fn has_deposit_slot_been_applied(&self, slot: U256) -> Result<bool> {
+    fn get_current_withdraw_slot(&self) -> Result<U256> {
         self.contract.query(
-            "hasDepositBeenApplied", slot, None, Options::default(), None,
-        ).wait().map_err(|e| DriverError::from(e))
-    }
-
-    fn deposit_hash_for_slot(&self, slot: U256) -> Result<H256> {
-        self.contract.query(
-            "getDepositHash", slot, None, Options::default(), None,
+            "withdrawIndex", (), None, Options::default(), None
         ).wait().map_err(|e| DriverError::from(e))
     }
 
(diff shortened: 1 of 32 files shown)


driver/src/order_driver.rs
```diff
@@ -0,0 +1,78 @@
...
+    let auction_slot = contract.get_current_auction_slot()?;
+
+    println!("Current top auction slot is {:?}", auction_slot);
+    let slot = util::find_first_unapplied_slot(
+        auction_slot + 1, 
+        Box::new(&|i| contract.has_auction_slot_been_applied(i))
+    )?;
+    if slot <= auction_slot {
+        println!("Highest unprocessed auction slot is {:?}", slot);
+        if util::can_process(slot, contract,
```
josojo: here, we might want to extend the  function can_process with an additional parametric function: slot_is_full: slot_nr => contract.batch_already_full()
which would check for deposits the nr of deposits and for orders the nr of orders and if they already reached the threshold, it can return true. 

I think it is important to accelerate processing.

josojo:
I can not unapprove it :( its good work, but its not yet ready :)

bh2smith:
Note that some of the changes included here have already been introduced in #85

josojo:
sorry, it is not yet ready for review. Just wanted to check whether Travis starts, if I am activating this...

driver/src/order_driver.rs
```diff
@@ -0,0 +1,94 @@
...
+            let orders = db.get_orders_of_slot(slot.low_u32())?;
+            let order_hash = orders.rolling_hash();
+            if order_hash != contract_order_hash {
+                return Err(DriverError::new(
+                    &format!("Pending order hash from contract ({}), didn't match the one found in db ({})", 
+                    contract_order_hash, order_hash), ErrorKind::StateError
+                ));
+            }
+
+            let solution = if orders.len() > 0 {
```
josojo: Instead of making the if case here, I would propose to check in find_prices if the orders.len()>0 and then return an error. The error is then turned into the fallback solution in the next line.

driver/src/order_driver.rs
fleupold: Sounds good. Can you submit a separate PR?

fleupold:
Looks great. Thanks for adding all the tests. 

I think we should hold off merging this until the listener code is landed and we can actually integrate this component.

bh2smith:
@fleupold does your comment suggest it is time to move back to Listener code and start handling Auction Settlement? Also, when was the last time this branch was synced with master? I am seeing a lot of changes that have already been applied here (typos and IDE warnings)

fleupold:
@bh2smith yes, if I understand correctly the listener should be the last functional piece missing for a full e2e Demo (which is very exciting)

bh2smith:
Except that the solver isn't actually called yet? I can get back into the Listener code for Auction Settlement first thing on Montag! Still need to pass account balances into the naive solver (So it doesn't try to fill an order with insufficient balance).

josojo:
@bh2smith The solver is already being called. Felix took care of everything :) So now, the listener is the blocking part. I think it has higher prio as balances in naive solver. Let's talk about it on Monday

driver/src/contract.rs
```diff
@@ -207,14 +207,17 @@ impl SnappContract for SnappContractImpl {
         prev_state: H256,
         new_state: H256,
         order_hash: H256,
-        prices: Vec<u8>, 
-        volumes: Vec<u8>) -> Result<()> {
+        prices_and_volumes: Vec<u8>) -> Result<()> {
+            println!("prices_and_volumes: {:?}", &prices_and_volumes);
```
bh2smith: Is there any way of "logging" these rather than printing?

bh2smith:
I think this should be merged after Auction Settlement branch.

driver/src/price_finding/naive_solver.rs
```diff
@@ -47,9 +51,8 @@ impl Order {
         exec_sell_amount: u128,
     ) -> U256 {
         // TODO - Refer to Alex's Lemma [ceil(p/float(q)) == (p + q - 1) // q]
```
bh2smith: I think this line can go. The fact that the integer division (p + q - 1) // q equals the Ceil it probably well known. However we should document this some how. Perhaps just as a note here.

```
// ceil(p/float(q)) == (p + q - 1) // q
```

bh2smith:
```suggestion
        // Note that: ceil(p / float(q)) == (p + q - 1) // q
```

driver/src/order_driver.rs
```diff
@@ -83,11 +83,11 @@ fn compute_updated_balances(
     let mut result = balances.clone();
     for (index, order) in orders.iter().enumerate() {
```
josojo: In case you might not be aware, I have a unit test for this function written in here:
https://github.com/gnosis/dex-services/pull/99/files#diff-37736b2e285edcecdc56bd9aacf3fceaL107

Since the old unit test was hopelessly broken

PR #82, opened 2019-04-05
Naive Solver

bh2smith:
Returns (hopefully) a feasible, non-trivial solution to the price matching algorithm whenever a direct match between two tokens can be made (used for integration testing)

Currently, solver is written in python, and will be translated into rust and integrated into the driver code.


tmp/simple_solver.py
```diff
@@ -0,0 +1,54 @@
...
+        match_conditions = [
+            x.buy_token == y.sell_token,
+            y.buy_token == x.sell_token,
+            x.buy_amount * y.buy_amount <= y.sell_amount * x.sell_amount
+        ]
+
+        if all(match_conditions):
+
+            if x.buy_amount <= y.sell_amount and x.sell_amount <= y.buy_amount:  # Type I-A (x <= y)
+                res_price[x.buy_token] = x.sell_amount
```
fleupold: What prevents an order from getting executed multiple times? E.g. order 2 could be matched first with order 1 and then order 2?

bh2smith:
@twalth3r could you help me compute the surplus at some point tomorrow?

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,74 @@
+use crate::models;
+
+use super::error::PriceFindingError;
+use super::price_finder_interface::Solution;
+use web3::types::U256;
+use crate::models::Order;
+
+impl Order {
```
josojo: Intuitively, I would have written these impl functions into the models:Order file, so that we do not forget about them and do not write these functions twice.

But I am not so sure here,

josojo:
sweet that you are rocking rust code!

driver/src/price_finding/naive_solver.rs
bh2smith: I was planning to put them in the appropriate place at some point, but at the moment I am just trying to make it work.. Good question though, maybe it actually makes sense to leave them here if they are only ever used here. We will discuss and think about this soon.

driver/src/models.rs
```diff
@@ -107,7 +107,7 @@ fn merkleize(leafs: Vec<Vec<u8>>) -> H256 {
     merkleize(next_layer)
 }
 
-#[derive(Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd, Debug)]
```
fleupold: Why is this needed? We should probably prefer to implement Display properly ourselves.

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn matches(&self, other: &Order) -> bool {
```
fleupold: How about having this return an `Option<enum>` which can be either TypeI-A TypeI-B or TypeII (maybe you can come up with some more descriptive names for them).

Then in your loop below you can use rusts awesome pattern matching feature.

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,74 @@
...
+                    prices[x.buy_token - 1] = y.sell_amount;
+                    prices[y.buy_token - 1] = x.sell_amount;
+
+                    sell_amount[i] = x.sell_amount;
+                    sell_amount[j] = y.sell_amount;
+
+                    buy_amount[i] = y.sell_amount;
+                    buy_amount[j] = x.sell_amount;
+                }
+                break;
```
fleupold: As with the python code, I find it somewhat counterintuitive that we set everything up to be looped but in reality we only want to find the first match and then return immediately. Maybe we can prepare the solution inside the loop and return there, and otherwise just return `trivialSolution()`

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+pub fn solve(orders: &Vec<models::Order>, num_tokens: u8) -> Result<Solution, PriceFindingError> {
+    // TODO - use mapping here
+    let mut prices: Vec<u128> = vec![0; num_tokens as usize];
+    let mut sell_amount: Vec<u128> = vec![0; orders.len()];
+    let mut buy_amount: Vec<u128> = vec![0; orders.len()];
+
+    for (i, x) in orders.iter().enumerate() {
```
fleupold: A slightly more functional way of writing this could use [combinations](https://docs.rs/itertools/0.7.6/itertools/trait.Itertools.html#method.tuple_combinations) and 

```
let match = orders.tuple_combinations().iter().find(|(o1, o2)| o1.matches(o2).is_some());
```

driver/src/price_finding/naive_solver.rs
bh2smith: OMG yes! The hardest part will be coming up with the better names.

driver/src/price_finding/naive_solver.rs
bh2smith: I really thought I had already replaces this break with a return statement...

driver/src/price_finding/naive_solver.rs
bh2smith: Yes, I was looking for something like this, but hadn't yet found anything. Thanks!

driver/src/models.rs
bh2smith: This was only temporary while I "debug" my code.

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn opposite_tokens(&self, other: &Order) -> bool {
+        self.buy_token == other.sell_token && self.sell_token == other.buy_token
+    }
+
+    fn have_price_overlap(&self, other: &Order) -> bool {
+        self.buy_amount * other.buy_amount <= other.sell_amount * self.sell_amount
+    }
+}
+
+pub fn solve(orders: &Vec<models::Order>, num_tokens: u8) -> Result<Solution, Error> {
```
josojo: I think it would be good to solve function, which is also processing the account balances. If I am not mistaken, then we also want to have a valid solution, which fits to our accounting system.

bh2smith: Yes, I suppose that would be true. However, I think I will first focus on getting a complete solution first. I haven't coded the surplus in Rust just yet. I would also like to somehow lose the x, y loop that I have here.

driver/src/price_finding/naive_solver.rs
fleupold: The calculation of balances is logic that is the same for all solvers. That’s why I’d argue it should live outside the actual solver component (it’s currently part of `run_auction_listener`)

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl Order {
+    fn attracts(&self, other: &Order) -> bool {
+        self.opposite_tokens(other) && self.have_price_overlap(other)
+    }
+    fn match_compare(&self, other: &Order) -> Option<OrderPairType> {
+        if self.attracts(other) {
+            if self.buy_amount <= other.sell_amount && self.sell_amount <= other.buy_amount {
+                Some(OrderPairType::TypeIa);
```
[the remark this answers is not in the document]
bh2smith: Me too! But the code won’t compile without it!  What to do.

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+    fn opposite_tokens(&self, other: &Order) -> bool {
+        self.buy_token == other.sell_token && self.sell_token == other.buy_token
+    }
+    fn have_price_overlap(&self, other: &Order) -> bool {
+        self.buy_amount * other.buy_amount <= other.sell_amount * self.sell_amount
+    }
+    fn surplus(
+        &self,
+        price: u128,
```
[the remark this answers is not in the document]
bh2smith: The price is needed to scale the amounts out of the “basis” they are represented in so they can all be added together.

driver/src/price_finding/naive_solver.rs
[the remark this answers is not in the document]
fleupold: Try `return Some(...);`

driver/src/price_finding/naive_solver.rs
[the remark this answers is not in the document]
fleupold: I might be misunderstanding. Is this the price from SellToken to BuyToken or something else? Maybe we can name it in a way that makes it clear.

driver/src/price_finding/naive_solver.rs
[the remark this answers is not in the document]
bh2smith: will do!

driver/src/price_finding/naive_solver.rs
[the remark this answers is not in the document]
bh2smith: Certainly will name this more appropriately.

driver/src/price_finding/naive_solver.rs
bh2smith: Right, yes, although I may have misunderstood @josojo's suggestion here is to pass account balances into this function so that we know what we are about to execute was actually a valid order (i.e. the user has sufficient balance to fulfill the executed sell amount).

driver/src/price_finding/naive_solver.rs
bh2smith: This is still a great idea to be implemented in the future.

driver/src/price_finding/naive_solver.rs
bh2smith: Yes, just to clarify, as I may have misunderstood your suggestion here is to pass account balances into this function so that we know what we are about to execute was actually a valid order (i.e. the user has sufficient balance to fulfill the executed sell amount).

driver/src/price_finding/naive_solver.rs
bh2smith: @fleupold  I am not sure what the reference to "balances" in your comment has to do with these Order implementations. Most of the functions implemented here are specific to the execution of the Naive Solver. However, the Order::surplus function might be useful to have outside of this file.

fleupold: The comment was meant as a  response to the suggestion below where Alex said we should do balance updates inside the solver.

josojo: @fleupold It's fine if we have the logic for balance updates outside of the solver. However, I was concerned that solvers need to enforce that all orders and their volumes are covered by sufficient balances of their participants. This needs to be in the solver logic itself, doesn't it? 
Also, each solver would hopefully have its own strategy to find a solution that is in accordance with the users' balances.

driver/src/price_finding/naive_solver.rs
```diff
@@ -0,0 +1,349 @@
...
+    BothPartiallyFulfilled,
+}
+
+impl Order {
+    fn attracts(&self, other: &Order) -> bool {
+        self.opposite_tokens(other) && self.have_price_overlap(other)
+    }
+
+    fn sufficient_seller_funds(&self, state: &State) -> bool {
+        let balance_index = self.sell_token as usize + (self.account_id - 1) as usize * TOKENS as usize;
```
bh2smith:
```suggestion
        let balance_index = (self.sell_token as usize) - 1 + (self.account_id - 1) as usize * TOKENS as usize;
```

bh2smith: I think we should somehow make a balance index function as a utility somewhere.

bh2smith:
@fleupold 

> If left.sellAmount == right.buyAmount and left.buyAmount == right.sellAmount the match_compare function would return LhsCompletelyFulfilled, wouldn't it? In this case both would be fully fulfilled.

This may be the case for the condition you have proposed, but that isn't an exhaustive analysis of what the last case describes. The first two cases accurately describe that Left and Right sides are completely filled. In the last scenario, there exists a price `P` between the stated limit prices `P_L and P_R` for which both order can be completely filled and the solution is as given in the code. This is actually some sort of Lemma, that we can provide a proof for if necessary. Although I think providing the solution (as is done here) counts as a proof.

driver/src/price_finding/naive_solver.rs
fleupold: Yes please. Let’s do a PR where balances is no longer public on the struct so that no one can directly access it, but instead has to go through helper functions.

bh2smith:
Closes #58

PR #84, opened 2019-04-08
[Bug] Typo in e2e auction data

bh2smith:
Order number 3 needed to be scaled by 10 in order to match the retrETH example. 

Easy fix.


bh2smith:
There is literally only one character of difference in the PR. I suspect the build is failing for some different reason. According to the logs there was no Ethereum client available for the contract migration. Has something changed there?

bh2smith:
@fleupold Yes, it was an accident.

Issue #83, opened 2019-04-08
Convert script input to WEI


fleupold:
The numbers we currently enter in our `.js` scripts (deposit/place order etc.) need to be scaled up by 10**18 to be in the right format for our driver (all values in the driver are expected to be wei and the optimizer doesn't work well with very tiny amounts). Since mongoDb doesn't seem to handle numbers >64bit we likely have to store values as strings, which will require some changes in the python code as well.

bh2smith:
> Since mongoDb doesn't seem to handle numbers >64bit we likely have to store values as strings, which will require some changes in the python code as well.

This is going to be a nightmare...

bh2smith:
I guess my question is, do we really need to change the scripts by multiplying amount by 10**18 or do we simply call the scripts with values like 

```js
truffle exec scripts/deposit.js  1 1 100000000000000000
```

bh2smith:
Isn't this also going to require changes to the driver... This is starting to seem like a complete overhaul.

- Listener; store amounts as string (this will involve conversion of the integer emited from the contract to string)

- Driver; Parse DB query results from string and interpret them at uintXX

- Tests; Update all deposit and withdraw requests in shell scripts from 1, 2, 3 to 100000000000000000, 200000000000000000, 300000000000000000 etc...

fleupold:
The driver part is already implemented as part of #80 (e.g. https://github.com/gnosis/dex-services/blob/38f977a5724973f0123a00ccfbfea4da64232ed8/driver/src/models.rs#L88). Feel free to use/copy that implementation.

I'd strongly suggest not to require the js scripts to take value with 18 zeros from the command-line, but instead use `web3.utils.toWei(amount_arg)` inside the script.

You are right that this touches a lot of components, but it turns out to be actually not that much code.

And it's really handy that we have a good set of e2e tests which allows finding the pieces that need adjustments quite quickly.

bh2smith:
@fleupold - Do you think that state_index and slot (both also using `uint` in the contract) should also be stored in the DB as strings?

fleupold:
Technically yes, but given that we can do 4 billions auctions before we would overflow them (with 3 minutes batch time that’s at least 12 billion minutes = ~25 years) we have some time before this becomes a practical issue.

bh2smith:
At that point it will be like another Y2K scramble!

PR #85, opened 2019-04-08
#83 - Store all large integers in DB as string

bh2smith:
1. Event Listener to write Strings to DB for all "large integers" (i.e. > 64 bits). Only change required here is to update the `to_dictionary` function for each model (since these dictionaries are what is written to the DB). Its really nice that we don't have to change the model definitions!

2. Driver will have to implement a different `from` method as which parses these amounts as strings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,8 @@ matrix:
       script:
         - cd dex-contracts && truffle migrate && cd -
         - sh ./test/e2e-tests-deposit.sh
-        - sh ./test/e2e-tests-auction.sh
         - sh ./test/e2e-tests-withdraw.sh
+        - sh ./test/e2e-tests-auction.sh
       after_script:
         - docker-compose logs
       before_cache:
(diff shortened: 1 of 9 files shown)


josojo:
Yes, this is the right direction to go!

bh2smith:
Maybe one of @josojo  or @fleupold can help with this (not very descriptive) error I am getting with the current state of the code.

```
driver_1 | Querying stateHash: { "stateHash": "0000000000000000000000000000000000000000000000000000000000000000" }
driver_1 | Deposit_driver error: JSON error
driver_1 | Current top withdraw_slot is 1
driver_1 | Highest unprocessed withdraw_slot is 0
driver_1 | Processing withdraw_slot 0
driver_1 | Querying stateHash: { "stateHash": "0000000000000000000000000000000000000000000000000000000000000000" }
driver_1 | Withdraw_driver error: JSON error
```

Best I can tell is that this is coming from `get_current_balances`

bh2smith:
@josojo & @fleupold. This is now working with e2e-tests and ready for review. Note that some parts have been taken from this other draft PR #80 (particularly the From implementations for the DB models).

bh2smith:
@fleupold unfortunately it is not going to be so easy to fix this last failing test case. The reason for failure is that the Merkle Roots do not agree. This is actually a bug in the contracts repo with the JS implementation if `encodePacked_16_8_128`. I will open an issue in contracts that will need to be taken care of before this can ever pass!

PR #91, opened 2019-04-12
[Driver, Easy] Refactoring deposit_driver for readability

josojo:
refactoring with hopefully the same beauty as Felix's code

### DIFF
--- a/driver/src/deposit_driver.rs
+++ b/driver/src/deposit_driver.rs
@@ -4,8 +4,10 @@ use crate::models::RollingHashable;
 use crate::db_interface::DbInterface;
 use crate::error::{DriverError, ErrorKind};
 use crate::contract::SnappContract;
+use crate::util;
 
-use web3::types::{H256, U256};
+use web3::types::H256;
+
 
 pub fn apply_deposits(
     state: &mut models::State,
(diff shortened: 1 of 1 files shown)


driver/src/deposit_driver.rs
```diff
@@ -21,74 +23,41 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Driver
...
+            let updated_balances = apply_deposits(&mut balances, &deposits);
+            let new_state_root = H256::from(updated_balances.rolling_hash());
+            
+            println!("New State_hash is {}", new_state_root);
+            contract.apply_deposits(slot, state_root, new_state_root, contract_deposit_hash)?;
             return Ok(true);
+        } else {
+            println!("Need to wait before processing deposit_slot {:?}", slot);
         }
-    } else {
```
fleupold: I kind of liked this piece of information

driver/src/deposit_driver.rs
```diff
@@ -21,74 +23,41 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Driver
...
+            let state_root = contract.get_current_state_root()?;
+            let contract_deposit_hash = contract.deposit_hash_for_slot(slot)?;
+            let mut balances = db.get_current_balances(&state_root)?;
+
+            let deposits = db.get_deposits_of_slot(slot.low_u32())?;
+            let deposit_hash = deposits.rolling_hash();
+            if deposit_hash != contract_deposit_hash {
+                return Err(DriverError::new(
+                    &format!("Pending deposit hash from contract ({}), didn't match the one found in db ({})", 
+                    deposit_hash, contract_deposit_hash), ErrorKind::StateError
```
fleupold: The order here is off. According to the string we have to print `contract_deposit_hash` first

driver/src/deposit_driver.rs
```diff
@@ -21,74 +23,41 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Driver
...
+                ));
+            }
+
+            let updated_balances = apply_deposits(&mut balances, &deposits);
```
fleupold: I find it weird that this method returns state *and* takes mutable state (hinting that the state that is input is changed and changed state is returned). 

I’d prefer if the input is immutable and we just return the updated state.

driver/src/deposit_driver.rs
bh2smith:
```suggestion
                    contract_deposit_hash, deposit_hash), ErrorKind::StateError
```

PR #92, opened 2019-04-13
[Listener] Auction Settlement

bh2smith:
- Listener should receive Auction Settlement Event, store it in the DB, and update account balances according to the `pricesAndVolumes` emited as part of the event. This involves implementation of

- Auction Settlement model (with same attributes as the event)
- to and from dictionary for DB storage
- serialize solution (byte array to prices and trade executions)
- update balances based on Orders and auction results

### DIFF
--- a/event_listener/dfusion_db/models.py
+++ b/event_listener/dfusion_db/models.py
@@ -141,7 +141,7 @@ class AuctionSettlement(NamedTuple):
     auction_id: int
     state_index: int
     state_hash: str
-    prices_and_volumes: str  # Stored as Hex String
+    prices_and_volumes: str  # Emitted as Hex String
 
     @classmethod
     def from_dictionary(cls, data: Dict[str, Any]) -> "AuctionSettlement":
@@ -162,14 +162,14 @@ class AuctionSettlement(NamedTuple):
             "pricesAndVolumes": self.prices_and_volumes,
         }
 
-    def serialize_solution(self) -> "AuctionResults":
+    def serialize_solution(self, num_tokens: int) -> "AuctionResults":
         """Transform Byte Code for prices_and_volumes into Prices & TradeExecution objects"""
         logging.info("Serializing Auction Results (from bytecode)")
         tmp = self.prices_and_volumes[2:]
         hex_str_array = [tmp[i] + tmp[i + 1] for i in range(0, len(tmp), 2)]
         byte_array = list(map(lambda t: int(t, 16), hex_str_array))
 
-        prices, volumes = byte_array[:30], byte_array[30:]
+        prices, volumes = byte_array[:num_tokens], byte_array[num_tokens:]
         buy_amounts = volumes[0::2]  # Even elements
         sell_amounts = volumes[1::2]  # Odd elements
 
--- a/event_listener/dfusion_db/models.py
+++ b/event_listener/dfusion_db/models.py
@@ -1,3 +1,4 @@
+import logging
 from enum import Enum
 from typing import NamedTuple, Dict, Any, List, Optional
 
@@ -36,7 +37,7 @@ class Deposit(NamedTuple):
     @classmethod
     def from_dictionary(cls, data: Dict[str, Any]) -> "Deposit":
         event_fields = ('accountId', 'tokenId', 'amount', 'slot', 'slotIndex')
-        assert all(k in data for k in event_fields), "Unexpected Event Keys"
+        assert all(k in data for k in event_fields), "Unexpected Event Keys: got {}".format(data.keys())
         return Deposit(
             int(data['accountId']),
             int(data['tokenId']),
@@ -67,7 +68,7 @@ class Withdraw(NamedTuple):
     @classmethod
     def from_dictionary(cls, data: Dict[str, Any]) -> "Withdraw":
         event_fields = ('accountId', 'tokenId', 'amount', 'slot', 'slotIndex')
-        assert all(k in data for k in event_fields), "Unexpected Event Keys"
+        assert all(k in data for k in event_fields), "Unexpected Event Keys: got {}".format(data.keys())
         return Withdraw(
             int(data['accountId']),
             int(data['tokenId']),
@@ -114,7 +115,7 @@ class Order(NamedTuple):
     @classmethod
     def from_dictionary(cls, data: Dict[str, Any]) -> "Order":
         event_fields = ('auctionId', 'slotIndex', 'accountId', 'buyToken', 'sellToken', 'buyAmount', 'sellAmount')
-        assert all(k in data for k in event_fields), "Unexpected Event Keys"
+        assert all(k in data for k in event_fields), "Unexpected Event Keys: got {}".format(data.keys())
         return Order(
             int(data['auctionId']),
             int(data['slotIndex']),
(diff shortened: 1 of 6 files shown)


event_listener/dfusion_db/snapp_event_receiver.py
```diff
@@ -132,4 +132,36 @@ def save_parsed(self, order: Order) -> None:
...
+        except AssertionError as exc:
+            logging.critical(
+                "Failed to record Settlement [{}] - {}".format(exc, settlement))
+
+    def __update_accounts(self, settlement: AuctionSettlement) -> None:
+        balances = self.database.get_account_state(settlement.state_index - 1).balances.copy()
+        num_tokens = self.database.get_num_tokens()
+
+        # Apply the balance updates according to settlement.prices_and_volumes
+        # Should we bother to fetch the corresponding orders?
```
josojo: We need to know the owners and the tokens involved in a given executed trading volume. I think orders are the only place where we can this information from.

event_listener/dfusion_db/snapp_event_receiver.py
```diff
@@ -132,4 +132,36 @@ def save_parsed(self, order: Order) -> None:
...
+        # We can assume the solution took balances into account?
```
josojo: For now, yes. 
But later we need for sure to make sure that only valid solutions are pushed into the database. But then we need a whole price submission verification tool, which goes beyond balance checking. Let's talk about this in our meeting today

event_listener/dfusion_db/tests/snapp_event_receiver_test.py
```diff
@@ -1,12 +1,13 @@
...
-from ..snapp_event_receiver import WithdrawRequestReceiver, DepositReceiver, StateTransitionReceiver, \
-    SnappInitializationReceiver, OrderReceiver
+from ..snapp_event_receiver import DepositReceiver, OrderReceiver, SnappInitializationReceiver
+from ..snapp_event_receiver import WithdrawRequestReceiver, StateTransitionReceiver
 from ..models import Deposit, StateTransition, TransitionType, Withdraw, AccountRecord, Order
 
 
 class DepositReceiverTest(unittest.TestCase):
-    def test_writes_deposit(self) -> None:
+    @staticmethod
```
fleupold: Somewhat unrelated changes ☹️

test/e2e-tests-deposit.sh
```diff
@@ -9,7 +9,7 @@ truffle exec scripts/setup_environment.js
 truffle exec scripts/deposit.js 3 3 18 
 truffle exec scripts/mine_blocks.js 21
 
-sleep 5
+sleep 10
```
fleupold: We should figure out a better way here than waiting a fixed amount of time. Maybe we can check repeatedly and fail if 10s have passed and it’s still not there.

event_listener/dfusion_db/models.py
```diff
@@ -135,3 +135,72 @@ def to_dictionary(self) -> Dict[str, Any]:
...
+        """Transform Byte Code for prices_and_volumes into Prices & TradeExecution objects"""
+        logging.info("Serializing Auction Results (from bytecode)")
+
+        # TODO - pass num_orders (as read from DB for solution verification)
+        tmp = self.prices_and_volumes[2:]
+        hex_str_array = [tmp[i: i+24] for i in range(0, len(tmp), 24)]
+        byte_array = list(map(lambda t: int(t, 16), hex_str_array))
+
+        prices, volumes = byte_array[:num_tokens], byte_array[num_tokens:]
+        buy_amounts = volumes[0::2]  # Even elements
```
fleupold: I believe this is incorrect. Can you check the implementation of `Solution.bytes` to see the exact encoding.

bh2smith: `Solution.bytes` is currently under reconstruction. The implementation found here is based on some discussions around possibly reducing the call data sent to the contract in the event that the order slots are not all full.

The Byte String encoding of pricesAndVolumes is now interpreted as follows

```
[p_1, p_2, ..., p_T | bA_1, sA_1, bA_2, sA_2, ...] 
```

where each entry denotes the hex representation of a 96 bit integer and

- `p_i` are prices
- `bA_i` are executed buy amounts
- `sA_i` are executed sell amounts

The reason for the interweaved buy-sell amounts is so that we can treat a zero-padded byte string as equivalent to the full length in the event that not all orders are filled.

bh2smith: @josojo  will push the changes to `Solution.bytes` in this PR

event_listener/dfusion_db/tests/snapp_event_receiver_test.py
bh2smith: Part of unit testing

test/e2e-tests-deposit.sh
bh2smith: Perhaps, but the driver sleep for 5 seconds anyway. So repeatedly trying will never succeed until at least 5 seconds has passed.

event_listener/dfusion_db/models.py
bh2smith: So essentially, one will be able to submit a Solution for an auction that only had 50 orders, by only submitting the bytes for each of the prices and only the trade executions of those 50 orders. All the remaining bytes will be treated as zeros.

fleupold: The same is possible in the existing solution by splitting the array in half and treating the first half as buy and second half as sell orders.

event_listener/dfusion_db/tests/snapp_event_receiver_test.py
fleupold: But these tests are not related to this change. You would still want to make this change if this PR was to be reverted.

test/e2e-tests-deposit.sh
fleupold: Theoretically the driver could pick up the change 1ms after the 21 blocks have been mined (if its personal 5s had passed in that moment). That means we might be waiting 10s too long in the worst case. 

Having these sleeps in tests is generally an anti pattern which we should try to solve more elegantly than just increasing the time to wait.

event_listener/dfusion_db/models.py
bh2smith: Yes I suppose that is true. Will talk to Alex, if he hasn't pushed this already.

bh2smith: This is true, but will result in more operations on the Smart Contract. Since we will need to "pad" the bytes coming in with zeros (so that the sha-hash agrees).

bh2smith: Right here, in the contract

https://github.com/gnosis/dex-contracts/blob/c8f452de6a2a5c4a2858df1bcabfec0494923770/contracts/SnappAuction.sol#L119

fleupold: I still don’t see how we don’t have to pad the other solution as well. The only difference I see is that the 0 bytes could be cheaper in the solution suggested here if the last orders happen to not be fulfilled. 

This is probably something that makes sense discussion synchronously.

josojo: Yes, we have to pad it as well, the only minor advantage of our solution is that it costs a little bit less onchain. I am personally fine with any solution, we just have to converge to one solution

fleupold: Can you explain how it is cheaper with a small example?

josojo: In your solution, we have to pad it twice - on two separate places as in our solution we only have to pad the ending. It does not really

bh2smith: Small example, 

Consider an auction which settles with the 6-order retrETH example (EUR, WINE, OIL). The solution would then be 
```
[p_E, p_O, p_W, 0, ..., 0 | bA1, sA1, bA2, sA2, ..., bA6, sA6, 0, 0, ..., 0, 0]
```
In this case we would only pass `[p_E, p_O, p_W, 0, ..., 0 | bA1, sA1, bA2, sA2, ..., bA6, sA6]` to the contract (as bytes), which would then pad these bytes when storing the sha-hash of the solution information as

```
auctions[slot].shaHash = sha256(pricesAndVolumes + "BUNCH_OF_ZERO_BYTES");
```

since, I assume, the Snark requires the hash to be of all information (even the non-execution of empty order slots).

PR #96, opened 2019-04-15
Closes #95 - Balance Index as utility function

### LINKED ISSUE #95: Balance index as utility function
Balance index, computed as 

```js
balance_index = num_tokens * (account_id - 1) + token_id - 1
```
should have a utility function to avoid possible errors

```js
fn balance_index(token_id: u8, account_id: u16) -> usize {
    TOKENS as usize * (account_id - 1) as usize  + (token_id - 1) as usize
}
```

bh2smith:


bh2smith:
@fleupold I know you mentioned making the balances privarte inside the state, but it appears (in about  half the cases) that these balances are sometimes updated. This is why I have refrained from doing something like the following

```js
impl State {
    fn balance_at(&self, token_id: u8, account_id: u16) -> u128 {
        compute index and return balance
    }
}
```

josojo:
It's only an approve under the assumption that the test is fixed for sure ;)

PR #107, opened 2019-04-25
Adjusting driver to deal with 3 minute batch times instead of 20 blocks

fleupold:
As title. E2E test should be passing.

### DIFF
--- a/driver/src/withdraw_driver.rs
+++ b/driver/src/withdraw_driver.rs
@@ -40,7 +40,7 @@ pub fn run_withdraw_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Drive
     if slot <= withdraw_slot {
         println!("Highest unprocessed withdraw_slot is {:?}", slot);
         if util::can_process(slot, contract,
-            Box::new(&|i| contract.creation_block_for_withdraw_slot(i))
+            Box::new(&|i| contract.creation_timestamp_for_withdraw_slot(i))
         )? {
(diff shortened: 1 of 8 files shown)


driver/src/withdraw_driver.rs
```diff
@@ -146,9 +146,9 @@ mod tests {
         contract.has_withdraw_slot_been_applied.given(slot).will_return(Ok(false));
         contract.has_withdraw_slot_been_applied.given(slot - 1).will_return(Ok(false));
 
-        contract.creation_block_for_withdraw_slot.given(slot-1).will_return(Ok(U256::from(10)));
+        contract.creation_timestamp_for_withdraw_slot.given(slot-1).will_return(Ok(U256::from(10)));
 
-        contract.get_current_block_number.given(()).will_return(Ok(U256::from(34)));
+        contract.get_current_block_timestamp.given(()).will_return(Ok(U256::from(200)));
```
bh2smith: This number, 200, seems arbitrary. What is it meant to be?

fleupold:
> Would it make sense in this context to use something like get_timestamp and pass which type we want as a parameter

I thought about this, and ended up with the thought that having three different methods or a method with a parameter, which can be three different types, is almost the same. Especially, since the implementation of this method would likely use three different calls and thus likely be a switch statement with three cases.

It might however be shorter for the mocked implementation to write. I would not object to such a change but am not sure if it's worth the hassle.

PR #110, opened 2019-04-27
Auction Initialization receiver, handler and tests

bh2smith:
### Change Summary

Although apparently large, this change merely includes `num_orders` as a constant in the database on SnappAuction deployment event. A few other parts of the code are changed as a result of this (such as additional validation for solution parsing on Auction Settlement).

### Test Plan
Event Listener Unit Tests should suffice here.


bh2smith:
This is failing because of bug #113 which should be resolved before proceeding here.

driver/src/price_finding/naive_solver.rs
```diff
@@ -65,7 +65,7 @@ impl PriceFinding for NaiveSolver {
         state: &State
     ) -> Result<Solution, PriceFindingError> {
         // Initialize trivial solution
-        let mut prices: Vec<u128> = vec![1; 1 + TOKENS as usize];
+        let mut prices: Vec<u128> = vec![1; TOKENS as usize];
```
bh2smith: @fleupold - Not sure how to get the commit has from this, but you could cherry-pick this if you haven't already.

event_listener/dfusion_db/tests/models_test.py
```diff
@@ -167,11 +168,21 @@ def test_to_dictionary(self) -> None:
 
 class AuctionResultsTest(unittest.TestCase):
 
+    def test_naive_retreth_results(self):
```
fleupold: This test kind of comes out of nowhere (as in I don't understand how it relates to the PR). I'm not entirely sure this is a valuable test given that the input string would be really hard to adjust by hand if anything in the encoding changed. We's likely just copy a new instance from the e2e test.

On the other hand I understand that it would help to catch the failing e2e faster by just running unit tests.

Anyway, I'd prefer it this could be a separate PR where we can discuss this.

bh2smith: Agree, included a new branch `retreath_listener_test` with only this test.

PR #111, opened 2019-04-27
settings.py still reading old env files

bh2smith:
I would like to avoid duplicate env var declaration in the docker-compose if possible. Any suggestions here would be cool!


docker-compose.yml
```diff
@@ -52,6 +52,9 @@ services:
     environment:
       - SNAPP_CONTRACT_ADDRESS=0xD833215cBcc3f914bD1C9ece3EE7BF8B14f841bb
       - ETHEREUM_NODE_URL=http://ganache-cli:8545
+      - DB_HOST=mongo
```
josojo: I think, we could write all env variables into .env and .env_db, etc. and then always read out of there. yml files can also read from there.

bh2smith: Could you show me how? I might prefer this over rewriting them twice everywhere.

josojo: Maybe 
```
while read LINE; do echo "export $LINE" | cut -f1 -d":"; done < /env
```

bh2smith: Yes, but then this line of shell script would have to be put into each container right?

josojo: no, we would only run it in the pre-install script

Then in the container we would put the envs via:
```
environment:
      - SNAPP_CONTRACT_ADDRESS=$SNAPP_CONTRACT_ADDRESS
      - ETHEREUM_NODE_URL=$ETHEREUM_NODE_URL
```

That is the most "elegant" method I am aware of. But maybe you want to google a little bit more

bh2smith: I am beginning to think that it might make the most sense to be very explicit about what environment variables are needed in which containers. For example, we should notice that the Listener does not require any special environment to interact with the solver (because it doesn't need to). With the code as it appears right now, we can clearly see that both the driver and listener require access to the contract and the database.

bh2smith: @josojo I fixed it as suggested!

PR #112, opened 2019-04-27
[Lint] Several clippy warnings by suggestion

bh2smith:
## Issue Description

Full report of `clippy` warnings can be found here

https://gist.github.com/bh2smith/af26bb187b1d1dbc1dc3f1963ac84052

Brief summary (i.e. some highlights) of those most commonly fixed

- warning: redundant closure found
```
 | .map_err(|e| DriverError::from(e))
 |          ^^^^^^^^^^^^^^^^^^^^^^^^ help: remove closure as shown: `DriverError::from`
```
- writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices.
```
 | fn serialize_balances(balances: &Vec<u128>, num_tokens: u8) -> serde_json::Value {
 |                                 ^^^^^^^^^^ help: change this to: `&[u128]`
```
- warning: length comparison to zero
```
 | if docs.len() == 0 {
 |    ^^^^^^^^^^^^^^^ help: using `is_empty` is clearer and more explicit: `docs.is_empty()`
```
- warning: identical conversion
```
 | let new_state_root = H256::from(updated_balances.rolling_hash());
 |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `H256::from()`: `updated_balances.rolling_hash()`
 
```

## Test Plan

```sh
git check master
cd driver
cargo clippy
```
Observe the warnings! 

```sh
git check cargo_clippy
cargo clippy
```
Be amazed!

### DIFF
--- a/driver/src/bin/main.rs
+++ b/driver/src/bin/main.rs
@@ -16,7 +16,8 @@ fn main() {
     let db_instance = MongoDB::new(db_host, db_port).unwrap();
     let contract = SnappContractImpl::new().unwrap();
 
-    let solver_env_var = env::var("LINEAR_OPTIMIZATION_SOLVER").unwrap_or("0".to_string());
+    let solver_env_var = env::var("LINEAR_OPTIMIZATION_SOLVER")
+        .unwrap_or_else(|_| "0".to_string());
     let mut price_finder : Box<PriceFinding> = if solver_env_var == "1" {
         Box::new(LinearOptimisationPriceFinder::new())
     } else {
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -93,85 +93,85 @@ impl SnappContract for SnappContractImpl {
                     None => Err(web3::Error::from("Current block not found"))
                 }
             })
-            .map_err(|e| DriverError::from(e))
+            .map_err(DriverError::from)
     }
 
     fn get_current_state_root(&self) -> Result<H256> {
         self.contract.query(
             "getCurrentStateRoot", (), None, Options::default(), None
-        ).wait().map_err(|e| DriverError::from(e))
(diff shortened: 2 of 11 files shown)


driver/src/bin/main.rs
```diff
@@ -16,7 +16,8 @@ fn main() {
     let db_instance = MongoDB::new(db_host, db_port).unwrap();
     let contract = SnappContractImpl::new().unwrap();
 
-    let solver_env_var = env::var("LINEAR_OPTIMIZATION_SOLVER").unwrap_or("0".to_string());
+    let solver_env_var = env::var("LINEAR_OPTIMIZATION_SOLVER")
+        .unwrap_or_else(|_| "0".to_string());
```
fleupold: Intuitively, I'd assume that using a closure here is less efficient than the constant we used to use before. Do you know why this version is supposed to be better?

driver/src/contract.rs
```diff
@@ -93,85 +93,85 @@ impl SnappContract for SnappContractImpl {
                     None => Err(web3::Error::from("Current block not found"))
                 }
             })
-            .map_err(|e| DriverError::from(e))
+            .map_err(DriverError::from)
```
fleupold: noice

driver/src/bin/main.rs
fleupold: found it: https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call

bh2smith: I can't tell if this link is for or against the change proposed here. Looks like an argument for the lint warning.

PR #117, opened 2019-04-29
[E2E] replacing sleep with retry statement

### LINKED ISSUE #100: [Refactor] Remove Sleep Statements
This includes integration tests (in bash scripts),  `run_listener` and `run_driver`.

fleupold:
I'm not entirely sure if this is a much better approach, since it requires a lot of thinking on which commands need to be retried and which ones don't.

Yet, it will make sure that we are not waiting unnecessarily and also not subject to flaky timing issues.

Looking forward to a discussion...

**Testplan**
Install retry and run e2e tests locally:

```
brew pull 27283
brew install retry
echo 'export PATH="/usr/local/opt/gnu-getopt/bin:$PATH"' >> ~/.bash_profile
./test/e2e-tests-deposit-withdraw.sh && ./test/e2e-tests-auction.sh
```

Closes #100

### DIFF
--- a/test/e2e-tests-auction.sh
+++ b/test/e2e-tests-auction.sh
@@ -25,17 +25,17 @@ truffle exec scripts/sell_order.js 4 1 0 4 52 # executed sell: 52
 truffle exec scripts/sell_order.js 5 2 0 20 280
 
 # Test Listener: There are now 6 orders in auction slot 1 and sellAmount for accountId = 5 is 280000000000000000000
-retry "mongo dfusion2 --eval \"db.orders.find({'auctionId': ${EXPECTED_AUCTION}}).size()\" | grep -w 6"
-retry "mongo dfusion2 --eval \"db.orders.findOne({'auctionId': ${EXPECTED_AUCTION}, 'accountId': 5}).sellAmount\" | grep -w 280000000000000000000"
+retry -t 5 "mongo dfusion2 --eval \"db.orders.find({'auctionId': ${EXPECTED_AUCTION}}).size()\" | grep -w 6"
+retry -t 5 "mongo dfusion2 --eval \"db.orders.findOne({'auctionId': ${EXPECTED_AUCTION}, 'accountId': 5}).sellAmount\" | grep -w 280000000000000000000"
 
 truffle exec scripts/wait_seconds.js 181
 
 # Test balances have been updated
 EXPECTED_HASH="0e0369b8be154350fd09141a43d90a53427221001ba9fee5c97145e777420944"
-retry "truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot' | grep ${EXPECTED_HASH}"
+retry -t 5 "truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot' | grep ${EXPECTED_HASH}"
 
 # Account 4 has now 4 of token 1 
-retry "mongo dfusion2 --eval \"db.accounts.findOne({'stateHash': '$EXPECTED_HASH'}).balances[121]\" | grep -w 4000000000000000000"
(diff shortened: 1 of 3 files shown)


fleupold:
Checkout https://travis-ci.org/gnosis/dex-services/jobs/526043012

Usually succeeds after first retry.
In the last instance succeeded after two retries.

test/e2e-tests-auction.sh
```diff
@@ -6,7 +6,7 @@ cd dex-contracts/
 EXPECTED_AUCTION=0
 
 # Ensure no orders yet in auction slot 1
-mongo dfusion2 --eval "db.orders.find({'auctionId': ${EXPECTED_AUCTION}}).size()" | grep 0
+mongo dfusion2 --eval "db.orders.find({'auctionId': ${EXPECTED_AUCTION}}).size()" | grep -w 0
```
bh2smith: What does the -w mean?

bh2smith:
* Test plan might require a `docker-compose build` to install retry in the container.

bh2smith:
Overall, I must say I am not terribly excited about including this `retry` as a dependency. This would be the point at which we rely on OS specific dependencies (i.e. Linux and Mac OS install both require different install guides -- at least for dev related tasks). The fact that it is just as non-deterministic as the sleep statements makes the inclusion of new external dev-dependencies even less appealing. 

I do, however, really like that you can see how many retries are needed before the tests move on. This would also reduce the total time necessary for a unit test to complete.

Also, once we have better logging in the driver, we could reduce the sleep time for the driver to awake. from 5 to even 1 (Notice how the Listener logs only appear when something happens and not every time it checks for something.

bh2smith:
I would be willing to approve this if the general vibe is leaning towards yes.

bh2smith:
How long does it wait before giving up?

fleupold:
> Test plan might require a docker-compose build to install retry in the container.

e2e tests run on local machine and not in container

> How long does it wait before giving up?

Configurable. By default after 10 retries (exponentially backing off from 0.3). 5 retries (10s is probably enough)

> Overall, I must say I am not terribly excited about including this retry as a dependency.

If you look at what `retry` is, it's really just another shell script (similar to how we use assert in the trusted setup e2e test). Instead of downloading it, we could also include it's source code in the repo directly.

Researching more thoughts on "sleep in unit tests" I found the general sentiment is that it's a bad idea and this kind of time-bound retry is considered better - mostly since it's not trading off robustness vs. execution speed in the happy case (e.g. [[1]](https://testing.googleblog.com/2008/08/tott-sleeping-synchronization.html), [[2]](https://ardalis.com/thread-sleep-in-tests-is-evil), [[3]](https://automationintesting.com/selenium/java/lessons/waits.html))

My two major concerns are:
1) It now takes much longer to get notified if a test is failing (can be adjusted with less number of retries)
2) Retry is used on multiple commands and thus requires a thought of which test assertions need to be waited for. Maybe we can find a way to trigger a single "state advanced" event in the listener/driver and then just retry-wait for this single event.

test/e2e-tests-auction.sh
fleupold: Match whole word. You are very likely to find a 0 somewhere in the sessionID (and thus this command used to pass even if the auction size was not 0)

josojo:
>It now takes much longer to get notified if a test is failing (can be adjusted with less number of retries)

Yeah, this is true. But it is also only true, if the failing statement is a retry, which should not happen too often.


I also don't have a strong opinion. I don't know what will be more convenient in the long run. Currently, I am tending for merging.

bh2smith:
Any action on this since I was away? To merge or not to merge, that is the question.

fleupold:
I'm in favor of merging and trying this out unless there are strong opinions against.

PR #118, opened 2019-04-30
[Driver] Models submodule

fleupold:
The models module has become quite large and messy, so I'm splitting it up into a submodule where each model can be in it's own file.

This change just moves a bunch of code around, so reviewing it should despite the PR's size not be too hard.

**Test plan** continuous integration.


driver/src/models/flux.rs
```diff
@@ -0,0 +1,90 @@
+use byteorder::{BigEndian, WriteBytesExt};
+use serde_derive::{Deserialize, Serialize};
+use web3::types::H256;
+
+use crate::models::{Serializable, RootHashable, merkleize};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug, Ord, PartialOrd, Eq)]
+#[serde(rename_all = "camelCase")]
```
bh2smith: I still don't really understand what this is...?

driver/src/models/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub trait RootHashable {
+    fn root_hash(&self, valid_items: &[bool]) -> H256;
+}
+
+pub trait Serializable {
+    fn bytes(&self) -> Vec<u8>;
+}
+
+fn merkleize(leafs: Vec<Vec<u8>>) -> H256 {
```
bh2smith: How does anyone ever use this if it isn't public? What am I missing here?

driver/src/models/mod.rs
```diff
@@ -0,0 +1,47 @@
+pub mod state;
```
bh2smith: What is with the file name here?  What is mod? There is no struct called mod here...

driver/src/models/state.rs
```diff
@@ -1,26 +1,13 @@
...
-pub trait RootHashable {
-    fn root_hash(&self, valid_items: &[bool]) -> H256;
-}
 
-pub trait Serializable {
-    fn bytes(&self) -> Vec<u8>;
-}
+use crate::models::{TOKENS, RollingHashable, Order};
+use crate::price_finding::{Solution};
+use crate::price_finding::linear_optimization_price_finder::{account_id, token_id};
```
bh2smith: could we, at some point, put this account and token id into some utility file?

fleupold: In my next PR I will make those private again. The model should not know about how the linear optimization solver decides to convert IDs into strings.

driver/src/models/mod.rs
fleupold: It just says that the module state (code you can find in state.rs) is a public submodule of the models model

driver/src/models/state.rs
bh2smith: Cool beans!

driver/src/models/mod.rs
bh2smith: No, my question is in reference to the file name `mod.rs`

driver/src/models/flux.rs
fleupold: When serializing it to json it turns the snake_case field names into camelCase.

E.g. the struct
```
{field_a: "foo"}
```

would become the json

```
{"fieldA": "foo"}
```

This is needed, because python uses camel case and rust uses snake case by convention for variable names. Thus python writes camelCased json objects into our DB. In order to use snake_cased variable names in rust while still automatically converting what we find in the database, this transformation is needed.

driver/src/models/mod.rs
fleupold: It's a convenience method to help implement the RootHashable trait, which is public. External components should use the rootHashable trait and not this method directly.

driver/src/models/flux.rs
bh2smith: > python uses camel case and rust uses snake case by convention for variable names

In fact, both use snake case for variables (and functions)

Sweet I get it now. thanks.

driver/src/models/mod.rs
fleupold: `models/mod.rs` contains the code that is exposed by the `models` module (`use crate::models`). Every other file in this folder contains the code exposed by the submodulde (e.g. `models/foo.rs` would be accessed via `use crate::models::foo`)

If there is no other submodule in a folder having `models/mod.rs` or just `models.rs` is equivalent. See more: https://doc.rust-lang.org/rust-by-example/mod/split.html

driver/src/models/flux.rs
fleupold: Interesting, than just the conversion into json is done with camelCase in python. We could thus also change all of this code (https://github.com/gnosis/dex-services/blob/master/event_listener/dfusion_db/models.py#L51-L55) to use snake case in the db.

PR #121, opened 2019-05-04
[Driver, Easy]refactoring the hash consistency check

josojo:
After reading the book, I just had to do one small refactoring. More to come!


driver/src/util.rs
```diff
@@ -21,6 +22,16 @@ pub fn find_first_unapplied_slot(
     Ok(U256::zero())
 }
 
+pub fn check_consistency_of_hashes(hash_calculated: H256, hash_from_contract: H256, flux_type: &str) -> Result<(), DriverError>{
```
fleupold: Indentation in this method it off.

driver/src/deposit_driver.rs
```diff
@@ -41,12 +41,7 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Driver
 
             let deposits = db.get_deposits_of_slot(slot.low_u32())?;
             let deposit_hash = deposits.rolling_hash(0);
-            if deposit_hash != contract_deposit_hash {
-                return Err(DriverError::new(
-                    &format!("Pending deposit hash from contract ({}), didn't match the one found in db ({})", 
-                    contract_deposit_hash, deposit_hash), ErrorKind::StateError
-                ));
-            }
+            check_consistency_of_hashes(deposit_hash, contract_deposit_hash, &(String::from("deposit")) )?;
```
fleupold: Why do you need to create a string and then reference it? Shouldn't 

```
check_consistency_of_hashes(deposit_hash, contract_deposit_hash, "deposit")?;
```

work?

josojo: yes it does!

driver/src/util.rs
```diff
@@ -1,36 +1,54 @@
...
         if has_slot_been_applied(slot - 1)? {
-            return Ok(slot)
+            return Ok(slot);
         }
         slot = slot - 1;
     }
     Ok(U256::zero())
 }
 
+pub fn check_consistency_of_hashes(
```
bh2smith: I propose renaming this to `hash_consistency_check` (not fond of conjunctions or prepositions in function names)

driver/src/util.rs
```diff
@@ -1,36 +1,54 @@
...
 }
 
+pub fn hash_consistency_check(
+    hash_calculated: H256,
+    hash_from_contract: H256,
+    flux_type: &str,
+) -> Result<(), DriverError> {
+    if hash_calculated != hash_from_contract {
+        return Err(DriverError::new(
+            &format!(
```
bh2smith: Could we merge the logger first and then change this to a log error... Or does it matter here since an Err is being returned anyway?

Issue #102, opened 2019-04-17
[Logs] Use proper logger (rust)


bh2smith:
Remove `println!` statements and replace with proper logger. 

Would also be nice to see a log regarding when auction orders are being passed to optimization solver (and which solver)

PR #128, opened 2019-06-04
[Easy] replacing println with logger statements

fleupold:
Addresses #102. Uses https://crates.io/crates/log with https://crates.io/crates/simple_logger.

### Test Plan

Watch output from e2e test


driver/src/withdraw_driver.rs
```diff
@@ -57,11 +57,11 @@ pub fn run_withdraw_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Drive
             let withdrawal_merkle_root = withdraws.root_hash(&valid_withdraws);
             let new_state_root = updated_balances.rolling_hash(balances.state_index + 1);
             
-            println!("New State_hash is {}, Valid Withdraw Merkle Root is {}", new_state_root, withdrawal_merkle_root);
+            debug!("New State_hash is {}, Valid Withdraw Merkle Root is {}", new_state_root, withdrawal_merkle_root);
             contract.apply_withdraws(slot, withdrawal_merkle_root, state_root, new_state_root, contract_withdraw_hash)?;
             return Ok(true);
         } else {
-            println!("Need to wait before processing withdraw_slot {:?}", slot);
+            info!("Need to wait before processing withdraw_slot {:?}", slot);
```
bh2smith: I see only one `info` and I believe that many of the `debug` above should also be `info`.

fleupold: You are right. I inversed the "severeness" of DEBUG and INFO in my head.

bh2smith: Killer!

bh2smith:
@fleupold Conflicts have been resolved and this appears ready for merge!

fleupold:
Not ready for merge yet, since I can't see any output on travis.

fleupold:
Nice catch. Looks like `evm_mine` is not working as expected in their "coverage-enabled fork of the ganache-cli client".
Could you create an issue with a minimal example on their github page? https://github.com/sc-forks/solidity-coverage

PR #134, opened 2019-06-12
Updating services to work with refactored library contract

fleupold:
Depends on https://github.com/gnosis/dex-contracts/pull/127/files

We need to add the address for the Lib (and adjust the address of the main contract) as well as use the designated view functions inside the driver instead of accessing state directly.

### TestPlan:
E2E tests


bh2smith:
e2e testing is this right?

```
docker-compose down && docker-compose up
e2e-tests-deposit-withdraw.sh
e2e-tests-auction.sh
```

PR #138, opened 2019-06-14
[Driver] Exposing db interface to query standing orders

fleupold:
This PR exposes StandingOrders as a concept in the driver and will allow the order_driver to query for them in the current auction.

### Test Plan

In the next PR make sure standing orders are queried correctly.

### DIFF
--- a/driver/src/models/order.rs
+++ b/driver/src/models/order.rs
@@ -7,7 +7,6 @@ use crate::models::{Serializable, RollingHashable, iter_hash};
 #[derive(Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct Order {
-    pub slot_index: u32,
     pub account_id: u16,
     pub sell_token: u8,
     pub buy_token: u8,
@@ -36,7 +35,6 @@ impl Serializable for u128 {
 impl From<mongodb::ordered::OrderedDocument> for Order {
     fn from(document: mongodb::ordered::OrderedDocument) -> Self {
         Order {
-            slot_index: document.get_i32("slotIndex").unwrap() as u32,
             account_id: document.get_i32("accountId").unwrap() as u16,
             buy_token: document.get_i32("buyToken").unwrap() as u8,
             sell_token: document.get_i32("sellToken").unwrap() as u8,
@@ -61,7 +59,6 @@ pub mod tests {
   #[test]
   fn test_order_rolling_hash() {
     let order = Order {
-      slot_index: 0,
       account_id: 1,
       sell_token: 2,
       buy_token: 3,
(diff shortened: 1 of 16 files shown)


driver/src/models/order.rs
```diff
@@ -7,7 +7,6 @@ use crate::models::{Serializable, RollingHashable, iter_hash};
 #[derive(Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct Order {
-    pub slot_index: u32,
```
bh2smith: What was the reason for this change? is this being recorded elsewhere? Does it not matter? Should we stop emitting the event in the contract? I think the contract needs to know for the purpose of fees.

fleupold: The field is never read in the contract code. It's required to be stored in the database, because we need to be able to query in a specific order. However, the driver then never has to read this index again since it knows that all orders are for the queried one.

I noticed this when trying to create orders for a standing order batch. There it would have been weird to think about `slot_indexes`. It would be possible to create "fake" slot indices, but given we don't even use this field, I removed it.

bh2smith: I feel like the contract is going to need this at some point if the fee model depends on the order index. If it isn't needed here for now then I suppose this was a good find!

bh2smith: I just hope we are not relying on the insertion order of the events in the DB, since I am still unsure if we can rely on them to be emitted in the order that they are collected.

fleupold: > I feel like the contract is going to need this at some point if the fee model depends on the order index.

The contract and the db need it, but not the driver. I'd hope that the fee model doesn't affect the driver.

> I just hope we are not relying on the insertion order of the events in the DB

That's a good find, we are indeed relying on that order. I think we can rely on the SC events being processed in order (that's a guarantee from the listener), however we should just add the sort criteria. Ok to do that in another PR? Even today since we never sort by slot_index (it's not accessed anywhere in the code) this potential problem exists.

bh2smith: Also, technically (according to the formal specification) there should be no discrimination of an order's index in a given auction, and it is much easier to prove such a fact if the index is never passed to the solver in the first place. So in a sense, this could be considered a benefit.

driver/src/models/standing_order.rs
```diff
@@ -0,0 +1,30 @@
...
+impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
+    fn from(document: mongodb::ordered::OrderedDocument) -> Self {
+        let account_id = document.get_i32("_id").unwrap() as u16;
+        StandingOrder {
+            account_id,
+            orders: document
+                .get_array("orders")
+                .unwrap()
+                .iter()
+                .map(|o| o.as_document().unwrap())
```
bh2smith: picky situation are the `o` on this line and the next meant to be interpreted as the same thing? If not I might call them something different. Either way, I think we could expand/elaborate on what these are here (since we don't necessarily need to save characters here).

PR #139, opened 2019-06-14
Fetching StandingOrders in order_driver

fleupold:
This PR starts fetching standing orders as part of the driver's order processing cycle.

### Test Plan

Added unit test.
Also made sure that running the docker landscape and executing the following commands, shows that we are processing the StandingOrder correctly:

```
cd dex-contracts
truffle exec scripts/setup_environment.js 6
truffle exec scripts/standing_order.js 0 1 2 1 1
truffle exec scripts/wait_seconds.js 181
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,5 +41,6 @@ matrix:
         - cd dex-contracts && retry truffle migrate && cd -
         - sh ./test/e2e-tests-deposit-withdraw.sh
         - sh ./test/e2e-tests-auction.sh
+        - sh ./test/e2e-tests-standing-order.sh
       after_script:
         - docker-compose logs
--- a/README.md
+++ b/README.md
@@ -87,8 +87,8 @@ For end-to-end tests, run from the project root:
 
 ```bash
 docker-compose down && docker-compose up
-test/e2e-tests-deposit.sh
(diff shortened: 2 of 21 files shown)


josojo:
Great PRs, very clean, very easy to review. Thanks!

I think the current sql-query, is not yet sufficient. Probably even the way we store them in the database might not be sufficient. 
Probably, it is explained by the following test the easiest:
```
cd dex-contracts
truffle exec scripts/setup_environment.js 6
truffle exec scripts/standing_order.js 0 1 2 1 1
truffle exec scripts/standing_order.js 0 1 2 1 4
truffle exec scripts/wait_seconds.js 181
```
Now the driver is processing the order:
```
truffle exec scripts/standing_order.js 0 1 2 1 1
```
instead of:
```
truffle exec scripts/standing_order.js 0 1 2 1 4
```

I think the best solution would be to make the field _id of a standing order an incrementing field, allowing us sort by _id and get the newest element.

fleupold:
> I think the current sql-query, is not yet sufficient.

Great catch 🎣 

Every record already has an implicitly growing objectId, which I believe we can use for this purpose. I will also add this scenario to the e2e test cases.

PR #140, opened 2019-06-14
[Listener] Consider standing orders when computing balance after AuctionSettlement

fleupold:
When computing the updated balances after auction settlement, we now also need to consider the standing orders that were valid at the given time.

### Test plan
Unit test + e2e test in next PR

### DIFF
--- a/event_listener/dfusion_db/database_interface.py
+++ b/event_listener/dfusion_db/database_interface.py
@@ -140,7 +140,7 @@ class MongoDbInterface(DatabaseInterface):
 
         pipeline = [
             {"$match": {"validFromAuctionId": {"$lte": auction_id}}},
-            {"$sort": {"validFromAuctionId": -1}},
+            {"$sort": {"validFromAuctionId": -1, "_id": -1}},
             {"$group": {
                 "_id": "$accountId",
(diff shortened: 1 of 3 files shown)


event_listener/dfusion_db/database_interface.py
```diff
@@ -136,7 +136,22 @@ def write_order(self, order: Order) -> None:
             "Successfully included Order - {}".format(order_id))
 
     def get_orders(self, auction_id: int) -> List[Order]:
-        return list(map(Order.from_dictionary, self.database.orders.find({'auctionId': auction_id})))
+        orders = list(map(Order.from_dictionary, self.database.orders.find({'auctionId': auction_id})))
+
+        pipeline = [
```
bh2smith: This is very interesting. Is this a mongo specific filtration feature?

event_listener/dfusion_db/database_interface.py
```diff
@@ -136,7 +136,22 @@ def write_order(self, order: Order) -> None:
...
+            {"$match": {"validFromAuctionId": {"$lte": auction_id}}},
+            {"$sort": {"validFromAuctionId": -1}},
```
josojo: Probably, we have here the same issue as in PR #139

PR #141, opened 2019-06-14
[e2e] update standing order test with full logic

fleupold:
Adjusting the e2e test to make sure

a) Standing ordered gets carried over in full (getting matched again even if matched before)
b) Standing order can be canceled after which it will no longer get matched.

### DIFF
--- a/test/e2e-tests-standing-order.sh
+++ b/test/e2e-tests-standing-order.sh
@@ -21,7 +21,8 @@ truffle exec scripts/sell_order.js 1 2 1 1 1
 truffle exec scripts/wait_seconds.js 181
 retry -t 5 "mongo dfusion2 --eval \"db.accounts.findOne({'stateIndex': 6}).balances[1]\" | grep -2 2000000000000000000"
 
-# Cancel standing order
+# Update, then cancel standing order in same batch (make sure only cancel gets processed)
+truffle exec scripts/standing_order.js 0 1 2 1 2
 truffle exec scripts/standing_order.js 0 0 0 0 0
 
 # Make sure it's no longer traded
--- a/test/e2e-tests-standing-order.sh
+++ b/test/e2e-tests-standing-order.sh
@@ -3,8 +3,28 @@ set -e
 
 cd dex-contracts/
 
+truffle exec scripts/sell_order.js 1 2 1 1 1
+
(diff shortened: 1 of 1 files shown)


test/e2e-tests-standing-order.sh
```diff
@@ -3,8 +3,28 @@ set -e
 
 cd dex-contracts/
 
+truffle exec scripts/sell_order.js 1 2 1 1 1
+
 # Place standing order in current Auction (accountId, buyToken, sellToken, minBuy, maxSell)
-truffle exec scripts/standing_order.js 0 1 2 12 12
+truffle exec scripts/standing_order.js 0 1 2 1 1
```
bh2smith: If we are using the first 50 account slots for reserved accounts, does this mean we have adjusted all the regular orders in the other E2E tests?

I see the sell order directly above is referring to accountId = 1

fleupold: Reserved accounts can place normal sell orders as well, so I don't think it should affect the e2e tests.

bh2smith: Ahhh yes I see. I suppose I hadn't noticed that there is not yet an implementation for account reservation.

test/e2e-tests-standing-order.sh
```diff
@@ -21,7 +21,8 @@ truffle exec scripts/sell_order.js 1 2 1 1 1
 truffle exec scripts/wait_seconds.js 181
 retry -t 5 "mongo dfusion2 --eval \"db.accounts.findOne({'stateIndex': 6}).balances[1]\" | grep -2 2000000000000000000"
 
-# Cancel standing order
+# Update, then cancel standing order in same batch (make sure only cancel gets processed)
```
bh2smith: Interesting example!

josojo: I think the auction index need to be changed (0->2) to make the test effective

bh2smith: Where is the auction index being referred to in this setting?

fleupold: I also don't understand. The values are: `(accountId, buyToken, sellToken, minBuy, maxSell)`

josojo: sry

PR #143, opened 2019-06-17
Modifications for new web3 lib

josojo:
I think we should use the new web3 lib, as the error features are more advanced and the error implementation changed quite a bit.

--
Finally, PR is ready for review
Testplan:
only unit tests

### DIFF
--- a/driver/Cargo.toml
+++ b/driver/Cargo.toml
@@ -22,7 +22,7 @@ rustc-hex = "*"
 sha2 = "*"
 byteorder = "*"
 hex = "0.3.2"
-web3 = { git = "https://github.com/tomusdrw/rust-web3", rev = "56ad08e" }
+web3 = "*"
 log = "0.4.6"
 simple_logger = "1.2.0"
 
--- a/driver/Cargo.toml
+++ b/driver/Cargo.toml
@@ -9,6 +9,7 @@ name = "driver"
 path = "src/lib.rs"
 
 [dependencies]
+error-chain = "0.12"
 chrono = "0.4.6"
 ethabi = "6.1.0"
 mongodb = "0.3.12"
--- a/driver/Cargo.toml
+++ b/driver/Cargo.toml
@@ -9,7 +9,6 @@ name = "driver"
 path = "src/lib.rs"
 
 [dependencies]
-error-chain = "0.12"
 chrono = "0.4.6"
 ethabi = "6.1.0"
 mongodb = "0.3.12"
--- a/driver/Cargo.toml
+++ b/driver/Cargo.toml
@@ -10,7 +10,7 @@ path = "src/lib.rs"
 
 [dependencies]
 chrono = "0.4.6"
-ethabi = "6.1.0"
+ethabi = "7.0.0"
 mongodb = "0.3.12"
 serde_json = "1.0"
 serde_derive = "1.0"
@@ -21,7 +21,7 @@ rustc-hex = "*"
 sha2 = "*"
 byteorder = "*"
 hex = "0.3.2"
-web3 = "*"
+web3 = "0.7"
 log = "0.4.6"
 simple_logger = "1.2.0"
 
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -59,7 +59,7 @@ impl SnappContractImpl {
         let snapp_base: serde_json::Value = serde_json::from_str(&contents)?;
         let snapp_base_abi: String = snapp_base.get("abi").ok_or("No ABI for contract")?.to_string();
 
-        let address: Address = (env::var("SNAPP_CONTRACT_ADDRESS")?)[2..22].parse()?;
+        let address: Address = (env::var("SNAPP_CONTRACT_ADDRESS")?)[2..].parse()?;
         let contract = Contract::from_json(web3.eth(), address, snapp_base_abi.as_bytes())?;
         Ok(SnappContractImpl { contract, web3, event_loop })
     }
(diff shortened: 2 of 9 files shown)


driver/src/contract.rs
```diff
@@ -90,7 +89,7 @@ impl SnappContract for SnappContractImpl {
             .and_then(|block_option| {
                 match block_option {
                     Some(block) => Ok(block.timestamp),
-                    None => Err(web3::Error::from("Current block not found"))
+                    None => Err(web3::error::Error::Decoder(String::from("Current block not found")))
```
bh2smith: Is this the new error you were looking for!?

josojo: no ...

bh2smith: it looks like it's coming from web3...

driver/src/contract.rs
```diff
@@ -59,9 +59,8 @@ impl SnappContractImpl {
         let snapp_base: serde_json::Value = serde_json::from_str(&contents)?;
         let snapp_base_abi: String = snapp_base.get("abi").ok_or("No ABI for contract")?.to_string();
 
-        let snapp_address = hex::decode(&(env::var("SNAPP_CONTRACT_ADDRESS")?)[2..])?;
-        let address: Address = Address::from(&snapp_address[..]);
-        let contract = Contract::from_json(web3.eth(), address, snapp_base_abi.as_bytes())?;
+        let address: Address = (env::var("SNAPP_CONTRACT_ADDRESS")?)[2..22].parse()?;
+        let contract = Contract::from_json(web3.eth(), address, snapp_base_abi.as_bytes()).unwrap();
```
josojo: I want to avoid the unwrap here, but then I get:

```
the trait bound `error::DriverError: std::convert::From<ethabi::errors::Error>` is not satisfied
```
I am implementing the trait:
```
impl From<ethabi::errors::Error> for DriverError {
    fn from(error: ethabi::errors::Error) -> Self {
        DriverError::new(error.description(), ErrorKind::AbiError)
    }
}
```
it keeps on telling me that the trait ethabi::errors is private.

bh2smith: So you are trying to remove the `unwrap`. In this case we should expect to get a return value of type `Result<Contract<T>, ethabi::Error>`

bh2smith: Yes, I can tell you whats wrong, here... Might have something to do with a [bug report](https://github.com/intellij-rust/intellij-rust/issues/3654) I filed a while back.

Notice that my IDE thinks the return type is `Contract<http>` is this expected?

Screenshot 2019-06-17 at 9 40 19 AM

bh2smith: Actually, the only problem I am experiencing when removing the unwrap is the mismatched types in the next line

```
error[E0308]: mismatched types
  --> src/contract.rs:64:32
   |
64 |         Ok(SnappContractImpl { contract, web3, event_loop })
   |                                ^^^^^^^^ expected struct `web3::contract::Contract`, found enum `std::result::Result`
```

driver/Cargo.toml
```diff
@@ -21,7 +21,7 @@ rustc-hex = "*"
 sha2 = "*"
 byteorder = "*"
 hex = "0.3.2"
-web3 = { git = "https://github.com/tomusdrw/rust-web3", rev = "56ad08e" }
+web3 = "*"
```
fleupold: Let's 📌 to a specific version here. Otherwise we might get random failures when the library pushes a new version that is breaking in some way. I think we should control the upgrades ourselves.

driver/src/contract.rs
fleupold: ethabi::error seems to be public here: https://docs.rs/ethabi/8.0.0/ethabi/struct.Error.html

driver/src/models/state.rs
```diff
@@ -56,7 +57,7 @@ impl RollingHashable for State {
       let result = hasher.result();
       hash = result.to_vec();
     }
-    H256::from(hash.as_slice())
+    H256::from(from_slice(&hash))
```
fleupold: Why did this have to change?

driver/Cargo.toml
bh2smith: Doesn't pinning ourselves to a specific version this early in development mean that we might get stuck behind and miss updates?

fleupold: I'm not saying we should never update, but we should be in control of our update process, specifically if the dependent library seems to be unstable in that it introduces breaking changes on a regular basis.

driver/src/contract.rs
josojo: > o you are trying to remove the unwrap. In this case we should expect to get a return value of type Result<Contract<T>, ethabi::Error>

sorry, I wanted to replace unwrap with error propagation done with a ?

>ethabi::error seems to be public here: https://docs.rs/ethabi/8.0.0/ethabi/struct.Error.html
Nice, an update to a newer version helped :)

driver/src/models/state.rs
josojo: H256 is part of the new web3 library. And in this library, they enforce the from-operator to act on a [u8, 32] instead of &[u8].
I think they enforce it because it is safer.

fleupold: Does `H256::from_slice(hash.as_slice())` work?

driver/src/deposit_driver.rs
```diff
@@ -47,7 +47,7 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Driver
             let new_state_root = updated_balances.rolling_hash(balances.state_index + 1);
             
             info!("New State_hash is {}", new_state_root);
-            contract.apply_deposits(slot, state_root, new_state_root, contract_deposit_hash)?;
+            contract.apply_deposits(slot + 10, state_root, new_state_root, contract_deposit_hash)?;
```
fleupold: What is this?

bh2smith: This has got to be a mistake! Good find!

driver/src/models/state.rs
josojo: works... nice! I updated it

PR #150, opened 2019-06-19
125 apply auction with new reorg protection: [step 1] get standing_order_index array form database

josojo:
It contains the query for getting standing_order_index from the database.

PR not yet ready. But I have some questions in the code.

Testplan:
automatic e2e tests only.

### DIFF
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -21,6 +21,8 @@ pub trait SnappContract {
     fn get_current_deposit_slot(&self) -> Result<U256>;
     fn get_current_withdraw_slot(&self) -> Result<U256>;
     fn get_current_auction_slot(&self) -> Result<U256>;
+    fn calculate_order_hash(&self, slot: U256, standing_order_index: Vec<U128>) -> Result<H256>;
+
 
     // Deposit Slots
     fn creation_timestamp_for_deposit_slot(&self, slot: U256) -> Result<U256>;
@@ -41,9 +43,6 @@ pub trait SnappContract {
     fn apply_deposits(&self, slot: U256, prev_state: H256, new_state: H256, deposit_hash: H256) -> Result<()>;
     fn apply_withdraws(&self, slot: U256, merkle_root: H256, prev_state: H256, new_state: H256, withdraw_hash: H256) -> Result<()>;
     fn apply_auction(&self, slot: U256, prev_state: H256, new_state: H256, order_hash: H256, standing_order_index: Vec<U128>, prices_and_volumes: Vec<u8>) -> Result<()>;
-
-    //Temporary functions
-    fn calculate_order_hash(&self, slot: U256, standing_order_index: Vec<U128>) -> Result<H256>;
 }
 
 #[allow(dead_code)] // event_loop needs to be retained to keep web3 connection open
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -3,7 +3,7 @@ extern crate mock_it;
 
 use web3::contract::{Contract, Options};
 use web3::futures::Future;
-use web3::types::{Address, H256, U256, BlockId};
+use web3::types::{Address, H256, U256, U128, BlockId};
 
 use crate::error::DriverError;
 
@@ -40,7 +40,10 @@ pub trait SnappContract {
     // Write methods
     fn apply_deposits(&self, slot: U256, prev_state: H256, new_state: H256, deposit_hash: H256) -> Result<()>;
     fn apply_withdraws(&self, slot: U256, merkle_root: H256, prev_state: H256, new_state: H256, withdraw_hash: H256) -> Result<()>;
-    fn apply_auction(&self, slot: U256, prev_state: H256, new_state: H256, order_hash: H256, prices_and_volumes: Vec<u8>) -> Result<()>;
+    fn apply_auction(&self, slot: U256, prev_state: H256, new_state: H256, order_hash: H256, standing_order_index: Vec<U128>, prices_and_volumes: Vec<u8>) -> Result<()>;
+
+    //Temporary functions
+    fn calculate_order_hash(&self, slot: U256, standing_order_index: Vec<U128>) -> Result<H256>;
 }
 
 #[allow(dead_code)] // event_loop needs to be retained to keep web3 connection open
@@ -219,6 +222,7 @@ impl SnappContract for SnappContractImpl {
         prev_state: H256,
         new_state: H256,
         order_hash: H256,
+        standing_order_index: Vec<U128>,
         prices_and_volumes: Vec<u8>) -> Result<()> {
             info!("prices_and_volumes: {:?}", &prices_and_volumes);
             let account = self.account_with_sufficient_balance().ok_or("Not enough balance to send Txs")?;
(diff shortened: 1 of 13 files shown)


driver/src/db_interface.rs
```diff
@@ -131,6 +135,38 @@ impl DbInterface for MongoDB {
...
+            .db(models::DB_NAME)
+            .collection("standing_orders")
+            .aggregate(pipeline, None)?
+            .map(|doc| doc.unwrap())
+            .map(|doc| (doc.get_i32("_id").unwrap() as u16,U128::from_dec_str(&(doc.get_i32("batchIndex").unwrap() as u16).to_string()).unwrap()))
+            //batchIndex is stored as i32. Probably, we should then also go with u32 in contract instead of U128
+            .collect::<Vec<_>>();
+
+        let mut standing_order_index = Vec::<U128>::with_capacity(50);
+        for i in 0..models::NUM_RESERVED_ACCOUNTS {
```
josojo: What is the best practice? Can I have "so much" code in the db_interface? Should I pull it into order_driver? On the other hand its just inserting 0 entries...

josojo:
currently, https://github.com/gnosis/dex-contracts/pull/111 accepts u128 for the batchIndex.
I think this is over engineered. If we use u32, our exchange can still run for 2**32/360/24/20 = 25k years, assuming no auctions run shorter than 3 minutes.

Currently, in the database, we only store the batch index as u32

driver/src/contract.rs
```diff
@@ -227,13 +231,26 @@ impl SnappContract for SnappContractImpl {
...
-                (slot, prev_state, new_state, order_hash, prices_and_volumes),
+                (slot, prev_state, new_state, order_hash, standing_order_index, prices_and_volumes),
                 account,
                 options,
             ).wait()
             .map_err(DriverError::from)
             .map(|_|())
     }
+    
+    fn calculate_order_hash(
```
fleupold: Are you sure we want to use a smart contract function here? Shouldn't we compute this hash based on the information we got from the database ourselves? Otherwise we might pass invalid order data to the solver without noticing.

driver/src/db_interface.rs
fleupold: I'd suggest we expose the batchIndex on the `StandingOrder` struct that we are already parsing in the other method. The list of StandingOrders is already available in the driver and if we can access `batchIndex` on each of them we don't need this new method at all.

driver/src/contract.rs
josojo: Sure, this would be the second step. New Pr just arrived.

josojo: I wanted to split the PRs for an easier review...

driver/src/db_interface.rs
josojo: You wrote the sql query in such a way that `batchIndex` is not accessible. It only gives you back:
`Standing Orders: [StandingOrder { account_id: 3, orders: [Order { account_id: 3, sell_token: 2, buy_token: 2, sell_amount: 5000000000000000000, buy_amount: 4000000000000000000 }] }]`

I could change your query. But I think these kinds of optimizations of the query for general usage might be wasted love, as new features will require even different query optimizations.

How do you feel about it? I really don't have a strong opinion here.

josojo: Hm... I decided that I will refactor your query, now.

bh2smith: > wasted love

What could this be!?

fleupold: Sorry just saw this. I believe all you'd need is another element in the group clause like:

```rust
"batchIndex" => (doc!{"$first" =>"$batchIndex" })
```

driver/src/contract.rs
fleupold: Ok, I think you could have also put the other PR first (exposing the native order hash calculation) and then just use it here. I feel it's a bit of a waste to add a new function here, just to immediately remove it again.

driver/src/order_driver.rs
```diff
@@ -94,6 +97,22 @@ fn update_balances(state: &mut State, orders: &[Order], solution: &Solution) {
     }
 }
 
+fn get_standing_orders_index(standing_orders: &Vec<models::StandingOrder>) -> Vec<u32> {
```
bh2smith: My question is about what this function is doing? Is it trying to determine which (of all standing orders) is the correct one? Is the index not related to the length of the list of previous standing orders? I see there is a vector here? Shouldn't an index be a number? 

I must be missing something, could you please tell me what this is about?

bh2smith: does the vector represent the index of standing orders where auction id corresponds to the return vector's index?

josojo: This is the list of standing order indexes used in the solution calculation.

driver/src/error.rs
```diff
@@ -30,7 +31,7 @@ impl From<std::io::Error> for DriverError {
 }
 impl From<web3::contract::Error> for DriverError {
     fn from(error: web3::contract::Error) -> Self {
-        DriverError::new(error.description(), ErrorKind::ContractError)
+        DriverError::new(&format!("{:}", error), ErrorKind::ContractError)
```
fleupold: This change seems to be a little bit random. Should we change all error conversions like that in a separate PR?

driver/src/order_driver.rs
```diff
@@ -47,6 +47,11 @@ pub fn run_order_listener<D, C>(
             info!("Standing Orders: {:?}", standing_orders);
             info!("All Orders: {:?}", orders);
 
+            let standing_order_indexes_u32 = get_standing_orders_indexes(&standing_orders);
```
fleupold: Why does `get_standing_orders_indexes` not return `Vec<u128>`? Also I believe you should be able to convert a u32 into a u128 by just calling `x as u128`. No need to convert into a string and back.

driver/src/order_driver.rs
```diff
@@ -91,6 +97,19 @@ fn update_balances(state: &mut State, orders: &[Order], solution: &Solution) {
     }
 }
 
+fn get_standing_orders_indexes(standing_orders: &Vec<models::StandingOrder>) -> Vec<u32> {
```
fleupold: nit: `batch_index_from_standing_orders`?

driver/src/order_driver.rs
```diff
@@ -91,6 +97,19 @@ fn update_balances(state: &mut State, orders: &[Order], solution: &Solution) {
...
+    let mut standing_order_indexes = Vec::<u32>::with_capacity(models::NUM_RESERVED_ACCOUNTS as usize);
+        for i in 0..models::NUM_RESERVED_ACCOUNTS {
```
fleupold: nit: Would be more efficient prefill an array with 0s and then iterate over the standing orders, inserting a value into the 0 array where necessary?

driver/src/order_driver.rs
bh2smith: It appears the code here has been changed substantially since I last asked my questions here. Looks much cleaner!

driver/src/order_driver.rs
```diff
@@ -91,6 +95,14 @@ fn update_balances(state: &mut State, orders: &[Order], solution: &Solution) {
     }
 }
 
+fn batch_index_from_standing_orders(standing_orders: &Vec<models::StandingOrder>) -> Vec<U128> {
+    let mut standing_order_indexes = vec![U128::zero(); models::NUM_RESERVED_ACCOUNTS as usize];
+        for o in standing_orders {
```
fleupold: nit: indentation is off.

PR #151, opened 2019-06-19
125 apply auction with new reorg protection: [step2] calculate the order_hash offchain

josojo:
tackles issues 125

### DIFF
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -20,6 +20,10 @@ pub trait RollingHashable {
     fn rolling_hash(&self, nonce: i32) -> H256;
 }
 
+pub trait DataHashable {
+    fn data_hash(&self, init_hash: H256) -> H256;
+}
+
 pub trait RootHashable {
     fn root_hash(&self, valid_items: &[bool]) -> H256;
 }
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -28,7 +28,7 @@ pub trait Serializable {
 
 fn merkleize(leafs: Vec<Vec<u8>>) -> H256 {
     if leafs.len() == 1 {
-        return H256::from_slice(&leafs[0]);
+        return H256::from(leafs[0].as_slice());
     }
     let next_layer = leafs.chunks(2).map(|pair| {
         let mut hasher = Sha256::new();
@@ -45,5 +45,5 @@ fn iter_hash<T: Serializable>(item: &T, prev_hash: &H256) -> H256 {
     hasher.input(item.bytes());
     let result = hasher.result();
     let b: Vec<u8> = result.to_vec();
-    H256::from_slice(&b)
+    H256::from(b.as_slice())
 }
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -11,7 +11,9 @@ pub use crate::models::standing_order::StandingOrder;
 use sha2::{Digest, Sha256};
 use web3::types::H256;
 
+//ToDo: get variables from database
 pub const TOKENS: u8 = 30;
+pub const NUM_RESERVED_ACCOUNTS: u8 = 50;
 pub const DB_NAME: &str = "dfusion2";
 
 pub trait RollingHashable {
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -1,10 +1,12 @@
 pub mod state;
 pub mod flux;
 pub mod order;
+pub mod standing_order;
 
 pub use crate::models::state::State;
 pub use crate::models::flux::PendingFlux;
 pub use crate::models::order::Order;
+pub use crate::models::standing_order::StandingOrder;
 
 use sha2::{Digest, Sha256};
 use web3::types::H256;
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -20,8 +20,8 @@ pub trait RollingHashable {
     fn rolling_hash(&self, nonce: i32) -> H256;
 }
 
-pub trait DataHashable {
-    fn data_hash(&self, init_hash: H256) -> H256;
+pub trait ConcatingHashable {
+    fn concating_hash(&self, init_hash: H256) -> H256;
 }
 
 pub trait RootHashable {
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -26,7 +26,7 @@ pub trait Serializable {
 
 fn merkleize(leafs: Vec<Vec<u8>>) -> H256 {
     if leafs.len() == 1 {
-        return H256::from(leafs[0].as_slice());
+        return H256::from_slice(&leafs[0]);
     }
     let next_layer = leafs.chunks(2).map(|pair| {
         let mut hasher = Sha256::new();
@@ -43,5 +43,5 @@ fn iter_hash<T: Serializable>(item: &T, prev_hash: &H256) -> H256 {
     hasher.input(item.bytes());
     let result = hasher.result();
     let b: Vec<u8> = result.to_vec();
-    H256::from(b.as_slice())
+    H256::from_slice(&b)
 }
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -20,7 +20,7 @@ pub trait RollingHashable {
     fn rolling_hash(&self, nonce: i32) -> H256;
 }
 
-pub trait ConcatingHashable {
+pub trait ConcatenatingHashable {
     fn concating_hash(&self, init_hash: H256) -> H256;
 }
 
--- a/driver/src/models/mod.rs
+++ b/driver/src/models/mod.rs
@@ -21,7 +21,7 @@ pub trait RollingHashable {
 }
 
 pub trait ConcatenatingHashable {
-    fn concating_hash(&self, init_hash: H256) -> H256;
+    fn concatenating_hash(&self, init_hash: H256) -> H256;
 }
 
 pub trait RootHashable {
--- a/driver/src/models/standing_order.rs
+++ b/driver/src/models/standing_order.rs
@@ -62,3 +62,34 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
         }
     }
 }
+
+#[cfg(test)]
+pub mod tests {
+  use super::*;
+  use web3::types::{H256};
+  use std::str::FromStr;
+
+  #[test]
+  fn test_concating_hash() {
+    let standing_order = models::StandingOrder::new(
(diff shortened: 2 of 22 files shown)


driver/src/order_driver.rs
```diff
@@ -127,7 +146,7 @@ mod tests {
         contract.get_current_block_timestamp.given(()).will_return(Ok(U256::from(200)));
         contract.order_hash_for_slot.given(slot).will_return(Ok(orders.rolling_hash(0)));
         contract.get_current_state_root.given(()).will_return(Ok(state_hash));
-        contract.calculate_order_hash.given((slot, Any)).will_return(Ok(H256::zero()));
+        contract.calculate_order_hash.given((slot, Any)).will_return(Ok(H256::from("0x438d54b20a21fa0b2f8f176c86446d9db7067f6e68a1e58c22873544eb20d72c")));
```
bh2smith: Does this `0x438d54b20a21fa0b2f8f176c86446d9db7067f6e68a1e58c22873544eb20d72c` have a special meaning? Is it the hash of an empty order set?

driver/src/util.rs
```diff
@@ -30,7 +30,7 @@ pub fn hash_consistency_check(
         return Err(DriverError::new(
             &format!(
                 "Pending {} hash from contract ({}), didn't match the one found in db ({})",
-                flux_type, hash_from_contract, hash_calculated
+                flux_type, serde_json::to_string(&hash_from_contract).unwrap(), serde_json::to_string(&hash_calculated).unwrap()
```
bh2smith:
```suggestion
                flux_type, 
                serde_json::to_string(&hash_from_contract).unwrap(),
                serde_json::to_string(&hash_calculated).unwrap()
```

driver/src/order_driver.rs
```diff
@@ -94,6 +97,22 @@ fn update_balances(state: &mut State, orders: &[Order], solution: &Solution) {
     }
 }
 
+fn calculate_order_hash(nonreserved_order_hash: H256, standing_orders: Vec<models::StandingOrder>) -> H256 {
```
fleupold: This could also be exposed on the model directly? Technically, you might be able to reuse the `rollingHashable` trait that we support for `Vec<Order>`.

driver/src/order_driver.rs
```diff
@@ -36,10 +36,15 @@ pub fn run_order_listener<D, C>(
...
             let mut orders = db.get_orders_of_slot(slot.low_u32())?;
-            let mut order_hash = orders.rolling_hash(0);
+            let order_hash = orders.rolling_hash(0);
             hash_consistency_check(order_hash, contract_order_hash, "order")?;
 
             let standing_orders = db.get_standing_orders_of_slot(slot.low_u32())?;
+            let standing_order_index = db.get_standing_orders_index_of_slot(slot.low_u32())?;
+            let order_hash_calculated = calculate_order_hash(order_hash, standing_orders.clone());
+            let order_hash_from_contract = contract.calculate_order_hash(slot, standing_order_index.clone())?;
+            hash_consistency_check(order_hash_calculated, order_hash_from_contract, "order")?;
```
fleupold: I think we should be able to differentiate the two errors if we decide to do a consistency check both for non-reserved as well as for all orders.

So I think we should either wrap the error, or pass a custom string that gets reflected in the error string.

driver/src/order_driver.rs
fleupold: Yeah should maybe not be the default but set in the expectations of each test case.

bh2smith: Ahh, yes, I hadn't noticed that this was a test case!

driver/src/order_driver.rs
josojo: good suggestions

josojo:
Okay, I have implemented another trait DataHashable, instead of RollingHashable. I think RollingHash would be misleading as we are not calculating any rollingHash, we are just hashing one data blub
. Especially, we are not using any nonce with i32.
```
pub trait RollingHashable {
    fn rolling_hash(&self, nonce: i32) -> H256;
}
```
@fleupold Do you think your suggestion is implementable without changing the public trait? If not, how should I change it?

josojo:
Okay, now I see, we could change the trait to:
```
pub trait RollingHashable {
    fn rolling_hash(&self, init_hash: H256) -> H256;
}
```
and claculate the init_hash before the function call as input `H256::from()`.

This would work. It is a nice solution as it would reuse existing traits. But it might also be very confusing as it is not a rolling hash.

Undecided on what to do....

fleupold:
> Okay, now I see, we could change the trait to:

Yes let's do exactly that. This way we can reuse most of the code and the "nonce" just becomes the initial hash.

driver/src/util.rs
```diff
@@ -30,7 +30,9 @@ pub fn hash_consistency_check(
         return Err(DriverError::new(
             &format!(
                 "Pending {} hash from contract ({}), didn't match the one found in db ({})",
-                flux_type, hash_from_contract, hash_calculated
+                flux_type, 
+                serde_json::to_string(&hash_from_contract).unwrap(),
```
fleupold: You should be able to use `{:#x}` in the format string directly, since H256 implements LowerHex (https://doc.rust-lang.org/nightly/core/fmt/trait.LowerHex.html)

driver/src/deposit_driver.rs
```diff
@@ -40,11 +41,11 @@ pub fn run_deposit_listener<D, C>(db: &D, contract: &C) -> Result<(bool), Driver
             let balances = db.get_current_balances(&state_root)?;
 
             let deposits = db.get_deposits_of_slot(slot.low_u32())?;
-            let deposit_hash = deposits.rolling_hash(0);
+            let deposit_hash = deposits.rolling_hash(H256::from(0));
```
fleupold: nit: `H256::zero()` works too

driver/src/models/standing_order.rs
```diff
@@ -17,6 +21,24 @@ impl StandingOrder {
     }
 }
 
+impl RollingHashable for Vec<StandingOrder> {
```
fleupold: Rolling Hashable is already implemented for any Vector of type T as long as this type T implements `Serializable` (https://github.com/gnosis/dex-services/blob/master/driver/src/models/order.rs#L47). Therefore, instead of implementing `rolling_hash` again, all you need to do is implement Serializable for StandingOrder.

To deal with potentially missing accounts, I'd just create "empty" StandingOrder structs in the Vector for them. This way you can reuse the vector when you try to extract all the batch indexes later (instead of rewriting this relatively complex `.iter.position.map.map` again).

driver/src/models/mod.rs
```diff
@@ -20,6 +20,10 @@ pub trait RollingHashable {
     fn rolling_hash(&self, nonce: i32) -> H256;
 }
 
+pub trait ConcatingHashable {
```
fleupold: *Concat**enat**ing*

driver/src/models/standing_order.rs
```diff
@@ -17,6 +21,24 @@ impl StandingOrder {
     }
 }
 
+impl ConcatingHashable for Vec<StandingOrder> {
+    fn concating_hash(&self, init_hash: H256) -> H256 {
+       let mut hasher = Sha256::new();
+        hasher.input(init_hash);
+        for i in 0..models::NUM_RESERVED_ACCOUNTS {
```
fleupold: Same argument about efficiency as in the other PR applies. I'm starting to think it would be easier if the db_interface always returned a vector of 50 elements with 0 orders for the ones that don't have any standing orders at the moment.

But that could also be done in another PR.

driver/src/order_driver.rs
```diff
@@ -49,6 +48,8 @@ pub fn run_order_listener<D, C>(
 
             let standing_order_indexes = batch_index_from_standing_orders(&standing_orders);
             let order_hash_from_contract = contract.calculate_order_hash(slot, standing_order_indexes.clone())?;
```
fleupold: I'd rename these variables to make the difference between the above `order_hash` and `contract_order_hash` more clear.

I don't really see what the difference between `contract_order_hash` and `order_hash_from_contract ` is.

driver/src/util.rs
fleupold: Does LowerHex representation instead of casting into a json string not work here?

driver/src/models/standing_order.rs
josojo: Made an issue

driver/src/util.rs
josojo: worked well. For some reason, it got lost here.

driver/src/models/mod.rs
```diff
@@ -20,6 +20,10 @@ pub trait RollingHashable {
     fn rolling_hash(&self, nonce: i32) -> H256;
 }
 
+pub trait ConcatenatingHashable {
```
bh2smith: nit: the "ing" in this term seems awkward. I think it would be more appropriate to use present tense "ConcatenateHashable"

driver/src/models/mod.rs
```diff
@@ -20,6 +20,10 @@ pub trait RollingHashable {
...
+    fn concating_hash(&self, init_hash: H256) -> H256;
```
bh2smith: here too; although this is somewhat different. My best guess here as a suggestion would be `concatenate_hash` or simply `concat_hash`.

driver/src/models/mod.rs
bh2smith: or just "Concatenate"

driver/src/models/mod.rs
fleupold: The goal of the term is to describe how the items are hashed together. In the rolling hashable trait this happens as `H(...H(H(i1), i2)...i_n)` whereas with concatenation it happens as `H(i1, ..., i_n)`

No strong feeling about removing the `ing` but then we should also adjust Rolling hashable. Might fit better to `RootHashable` which doesn't have an ing either.

fleupold:
I can get behind ListHashable (then RootHashable should maybe become TreeHashable). But then also RollHashable would be on a list as well, so it wouldn't be clear from the name how the two are different.

bh2smith:
is there any reason for choosing `ListHashable` over `HashableList`? Perhaps this is a naming convention for _Traits_ that I am yet to become comfortable/familiar with.

fleupold:
The main purpose of this Trait is to say it's Hashable. And then I wanted to add a qualifier that describes the different ways to hash a datatype (e.g. into a merkle root, rolling each elements hash to the next, concatenating all elements and hash the resulting blob).

josojo:
Thanks for discussing the names. Sorry, I was lagging energy yesterday evening.

I interpret the thumbs up as an agreement and will merge it.

Issue #153, opened 2019-06-20
[test non-deterministic]


josojo:
The test e2e-tests-auctions sometimes times out.
```

82.22s$ sh ./test/e2e-tests-auction.sh
MongoDB shell version v4.0.7
connecting to: mongodb://127.0.0.1:27017/dfusion2?gssapiServiceName=mongodb
MongoDB server version: 4.0.10
0
Using network 'development'.
Deposit successful: Slot 1 - Index 0 - Hash 0x105b8eeca0c4f7c035ad1057ae2115eacdbd1c956ea0284c89f6bb7c56cb9104
Using network 'development'.
Deposit successful: Slot 1 - Index 1 - Hash 0x1a3a169be7da1d8fcdf64bb5dca03ee15fe13dfc5b31d6c681ba811f1a384c87
Using network 'development'.
Deposit successful: Slot 1 - Index 2 - Hash 0xe17e3d0a3e3cb07c32145ac0df860dbf7ff1a0e9a20238788a09003fedfc4db8
Using network 'development'.
Deposit successful: Slot 1 - Index 3 - Hash 0x882e6011638ddc54cde9632ad264f79070f721ba49630d328c2f3b4844629339
Using network 'development'.
Deposit successful: Slot 1 - Index 4 - Hash 0xc0e471682b5d2fcd757d5941242dea5a1dd300c1b7a98273722248977d2b7fd2
Using network 'development'.
Deposit successful: Slot 1 - Index 5 - Hash 0x37255d9e4e537066de7c706ac93af4f37f6c9ffa9fbacfc44d5b4039352bafeb
Using network 'development'.
Limit Sell Order successful: Auction 0 - Index 0 - Hash 0x94b5f073930872b0b24d79c4408df094c6cb6721cc8346dcf568179e4a7579bd
Using network 'development'.
Limit Sell Order successful: Auction 0 - Index 1 - Hash 0x4a4310835874fab3dfcf8db28d82830fa2048c39480d96879b49aeda29327433
Using network 'development'.
Limit Sell Order successful: Auction 0 - Index 2 - Hash 0x6af9114cc601525531aff894aed96389106c2bc0997d8993e1a5e72e034b6edc
Using network 'development'.
Limit Sell Order successful: Auction 0 - Index 3 - Hash 0x80e17ce1f17855ee0e194862a46a004a2991f3bc7b2cfa9368090a1d218eec64
Using network 'development'.
[...]

PR #154, opened 2019-06-21
Increase number of retries in e2e tests; Prevent failing e2e tests because of time delays

josojo:
Tires to mitigate the issue #153


josojo:
I hadn't had issues for quite some time. Hence, I will close it

Issue #152, opened 2019-06-20
[driver] Change db_query for better efficiency.


josojo:
Felix comment of PR:
 I'm starting to think it would be easier if the db_interface always returned a vector of 50 elements with 0 orders for the ones that don't have any standing orders at the moment.

josojo:
done by the referenced PR

PR #155, opened 2019-06-21
Change get_standing_orders_of_slot to return always all standing_orders, including non-placed ones

josojo:
tackels #152 

testplan:
unit tests only


driver/src/models/standing_order.rs
```diff
@@ -19,19 +19,17 @@ impl StandingOrder {
     pub fn get_orders(&self) -> &Vec<super::Order> {
         &self.orders
     }
+    pub fn get_orders_length(&self) -> usize {
```
bh2smith: could we call this `num_order`? Then using it would be `standing_order.num_orders()`

driver/src/db_interface.rs
```diff
@@ -124,12 +124,21 @@ impl DbInterface for MongoDB {
...
             .map(|d| d.map(models::StandingOrder::from).map_err(DriverError::from))
-            .collect::<Result<Vec<_>, _>>()
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let empty_standing_order = models::StandingOrder::new(
+            0, 0, vec![]
+        );
+        let mut standing_orders = vec![empty_standing_order; models::NUM_RESERVED_ACCOUNTS as usize];
+        for standing_order in non_zero_standing_orders {
+            standing_orders[standing_order.account_id as usize] = standing_order.clone();
```
josojo: I tried to be less clonely(), but did not find a way to do it. I have seen people switching to references or using std::mem::swap. But I think neither of these methods is appropriated here.

driver/src/models/standing_order.rs
josojo: yeah, that is better...

driver/src/db_interface.rs
```diff
@@ -124,12 +124,21 @@ impl DbInterface for MongoDB {
...
-        self.client
+        let non_zero_standing_orders = self.client
             .db(models::DB_NAME)
             .collection("standing_orders")
             .aggregate(pipeline, None)?
             .map(|d| d.map(models::StandingOrder::from).map_err(DriverError::from))
-            .collect::<Result<Vec<_>, _>>()
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let empty_standing_order = models::StandingOrder::new(
```
fleupold: How about implementing `StandingOrder::empty()` ?

driver/src/db_interface.rs
fleupold: Does foreach work (https://docs.rs/foreach/0.3.0/foreach/)?

> ForEach trait and for_each! macro allow you to use iterator inside iteration loop, which is not posible when using for-in loop.

driver/src/models/standing_order.rs
```diff
@@ -19,19 +19,17 @@ impl StandingOrder {
...
        let mut hasher = Sha256::new();
         hasher.input(init_hash);
         for i in 0..models::NUM_RESERVED_ACCOUNTS {
-            hasher.input(self
-                .iter()
-                .position(|o| o.account_id == i as u16) 
-                .map(|k| self[k].get_orders())
-                .map(|o| o.rolling_hash(0))
-                .unwrap_or(H256::zero()));
+            hasher.input(self[i as usize].get_orders().rolling_hash(0));
```
fleupold: nit: 

```rust
for standing_order in self {
   hasher.input(standing_order.get_orders().rolling_hash(0));
}
```

driver/src/db_interface.rs
fleupold: Also, accountId shouldn't be always 0 but increase with each index, shouldn't it?

driver/src/order_driver.rs
```diff
@@ -97,10 +98,9 @@ fn update_balances(state: &mut State, orders: &[Order], solution: &Solution) {
 }
 
 fn batch_index_from_standing_orders(standing_orders: &Vec<models::StandingOrder>) -> Vec<U128> {
-    let mut standing_order_indexes = vec![U128::zero(); models::NUM_RESERVED_ACCOUNTS as usize];
-
+    let mut standing_order_indexes = Vec::<U128>::new();
```
fleupold: nit: this can become a 1-liner

```rust
standing_orders.iter().map(|o| U128::from(o.batch_index)).collect()
```

driver/src/db_interface.rs
josojo: I use for_each now. It makes the code easier to understand. However, I was not able to remove the clone.

driver/src/db_interface.rs
```diff
@@ -124,12 +124,18 @@ impl DbInterface for MongoDB {
...
             .collection("standing_orders")
             .aggregate(pipeline, None)?
             .map(|d| d.map(models::StandingOrder::from).map_err(DriverError::from))
-            .collect::<Result<Vec<_>, _>>()
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut standing_orders = (0..models::NUM_RESERVED_ACCOUNTS)
+            .map(|k| models::StandingOrder::empty(k as u16))
+            .collect::<Vec<models::StandingOrder>>();
+        non_zero_standing_orders.iter().for_each(|k| standing_orders[k.account_id as usize] = k.clone());
```
fleupold: `into_iter` instead of `iter` will make it so that your `k` inside this closure is the value and not the reference. This way you don't need `k.clone()` anymore.

However, accessing the index (`k.account`) when setting an element of `Vec` is considered to happen after the value that is being set is consumed (in this case k, so k.account is accessing a value after it is consumed).
Therefore, you need to close k.account_id before using it (yet this is just cloning a single u32 instead of a full db record).

Here is the code:

```rust
non_zero_standing_orders.into_iter().for_each(|k| {
  let acc_id = k.account_id.clone() as usize;
  standing_orders[acc_id] = k;
});
```

bh2smith:
Tests appear to be failing here because of maximum retires:

```
waited 181 seconds
Before retry #1: sleeping 0.3 seconds
Before retry #2: sleeping 0.6 seconds
Before retry #3: sleeping 1.2 seconds
Before retry #4: sleeping 2.4 seconds
Before retry #5: sleeping 4.8 seconds
Retries exhausted
The command "sh ./test/e2e-tests-standing-order.sh" exited with 1.
```

Maybe it is a good time for us to try and fix this.

josojo:
I figured out that using :
`Result<[models::StandingOrder; models::NUM_RESERVED_ACCOUNTS as usize], DriverError>;`
is not a wise, as we can not differentiate nullified orders ( orders with just zeros as parameters) and non-placed orders at all. But for the rolling hash caclulation, it actually does make a difference, whether we have a non-placed order or a zero order.
Differentiating these two cases would require the introduction of a counter of orders placed. I think this is unjustified complexity. I will revert the latest changes on this PR.

fleupold:
> But for the rolling hash caclulation, it actually does make a difference, whether we have a non-placed order or a zero order.

You are right. I wonder if this is something we need to change in the future as in the snark we will likely fill up unoccupied slots with 0s.

josojo:
> You are right. I wonder if this is something we need to change in the future as in the snark we will likely fill up unoccupied slots with 0s.

Yeah, then many things might be needed to change. Also, our rolling hash of the normal orders would be required to be computed differently. It would always need to be a rolling hash of 500 orders, just x<500 would no longer be an option, as the snark would also always compute 500 rolling hashes.

josojo:
@fleupold I think the PR is ready. Changing the rolling_hash construction should be done in another pr, if we decided to do it.
Could you review it, please?

josojo:
I pushed the code for a potential solution for using arrays for standing orders.

Using arrays, which for types not implementing the Copy trait is a f... pain.
See the 
https://stackoverflow.com/questions/28656387/initialize-a-large-fixed-size-array-with-non-copy-types

I could not implement the macro solution, as rust does not implement a clean way to share a macro over several crates
I could not use the Default trait, as it would be required to be implemented in std::default::Default.
I can not write the 
```
impl Default for [StandingOrder; models::NUM_RESERVED_ACCOUNTS]
```
because only traits defined in the current crate can be implemented for arbitrary types.

Now, I came up with a solution, but it uses unsafe code.

driver/src/db_interface.rs
```diff
@@ -116,20 +115,29 @@ impl DbInterface for MongoDB {
...
+    ) -> Result<[models::StandingOrder; models::NUM_RESERVED_ACCOUNTS], DriverError> {
         let pipeline = vec![
             doc!{"$match" => (doc!{"validFromAuctionId" => (doc!{ "$lte" => slot})})},
             doc!{"$sort" => (doc!{"validFromAuctionId" => -1, "_id" => -1})},
             doc!{"$group" => (doc!{"_id" => "$accountId", "orders" => (doc!{"$first" =>"$orders" }), "batchIndex" => (doc!{"$first" => "$batchIndex" })})}
         ];
 
-        info!("Querying standing_orders: {:?}", pipeline);
-        self.client
+               info!("Querying standing_orders: {:?}", pipeline);
```
fleupold: nit: indentation

driver/src/db_interface.rs
```diff
@@ -116,20 +115,29 @@ impl DbInterface for MongoDB {
     fn get_standing_orders_of_slot(
         &self,
         slot: u32,
-    ) -> Result<Vec<models::StandingOrder>, DriverError> {
+    ) -> Result<[models::StandingOrder; models::NUM_RESERVED_ACCOUNTS], DriverError> {
```
fleupold: 👍

driver/src/models/standing_order.rs
```diff
@@ -19,20 +19,26 @@ impl StandingOrder {
     pub fn get_orders(&self) -> &Vec<super::Order> {
         &self.orders
     }
+    pub fn num_orders(&self) -> usize {
+        self.orders.len()
+    }
+    pub fn empty(account_id: u16) -> StandingOrder {
+        models::StandingOrder::new(account_id, 0, vec![])
+    }
+    pub fn initialize_array(&self) -> [StandingOrder; models::NUM_RESERVED_ACCOUNTS] {
```
fleupold: I liked the idea of having a macro that can be used for all kinds of types more than having `unsafe` specifically for StandingOrder, but I guess it's not a big issue. 

https://github.com/JoshMcguigan/arr_macro is a crate that has this feature claiming 100% safe rust code.

driver/src/db_interface.rs
bh2smith: I guess clippy doesn't catch these things....

driver/src/db_interface.rs
```diff
@@ -116,24 +115,24 @@ impl DbInterface for MongoDB {
...
             .collection("standing_orders")
             .aggregate(pipeline, None)?
             .map(|d| d.map(models::StandingOrder::from).map_err(DriverError::from))
             .collect::<Result<Vec<_>, _>>()?;
 
-        let mut standing_orders = (0..models::NUM_RESERVED_ACCOUNTS)
-            .map(|k| models::StandingOrder::empty(u16::from(k)))
-            .collect::<Vec<models::StandingOrder>>();
+        let mut standing_orders = models::StandingOrder::empty(0).initialize_array();
+        (0..models::NUM_RESERVED_ACCOUNTS)
```
bh2smith: I am not sure I understand this line here... Does it connect to the previous line? If so, then this appears to be an indentation issue, but if not what is it?

driver/src/models/standing_order.rs
bh2smith: Could you elaborate here on the meaning of `unsafe`?

driver/src/models/mod.rs
```diff
@@ -1,6 +1,7 @@
 pub mod state;
 pub mod flux;
 pub mod order;
+#[macro_use]
```
bh2smith: This appears out of nowhere? I also don't see any changes in this file that relate to the inclusion of this line. Was this intentional?

driver/src/models/standing_order.rs
```diff
@@ -25,9 +25,16 @@ impl StandingOrder {
     pub fn empty(account_id: u16) -> StandingOrder {
         models::StandingOrder::new(account_id, 0, vec![])
     }
+    pub fn initialize_array(&self) -> [StandingOrder; models::NUM_RESERVED_ACCOUNTS] {
+        let mut array: [StandingOrder; models::NUM_RESERVED_ACCOUNTS] = unsafe { std::mem::uninitialized() };
```
bh2smith: This seems... somehow unsafe.

driver/src/models/standing_order.rs
bh2smith: Never mind, I will read about it here --> [Unsafe Rust](https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html)

josojo: Well, generally, I would agree. But I am not so sure whether I am happy about pulling another package in there with an 0.* release. If we need it this initialization in two places, I would probably also use the package, but for just one place, I would stick with the unsafe code

I think having this small unsafe code is not much of an issue. Unsafe just means that the compiler can't run all the usual checks.

driver/src/db_interface.rs
josojo: It just creates an iterator for the numbers 0 to NUM_RESEVER_ACCOUNTS. Yeah it looks a little bit unsual

driver/src/models/mod.rs
josojo: How did you find this one? I had already pushed the cleaning commit a long time before your comment,

driver/src/models/standing_order.rs
fleupold: I don't think the version number is a good argument as they are completely arbitrary and have very little meaning.

> I think having this small unsafe code is not much of an issue.

It can easily be an argument next time that there is already a place with unsafe code so it's fine to add another (cf. [The Broken Window Theory](https://www.freecodecamp.org/news/pragmatic-programmer-broken-windows-6916998eecbe/))

I don't insist on changing it but wanted to point this out.

bh2smith: Cute. I enjoyed reading about the _Broken Window Theory_

docker/driver/base/Dockerfile
```diff
@@ -4,8 +4,8 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends curl ssh git libc-dev gcc openssl pkg-config libssl-dev \
     && rm -rf /var/lib/apt/lists/*
 
-#Install Rust
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+#Install Rust with version 1.36
```
fleupold: How is this enforced? Doesn't it always install the newest rust version?

.travis.yml
```diff
@@ -11,6 +11,8 @@ matrix:
         - rustup component add clippy
       script:
         - cd driver/
+        - ls /home/travis/.cargo/registry/src/
```
fleupold: this doesn't seem to do anything.

.travis.yml
```diff
@@ -11,6 +11,8 @@ matrix:
...
+        - echo "cognitive-complexity-threshold = 30" > '/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/mongodb-0.3.12/clippy.toml'
```
fleupold: Why is this needed?

docker/driver/base/Dockerfile
josojo: Yes it does. I tried to trigger a rebuild and only later realized that I have to push the docker image manully.
I will revert it

.travis.yml
josojo: was just for testing... sorry

josojo:
I could not find a solution for the clippy issue:
https://github.com/rust-lang/rust-clippy/issues/3874

`--cap-lints allow` would not work, as we want warnings also to trigger a failure of test

--- locally I can just run:
`
        - echo "cognitive-complexity-threshold = 30" > ~/.cargo/registry/src/github.com-1ecc6299db9ec823/mongodb-0.3.12/clippy.toml  
`
and everything works. But this is of course not a good solution.
Also, because the file is created in the moment of running clippy, this does not work in Travis.

Does anyone have a suggestion?

.travis.yml
```diff
@@ -11,7 +11,7 @@ matrix:
         - rustup component add clippy
       script:
         - cd driver/
-        - cargo clippy -- -D warnings
+        - cargo clippy -- cap-lints  # -- clippy has an issue: https://github.com/rust-lang/rust-clippy/issues/3874
```
bh2smith: I saw that one of the suggestions in the thread here was to use `--cap-lints`. Could you mentioned a little more about what it does?

bh2smith:
Hey @josojo - It appears there is a problem compiling a couple dependencies? Is this because of the rust version or clippy?

```
1.83s$ cargo clippy -- cap-lints
    Checking mongodb v0.3.12
    Checking tokio-uds v0.1.7
error: multiple input filenames provided (first two filenames are `/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-uds-0.1.7/src/lib.rs` and `cap-lints`)
error: Could not compile `tokio-uds`.
warning: build failed, waiting for other jobs to finish...
error: multiple input filenames provided (first two filenames are `/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/mongodb-0.3.12/src/lib.rs` and `cap-lints`)
error: Could not compile `mongodb`.
To learn more, run the command again with --verbose.
The command "cargo clippy -- cap-lints" exited with 101.
```

.travis.yml
josojo: I could really not find any good documentation of that.

If Clippy is linting all the code, it also needs to know the code of the dependencies, as it needs to infer some types. That is why Clippy is looking into the mongodb library, but in there the clippy.toml file is not valid with the new rust version. This is causing the issue. Cap-lints is supposed to cut off the linting so that especially we no longer have to compile the libraries. 

But obviously, it is not working and I can not find a solution right now.

fleupold: They pushed a fix to the mongodb repo yesterday: https://github.com/mongodb-labs/mongo-rust-driver-prototype/commit/9203ec3216976cfbfddb07856b25d061965810af

Pinning the dependency to that commit should fix the issue.

josojo: NICE!

Thanks a lot for unblocking

bh2smith:
Just a question here: 
Does pinning to a specific commit of a repo mean that we should keep an eye open for the next release which includes these changes?

josojo:
Yes, totally.

PR #156, opened 2019-06-24
[Easy] adding dyn infront of trait types

fleupold:
Removing a bunch of 

```
warning: trait objects without an explicit `dyn` are deprecated
```

on the newest rust version.

## Test plan
Compile without warnings.

### DIFF
--- a/driver/src/bin/main.rs
+++ b/driver/src/bin/main.rs
@@ -20,7 +20,7 @@ fn main() {
 
     let solver_env_var = env::var("LINEAR_OPTIMIZATION_SOLVER")
         .unwrap_or_else(|_| "0".to_string());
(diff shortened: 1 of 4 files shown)


fleupold:
The issue you linked to is completely different. It has something to do with `cognitive-complexity-threshold` and not with dyn types.

fleupold:
This PR fixes compiler warnings on the most recent rust nightly (regardless of clippy). To reproduce the warnings do:

```
git checkout master
rustup update
cargo +nightly build
```

fleupold:
> Provided these changes don't require upgrading rust version, then this is fine.

It does not require an update. The current version of rust doesn't warn about these issues. Only the newest nightly warns about them, but it's probably still good to fix, since stable will start warning soon.

Issue #22, opened 2019-01-31
[Driver] linter in travis


josojo:
include a linter in the travis build?

PR #158, opened 2019-06-24
Clippy - Rust Linting 

### LINKED ISSUE #63: Lint all code
Prepare and include `.pylintrc` file and include as part of Travis build scripts. Sharing pre-commit hooks amongst the team will be beneficial.

bh2smith:
Closes #63 and #22. Include strict linting (for the purpose of improving our rust)

**Test Plan**
See that travis passes.

### DIFF
--- a/driver/src/bin/main.rs
+++ b/driver/src/bin/main.rs
@@ -27,7 +27,7 @@ fn main() {
     };
     loop {
         // Start driver_components
-        run_driver_components(&db_instance, &contract, &mut price_finder);
+        run_driver_components(&db_instance, &contract, &mut *price_finder);
         thread::sleep(Duration::from_secs(5));
     }
 }
--- a/driver/src/models/standing_order.rs
+++ b/driver/src/models/standing_order.rs
@@ -28,10 +28,10 @@ impl ConcatenatingHashable for Vec<StandingOrder> {
         for i in 0..models::NUM_RESERVED_ACCOUNTS {
             hasher.input(self
                 .iter()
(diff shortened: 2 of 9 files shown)


driver/src/bin/main.rs
```diff
@@ -27,7 +27,7 @@ fn main() {
     };
     loop {
         // Start driver_components
-        run_driver_components(&db_instance, &contract, &mut price_finder);
+        run_driver_components(&db_instance, &contract, &mut *price_finder);
```
fleupold: It's a bit ugly to reference and dereference in the same line, but I also don't know a better way of doing this.

driver/src/models/standing_order.rs
```diff
@@ -28,10 +28,10 @@ impl ConcatenatingHashable for Vec<StandingOrder> {
         for i in 0..models::NUM_RESERVED_ACCOUNTS {
             hasher.input(self
                 .iter()
-                .position(|o| o.account_id == i as u16) 
+                .position(|o| o.account_id == u16::from(i))
```
fleupold: This one is a really good find. Turns out that using as u16 would also work if i was e.g. a u128 (and then just cut off 112 bits)

bh2smith: Yes, from what the error message was saying, this was unsafe because it would truncate the number.

driver/src/bin/main.rs
bh2smith: This was what the warning suggested as a solution. I also don't know another way.

PR #160, opened 2019-06-25
[Hooks] Pre-Commit and Pre-Push

bh2smith:
_pre-commit_ runs Python (`pylint`) and Rust (`clippy`) Linter

_pre-push_ runs unit tests for rust and python

**User Guide:**

```bash
[optional] rm .git/hooks/pre-commit
cp hooks/pre-commit .git/hooks/ && chmod +x .git/hooks/pre-commit
[optional] rm .git/hooks/pre-push
cp hooks/pre-push .git/hooks/ && chmod +x .git/hooks/pre-push
```

Notice that we are not using [cargo-husky](https://github.com/rhysd/cargo-husky) for the time being because this is not (yet) a pure rust project. Furthermore, this project is not currently ready for use of customized features (such as linting on pre-commit). Basically, the pre-commit hook checks all files rather than the diff. So, at the moment, our hooks have the upper hand.


hooks/pre-push
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -e
+
+echo '+cargo test'
```
fleupold: Is there a way to only run this if at least one .rs file was touched since the last push?

hooks/pre-commit
```diff
@@ -0,0 +1,52 @@
+#!/bin/sh
+
+STAGED_PY_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep ".py$")
+STAGED_RS_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep ".rs$")
+
+
+if [[ ("$STAGED_PY_FILES" = "") && ("$STAGED_RS_FILES" = "") ]]; then
```
fleupold: Can we split this into only running the rust linter only when rust files were touched and pylint only when python was touched.

hooks/pre-push
bh2smith: Yes sure. Will do tomorrow.

bh2smith: Unfortunately not, I have not been able to run `clippy` on specific files.

hooks/pre-commit
bh2smith: The linter already only runs on python files that are were changed as part of the commit. Unfortunately, I have not been able to run clippy on individual files (so certainly not only staged files). Best I can do so far is to avoid running `clippy` if no rust files were touched. HOw does that sound?

hooks/pre-push
bh2smith: My proposed solution here is to get the file names from the diff with master and only run cargo test when there are rust changes and only run pytest when there are python changes. Keeping in mind that this is no "smart-script" so the branch diff check will always be in comparison with master. Also not that the `pre-push` hook is not meant to be necessary (only intended to make the dev's life easier).

hooks/pre-commit
bh2smith: Check out the new and improved pre-push!

```
➜  dex-services git:(hooks) git push
Skipping cargo test; Branch does not differ from master in a rusty way!
Skipping pytest; Branch does not differ from master in a pythonic way!
PUSH SUCCEEDED
```

driver/Cargo.toml
```diff
@@ -26,6 +26,4 @@ log = "0.4.6"
 simple_logger = "1.2.0"
 
 [dev-dependencies]
-mock-it = "0.2.0"
-
-
+mock-it = "0.2.0"
```
bh2smith:
```suggestion
mock-it = "0.2.0"

```

hooks/pre-push
bh2smith: Check out the new and improved pre-push!

```
➜  dex-services git:(hooks) git push
Skipping cargo test; Branch does not differ from master in a rusty way!
Skipping pytest; Branch does not differ from master in a pythonic way!
PUSH SUCCEEDED
```

hooks/pre-commit
josojo: Yes, I think this is okay. Especially, since clippy seems to have its own cache and is not running on all files.

fleupold: nit: I’d suggest to remove this if/else branch altogether now since it feels unnecessary (covered by if and loop below)

bh2smith: Yes, I could do that I just thought it would break earlier if we left it here... I have no personal preference.

PR #162, opened 2019-06-27
Independent End to End Testing

### LINKED ISSUE #161: e2e-tests should not be dependent
Currently, the e2e-tests-auction, e2e-tests-standing-order, etc have to be run in a specific order. We should change this by using reverts in ganache-cli.

bh2smith:
Closes #161 

Now each of the e2e-tests can be executed in any order with full independence.

**TestPlan**

Execute the following in any, say alphabetical, order you like

```bash
./test/e2e-tests-auction.sh
./test/e2e-tests-deposit-withdraw.sh
./test/e2e-tests-standing-order.sh
```

Note that you can always check that the database has been reset to its "contract initialization" state after each test.


test/snap-and-setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+cd dex-contracts/
```
fleupold: Could this script and the functionality of revert-cleanup be exposed as functions in a utility script instead of creating multiple files which are automatically executed when sourced?

driver/src/price_finding/naive_solver.rs
```diff
@@ -340,4 +340,41 @@ pub mod tests {
         let res = solver.find_prices(&orders, &state);
         assert_eq!(U256::from(0), res.unwrap().surplus);
     }
+
+    #[test]
+    fn test_e2e_example() {
```
fleupold: Why do we need this test case? Should this be a separate PR?

test/revert-cleanup.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+echo "Revert Ganache to Pre-Setup"
+cd dex-contracts
```
fleupold: I don't really like that this `cd` is matched by a `cd` in another file. It will be hard to reason about which script starts in which folder it is expecting to be and in which folder it will end up. It would be nice if all scripts can expect to be called from some folder and will return to that folder at the end.

test/revert-cleanup.sh
```diff
@@ -0,0 +1,24 @@
...
+truffle exec scripts/ganache/revertSnapshot.js $1
+
+echo "Dropping Deposits"
+mongo dfusion2 --eval "db.deposits.drop()" | grep -w true
```
fleupold: Should these only drop deposits up to a certain index? Otherwise the snapshot and database become out of sync.

bh2smith: Not sure I understand. All the unit tests have been updated to include any necessary deposits for themselves. This script basically resets the DB to its "contract deployment" state. I suppose it could be more refined, the the `drop` should be general enough to handle all test cases.

test/revert-cleanup.sh
bh2smith: I agree. I also never really liked that. I will take you suggestion and try to make that happen!

driver/src/price_finding/naive_solver.rs
bh2smith: I introduced it to figure out what we needed in the e2e-test and figured it made sense to leave here since it came in handy for these changes. However, I discovered only later that I only needed this unit test because I hadn't included the appropriate deposits. I then found it interesting to have an example with zero surplus.

bh2smith: Would be happy to remove this if you insist, but might not bother including it in a separate PR in that case...

test/snap-and-setup.sh
bh2smith: Not sure I understand. Are you referring to the `ganache/makeSnapshot` and `ganache/revertSnapshot`  being in the same file? This was the case at one point, but @josojo mentioned a reason I can't recall about making them in two two different ones. Maybe we can all put some ideas together here and come up with something great! I have been tinkering with your suggestion to do something like this


```js
// make ganache snapshot 
const makeSnapshot = async function(web3Provider=web3) {
  const res = await send("evm_snapshot", web3Provider)
  return res
}

const revertSnapshot = async function(snapID, web3Provider=web3) {
  await send("evm_revert", [snapID], web3Provider)
}
```

bh2smith:
@fleupold  I agree that the shell scripts are not the most ideal and would like to work toward something more robust. 

> Would running docker-compose up/down in between test cases instead of resetting ganache manually be much slower?

I don't think that starting and stopping the containers is the right way to go and my first thoughts are that it would certainly not be a faster approach. Of course this is jut a gut-feeling.

fleupold:
> I don't think that starting and stopping the containers is the right way to go 

Why is that? To me it feels like resetting the environment would be the ultimate way of ensuring independence. Otherwise, you need to ensure that all components (database, ganache, etc.) are reset, which requires more thought.

I'd be interested to see the time it would take to docker up/down in between.

josojo:
Ben and I compared the run times with docker-container resets and just resetting the database and ganache. To sum it up the container resetting was just taking 5 seconds longer (144s compared to 139), so Ben and I decided to go for it.

test/revert-cleanup.sh
bh2smith: this is removed - some of the lost, but useful commands have been moved into the wiki.

bh2smith:
> From a usability perspective I'd want to start my docker environment and be able to execute any of the e2e tests (without it resetting the environment again every time at the beginning).

The purpose of this PR is to make the e2e-tests independent, if each test doesn't ensure that it is beginning from scratch, there is no way we can do this.

fleupold:
> If each test doesn't ensure that ganache and mongo are starting from scratch, there is no way we can do this

Each test can assume that it is starting on a clean state and thus will be independent from one another.  When running the test on travis, travis would make sure to reset the docker container between each step.
If we want to keep all logic inside the e2e test, I'd suggest rather than stopping a potentially existing container we assume nothing is running at all, start a container and shut it down at the end.

josojo:
>Each test can assume that it is starting on a clean state and thus will be independent from one another. 

I was under the assumption that during development we want to run e2e tests, restart docker, run e2e test...so restarting everything within the e2e might be also a good solution. But now, I am reevaluating it again and yes, usually, during development we want to also restart other containers as the driver and would anyway have to run docker-compose down and up...

Let's change it. I will put the revert-container script into the travis instead of the e2e tests itself

bh2smith:
> If we want to keep all logic inside the e2e test, I'd suggest rather than stopping a potentially existing container we assume nothing is running at all, start a container and shut it down at the end.

I could get into this idea.

josojo:
> > If we want to keep all logic inside the e2e test, I'd suggest rather than stopping a potentially existing container we assume nothing is running at all, start a container and shut it down at the end.
> 
> I could get into this idea.

Then testing will take even longer... That is unnecessary waiting time for me.

bh2smith:
> Then testing will take even longer... That is unnecessary waiting time for me.


Right, so then we leave it as it currently is which is to run the restart scripts between each test. Is good everyone @fleupold & @josojo ?

josojo:
> > Then testing will take even longer... That is unnecessary waiting time for me.
> 
> Right, so then we leave it as it currently is which is to run the restart scripts between each test. Is good everyone @fleupold & @josojo ?

yes!

test/restart-containers.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+docker-compose kill mongo
+docker-compose rm -f mongo
```
fleupold: We should be able to use `docker-compose rm -fs` instead of kill and rm

bh2smith:
This was fun 🏒

PR #163, opened 2019-07-03
[Easy] Travis to fail whenever any test-script fails

bh2smith:
Test Plan - see that Travis build is successful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,9 @@ matrix:
         # driver/listener require compiled contracts
         - cd dex-contracts && truffle compile && cd -
         - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
+        - cd dex-contracts && retry truffle migrate && cd -
       script:
         - set -e # make this script fail as soon as any individual command fail
-        - sh ./test/restart-containers.sh
         - sh ./test/e2e-tests-deposit-withdraw.sh
         - sh ./test/restart-containers.sh
         - sh ./test/e2e-tests-auction.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ matrix:
         - cd dex-contracts && truffle compile && cd -
         - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
       script:
+        - set -e # make this script fail as soon as any individual command fail
         - cd dex-contracts && retry truffle migrate && cd -
         - sh ./test/e2e-tests-deposit-withdraw.sh
         - sh ./test/e2e-tests-auction.sh
--- a/test/e2e-tests-deposit-withdraw.sh
+++ b/test/e2e-tests-deposit-withdraw.sh
@@ -15,7 +15,7 @@ truffle exec scripts/deposit.js 2 2 18
 truffle exec scripts/wait_seconds.js 181
 
 # Expect that driver has processed deposit slot and ensure updated balances are as expected
-EXPECTED_HASH="73815c173218e6025f7cb12d0add44354c4671e261a34a360943007ff6ac7af5"
+EXPECTED_HASH="43815c173218e6025f7cb12d0add44354c4671e261a34a360943007ff6ac7af5"
 retry -t 5 "truffle exec scripts/invokeViewFunction.js 'getCurrentStateRoot' | grep ${EXPECTED_HASH}"
 retry -t 5 "mongo dfusion2 --eval \"db.accounts.findOne({'stateHash': '${EXPECTED_HASH}'}).balances[62]\" | grep -w 18000000000000000000"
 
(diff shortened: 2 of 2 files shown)


josojo:
Isn't the set -e part of all scripts, and so we don't need it? I am unsure here...

.travis.yml
```diff
@@ -41,6 +41,7 @@ matrix:
         - cd dex-contracts && truffle compile && cd -
         - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
       script:
+        - set -e # make this script fail as soon as any individual command fail
```
fleupold: Can you test this by changing the expected hash in one of the first tests to some wrong value and check if travis stops after that test instead of running through all test cases.

.travis.yml
```diff
@@ -41,6 +41,7 @@ matrix:
...
         - sh ./test/restart-containers.sh
```
fleupold: Unrelated, but do we need to restart the container before the very first test?

.travis.yml
bh2smith: Sure thing! Just a moment

.travis.yml
bh2smith: Yes, because the restart script has the `truffle migrate` in it.

fleupold: Can we make this part of the before script? It feels a bit 🤡 to start a container just to restart it right afterwards.

.travis.yml
bh2smith: [Here](https://github.com/gnosis/dex-services/pull/163/commits/07fb46a242d6c7b21d39bd8b18de892e57c89d36) we are.

.travis.yml
bh2smith: Ya, sure... I will change it here and see how you like it.

.travis.yml
bh2smith: Build failed as expected. [Check it out](https://travis-ci.org/gnosis/dex-services/builds/554130831?utm_source=github_status&utm_medium=notification)

.travis.yml
bh2smith: Also, we could consider taking the truffle migrate out of the restart. here are a few options.

A: I think this looks stupid
```
      install:
        - git submodule update --init
        - npm install -C dex-contracts
        # driver/listener require compiled contracts
        - cd dex-contracts && truffle compile && cd -
        - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
*       - sh ./test/restart-containers.sh
      script:
        - set -e # make this script fail as soon as any individual command fail
*
        - sh ./test/e2e-tests-deposit-withdraw.sh
        - sh ./test/restart-containers.sh
        - sh ./test/e2e-tests-auction.sh
        - sh ./test/restart-containers.sh
        - sh ./test/e2e-tests-standing-order.sh
```


B: This seems like a reasonable option
```
      install:
        - git submodule update --init
        - npm install -C dex-contracts
        # driver/listener require compiled contracts
        - cd dex-contracts && truffle compile && cd -
        - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
      script:
        - set -e # make this script fail as soon as any individual command fail
*       - cd dex-contracts && retry truffle migrate && cd -
        - sh ./test/e2e-tests-deposit-withdraw.sh
        - sh ./test/restart-containers.sh
        - sh ./test/e2e-tests-auction.sh
        - sh ./test/restart-containers.sh
        - sh ./test/e2e-tests-standing-order.sh
```



C: We remove the last line from `restart-containers` (which is `cd dex-contracts && retry truffle migrate && cd -`) and do the following:
```
      install:
        - git submodule update --init
        - npm install -C dex-contracts
        # driver/listener require compiled contracts
        - cd dex-contracts && truffle compile && cd -
        - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
      script:
        - set -e # make this script fail as soon as any individual command fail
*       - cd dex-contracts && retry truffle migrate && cd -
        - sh ./test/e2e-tests-deposit-withdraw.sh
        - sh ./test/restart-containers.sh
*       - cd dex-contracts && retry truffle migrate && cd -
        - sh ./test/e2e-tests-auction.sh
        - sh ./test/restart-containers.sh
*       - cd dex-contracts && retry truffle migrate && cd -
        - sh ./test/e2e-tests-standing-order.sh
```

I think option C ruins the purpose of making the tests independent.

josojo: I like B the best

bh2smith: Have decided to go with a variant of option `B`

bh2smith:
@josojo 

> Isn't the set -e part of all scripts, and so we don't need it? I am unsure here...

Yes, each script has one of those, but I noticed that the travis continues to finish the `script` commands even when one of them fails. So `set -e` exits script on failed command, but we also want to stop the collection of scripts being executed by putting this in the travis file as well. As a simple example, check this out...


```bash
# file.a
set -e
echo "hello - A"
some-failed-command
echo "Goodbye A"

# b.file 
set -e
echo "hello - B"
some-failed-command
echo "Goodbye - B"

# parent.file 
./a.file
./b.file

# better.parent 
set -e
./a.file
./b.file
```

The output is as follows:

```bash
./parent.file 
hello - A
./a.file: line 3: some-failed-command: command not found
hello - B
./b.file: line 3: some-failed-command: command not found
```

```bash
./better.parent 
hello - A
./a.file: line 3: some-failed-command: command not found
```

and the better parent is the desired outcome of this PR.

PR #165, opened 2019-07-15
David-Fix: README & .GITIGNORE

W3stside:
1. `truffle compile` will use global install - `npx truffle compile` uses local version correctly (my install was failing)
2. added `.history` to `.gitignore` for users using it in VS

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ git clone git@github.com:gnosis/dex-services.git
 cd dex-services
 git submodule update --init
 cd dex-contracts 
-npm install && truffle compile 
+npm install && npx truffle compile 
 cd ../
 docker-compose up
 ```
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ venv.bak/
 .idea
 .vscode
 .DS_Store
+
+# VS
+.history


README.md
```diff
@@ -46,7 +46,7 @@ git clone git@github.com:gnosis/dex-services.git
 cd dex-services
 git submodule update --init
 cd dex-contracts 
-npm install && truffle compile 
+npm install && npx truffle compile
```
fleupold: This assumes people are using npx (I am not). I'd argue people who use npx know how to execute this command, whereas people who don't might think they are required to install it.

W3stside:
```
By default, npx will check whether <command> exists in $PATH, or in the local project binaries, and execute that. If <command> is not found, it will be installed prior to execution.
```

then i suggest adding a script to run via `npm run truffle compile` in the `package.json` or `npm i -g truffle@<package.json version>` since that command does not run the local package

W3stside: `npm@5.2.0` installs `npx` alongside automatically

fleupold: 👍 You are right that otherwise we need to tell people to install truffle globally. I agree that using npx is the better option then.

PR #180, opened 2019-07-16
[TheGraph] bootstrapping rust listener binary

fleupold:
This PR creates a new rust binary which should eventually replace the existing python listener. It uses [The Graph](https://thegraph.com/) as a dependency.

In this PR we basically just create a graph instance (without any subgraph definition) and connect it to the database (postgres) and the ethereum chain (via ganache). We do not yet start a subgraph instance, which is why the code in the `RuntimeHost` and `LinkResolver` component can be unimplemented for now.

We do however setup travis and docker-compose to make sure the binary is properly integrated into CI from the very beginning.

In the next PRs I will implement the two custom components (local LinkResovler instead of an IPFS based one, Rust RuntimeHost instead of WASM RuntimeHost), before we can start implementing our custom handler code in rust.

### Test Case

`> docker-compose up listener`

See that the process connects successfully

```
listener_1     | Jul 16 17:23:47.747 INFO Connected to Ethereum
listener_1     | Jul 16 17:23:47.903 INFO Connected to Postgres, url: postgresql://dfusion:HIDDEN_PASSWORD@postgres/dfusion, component: Store
listener_1     | Jul 16 17:23:47.929 INFO Waiting for other graph-node instances to finish migrating, component: Store
listener_1     | Jul 16 17:23:47.936 INFO Running migrations, component: Store
listener_1     | Jul 16 17:23:47.965 INFO Migrations finished, component: Store
listener_1     | Jul 16 17:23:47.973 INFO Completed pending Postgres schema migrations, component: Store
listener_1     | Jul 16 17:23:48.021 INFO Starting GraphQL HTTP server at: http://localhost:8000, component: GraphQLServer
listener_1     | Jul 16 17:23:48.027 INFO Starting GraphQL WebSocket server at: ws://localhost:8001, component: SubscriptionServer
listener_1     | Jul 16 17:23:48.038 INFO Downloading latest blocks from Ethereum. This may take a few minutes..., component: BlockIngestor
```


.travis.yml
```diff
@@ -2,18 +2,23 @@ if: (branch = master) OR (type = pull_request) OR (tag IS present)
...
       cache: 
         cargo: true
         directories:
           - driver/target
+          - listener/target
       before_script:
         - rustup component add clippy
       script:
         - cd driver/
         - cargo clippy -- -D warnings
```
josojo: Over here, you get the same issue, which I am having in the other PR:
```
error: error reading Clippy's configuration file `/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/mongodb-0.3.12/clippy.toml`: found deprecated field `cyclomatic-complexity-threshold`. Please use `cognitive-complexity-threshold` instead.
    Checking crossbeam v0.6.0
    Checking tokio-reactor v0.1.8
error: aborting due to previous error
```
I could not yet find a solution for that :(

docker/rust_base/Dockerfile
```diff
@@ -1,14 +1,20 @@
 FROM fleupold/dfusion-solver-base:v2
 
 RUN apt-get update \ 
-    && apt-get install -y --no-install-recommends curl ssh git libc-dev gcc openssl pkg-config libssl-dev \
+    && apt-get install -y --no-install-recommends curl ssh git libc-dev gcc openssl pkg-config libssl-dev libpq-dev \
```
bh2smith: This is a dependency for communicating with PSQL?

Do we think it would make sense to put these dependencies on separate lines so we can include a brief comment about their necessity? Or is this too much more than necessary?

listener/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+name = "listener"
+version = "0.1.0"
+authors = ["Alexander Herrmann <alex@gnosis.io>", "Ben Smith <ben@gnosis.io>", "Felix Leupold <felix@gnosis.io>"]
+edition = "2018"
+
+[[bin]]
+name = "listener"
+path = "src/main.rs"
+
+[dependencies]
```
bh2smith: I see there are a lot of pinned dependencies related to the graph. Do we expect to be permanently pinned to this version (if it suffices)?

listener/src/link_resolver.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct LocalLinkResolver {}
+
+impl LinkResolverTrait for LocalLinkResolver {
+    /// Fetches the link contents as bytes.
+    fn cat(
+        &self,
+        _logger: &Logger,
+        _link: &Link,
+    ) -> Box<Future<Item = Vec<u8>, Error = failure::Error> + Send> {
+        unimplemented!();
```
bh2smith: This is nifty! I wasn't aware of this macro. 💯

docker/rust_base/Dockerfile
fleupold: I agree with your point/concern.

Initially I installed the packages further below when I also create the empty packet structure. However this required to do another command which is not great for the docker image size.

Maybe we can land on a middle ground where I put them in a separate line in the same command and add a comment about that they are only for the graph listener.

listener/Cargo.toml
fleupold: If appropriate/necessary we can upgrade them all at the same time. Given that they point at a tagged version and not a random commit I'm pretty confident this version is somewhat stable.

docker/rust_base/Dockerfile
bh2smith: Works for me!

listener/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    // Create components for tokio context: multi-threaded runtime, executor
+    // context on the runtime, and Timer handle.
+    //
+    // Configure the runtime to shutdown after a panic.
+    let runtime: Arc<Mutex<Option<runtime::Runtime>>> = Arc::new(Mutex::new(None));
+    let handler_runtime = runtime.clone();
+    *runtime.lock().unwrap() = Some(
+        runtime::Builder::new()
+            .core_threads(100)
```
bh2smith: How did we decide on this 100?

listener/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+            .panic_handler(move |_| {
+                let runtime = handler_runtime.clone();
+                std::thread::spawn(move || {
+                    if let Some(runtime) = runtime.lock().unwrap().take() {
+                        // Try to cleanly shutdown the runtime, but
+                        // unconditionally exit after a while.
+                        std::thread::spawn(|| {
+                            std::thread::sleep(Duration::from_millis(3000));
```
bh2smith: ditto. Is this a common number of ms to sleep?

listener/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+                            std::process::exit(1);
+                        });
+                        runtime
+                            .shutdown_now()
+                            .wait()
+                            .expect("Failed to shutdown Tokio Runtime");
+                        println!("Runtime cleaned up and shutdown successfully");
```
bh2smith: `println!` - could we make this a log instead!

bh2smith:
```suggestion
                        info!("Runtime cleaned up and shutdown successfully");
```

listener/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+    let http_port = env::var("GRAPHQL_PORT")
+        .expect("Specify GRAPHQL_PORT variable")
+        .parse::<u16>()
+        .expect("Couldn't parse GRAPHQL_PORT variable as u16");
+    let ws_port = env::var("WS_PORT")
+        .expect("Specify WS_PORT variable")
+        .parse::<u16>()
+        .expect("Couldn't parse WS_PORT variable as u16");
+
+    let logger = logger(false);
```
bh2smith: hopefully having the logger declared below the `log!` macro is used doesn't cause any problems.

listener/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+    // Set up Ethereum transport
+    let (transport_event_loop, transport) = Transport::new_rpc(&ethereum_node_url);
+
+    // If we drop the event loop the transport will stop working.
+    // For now it's fine to just leak it.
+    std::mem::forget(transport_event_loop);
+
+    let eth_adapter = Arc::new(graph_datasource_ethereum::EthereumAdapter::new(
+        transport,
+        0,
```
bh2smith: What is this Ethereum adapter parameter? Is it network id?

listener/src/main.rs
fleupold: tbh, most of this main code is bluntly stolen from the main graph node repo https://github.com/graphprotocol/graph-node/blob/v0.13.2/node/src/main.rs#L84

listener/src/main.rs
fleupold: it's the start block

listener/src/main.rs
fleupold: There is no logger available in this scope. I hope having this one off print statement is ok? We could try to instantiate it outside async_main and pass it in but I'm not sure if it's that important. Again, most of this setup glue code is taken from the graph node.

fleupold:
I'm usually in favor of having small readable components. Here I think we have a lot of wiring code, which doesn't really do much logic. Also as you stated most of the statements already do a unique thing (creating, starting or wiring up a component) and the concrete logic lives inside each sub-component.
Therefore, I'm not sure if we can actually gain much by creating more sub-methods

listener/src/main.rs
bh2smith: Good enough... just wanted to point it out. If I recall correctly, the python linter would fail if there were any dangling print statements.

PR #183, opened 2019-07-17
[The Graph] Implementing a local link resolver

fleupold:
This PR implements a local link resolver that can look up specification files required for the subgraph definition from disk. It also creates a basic subgraph definition for our project.

### Test plan

In a first shell: 
```
```

In another shell
```
cd dex-contracts
truffle migrate
```

This should triffer a yet unimplemented handler and cause the process in terminal 1 to panic. Defining a handler trait for handling events will be part of the next PR.


listener/src/main.rs
```diff
@@ -3,6 +3,8 @@ extern crate graph;
 extern crate graph_core;
 extern crate graph_datasource_ethereum;
 extern crate lazy_static;
+#[macro_use]
+extern crate slog;
```
bh2smith: Is this used somewhere? If so, I can't tell.

listener/src/main.rs
```diff
@@ -13,7 +15,11 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use graph::components::forward;
-use graph::prelude::*;
+use graph::prelude::{
+    SubgraphRegistrar as SubgraphRegistrarTrait,
+    SubgraphAssignmentProvider as SubgraphAssignmentProviderTrait,
+    *
```
bh2smith: Looks like we only need `SubgraphName` and `SubgraphDeploymentId` in addition to those already explicitly imported. Do we expect to use other graph preludes later?

listener/subgraph_definition/dfusion
```diff
@@ -0,0 +1,22 @@
...
+      kind: Mapping
+      apiVersion: 0.0.1
+      language: rust
+      entities: []
+      abis: []
+      eventHandlers:
+        - event: Deposit(uint16,uint8,uint128,uint256,uint16)
+          handler: rust
+      file: 
+        /: rustHandler.wasm
```
bh2smith: Which part of our existing code base generates this file? Is this created during docker build? Will this file remain constant for ever onward?

fleupold: At this point this file is manually created. We will need to add a new (event, handler) item pair for every event we want to handle. 
Eventually, we could consider creating this manifest in code rather than reading it from a file but this way allowed us to reuse the most of the existing graph components.

I'd hope that at some point the graph might actually support external rust handlers in the core system.

listener/src/main.rs
fleupold: Sure, I can make dependencies explicit.

listener/src/link_resolver.rs
```diff
@@ -3,18 +3,40 @@ use graph::data::subgraph::Link;
 use graph::prelude::LinkResolver as LinkResolverTrait;
 
 use futures::prelude::*;
+use futures::future::*;
 use slog::Logger;
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::Path;
+
+fn read_file(file: &str) -> Result<Vec<u8>, failure::Error> {
```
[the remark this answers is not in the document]
fleupold: Vec<u8> is just a byte buffer. It's basically the data that has been read from disk (json, yaml or binary in the case of the wasm code).

fleupold:
> Does this mean that the event is actually heard, but then the service only tries to access the abi after catching the event and only then fails?

At the moment we are not accessing the abi at all (and I'm not sure we ever will). We provide the event name which we want to listen to explicitly in the subgraph definition. This name will be hashed and compared with events found in any of the parsed blocks (also filtering on the contract address).

I believe the old event listener used the ABI to look up the arguments and thus compute the correct topic hash. The Graph requires the event name to also list the argument types directly (thus no abi lookup is needed).

If it became necessary, we could create a script that automatically extracts the abi from the truffle build artifacts.

listener/subgraph_definition/dfusion
bh2smith: Perhaps, we could include something in the readme or wiki with instructions on this for the time being. If I remember correctly there is an outdated WIKI article with instructions on how to configure the listener to catch new events. [Here](https://github.com/gnosis/dex-services/wiki/Adding-Event-Listener)

fleupold: Good point. Added a section for the Graph (still with a few TODOs to be filled out throughout the course of this stack). Will eventually remove the Python tutorial once we remove that code.

PR #185, opened 2019-07-18
[The Graph] Implementing a runtime to route events to specific handlers

fleupold:
This PR implements an interface which each event handler (e.g. Deposit, Withdraw, etc.) will implement. It also implements the RustRuntimeHandler which takes the events that are supposed to be processed according to the subgraph definition and forwards them to the specific event handler.

The next PR will implement the first concrete EventHandler (listening to Deposit events).

### Test Plan

In one console run:

```sh
```

In another one run:
```sh
cd dex-contracts
truffle migrate
```

Observe the following print statements (and that the process no longer crashes)
```
INFO Received event, block_hash: ...
WARN Unhandled event with topic 0xba69…11ee, block_hash: ...
```

### DIFF
--- a/listener/src/runtime_host.rs
+++ b/listener/src/runtime_host.rs
@@ -1,6 +1,7 @@
 use failure::Error;
 use futures::future::*;
 use slog::Logger;
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use graph::components::ethereum::{EthereumCall, EthereumBlockTriggerType, EthereumBlock};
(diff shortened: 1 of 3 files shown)


listener/src/runtime_host.rs
```diff
@@ -1,33 +1,48 @@
...
         _data_source: DataSource,
     ) -> Result<Self::Host, Error> {
-        Ok(RustRuntimeHost {})
+        Ok(RustRuntimeHost::new())
+    }
+}
+
+#[derive(Debug)]
+pub struct RustRuntimeHost {
+    handlers: HashMap<H256, Box<EventHandler>>
```
josojo: I really like this rustic polymorphism structure. ❤️

PR #186, opened 2019-07-19
[Refactoring] Moving models code into reusable core library

fleupold:
This PR moves the models submodule into a separate rust library which can be reused by other components (particularly the Graph). The intention for this new library is to capture all common code that will be shared among the different user facing rust binaries.

### Test plan

CI


josojo:
good PR for code sharing between the different components.

Way cool!

PR #188, opened 2019-07-19
Refactor data types in rust models to make them consistent with SC

fleupold:
I noticed that the datatypes we use in our rust models and the datatypes used in the events that the smart contract emits are different. This PR attempts at unifying the two data models.

### Test plan
compilation and rust unit tests.


fleupold:
> Not so sure about nonce on rolling hash though.

Yeah, that one should ideally be a H256 as well. However, it's currently used in a slightly weird way when computing the state root hash, which I didn't want to touch in this PR. I assume this will change anyways when we go to 2**24 accounts.

PR #189, opened 2019-07-19
Remove `set -e` in travis script

fleupold:
We still don't see the error logs when a travis build fails (e.g. https://travis-ci.org/gnosis/dex-services/jobs/560457456).

This is critical to find the reason for our flakyness. I believe that the reason could be the `set -e` command inside the script, which I assume makes travis immediately stop the build and all future commands.

Running an expectedly failing build to verify.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,6 @@ matrix:
         - docker-compose up -d ganache-cli driver listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
         - cd dex-contracts && retry truffle migrate && cd -
       script:
-        - set -e # make this script fail as soon as any individual command fail
         - sh ./test/e2e-tests-deposit-withdraw.sh
         - sh ./test/restart-containers.sh
         - sh ./test/e2e-tests-auction.sh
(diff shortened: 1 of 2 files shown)


fleupold:
cf. https://travis-ci.org/gnosis/dex-services/jobs/561032642 is showing the docker logs after failure. Thus removing the set -e is necessary to show the logs at the end.

josojo:
Hm...

In your intentionally broken travis report, right after
```
waited 181 seconds
Before retry #1: sleeping 0.3 seconds
Before retry #2: sleeping 0.6 seconds
Before retry #3: sleeping 1.2 seconds
Before retry #4: sleeping 2.4 seconds
Before retry #5: sleeping 4.8 seconds
Retries exhausted
The command "sh ./test/e2e-tests-deposit-withdraw.sh" exited with 1.
```
Why do I only see the truffle logs, but not all the other logs from driver and listener?

fleupold:
> Why do I only see the truffle logs, but not all the other logs from driver and listener?

Check line 1034 (and expand). There you should see all the logs from the driver and the listener.

PR #190, opened 2019-07-22
[The Graph] Deposit Handler

fleupold:
This PR implements the first handler (for deposits). It basically creates a way to convert an ethereum Log into the data model already defined for the driver and another method to convert the data model into an Entity (needed for writing it to the DB).

### Test Plan

Demo Video can be found in dex-execution slack channel.

In one terminal run:

```
```

In another one run:
```sh
cd dex-contracts
truffle migrate
```

Open a browser at `localhost:8000` and type the following query:

```
query {
  deposits {
    id
    amount
    slot
    tokenId
    slotIndex
    accountId
  } 
}
```

You should the newly created deposit.


dfusion_rust_core/src/models/flux.rs
```diff
@@ -36,6 +41,31 @@ impl From<mongodb::ordered::OrderedDocument> for PendingFlux {
...
 
+impl From<Arc<Log>> for PendingFlux {
+    fn from(log: Arc<Log>) -> Self {
+        let mut bytes: Vec<u8> = log.data.0.clone();
+        PendingFlux {
+            account_id: pop_u16_from_log_data(&mut bytes),
+            token_id: pop_u8_from_log_data(&mut bytes),
+            amount: pop_u128_from_log_data(&mut bytes),
+            slot: pop_u256_from_log_data(&mut bytes),
+            slot_index: pop_u16_from_log_data(&mut bytes),
```
bh2smith: I see `pop_u16_from_log_data` twice in here. Is it safe to assume that `pop` literally means pop (like in computer lingo) so that the information is forgotten after it is popped and the second call of this is just taking the next 16 bytes off the front?

dfusion_rust_core/src/models/util.rs
```diff
@@ -0,0 +1,20 @@
+use std::convert::TryInto;
+use web3::types::U256;
+
+pub fn pop_u8_from_log_data(bytes: &mut Vec<u8>) -> u8 {
+    pop_u256_from_log_data(bytes).as_u32().try_into().unwrap()
```
bh2smith: Are we at all worried that popping 256 and only needing 8 wouldn't forget more information than intended?

docker/listener/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM fleupold/dfusion-rust-base:v1
 # Copy requirements and install
 WORKDIR /app/listener
 COPY listener/Cargo.toml ./
+COPY dfusion_rust_core /app/dfusion_rust_core
 RUN cargo build
 
 # Copy smart contract build artefacts
```
bh2smith:
```suggestion
# Copy smart contract build artifacts
```

dfusion_rust_core/src/models/flux.rs
fleupold: Yes, happy to rename to something that makes that more clear.

listener/Cargo.toml
```diff
@@ -16,6 +16,8 @@ graph-store-postgres = { git = "https://github.com/graphprotocol/graph-node", ta
 graph-server-http = { git = "https://github.com/graphprotocol/graph-node", tag = "v0.13.2" }
 graph-server-websocket = { git = "https://github.com/graphprotocol/graph-node", tag = "v0.13.2" }
 
+dfusion_core = { path = "../dfusion_rust_core" }
```
bh2smith: I feel like these kind of imports would have already been included? Is this part of a different PR?

dfusion_rust_core/src/models/util.rs
fleupold: For some reason the data is not compressed, so even though we are only emitting a e.g. u16 from the SC, the event log reserves 32 bytes (u256) for the value. 
These methods are so that fact the developer doesn't have to know about that fact (they should only care how big the type is they are reading/popping).

listener/Cargo.toml
fleupold: This is the first PR where the listener actually uses the common library (before I only created it and made the driver use it).

listener/src/event_handler/deposit_handler.rs
```diff
@@ -0,0 +1,44 @@
...
+        &self,
+        logger: Logger,
+        _block: Arc<EthereumBlock>,
+        _transaction: Arc<Transaction>,
+        log: Arc<Log>
+    ) -> Result<Vec<EntityOperation>, Error> {
+        let entity_id = util::entity_id_from_log(&log);
+        let flux = PendingFlux::from(log);
+        
+        info!(logger, "Processing Deposit {:?}", &flux);
```
bh2smith: This is a cool log. Just a question here, is there another (more generic) log statement like 

```
"{} - event received" % event.name
```

listener/src/event_handler/deposit_handler.rs
```diff
@@ -0,0 +1,44 @@
...
+        
+        let mut entity: Entity = flux.into();
+        // We do not care about the ID inside the flux data model,
+        // so we have to set them later.
+        entity.set("id", &entity_id);
```
bh2smith: I am not sure I understand this "entity id" at first glance? Is this for our records or is it coming from the event? I suppose this is the unique id as stored by `postgres`

listener/subgraph_definition/schema.graphql
```diff
@@ -1,3 +1,8 @@
 type Deposit @entity {
     id: ID!
+    accountId: Int!
+    tokenId: Int!
+    amount: BigInt!
```
bh2smith: Safe to assume the graph treats this as a string?

listener/src/event_handler/util.rs
```diff
@@ -0,0 +1,20 @@
+use std::sync::Arc;
+use graph::components::store::EntityKey;
+use web3::types::Log;
+
+use crate::SUBGRAPH_ID;
+
+pub fn entity_id_from_log(log: &Arc<Log>) -> String {
```
bh2smith: Is the entity id the transaction hash?

listener/src/event_handler/deposit_handler.rs
fleupold: Unfortunately we only know the keccak hash of the log (not the reverse plain text name). Therefore I'm expecting each handler to know which plaintext event they are listening to.

listener/src/event_handler/deposit_handler.rs
fleupold: >  I suppose this is the unique id as stored by postgres

Exactly and unlike in mongo db we actual have to create it. At the moment I'm taking the transaction hash and the index of this event inside the transaction, as this should hopefully be unique.

listener/src/event_handler/util.rs
fleupold: it's tx_hash appended by the index of the event inside the block (which should be unique)

listener/subgraph_definition/schema.graphql
fleupold: We don't really care how they treat it on byte level as on a code level we work with the BigDecimal abstraction (but yes, turns out they store it as a string similar to how we do it atm in mongo).

PR #191, opened 2019-07-23
[The Graph] Adding Initialization event listener

fleupold:
This PR adds a new Graph handler for the SnappBase initialisation event. It creates the genesis state and writes it into the graph database.

### Test Plan

In one terminal run:
```
```

In another terminal run:
```
cd dex-contracts
truffle migrate
```

See how the event is being handles in the log statements. To verify go to localhost:8000 and run the following query:

```graphql
query {
  accountStates {
    id
    numTokens 
    balances
  } 
}
```

### DIFF
--- a/dfusion_rust_core/src/models/state.rs
+++ b/dfusion_rust_core/src/models/state.rs
@@ -1,9 +1,13 @@
 use byteorder::{BigEndian, WriteBytesExt};
+use graph::data::store::{Entity, Value};
 use serde_derive::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
-use web3::types::{H256, U256};
+use std::sync::Arc;
+use web3::types::{H256, U256, Log};
 
 use crate::models::{TOKENS, RollingHashable};
+
+use super::util::{pop_u8_from_log_data, pop_u16_from_log_data, pop_h256_from_log_data, to_value};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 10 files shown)


dfusion_rust_core/src/models/state.rs
```diff
@@ -1,10 +1,14 @@
...
+use std::sync::Arc;
+use web3::types::{H256, U256, Log};
 
 use crate::models::{TOKENS, RollingHashable};
 
+use super::util::{pop_u8_from_log_data, pop_u16_from_log_data, pop_h256_from_log_data, to_value};
+
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct State {
```
bh2smith: Maybe its a little too late, but would it make more sense to call this AccountState?

listener/src/event_handler/initialization_handler.rs
```diff
@@ -0,0 +1,34 @@
...
+        log: Arc<Log>
+    ) -> Result<Vec<EntityOperation>, Error> {
+        info!(logger, "Processing SnappBase Initialization Event");
+        let state = State::from(log);
+        let entity: Entity = state.into();
+        
+        Ok(vec![
+            EntityOperation::Set {
+                key: util::entity_key(
+                    "AccountState",
```
bh2smith: I see AccountState is used here. 👍

dfusion_rust_core/src/models/state.rs
fleupold: Yes, will make this a separate PR though.

PR #193, opened 2019-07-24
[TheGraph] applyDepositHandler

fleupold:
Apologies for the large PR, I already extracted the reorg of driver logic into the shared library in a separate PR but it still turned out quite big. I guess I could factor each of the things that's going on here into separate PRs which by themselves wouldn't make a ton of sense. Let me know if you prefer that.

What's going on here?

1. We register the event and event handler
2. We provide conversion methods to parse an Entity (read result from the GraphStore) into the respective data models (State, PendingFlux)
3. We implement the FluxTransitionHandler for deposits (withdraws will be handled later)
4. Unit + E2E tests

### Test Plan

Run the docker-compose in one terminal. Run the `e2e-tests-deposit-withdraw.sh` in another (after `truffle migrate`). At the end run the following query in the graphiql interface (localhost:8000):

```
query {
  accountStates {
    id
    numTokens 
  } 
}
```

You can also query for balances if you are ready to scroll a lot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ matrix:
           - $HOME/.nvm/versions/node/v8.12.0/lib/node_modules
           - dex-contracts/node_modules/
       before_install:
-        - npm install -g truffle
+        - npm install -g truffle node-jq
         - sudo sh -c "curl https://raw.githubusercontent.com/kadwanev/retry/a1b1826bdb0a78189d5c70c858dc676f5133b1d7/retry -o /usr/local/bin/retry && chmod +x /usr/local/bin/retry"
       install:
         - git submodule update --init
--- a/dfusion_rust_core/src/models/flux.rs
+++ b/dfusion_rust_core/src/models/flux.rs
@@ -1,11 +1,11 @@
 use byteorder::{BigEndian, WriteBytesExt};
-use graph::data::store::{Entity};
+use graph::data::store::Entity;
 use serde_derive::{Deserialize, Serialize};
 use std::sync::Arc;
 use web3::types::{H256, U256, Log};
 
 use crate::models::{Serializable, RootHashable, merkleize};
-use super::util::{pop_u8_from_log_data, pop_u16_from_log_data, pop_u128_from_log_data, pop_u256_from_log_data, to_value};
+use super::util::*;
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, Debug, Ord, PartialOrd, Eq)]
 #[serde(rename_all = "camelCase")]
@@ -48,6 +48,18 @@ impl From<Arc<Log>> for PendingFlux {
             amount: pop_u128_from_log_data(&mut bytes),
             slot: pop_u256_from_log_data(&mut bytes),
             slot_index: pop_u16_from_log_data(&mut bytes),
+        }
+    }
+}
+
+impl From<Entity> for PendingFlux {
+    fn from(entity: Entity) -> Self {
+        PendingFlux {
+            account_id: u16::from_entity(&entity, "accountId"),
+            token_id: u8::from_entity(&entity, "tokenId"),
+            amount: u128::from_entity(&entity, "amount"),
+            slot: U256::from_entity(&entity, "slot"),
+            slot_index: u16::from_entity(&entity, "slotIndex"),
         }
     }
 }
@@ -156,7 +168,7 @@ pub mod unit_test {
   }
 
   #[test]
-  fn test_to_entity() {
+  fn test_to_and_from_entity() {
       let flux = PendingFlux {
             account_id: 1,
             token_id: 1,
@@ -172,6 +184,7 @@ pub mod unit_test {
         entity.set("slot", BigDecimal::from(0));
         entity.set("slotIndex", 0);
 
-        assert_eq!(entity, flux.into());
+        assert_eq!(entity, flux.clone().into());
+        assert_eq!(flux, PendingFlux::from(entity));
   }
 }
--- a/dfusion_rust_core/src/models/state.rs
+++ b/dfusion_rust_core/src/models/state.rs
@@ -7,7 +7,7 @@ use web3::types::{H256, U256, Log};
 
 use crate::models::{TOKENS, RollingHashable, PendingFlux};
 
-use super::util::{pop_u8_from_log_data, pop_u16_from_log_data, pop_h256_from_log_data, to_value};
+use super::util::*;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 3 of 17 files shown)


listener/src/event_handler/util.rs
```diff
@@ -11,10 +12,88 @@ pub fn entity_id_from_log(log: &Arc<Log>) -> String {
...
+        order_direction: None,
+        range: EntityRange {
+            first: None,
+            skip: 0
+        }
+    }
+}
+
+#[cfg(test)]
+pub mod test {
```
bh2smith: I don't see any assertions in these tests? Is this normal?

listener/src/runtime_host.rs
```diff
@@ -35,7 +47,7 @@ impl RuntimeHostBuilder for RustRuntimeHostBuilder {
         _subgraph_id: SubgraphDeploymentId,
         _data_source: DataSource,
     ) -> Result<Self::Host, Error> {
-        Ok(RustRuntimeHost::new())
+        Ok(RustRuntimeHost::new(self.store.clone()))
```
bh2smith: What is this `self.store.clone()` doing for us?

listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -0,0 +1,209 @@
...
+            FluxTransitionType::Deposit => {
+                let deposit_query = util::entity_query(
+                    "Deposit", EntityFilter::Equal("slot".to_string(), to_value(&slot))
+                );
+                let deposits = self.store
+                    .find(deposit_query)?
+                    .into_iter()
+                    .map(PendingFlux::from)
+                    .collect::<Vec<PendingFlux>>();
+                account_state.apply_deposits(&deposits);
```
bh2smith: sexy!

dfusion_rust_core/src/models/util.rs
```diff
@@ -32,4 +33,78 @@ pub fn to_value<T: ToString>(value: &T) -> Value {
...
+}
+
+impl EntityParsing for Vec<u128> {
+    fn from_entity(entity: &Entity, field: &str) -> Self {
+        entity
+            .get(field)
+            .and_then(|value| value.clone().as_list())
+            .map(|list| list
+                .into_iter()
+                .map(|value| u128::from_str(
```
josojo: Can we reuse the from_entity code for u128 here?

listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    fn create_state_transition_event(
+        transition_type: FluxTransitionType, 
+        new_state_index: u8, 
+        new_state_root: H256,
+        applied_slot: u8
+    ) -> Arc<Log> {
+        let bytes: Vec<Vec<u8>> = vec![
+            /* transition_type */ vec![0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, transition_type as u8],
```
bh2smith: Is there no cleaner way to construct these arrays?

listener/src/event_handler/flux_transition_handler.rs
bh2smith: However, I can't seem to find `apply_deposits`... I assume it is implemented for state and was done in a different PR.

dfusion_rust_core/src/models/util.rs
fleupold: It would have to be from_value, since the item is a `Vec<Value>`.

.travis.yml
```diff
@@ -31,7 +31,7 @@ matrix:
           - $HOME/.nvm/versions/node/v8.12.0/lib/node_modules
           - dex-contracts/node_modules/
       before_install:
-        - npm install -g truffle
+        - npm install -g truffle node-jq
```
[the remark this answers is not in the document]
fleupold: Yes, will add jq, retry will be gone with the e2e refactoring

dfusion_rust_core/src/models/flux.rs
```diff
@@ -52,6 +52,18 @@ impl From<Arc<Log>> for PendingFlux {
     }
 }
 
+impl From<Entity> for PendingFlux {
```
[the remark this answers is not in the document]
fleupold: This is an individual Flux, not a Batch of Fluxes

dfusion_rust_core/src/models/state.rs
```diff
@@ -7,7 +7,7 @@ use web3::types::{H256, U256, Log};
 
 use crate::models::{TOKENS, RollingHashable, PendingFlux};
 
-use super::util::{pop_u8_from_log_data, pop_u16_from_log_data, pop_h256_from_log_data, to_value};
+use super::util::*;
```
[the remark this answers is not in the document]
fleupold: yes, I almost used all the utils, so I think it's fine importing them all

dfusion_rust_core/src/models/state.rs
```diff
@@ -105,10 +105,21 @@ impl From<Arc<Log>> for State {
     }
 }
 
+impl From<Entity> for State {
```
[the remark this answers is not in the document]
fleupold: The renaming diff depends on this one.

dfusion_rust_core/src/models/state.rs
```diff
@@ -196,11 +207,12 @@ pub mod tests {
         };
         
         let mut entity = Entity::new();
-        entity.set("id", "0x0000000000000000000000000000000000000000000000000000000000000000");
+        entity.set("id", "0000000000000000000000000000000000000000000000000000000000000000");
```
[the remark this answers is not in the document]
fleupold: nope, I had to get rid of it in order to make the reverse parsing possible (H256::from_str expects a no prefixed hex string).

listener/src/runtime_host.rs
fleupold: It makes a copy of the store arc and passes it into the RustRuntimeHost. We need to pass a value rather than a reference since the RuntimeHost might run on a different thread and otherwise multiple of these threads might be accessing the same instance. 
If we just passed by value without clone, we could only ever build one RustRuntimeHost (which might be what we are doing atm but feels like an unnecessary restriction).

listener/src/event_handler/util.rs
fleupold: It doesn't test anything but just provides test data for other threads.

listener/src/event_handler/flux_transition_handler.rs
fleupold: Yes, it was https://github.com/gnosis/dex-services/pull/192

listener/src/event_handler/flux_transition_handler.rs
fleupold: Happy to suggestions. If I assign each vector to it's own variable it become really annoying merging them (it requires one extra line for each variable). I found this flatMap style the most clean while still expressive, but agree it could be better.

PR #198, opened 2019-07-27
[TheGraph] Apply Withdraws

### LINKED ISSUE #171: Apply Withdrawals handler [Rust]

bh2smith:
This depends on #197 and closes #171.

**Test Plan**

Run the `docker-compose` in one terminal. Run the `e2e-tests-deposit-withdraw.sh` in another (after truffle migrate). At the end run the following query in the graphiql interface (localhost:8000):

```
query {
  accountStates {
    id
    numTokens 
  } 
}
```

### DIFF
--- a/dfusion_rust_core/src/models/account_state.rs
+++ b/dfusion_rust_core/src/models/account_state.rs
@@ -1,5 +1,5 @@
 use byteorder::{BigEndian, WriteBytesExt};
-use graph::data::store::{Entity, Value};
+use graph::data::store::Entity;
 use serde_derive::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::sync::Arc;
@@ -11,16 +11,16 @@ use super::util::*;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
-pub struct State {
+pub struct AccountState {
     pub state_hash: H256,
     pub state_index: U256,
     balances: Vec<u128>,
     pub num_tokens: u8,
 }
 
-impl State {
+impl AccountState {
     pub fn new(state_hash: H256, state_index: U256, balances: Vec<u128>, num_tokens: u8) -> Self {
-        State { state_hash, state_index, balances, num_tokens }
+        AccountState { state_hash, state_index, balances, num_tokens }
     }
     fn balance_index(&self, token_id: u8, account_id: u16) -> usize {
         self.num_tokens as usize * account_id as usize  + token_id as usize
@@ -53,7 +53,7 @@ impl State {
     }
 }
 
-impl RollingHashable for State {
+impl RollingHashable for AccountState {
   //Todo: Exchange sha with pederson hash
   fn rolling_hash(&self, nonce: u32) -> H256 {
     let mut hash = vec![0u8; 28];
@@ -72,9 +72,9 @@ impl RollingHashable for State {
   }
 }
 
-impl From<mongodb::ordered::OrderedDocument> for State {
+impl From<mongodb::ordered::OrderedDocument> for AccountState {
     fn from(document: mongodb::ordered::OrderedDocument) -> Self {
-        State {
+        AccountState {
             state_hash: document.get_str("stateHash")
                 .unwrap()
                 .parse::<H256>()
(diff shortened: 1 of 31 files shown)


josojo:
sweet, looks also already great. Looking forward to the unit tests

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -51,6 +51,14 @@ impl AccountState {
         self.state_index = self.state_index.saturating_add(U256::one());
         self.state_hash = self.rolling_hash(self.state_index.low_u32());
     }
+
+    pub fn apply_withdraws(&mut self, withdraws: &[PendingFlux]) {
```
fleupold: Ideally this logic is moved from wherever it already exists in the driver (potentially a separate PR which has to come first).

listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -89,8 +89,25 @@ impl EventHandler for FluxTransitionHandler {
...
-                Ok(vec![])
+                let withdraw_query = util::entity_query(
+                    "Withdraw", EntityFilter::Equal("slot".to_string(), slot.to_value())
+                );
+                let withdraws = self.store
+                    .find(withdraw_query)?
+                    .into_iter()
+                    .map(PendingFlux::from)
+                    .collect::<Vec<PendingFlux>>();
+                account_state.apply_withdraws(&withdraws);
```
fleupold: Maybe we can return the account state from the match statement and have the logic to generate the entity operation vector shared between deposits and withdraws.

bh2smith:
> Would be awesome if this could be based on the branch that implements the basic WithdrawHandler

This is an always has been based on the withdraw handler branch.

dfusion_rust_core/src/models/account_state.rs
bh2smith: I am pretty sure this is the appropriate place in the code for such a function (considering apply deposits is right here in the same file).

listener/src/event_handler/flux_transition_handler.rs
bh2smith: Very good point! I will move this part out and learn how to return stuff from a match!

dfusion_rust_core/src/models/account_state.rs
fleupold: I agree, my point is it must already exist somewhere else today. And my ask would be to move the logic from there into here (cf. #192)

fleupold:
> This is an always has been based on the withdraw handler branch.

Currently this PR is based off master, meaning that whenever I read the changes, I also read the ones from #197 . It'd be easier to review if the base of this PR was #197 instead of master.

dfusion_rust_core/src/models/account_state.rs
bh2smith: Ahh! Yes! I see. Interesting! I will find it and deal with that!

bh2smith:
Please note that #199  should be merged before this (as it is an independent change, directly relevant to this).

listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -179,6 +185,79 @@ pub mod test {
         assert!(result.is_err());
     }
 
+    #[test]
```
fleupold: Could add a test for when no state exists?

listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -179,6 +185,79 @@ pub mod test {
...
+            amount: 10,
+        };
+        let mut entity: Entity = first_withdraw.into();
+        entity.set("id", "1");
+        store.apply_entity_operations(vec![EntityOperation::Set {
+            key: util::entity_key("Withdraw", &entity),
+            data: entity
+        }], None).unwrap();
+
+        let second_withdraw = PendingFlux {
```
fleupold: Maybe include one withdraw that's larger than the balance and thus should get ignored?

bh2smith: Yes this is certainly something that should be tested for!

listener/src/event_handler/flux_transition_handler.rs
bh2smith: Sure thing! I'm on it!

bh2smith: Actually, I am not sure this is going to work here since a pre-existing state is needed in order to "process_event"

https://github.com/gnosis/dex-services/blob/7849c453507b7fe359b782fab76e4a4417f687ce/listener/src/event_handler/flux_transition_handler.rs#L64-L67

Unless you are hinting at the fact that this `ok_or_else` should be tested.. If so, I would argue this is not part of this issue.

listener/src/event_handler/flux_transition_handler.rs
bh2smith: Done!

listener/src/event_handler/flux_transition_handler.rs
fleupold: I think it would be useful to document/test what happens if we receive such an event but find no state for the specified hash (or there is no state at all in the DB)

The test should make sure that we return an error in that case. 

The fact that it’s is just testing an ok_or is an implementation detail that might change in the future.

bh2smith: Yes I like this will make a separate PR with this and any other tests like this today!

PR #201, opened 2019-07-30
[theGraph] SellOrder Handler

### LINKED ISSUE #169: Order event handler [Rust]

W3stside:
Closes #169 

[WIP]

~Code seems correct, however logs in graphQL not showing `sellOrder` events. Maybe I am not building properly with rust before testing. Would love help here~

Locally merging in PR Order Model Extension (#202) successfull captured the SellOrder events into the GraphQL layer as seen here:
Screenshot from 2019-07-31 11-06-56

GraphiQL:
Screenshot from 2019-07-31 11-10-38

* Issues: None
~`auctionId` not properly saving - taking a look~


fleupold:
`uint` in the event name has to be uint256 (cf. note in the [wiki](https://github.com/gnosis/dex-services/wiki/Adding-Event-Listener#rust-event-listener-the-graph))

listener/src/event_handler/sellorder_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+impl EventHandler for SellOrderHandler {
+    fn process_event(
+        &self,
+        logger: Logger,
+        _block: Arc<EthereumBlock>,
+        _transaction: Arc<Transaction>,
+        log: Arc<Log>
+    ) -> Result<Vec<EntityOperation>, Error> {
+        let entity_id = util::entity_id_from_log(&log);
+        let flux = PendingFlux::from(log);
```
fleupold: You are creating a PendingFlux, where you actually should create an Order?

listener/src/event_handler/mod.rs
```diff
@@ -20,6 +20,9 @@ pub use initialization_handler::InitializationHandler;
 mod flux_transition_handler;
 pub use flux_transition_handler::FluxTransitionHandler;
 
+mod sellorder_handler;
```
fleupold: nit: would call the module just order_handler

listener/subgraph_definition/dfusion
```diff
@@ -24,5 +24,7 @@ dataSources:
           handler: rust
         - event: WithdrawRequest(uint16,uint8,uint128,uint256,uint16)
           handler: rust
+        - event: SellOrder(uint,uint16,uint16,uint8,uint8,uint96,uint96)
```
bh2smith: This could be your problem! Gotta use uint256

listener/src/event_handler/sellorder_handler.rs
W3stside: Yes, PR Order extension is handling this

listener/src/event_handler/mod.rs
W3stside: done

dfusion_rust_core/src/models/order.rs
```diff
@@ -8,10 +8,10 @@ use crate::models::{Serializable, RollingHashable, iter_hash};
 #[serde(rename_all = "camelCase")]
 pub struct Order {
     pub account_id: u16,
-    pub sell_token: u8,
     pub buy_token: u8,
-    pub sell_amount: u128,
+    pub sell_token: u8,
```
fleupold: Is this change really required? Usually you can instantiate structs with field names out of order.

W3stside: no i just got annoyed by the order and felt strongly about it in the moment :man_shrugging:

fleupold:
E2E tests should pass after rebase on master (last iteration failed because sleep timed out, which is now replaced by wait_for).
Unit tests fail, because #207 needs to be merged first. What I usually do in this case is basing this PR on top of #207 until #207 is merged (this way I can see that the tests are passing, etc). It's just important to remember changing the base branch back to master before merging.

W3stside:
`graph-listener` after first `restart-containers.sh` is called doesn't seem to be picking up any orders, thus failing tests... investigating

W3stside:
> Alternative (or possibly the same) to @fleupold's rebasing suggestion I would usually do the following,
> 
> * Wait for other branch that should be merged first to be merged to master,
> * then merge master into this
> 
> Looks good to me! Awesome stuff. Not super keen on the _force-push_ in the commit history... There is usually another way.

sure, just merging it works, but is a bit ugly in the commit history. `force-push` in individually made PRs are usually fine as it's a `squash-merge` back into `master` anyways and comments remain.

Though there are definitely times when it would not be beneficial, I agree. For example, in a PR where 2 ppl are working together

test/e2e-tests-auction.sh
```diff
@@ -30,6 +30,14 @@ step "Place 6 orders in current Auction" \
  truffle exec scripts/sell_order.js 4 1 0 4 52  && \
  truffle exec scripts/sell_order.js 5 2 0 20 280"
 
+step_with_retry "[theGraph]: SellOrder was added to graph db - accountId 5's sellOrder === 280" \
```
bh2smith: nit: "[theGraph]:" looks kinda awkward next to all the other tests...

bh2smith:
Have we considered the fact that SellOrders are emitted from the `SnappAuction` contract and not the `SnappBase`? This could easily be a reason why the event is not being picked up...

test/e2e-tests-auction.sh
W3stside: will remove and retry build

W3stside:
> Have we considered the fact that SellOrders are emitted from the `SnappAuction` contract and not the `SnappBase`? This could easily be a reason why the event is not being picked up...

addressed together @bh2smith - it was a `restart-containers` issue

PR #203, opened 2019-07-31
[theGraph/dfusion Model] Order Extend

### LINKED ISSUE #202: Order struct extension - SellOrder/BuyOrder
dfusion rust core model Order needs to be extended to include:
1. `slot`
2. `slot_index`

W3stside:
Would fix #202 

Extension of Order model to include:
1. `slot`
2. `slot_index`

### DIFF
--- a/dfusion_rust_core/src/models/order.rs
+++ b/dfusion_rust_core/src/models/order.rs
@@ -11,8 +11,8 @@ use super::util::*;
 #[derive(Debug, Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct Order {
-    pub slot: U256,
-    pub slot_index: u16,
+    pub slot: Option<U256>,
+    pub slot_index: Option<u16>,
     pub account_id: u16,
     pub sell_token: u8,
     pub buy_token: u8,
@@ -42,13 +42,13 @@ impl From<Arc<Log>> for Order {
     fn from(log: Arc<Log>) -> Self {
         let mut bytes: Vec<u8> = log.data.0.clone();
         Order {
-            slot: U256::pop_from_log_data(&mut bytes),
-            slot_index: u16::pop_from_log_data(&mut bytes),
+            slot: Some(U256::pop_from_log_data(&mut bytes)),
+            slot_index: Some(u16::pop_from_log_data(&mut bytes)),
             account_id: u16::pop_from_log_data(&mut bytes),
-            sell_token: u8::pop_from_log_data(&mut bytes),
             buy_token: u8::pop_from_log_data(&mut bytes),
-            sell_amount: u128::pop_from_log_data(&mut bytes),
+            sell_token: u8::pop_from_log_data(&mut bytes),
             buy_amount: u128::pop_from_log_data(&mut bytes),
+            sell_amount: u128::pop_from_log_data(&mut bytes),
         }
     }
 }
@@ -56,13 +56,13 @@ impl From<Arc<Log>> for Order {
 impl From<Entity> for Order {
     fn from(entity: Entity) -> Self {
         Order {
-            slot: U256::from_entity(&entity, "auctionId"),
-            slot_index: u16::from_entity(&entity, "slotIndex"),
+            slot: Some(U256::from_entity(&entity, "auctionId")),
+            slot_index: Some(u16::from_entity(&entity, "slotIndex")),
             account_id: u16::from_entity(&entity, "accountId"),
-            sell_token: u8::from_entity(&entity, "sellToken"),
             buy_token: u8::from_entity(&entity, "buyToken"),
-            sell_amount: u128::from_entity(&entity, "sellAmount"),
+            sell_token: u8::from_entity(&entity, "sellToken"),
             buy_amount: u128::from_entity(&entity, "buyAmount"),
+            sell_amount: u128::from_entity(&entity, "sellAmount"),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -57,6 +57,8 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
                 .iter()
                 .map(|raw_order| raw_order.as_document().unwrap())
                 .map(|order_doc| super::Order {
+                        slot: order_doc.get_str("slot").unwrap().parse().unwrap(),
```
fleupold: Can you check the representation inside mongo? Standing orders don't have a `slot` or `slot_index` field so you should be crashing here (because of the unwrap). Maybe we should make slot and slot_index an Optional so that it can be None?

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,9 +38,40 @@ impl Serializable for u128 {
...
 }
 
+impl From<Arc<Log>> for Order {
+    fn from(log: Arc<Log>) -> Self {
+        let mut bytes: Vec<u8> = log.data.0.clone();
+        Order {
+            slot: U256::pop_from_log_data(&mut bytes),
+            slot_index: u16::pop_from_log_data(&mut bytes),
+            account_id: u16::pop_from_log_data(&mut bytes),
+            sell_token: u8::pop_from_log_data(&mut bytes),
```
fleupold: buyToken comes before sellToken in the event data. Thus it needs to be popped first.

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,9 +38,40 @@ impl Serializable for u128 {
...
+            buy_token: u8::pop_from_log_data(&mut bytes),
+            sell_amount: u128::pop_from_log_data(&mut bytes),
```
fleupold: dito here.

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,9 +38,40 @@ impl Serializable for u128 {
...
+            sell_amount: u128::from_entity(&entity, "sellAmount"),
+            buy_amount: u128::from_entity(&entity, "buyAmount"),
+        }
+    }
+}
+
 impl From<mongodb::ordered::OrderedDocument> for Order {
     fn from(document: mongodb::ordered::OrderedDocument) -> Self {
         Order {
+            slot: U256::from(document.get_i32("slot").unwrap()),
```
fleupold: Did you check the record inside mongo db? There is no field slot on order. For some reason it's called auctionId there.

dfusion_rust_core/src/models/order.rs
```diff
@@ -1,12 +1,18 @@
...
+use web3::types::{H256, U256, Log};
 
 use crate::models::{Serializable, RollingHashable, iter_hash};
 
+use super::util::*;
+
 #[derive(Debug, Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct Order {
+    pub slot: U256,
```
fleupold: Maybe this should be an Optional struct auctionInformation which contains two fields (slot and slotIndex) but can be None for StandingOrders?

josojo:
This really looks already pretty rustic. Nice!
Yeah, if you follow Felix's comments, that should fix things and make it work.

dfusion_rust_core/src/models/order.rs
```diff
@@ -1,12 +1,18 @@
...
+use web3::types::{H256, U256, Log};
 
 use crate::models::{Serializable, RollingHashable, iter_hash};
 
+use super::util::*;
+
 #[derive(Debug, Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct Order {
+    pub slot: Option<U256>,
```
fleupold: Given that these can semantically only be either both Some or both None, I'd prefer if this was a single field. 

The field would be an optional of a struct with two fields. It could be called something like `batch_information` or something like that. One the database level it's fine to have the datatype flat (as in not two different type of records).

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,9 +38,40 @@ impl Serializable for u128 {
...
+            buy_amount: u128::from_entity(&entity, "buyAmount"),
+            sell_amount: u128::from_entity(&entity, "sellAmount"),
+        }
+    }
+}
+
 impl From<mongodb::ordered::OrderedDocument> for Order {
     fn from(document: mongodb::ordered::OrderedDocument) -> Self {
         Order {
+            slot: None,
```
fleupold: Let's use the info that we can find on the record here.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -227,6 +227,8 @@ pub mod tests {
     #[test]
     fn test_serialize_order() {
         let order = models::Order {
+            slot: U256::zero(),
```
fleupold: Feel free to leave all None for the tests that don't care.

dfusion_rust_core/src/models/order.rs
W3stside: but wouldn't that mean the graph query would need to be nested? 
e.g 
```
sellOrders ( 
  where: {
    batch_information {
      slot: 123
      slot_index: 2
    }
  }
}
```
which isnt possible yet?

fleupold: On the db level, order could still be flat (as in have 7 fields: slot, slot_index, accountId, buy/sell token, buy/sell amount).

W3stside: i understand now

W3stside:
closing in favour of #206

bh2smith:
I really like to see that use of rust's `Optional` feature. It makes perfect sense here and now I finally understand its purpose! Not sure exactly why this was closed, but I see its was in favour of #206, so I will go look at that then.

PR #205, opened 2019-07-31
[Build Fix] fix master build

fleupold:
Master is currently failing due to an issue with a dependency described here: https://github.com/graphprotocol/graph-node/issues/1090

The issue suggests to pin the dependency to the last good revision using

```
cargo update -p https://github.com/graphprotocol/ethabi.git#ethabi --precise b2e44d8439d9712a5d911f9992112a4509f8a633
```

This PR does this for the local listener and driver Cargo.lock file. It then makes sure those lock files are also copied inside the docker container so that the pinned versions are respected there.

### Test Plan

See travis pass


fleupold:
Feel free to merge/commandeer while I'm on the plane tomorrow.

PR #207, opened 2019-08-01
206: Slot & Slot Index >> Batch Information optional struct

W3stside:
Branched off of (now closed) #202 

Closes #203 

As suggested in #203, make `slot` and `slot_index` (optional props) into an optional struct named `batch_information` instead

Fixes tests

### DIFF
--- a/dfusion_rust_core/src/models/order.rs
+++ b/dfusion_rust_core/src/models/order.rs
@@ -64,7 +64,7 @@ impl From<Entity> for Order {
     fn from(entity: Entity) -> Self {
         Order {
             batch_information: Some(BatchInformation{
-                slot: U256::from_entity(&entity, "slot"),
+                slot: U256::from_entity(&entity, "auctionId"),
                 slot_index: u16::from_entity(&entity, "slotIndex"),
             }),
             account_id: u16::from_entity(&entity, "accountId"),
@@ -80,7 +80,7 @@ impl From<mongodb::ordered::OrderedDocument> for Order {
     fn from(document: mongodb::ordered::OrderedDocument) -> Self {
         Order {
             batch_information: Some(BatchInformation{
-                slot: U256::from(document.get_i32("slot").unwrap()),
+                slot: U256::from(document.get_i32("auctionId").unwrap()),
                 slot_index: document.get_i32("slotIndex").unwrap() as u16,
             }),
             account_id: document.get_i32("accountId").unwrap() as u16,
--- a/dfusion_rust_core/src/models/order.rs
+++ b/dfusion_rust_core/src/models/order.rs
@@ -51,8 +51,6 @@ impl From<Arc<Log>> for Order {
                 slot: U256::pop_from_log_data(&mut bytes),
                 slot_index: u16::pop_from_log_data(&mut bytes),
             }),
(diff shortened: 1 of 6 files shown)


dfusion_rust_core/src/models/order.rs
```diff
@@ -44,6 +91,21 @@ impl From<mongodb::ordered::OrderedDocument> for Order {
     }
 }
 
+impl Into<Entity> for Order {
+    fn into(self) -> Entity {
+        let batch_info = self.batch_information.unwrap();
```
fleupold: Maybe just set the fields if the value is some? Right now, we crash if the value is not there.

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,9 +43,45 @@ impl Serializable for u128 {
...
+            buy_amount: u128::from_entity(&entity, "buyAmount"),
+            sell_amount: u128::from_entity(&entity, "sellAmount"),
+        }
+    }
+}
+
 impl From<mongodb::ordered::OrderedDocument> for Order {
     fn from(document: mongodb::ordered::OrderedDocument) -> Self {
         Order {
+            batch_information: None,
```
fleupold: I think this should still be Some given that the data is there.

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,9 +43,45 @@ impl Serializable for u128 {
...
 
+impl From<Arc<Log>> for Order {
+    fn from(log: Arc<Log>) -> Self {
+        let mut bytes: Vec<u8> = log.data.0.clone();
+        Order {
+            batch_information: Some(BatchInformation{
+                slot: U256::pop_from_log_data(&mut bytes),
+                slot_index: u16::pop_from_log_data(&mut bytes),
+            }),
+            // slot: Some(U256::pop_from_log_data(&mut bytes)),
```
fleupold: nit: remove.

W3stside:
@fleupold added unit tests

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,9 +43,46 @@ impl Serializable for u128 {
     }
 }
 
+impl From<Arc<Log>> for Order {
+    fn from(log: Arc<Log>) -> Self {
+        let mut bytes: Vec<u8> = log.data.0.clone();
```
fleupold: ubernit: we shouldn't need to clone here. But can fix that in a forward PR given that this one already passed CI.

W3stside: sounds good, merging

fleupold: Maybe a few more words on when we need to clone, since I also saw it in @anxolin's PR.

Clone makes a copy of the variable and leaves the existing instance untouched and thus "consumable". Whenever the original variable is not used later, this is unnecessary since we could have also "consumed" (aka moved) it.
If the variable is used later we might have to use clone, but only if the current use of it is "consuming". That is, e.g. an assignment or if the variable passed by value into another function (if we wrote the function it can be worthwhile checking if it has to be passed by value).
If we can use the value by reference (e.g. in a print statement), we don't consume the variable and thus doesn't require a clone (just an & before the variable).

PR #209, opened 2019-08-05
Use npx for truffle instead of global system one

anxolin:
The services depends on truffle and other dependencies being install globally.

Truffle can have different versions depending on the project, so it's better not to rely on the global truffle and use npx instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,22 +66,22 @@ In order to setup some testing accounts and make the first deposits (from accoun
 
 ```bash
 cd dex-contracts
-truffle exec scripts/setup_environment.js
-truffle exec scripts/deposit.js 1 1 18
-truffle exec scripts/wait_seconds.js 181
+npx truffle exec scripts/setup_environment.js
+npx truffle exec scripts/deposit.js 1 1 18
+npx truffle exec scripts/wait_seconds.js 181
 ```
 
 To claim back the deposit, submit a withdraw request:
 
 ```bash
-truffle exec scripts/request_withdraw.js 1 1 18
+npx truffle exec scripts/request_withdraw.js 1 1 18
 ```
 
 After 20 blocks have passed, the driver will apply the state transition and you should be able to claim back your funds:
 
 ```bash
-truffle exec scripts/wait_seconds.js 181
-truffle exec scripts/claim_withdraw.js 0 1 1
+npx truffle exec scripts/wait_seconds.js 181
+npx truffle exec scripts/claim_withdraw.js 0 1 1
 ```
 
 Tests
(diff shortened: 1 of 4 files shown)


W3stside:
npx requires npm +5.6 - this should be added to readme

test/restart-containers.sh
```diff
@@ -6,4 +6,4 @@ docker-compose up -d mongo
 docker-compose rm -sf ganache-cli
 docker-compose up -d ganache-cli
 docker-compose restart listener
-cd dex-contracts && wait-port -t 30000 8545 && truffle migrate && cd -
\ No newline at end of file
+cd dex-contracts && npx wait-port -t 30000 8545 && npx truffle migrate && cd -
```
bh2smith: what is wait-port? Is this a truffle command?

bh2smith:
I only wonder what the real benefit is here. Having a single installation of truffle shouldn't make a difference whether it is local or global. If i understand the differences correctly, local installations are only meaningful in a developer's environment where there are several projects on the go at the same time and each of them requires different versions of same thing. When it comes to a single project, there is only need for one. For example, we don't activate a virtual environment for our python 3.6 dependency when running the travis build.

test/restart-containers.sh
anxolin: wait-port is something that allows you to wait for one service to be up.
It was added by Felix to coordinate better when the migration is executed.

I just use npx so we don't forze users to have it installed in their computer:
https://www.npmjs.com/package/wait-port

anxolin:
> I only wonder what the real benefit is here. Having a single installation of truffle shouldn't make a difference whether it is local or global. If I understand the differences correctly, local installations are only meaningful in a developer's environment where there are several projects on the go at the same time and each of them requires different versions of same thing. When it comes to a single project, there is only need for one. For example, we don't activate a virtual environment for our python 3.6 dependency when running the travis build.

Yes, this is the main reason. That's right, in python you usually don't create projects and install the dependencies as global, this could be a source of problems, instead you create this virtual env to make sure there's no conflicts. This would be similar.

Additionally adds one extra step or dependency for people trying to run it. So they need to install a global package on their system before running these commands.

Having said that, it's not the end of the world to work with a global truffle, I just did it locally to solve my issues and the create this draft PR (in case I founded more similar things).

anxolin:
> npx requires npm +5.6 - this should be added to readme

I can add it, although I think npx was release in node 5.2 in July 2017. I would be surprised if anyone is using this version.

Anyways, there other requirements and we are not detailing them. Note that this PR even works in the direction of removing some of them (truffle and wait-port) :)

W3stside:
I think this PR is better covered in #210 - thoughts on closing in favour of #210 ? @anxolin @fleupold

fleupold:
Yes, sounds good.

anxolin:
> I think this PR is better covered in #210 - thoughts on closing in favour of #210 ? @anxolin @fleupold


I created this as a draft, cause I was not sure it was complete, anyways it was good feedback here.

Well, this one also have the e2e tests. I'll just rebase with master, and ask for merging. It should only add the e2e part now.

PR #210, opened 2019-08-05
Restart all services in restart-containers.sh

anxolin:
Restarting the containers using `test/restart-containers.sh` script won't restart services like `graph-listener`.

Modified the script so all services are restarted, so it takes for example the new mappings/handlers for graph-listener

### DIFF
--- a/test/restart-containers.sh
+++ b/test/restart-containers.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -e
 
-docker-compose rm -sf mongo
-docker-compose up -d mongo
-docker-compose rm -sf ganache-cli
-docker-compose up -d ganache-cli
-docker-compose restart listener
+docker-compose rm -sf
+docker-compose up -d
 cd dex-contracts && wait-port -t 30000 8545 && truffle migrate && cd -
--- a/test/restart-containers.sh
+++ b/test/restart-containers.sh
@@ -1,6 +1,17 @@
 #!/bin/bash
 set -e
 
-docker-compose rm -sf
-docker-compose up -d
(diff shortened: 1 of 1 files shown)


test/restart-containers.sh
```diff
@@ -1,9 +1,6 @@
...
 set -e
 
-docker-compose rm -sf mongo
-docker-compose up -d mongo
-docker-compose rm -sf ganache-cli
-docker-compose up -d ganache-cli
-docker-compose restart listener
-cd dex-contracts && wait-port -t 30000 8545 && truffle migrate && cd -
\ No newline at end of file
+docker-compose rm -sf
```
fleupold: I think it should be fine to use `docker-compose down && docker-compose up -d` We shouldn't have to remove/rebuild the containers.

Moreover, we have a truffle image inside docker compose, which I believe is not needed anymore (since we migrate from outside the container). However `docker-compose up -d` starts all containers, so we should probably remove the truffle one or explicitly specify the containers we need.

This is what's causing the e2e test to fail atm - cf. that the contracts are being deployed twice and thus the setup environment script registers tokens on the wrong address (`SnappBase.deployed()` points at the second instance, but the listener is configured to observe the first)

W3stside: I second what @fleupold says #169 won't pass unit tests until this is fixed - works locally for me

W3stside: the latest commit, authored with @bh2smith passed my tests. It required resetting postgres and restarting the graph-listener. Uses `npx` for both `wait-port` and `truffle`

bh2smith: Authored by both of us! Pair Programming FTW!

PR #211, opened 2019-08-06
Standing order listener

### LINKED ISSUE #200: Standing sell orders handler [rust]

event: `StandingSellOrderBatch`

anxolin:
Closes #200 once it's completed.

This PR main goals are to:
[] Add event listener and entities for `graph-listener` for the standing order events (`StandingSellOrderBatch`)
[] Add any missing field or entity in the models of `dfusion_rust_core`
[] Create tests for the model of `dfusion_rust_core`
[] Adapt e2e tests to verify the indexed events


anxolin:
I started doing the previous commits, as you see I left the handler unimplemented, since I'm going to work first in the model tests to make sure I can create the model from the log entry.

Also, it's not loading the info from the packed orders yet from the bytes (I'll continue with that), but you can see already how it's creating extracting the main data from the event, and then iterating over the packed orders to create the `Order`.

Let me know if this goes in the right direction. 
Also, I'm not sure about the formatting that it's being used. Some changes in `standing_order.rs` are only formatting. I used `rustcfmt` that uses 4 spaces.

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,6 +34,27 @@ impl Serializable for u128 {
     }
 }
 
+impl From<Vec<u8>> for Order {
+    fn from (bytes: Vec<u8>) -> Self {
+        let mut bytes = bytes.clone();
```
fleupold: I think we should not need to clone here.

dfusion_rust_core/src/models/order.rs
```diff
@@ -32,6 +34,27 @@ impl Serializable for u128 {
...
+        println!("TODO: Read order bytes: {:?}", bytes);
+        // TODO: Extract order from bytes
+        
+        Order {
```
fleupold: If the format is the same as for the non standing order events, we should try to use the same code for parsing both. But slot_index might be missing here?

listener/subgraph_definition/schema.graphql
```diff
@@ -21,4 +21,20 @@ type Withdraw @entity {
...
+
+type StandingSellOrderBatch @entity {
+    id: ID!
+    accountId: Int!
+    batchIndex: Int!
+    validFromAuctionId: Int!
+    orders: [OrderDetails!]
+}
+
+type OrderDetails @entity {
```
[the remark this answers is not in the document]
anxolin: I think so, I guess we can add a None and use the same model.
Once I merge the PR from David, I'll make sure to review this part (I'll leave this unsolved for now)

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -48,54 +63,133 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
...
+        // Get basic data from event
+        let account_id = u16::pop_from_log_data(&mut bytes);
+        let batch_index = U256::pop_from_log_data(&mut bytes);
+        let valid_from_auction_id = U256::pop_from_log_data(&mut bytes);
+        let mut packed_orders_bytes = bytes.clone();
+        assert!(packed_orders_bytes.len() % 26 == 0, "Each order should be packed in 26 bytes");
+        
+
+        // Extract packed order info
+        let mut orders : Vec<models::Order> = Vec::new();
```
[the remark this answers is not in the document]
anxolin: nice! I haven't got this far in the book. I had something like that in mind, but didn't know how to write it

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -48,54 +63,133 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
     fn from(document: mongodb::ordered::OrderedDocument) -> Self {
         let account_id = document.get_i32("_id").unwrap() as u16;
         let batch_index = U256::from(document.get_i32("batchIndex").unwrap());
+        let valid_from_auction_id = U256::from(document.get_i32("validFromAuctionId").unwrap());
```
fleupold: This field doesn't seem to be present. I assume you have to add it to the group clause here: https://github.com/gnosis/dex-services/blob/master/driver/src/db_interface.rs#L123

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -48,54 +63,133 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
...
+}
+
+impl From<Arc<Log>> for StandingOrder {
+    fn from(log: Arc<Log>) -> Self {
+        let mut bytes: Vec<u8> = log.data.0.clone();
+        // Get basic data from event
+        let account_id = u16::pop_from_log_data(&mut bytes);
+        let batch_index = U256::pop_from_log_data(&mut bytes);
+        let valid_from_auction_id = U256::pop_from_log_data(&mut bytes);
+        let mut packed_orders_bytes = bytes.clone();
```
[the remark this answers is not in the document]
anxolin: No need for now since it's not implemented. 

I was going to use the `pop_from_log_data` methods that mutate the vars. So this lines shadow the `bytes` and make it mutable, so we can extract the params of the orders.

The thing I found doing this is that we don't have a `uint96` version of those pop methods...
Maybe is best just to take the slices from the Vec and getting the values in an immutable way without using those utils

dfusion_rust_core/src/models/order.rs
anxolin: Yep, there's no need for it now. 

It would haven been required for how I was going to implement it; by using the pop_from_log_data methods that mutate the vars. So this lines shadow the bytes and make it mutable, so we can extract the params of the orders.

The thing I found doing this is that we don't have a uint96 version of those pop methods...
Maybe is best just to take the slices from the Vec and getting the values in an immutable way without using those utils

dfusion_rust_core/src/models/order.rs
anxolin: Yep, I haven't incorporated this PR yet: https://github.com/gnosis/dex-services/pull/207/files

I see on it a `Option<BatchInformation>`. Once it's merged to master, I'll bring those changes to this branch and adapt it so I add a `None` in there.

In that PR I don't see a parsing from the packed form of the events, that is what I implemented in this one (`From<Vec<u8>>`). 

David created the `From<Entity` and `From<Arc<Log>>`.

dfusion_rust_core/src/models/order.rs
fleupold: Even if you need this variable to be mutable, you don't need to clone. You could move bytes into the mutable version:

```
let mut bytes = bytes
```

This will consume the immutable version of bytes and make it available under the mutable variable.

anxolin:
Rebased the master branch to incorporate David's changes

listener/subgraph_definition/schema.graphql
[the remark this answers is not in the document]
bh2smith: > reuse the existing Order entity

I really hope this is the way we go, because the Auction Settlement Listener is going to have to fetch all orders and work with them. In fact, as I see it, `slot` and `slot_index` will be necessary here just as well...

bh2smith:
On second thought, I wonder if we have actually handled the standing orders in the existing listener, because I don't recall having to sort out the orders from the standing orders there... Will look into this further. 

Also, based on the content of failed build here, it looks to me like somebody isn't using the hooks provided!

bh2smith:
Observe that our current implementation of Auction Settlement does not explicitly take standing order into account:

https://github.com/gnosis/dex-services/blob/ef7e011390a3db1ed71befbab5d8130c6d63f938/event_listener/dfusion_db/snapp_events/auction_settlement.py#L21-L40

This could mean that we have missed some aspect of our current listener implementation. I would suspect that we could have an auction with one regular order and one standing order. If any non-trivial amount of that standing order is executed, I would expect that this would break the expected results of an end to end test. Perhaps this is better to be left as a bug report... @fleupold or @josojo should we include an e2e-test regarding "Mixed Auctions"?

listener/subgraph_definition/schema.graphql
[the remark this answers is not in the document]
anxolin: Thanks Ben!

After your nice explanation about what are these slots for standing orders, I changed how the order is "unpacked" from the event. Now it hands to the "`From<bytes> --> Order`" a bit more info, so I can deduct these fields.

The changes are in this commit:
https://github.com/gnosis/dex-services/commit/940943487a0ddff0c9cb6145f8f874222bbb0e6c

I adapted the test for `test_from_log` and it's passing now. 
I know the test `test_to_and_from_entity` it's still failing, cause I don't have finished that part (I still have to figure out how to save entities containing other entities)

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -69,35 +110,199 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
...
+    fn from(log: Arc<Log>) -> Self {
+        let mut bytes: Vec<u8> = log.data.0.clone();
+        // Get basic data from event
+        let batch_index = U256::pop_from_log_data(&mut bytes);
+        let valid_from_auction_id = U256::pop_from_log_data(&mut bytes);
+        let account_id = u16::pop_from_log_data(&mut bytes);
+        // let packed_orders_bytes = bytes;
+
+        println!("Parsing orders from packedOrders. {} bytes. {:?}", bytes.len(), bytes);
+        assert!(bytes.len() % 26 == 0, "Each order should be packed in 26 bytes");
```
anxolin: I guess these two lines are not super good practice.
Not sure what's the best way to log messages. I added that cause it was crashing the assert, for some reason I'm getting a different number of bytes there.

The `assert` I guess it's also not beautiful because it makes the code to "panic" hehe
I saw the usual approach is to use the trait `Result` but it just didn't see how it fits here. Maybe, the validations should be done before calling this function, not sure.

anxolin:
> Also, based on the content of failed build here, it looks to me like somebody isn't using the hooks provided!

I had the hooks until I re-installed the project from scratch when I had issues with docker, then forgot to add them again :( 

They are back in, I solved also the issues it was complaining about.

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -126,7 +126,7 @@ impl From<Arc<Log>> for StandingOrder {
         let orders: Vec<models::Order> = bytes.chunks(26)
             .enumerate()
             .map(|(order_number, chunk)| models::Order::from(EncodedOrder {
-                account_id: account_id,
+                account_id,
```
bh2smith: 👍

fleupold:
> Observe that our current implementation of Auction Settlement does not explicitly take standing orders into account:

@bh2smith it does. See that [`get_orders`](https://github.com/gnosis/dex-services/blob/ef7e011390a3db1ed71befbab5d8130c6d63f938/event_listener/dfusion_db/database_interface.py#L151) also fetches all standing orders that are valid for this batch.

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -1,24 +1,46 @@
...
 
 #[derive(Debug, Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct StandingOrder {
     pub account_id: u16,
     pub batch_index: U256,
+    pub valid_from_auction_id: U256,
     orders: Vec<super::Order>,
 }
+pub struct EncodedOrder {
```
fleupold: Not sure it's worthwhile creating a public struct for this. It might be a bit confusing what it is used for, when really it's just a way to convert aggregated information we receive in a Log to an order.

I'd just create a method on `Order` directly that is something like `from_standing_order_event(account_id: u16, ..., bytes: Vec<u8>)`

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -69,35 +110,214 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
...
+        let packed_orders_bytes = &bytes[0..byte_size];
+        println!("Parsing orders from packedOrders. {} bytes. {:?}", packed_orders_bytes.len(), packed_orders_bytes);
+        assert!(packed_orders_bytes.len() % 26 == 0, "Each order should be packed in 26 bytes");
+                
+        // Extract packed order info
+        let orders: Vec<models::Order> = packed_orders_bytes.chunks(26)
+            .enumerate()
+            .map(|(order_number, chunk)| models::Order::from(EncodedOrder {
+                account_id,
+                auction_id: batch_index,
```
fleupold: This is not really correct. `batch_index` is the auction_id from which point onwards the standing order is valid. However, the way we query for auction_id is via exact match (where `auction_id == x`). A standing order would therefore only appear in the batch it was created in (but it should in all following ones until it was updated).

Check [here](https://github.com/gnosis/dex-services/blob/2143f1f808411528f976c91082509dc4cdc23af3/event_listener/dfusion_db/database_interface.py#L153) how we combine orders and standing orders in the current python code.

I think the Order entities on a StandingOrderBatch should not have an auction ID on them (at least not when they are written to an entity). This way they are ignored on the one-off order fetching (where auction_id==x) and can be separately fetched.

dfusion_rust_core/src/models/order.rs
```diff
@@ -1,19 +1,24 @@
...
 use std::sync::Arc;
 use web3::types::{H256, U256, Log};
 
+use super::util::*;
+use super::EncodedOrder;
+
 use crate::models::{Serializable, RollingHashable, iter_hash};
 
-use super::util::*;
+pub const NUM_SLOTS_REGULAR_ORDERS: u16 = 500;
```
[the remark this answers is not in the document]
anxolin: Nice, I'll move the constants there.

@fleupold  I'm not sure how you suggest calculating NUM_SLOTS_REGULAR_ORDERS or NUM_SLOTS_PER_STANING_ORDER_ACCOUNT using NUM_RESERVED_ACCOUNTS

As I understood:
* there's 1000 slots
* 500 is for normal trades, they are served in FIFO order
* The other 500 are for the 50 accounts:
   * 500-509 are for the 10 slots of account 0
   * 510-519 are for the 10 slots of account 1
   * ...and so forth

If this is true (pls confirm) We would need the 3 of them:
```
pub const NUM_RESERVED_ACCOUNTS: usize = 50;
pub const NUM_SLOTS_REGULAR_ORDERS: u16 = 500;
pub const NUM_SLOTS_PER_STANING_ORDER_ACCOUNT: u16 = 10;
```
`NUM_SLOTS_REGULAR_ORDERS + NUM_SLOTS_PER_STANING_ORDER_ACCOUNT * account_id + num of order for this user`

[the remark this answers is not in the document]
anxolin: https://github.com/gnosis/dex-services/pull/211/commits/87a5a562c1aad4e7317ebfbc7ec1cdf7d7abf431

dfusion_rust_core/src/models/standing_order.rs
anxolin: https://github.com/gnosis/dex-services/pull/211/commits/9b3d3412703f4263408bb98e27d484885942cfbd

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -69,35 +89,205 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
     }
 }
 
+impl From<&Arc<Log>> for StandingOrder {
+    fn from(log: &Arc<Log>) -> Self {
+        let mut bytes: Vec<u8> = log.data.0.clone();
+        println!("Parsing StandingOrder from bytes. {} bytes. {:?}", bytes.len(), bytes);
```
bh2smith: can we make this an `info!(logger, ...)` instead of `println!`?

listener/subgraph_definition/schema.graphql
```diff
@@ -23,6 +23,14 @@ type Withdraw @entity {
     slotIndex: Int!
 }
 
+type StandingSellOrderBatch @entity {
+    id: ID!
+    accountId: Int!
+    batchIndex: BigInt!
+    validFromAuctionId: BigInt!
+    orders: [SellOrder!]
```
fleupold: auctionId and slotIndex on the SellOrder entity might have to become optional (as in we need to remove the !)

listener/src/event_handler/order_handler.rs
```diff
@@ -29,7 +29,7 @@ impl EventHandler for SellOrderHandler {
         info!(logger, "Processing SellOrder {:?}", &order);
 
         let mut entity: Entity = order.into();
-        entity.set("id", &entity_id);
+        entity.set("id", format!("{}_0", &entity_id));
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2F3ohzdYJK1wAdPWVk88%2Fgiphy.gif&token=oops" data-gitmeme-token="oops"></a>

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -1,24 +1,41 @@
...
 
+use super::util::*;
 
 #[derive(Debug, Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct StandingOrder {
     pub account_id: u16,
     pub batch_index: U256,
-    orders: Vec<super::Order>,
+    pub valid_from_auction_id: U256,
```
fleupold: unrelated to this PR: can we make these non public?

dfusion_rust_core/src/models/standing_order.rs
anxolin: Done!, I had to add the lib too.

listener/src/event_handler/order_handler.rs
anxolin: @fleupold this was on purpose. 

https://github.com/gnosis/dex-services/pull/211/commits/adbfcec07d4747d734f22668576de3ff901d32c5

The thing is the standing orders use an id of: `block_hash + log_index + order number`

For coherence between all orders, I decided to follow a naming convention, I just add a `_0` for normal orders, since they are the first and only order.

I don't mind to remove it, I don't have a strong opinion about it.

dfusion_rust_core/src/models/standing_order.rs
anxolin: Now I see there's a `get_orders` method that borrows the orders. Is this a standard practice in rust?
I'll make it private again

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -69,35 +89,197 @@ impl From<mongodb::ordered::OrderedDocument> for StandingOrder {
...
+            account_id,
+            batch_index,
+            valid_from_auction_id,
+            orders
+        }
+    }
+}
+
+
+impl From<Entity> for StandingOrder {
```
fleupold: I'd argue that the signature should look more like

```rs
impl From<(Entity, Vec<Entity>)> ...
impl Into<(Entity, Vec<Entity>)> ...
```

where the first entity is of type StandingOrder and the vector is the orders associated with the StandingOrder.

It would be nice if we got a coherent entity set back from the into method instead of relying on the handler to mutate the resulting entity with the order IDs.

listener/src/event_handler/order_handler.rs
fleupold: Ok I didn't see that. I'm fine with that.

test/e2e-tests-standing-order.sh
```diff
@@ -17,8 +17,36 @@ step "Place Sell Order" \
 step "Place standing order in current Auction" \
 "npx truffle exec scripts/standing_order.js 0 1 2 1 1"
 
-step_with_retry "Check standing order has been recorded" \
-"mongo dfusion2 --eval \"db.standing_orders.findOne({accountId:0, batchIndex:0, orders: [ { buyToken:1, sellToken:2, buyAmount:'1000000000000000000', sellAmount:'1000000000000000000' }]})\" | grep ObjectId"
+step_with_retry "Check mongo standing order has been recorded" \
+"mongo dfusion2 --eval \"db.standing_orders.findOne({accountId:0, batchIndex:0, validFromAuctionId:0, orders: [ { buyToken:1, sellToken:2, buyAmount:'1000000000000000000', sellAmount:'1000000000000000000' }]})\" | grep ObjectId"
+
+step_with_retry "Check graph standing order batch has been recorded" \
+"source ../test/utils.sh && query_graphql \
```
fleupold: Wouldn't this check be enough, since you are already listing the order here? Technically you could grep for multiple fields at the same time.

Might make the tests a little bit more concise if we don't have to check two queries per standing order.

test/e2e-tests-standing-order.sh
```diff
@@ -38,9 +66,67 @@ step_with_retry "Make sure standing order is still traded" \
...
+          sellAmount: \"2000000000000000000\" \
+        }) { \
+          buyAmount \
+        } \
+    }\" | grep 1000000000000000000"
+
 step "Cancel standing order in same batch (make sure only cancel gets processed)" \
 "npx truffle exec scripts/standing_order.js 0 0 0 0 0"
 
+# TODO: Review after the PR that discard orders with sellVolume 0
```
fleupold: Can you create an issue for Todos you intend to merge and reference the issue in the comment?

dfusion_rust_core/src/models/order.rs
```diff
@@ -25,6 +26,27 @@ pub struct Order {
     pub sell_amount: u128,
 }
 
+impl Order {
+    pub fn from_encoded_order (
+        account_id: u16,
+        bytes: Vec<u8>
```
fleupold: Could this be a fixed length slice also? We should only ever be able to convert vectors with 26 elements and it would be nice if the compiler can enforce that. 
I guess it might be a problem converting a chunk into a fixed size slice, but maybe we can do that in the call-site with some form of unwrap.

anxolin:
> StandingOrder

anxolin:
> Just a suggestion around converting StandingOrder structs into entities, I think it would be nice to have all the conversion logic in one place if possible.

@fleupold I think it's a good suggestion. I was doing it like that, BUT then I realised that the standing order `Entity` doesn't have orders as subentities, It can have that. It need to be a List of ids referencing the orders.

The issue is that, in coherence with the other handlers, I'm handling the `Ids` in the handler, and the `Into` entities in each model file.

The solutions I see is:
* Leave it like this. I don't think is that bad :P 
* Model the ID in the standing order batch and orders, so the handler adds the ids before calling the `into()`. And the `into()` creates the standing order + the list of ids. `----->` PROBLEM: To be really coherence we should introduce the ID also to the other entities, and adapt a lot of code.

test/e2e-tests-standing-order.sh
anxolin: You cannot filter by subentities right now. So the second query verifies the values of the order, I don't see how I can do that easily with grep.

Also, I think it's good to check the query for both entities

fleupold: This should do the same as the two queries in one. We basically ask for all orders on the standing order, select all relevant fields and then match the expected json output.

```sh
step_with_retry "Check graph standing order batch has been updated" \
"source ../test/utils.sh && query_graphql \
    \"query { \
        standingSellOrderBatches(where: { \
          accountId: 0, \
          batchIndex: 1, \
          validFromAuctionId: 2 \
        }) { \
        orders { \
            buyToken \
            sellToken \
            buyAmount \
            sellAmount \
        } \
      } \
    }\" | grep \"buyAmount.:.1000000000000000000.,.buyToken.:2,.sellAmount.:.2000000000000000000.,.sellToken.:1\""
```

dfusion_rust_core/src/models/standing_order.rs
anxolin: As we agreed later, to avoid having to include ids in the model and adapting the code (including the mongo impl), I've implemented just the `From<(Entity, Vec<Entity>)>` for now

test/e2e-tests-standing-order.sh
anxolin: You are a grep virtuoso. I'll make the change

dfusion_rust_core/src/models/order.rs
```diff
@@ -26,15 +26,20 @@ pub struct Order {
     pub sell_amount: u128,
 }
 
+
 impl Order {
     pub fn from_encoded_order (
         account_id: u16,
-        bytes: Vec<u8>
+        bytes: &[u8; 26]
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FQh6NZWsFx1G1O%2Fgiphy.gif&token=awesome" data-gitmeme-token="awesome"></a>

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -123,31 +128,67 @@ impl From<&Arc<Log>> for StandingOrder {
...
 impl Into<Entity> for StandingOrder {
     fn into(self) -> Entity {
-        let mut entity = Entity::new();
-                
+        let mut entity = Entity::new();                
         entity.set("accountId", self.account_id.to_value());
         entity.set("batchIndex", self.batch_index.to_value());
         entity.set("validFromAuctionId", self.valid_from_auction_id.to_value());
+
+        // let orders = standing_order.get_orders().clone();
```
fleupold: Should we remove and replace with a comment/task?

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -123,31 +128,67 @@ impl From<&Arc<Log>> for StandingOrder {
...
-            batch_index,
-            valid_from_auction_id,            
-            orders: vec![] // Orders cannot be reconstructed just from the entity
+impl From<(Entity, Vec<Entity>)> for StandingOrder {
+    fn from(entities: (Entity, Vec<Entity>)) -> Self {
+        let batch_entity = entities.0;
+        let mut order_entities_iter = entities.1.into_iter();
+        let order_ids_value = batch_entity.get("orders").expect("it should contain order ids");
+
+        if let Value::List(order_ids) = order_ids_value {
```
fleupold: nit: to avoid the if/else block and keep the indentation level by linear you can do:

`let order_ids = order_ids_value.as_list.expect("orders should be a List")`

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -123,31 +128,67 @@ impl From<&Arc<Log>> for StandingOrder {
...
+
+            let account_id = u16::from_entity(&batch_entity, "accountId");
+            let batch_index = U256::from_entity(&batch_entity, "batchIndex");
+            let valid_from_auction_id = U256::from_entity(&batch_entity, "validFromAuctionId");
+
+            let orders = order_ids.iter()
+                .map(|order_id| {
+                    // Get matching order for the id
+                    let order_entity: Entity = order_entities_iter
+                        .find(|current_order| {
```
fleupold: Unnecessary O(n^2) but who cares when n is 10.

dfusion_rust_core/src/models/standing_order.rs
```diff
@@ -123,31 +128,67 @@ impl From<&Arc<Log>> for StandingOrder {
...
+        let order_ids_value = batch_entity.get("orders").expect("it should contain order ids");
+
+        if let Value::List(order_ids) = order_ids_value {
+            assert_eq!(
+                order_ids.len(),
+                order_entities_iter.len(),
+                "The entity should have the same number of orders as Vec<Entity>"
+            );
+
+            let account_id = u16::from_entity(&batch_entity, "accountId");
```
fleupold: ubernit: since not read anywhere else, maybe inline below?

dfusion_rust_core/src/models/order.rs
```diff
@@ -26,15 +26,20 @@ pub struct Order {
...
     ) -> Self {
-        let sell_token =  u8::from_le_bytes([bytes[25]]);
-        let buy_token = u8::from_le_bytes([bytes[24]]);
-        let sell_amount = read_amount(&bytes[12..24]);
-        let buy_amount = read_amount(&bytes[0..12]);
+        let sell_token =  u8::from_le_bytes([bytes[25]]); // 1 byte
+        let buy_token = u8::from_le_bytes([bytes[24]]);   // 1 byte
+        let sell_amount = read_amount(
+            &get_amount_from_slice(&bytes[12..24])         // 12 bytes
```
fleupold: Ok, I see how this is not the biggest win here, but still thanks for changing.

dfusion_rust_core/src/models/standing_order.rs
anxolin: I'm confused not sure which other into you are referring :) 

Also, super weird the order of the comments here, hahaha:

image

dfusion_rust_core/src/models/standing_order.rs
anxolin: ups..

dfusion_rust_core/src/models/standing_order.rs
anxolin: the cyclomatic complexity of this methods says:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FnEMmyUp4hl1QOzovKh%2Fgiphy.gif&token=thanks" data-gitmeme-token="thanks"></a>

dfusion_rust_core/src/models/standing_order.rs
anxolin: Why do you say unnecessary:
- Because we can do it linear, i.e. using hash map
- Because we think is enough with assuming the orders match the ids and in the same order, so it's just linear
- For any other reason

I think a good principle to keep in mind is:
image

I don't think this will be the bottleneck in this system

dfusion_rust_core/src/models/standing_order.rs
anxolin: I think this can be subjective

For me is more clear like this, this way, you separate the part where you extract the data from the construction of the object. 

I don't have a strong opinion here

dfusion_rust_core/src/models/order.rs
```diff
@@ -25,6 +26,27 @@ pub struct Order {
     pub sell_amount: u128,
 }
 
+impl Order {
+    pub fn from_encoded_order (
+        account_id: u16,
+        bytes: Vec<u8>
+    ) -> Self {
+        let sell_token =  u8::from_le_bytes([bytes[25]]);
```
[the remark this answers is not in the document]
anxolin: @bh2smith  I'll re-check this now, if it's true is not a Nit at all. It would be a big issue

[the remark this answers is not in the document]
anxolin: I think that was a super good catch! 
but, now I'm confused!

In theory `bytes[25]` is the buy token, not the sell token
image

However in the python implementation, we get the same results:
image


As the graph one:
image

But it seams the order in python parsing is fine:
image


I'll have to look more into this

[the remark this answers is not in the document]
anxolin: NVM: I was looking at it, to tired. Yep, there was a mistake.
Super good catch!

PR #213, opened 2019-08-08
GraphNode reader

fleupold:
At the moment, the only way to interact with the low level postgres representation from The Graph is via the [Store trait](https://github.com/graphprotocol/graph-node/blob/master/graph/src/components/store.rs#L650).

This trait is very elaborative in that it not only handles both read and write operations, it also covers both chain related storage as well as entity related storage.

The only thing we are interested on the driver side (and long term also on the listener side) is to read Entities. This means, we don't want to setup the entire dependency graph that the main graph component requires (GQL adapter, ethereum adapter, etc), but instead want to be able to instantiate a reader component just from a postgres connection (with a read-only user).

This PR takes the large store component from the graph repo and removes all write and chain related parts, leaving only a simple interface to query entities (code is copied). This interface will be used to implement database access inside the driver.

Long term, it would be nice if we could push such a separation of concerns inside the graph's main repository as their monolithic storage design seems quite inflexible.

### TestPlan

Make sure we can instantiate a version of this reader in the driver's main function just using a postgres connection.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
       before_script:
         - rustup component add clippy
       script:
-        - cargo clippy --all -- -D warnings
+        - cargo clippy --all --exclude graph-node-reader -- -D warnings
         - cargo check --verbose --all
         - cargo test --verbose --all
     - language: python
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -950,6 +950,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "graph-node-reader"
+version = "0.13.2"
+dependencies = [
+ "diesel 1.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "diesel-derive-enum 0.4.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "diesel-dynamic-schema 1.0.0 (git+https://github.com/diesel-rs/diesel-dynamic-schema?rev=a8ec4fb1)",
+ "graph 0.13.2 (git+https://github.com/graphprotocol/graph-node?tag=v0.13.2)",
+]
+
+[[package]]
 name = "graph-runtime-derive"
 version = "0.13.2"
 source = "git+https://github.com/graphprotocol/graph-node?tag=v0.13.2#602fba1af5322ceeb58cc09c7120bdacbaa5cbea"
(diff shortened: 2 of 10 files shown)


anxolin:
Which kind of information would need to get the driver?

I understand that basic queries about the indexed events. If that's the case, wouldn't be enough to just query using graphql?

Also, what would you think about creating just a `trait` abstracting the data access, with returns the info you need (model), following the Repository Pattern. 
The trivial implementation just would be instantiated with the URL of the graphql service.

Following this repository pattern, we could create a second repository that submit the solutions to the blockchain. This other would have also a trivial implementation, that would send the tx to the blockchain.

fleupold:
@anxolin 

> Which kind of information would need to get from the driver?

The information the driver needs in order to operate is abstracted away behind the [`DbInterface`](https://github.com/gnosis/dex-services/blob/0e939182673a015ea5fe07be75784fc0856ba42a/driver/src/db_interface.rs#L15-L36) trait. I'm not super familiar with the Repository Pattern, but maybe this is what you mean by it? We already have an implementation of this trait for the MongoDb and the series of PRs I'm working on will add an implementation for the Graph Postgres DB. The smart contract write operations are abstracted behind the [`SnappContract`](https://github.com/gnosis/dex-services/blob/7849c453507b7fe359b782fab76e4a4417f687ce/driver/src/contract.rs#L15-L46) trait.

> wouldn't be enough to just query using graphql?

One of the major points of this refactoring was to allow driver and listener to reuse code. In the listener I would prefer to use the postgres connection for reads as I'm not sure what consistency guarantees the GraphQl connection gives. Also, once we query for 2^24 account balances we might have to use some significant multi-query logic (traversing the merkle tree structure in the database) and using a native postgres driver is likely going to be more efficient than sending a GraphQL API network call for every fetch.

anxolin:
Yes, I meant not exposing the data layer complexities to the solver/driver. Yes I think abstractions like `DbInterface` is the way to go.

My comment was in the direction of, do we really need if we are doing basic queries. If you foresee complex queries or lots of overhead, it might be a good idea.  You know better.

Querying to the graph instead of directly to the DB has also it's advantages; like the API and the driver are more decoupled, it uses just a REST API and therefore is simpler to increase the number of instances in any of the ends: 
i.e. create 20 listeners and 3 solvers/drivers
i.e. 1 listener and 10 solver/drivers

PR #214, opened 2019-08-08
[Driver] Move DBInterface into shared library

fleupold:
This PR moves the db interface into the dfusion_core so that it can long term be reused by the listener and the driver. Moreover, it seems to make more sense to implement the Graph based database reader in the `core` workspace project since the `driver` project would otherwise have to depend on a lot of graph specific components (EntityFilter, Query, etc), which the core component already has.

Note, that I reimplemented the Error class for the Database (making it its own sub-error class). I'm also experimenting using the [error_chain](https://docs.rs/error-chain/0.12.1/error_chain/) for more convenient error handling across crates but am running into problems with our mocking framework (it requires all datatypes to be cloneable, yet error_chain errors are not easily cloneable)

### Test Plan
Since this is a refactor, travis should suffice.


anxolin:
Nice! That's a good move.

Also positive for allowing for example a CLI that depends on the core.

> Moreover, it seems to make more sense to implement the Graph based database reader in the core workspace project since the driver project would otherwise have to depend on a lot of graph specific components (EntityFilter, Query, etc), which the core component already has.

I made a comment in the other PR about the Graph node reader that relates to this.

Good stuff!

driver/src/db_interface.rs
```diff
@@ -1,200 +0,0 @@
-#[cfg(test)]
```
anxolin: I think sometimes git is not smart enough. In this situations, is better to instruct him with `git mv` so it knows that a file was moved, so it will show only the changes.

driver/src/mongo_db.rs
```diff
@@ -0,0 +1,138 @@
...
+pub struct MongoDB {
+    pub client: Client,
+}
+impl MongoDB {
+    pub fn new(db_host: String, db_port: String) -> Result<MongoDB, DatabaseError> {
+        let port = db_port
+            .parse::<u16>()
+            .map_err(
+                |e| DatabaseError::chain(ConfigurationError, "Couldn't parse port", e)
+            )?;
```
anxolin: Nice! love this "map_err" + "? operator"

PR #216, opened 2019-08-13
Minor changes to the db interface

fleupold:
This PR adds a method to obtain the AccountState by state index (previously only possible by state root). This is needed in the listener whenever we apply a state transition.

I also cleaned up the interface to use the full width U256 instead of using only the low u32 of our slot/state index.

### Test Plan

CI

### DIFF
--- a/driver/src/mongo_db.rs
+++ b/driver/src/mongo_db.rs
@@ -95,7 +95,7 @@ impl DbInterface for MongoDB {
         &self,
         state_index: &U256,
     ) -> Result<models::AccountState, DatabaseError> {
-        let query = doc!{ "stateIndex" => state_index.to_string() };
+        let query = doc!{ "stateIndex" => state_index.low_u64() };
         info!("Querying stateIndex: {}", query);
         self.get_balances_for_query(query)
     }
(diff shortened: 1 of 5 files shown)


driver/src/mongo_db.rs
```diff
@@ -84,36 +79,56 @@ impl DbInterface for MongoDB {
...
+        let query = doc!{ "stateHash" => format!("{:x}", state_root) };
+        info!("Querying stateHash: {}", query);
+        self.get_balances_for_query(query)
+    }
+
+    fn get_balances_for_state_index(
+        &self,
+        state_index: &U256,
+    ) -> Result<models::AccountState, DatabaseError> {
+        let query = doc!{ "stateIndex" => state_index.low_u64() };
```
anxolin: I'm curious, about this part.

I understand this would panic if the `state_index` don't fit in a u64. If this is the case, shouldn't it be u64 the argument type, so this restriction is part of the signature?

fleupold: The documentation just says

> Conversion to u64

I assume it takes to lowest 64 bits and truncates the rest (as in it doesn't panic if it overflows). The problem here is that in mongoDb we are storing state_index as a number and it can only represent up to 64 bits. For account balances we are using strings, which we should have used here too.
However, given that mongoDb will die at the end of this series of PRs it's probably not worth changing the schema.

anxolin: Yes, it's no worth changing it.

And I think you are also right about the panic, I didn't look it ok, I took a quick look and saw the panic from the next method in the docs:

image

PR #217, opened 2019-08-13
[The Graph] Postgres db interface implementation

fleupold:
This PR implements the DBInterface trait for the postgres backed graph reader. Once fully implemented, the driver should be able to switch from MongoDB to Postgres DB by just instantiating the other implementation [here](https://github.com/gnosis/dex-services/blob/master/driver/src/bin/main.rs#L18).

For now, I'm using this component in the existing graph listeners that are reading state from the database. I'd hope we can use the same component in the AuctionApplication handler that @bh2smith is working on. This will require finishing the two remaining unimplemented methods.

### Test Plan

Make sure E2E tests covering the exiting graph listeners are still running. 

Also, for withdraw & deposit tests we can already make the adjustment outlined above and have the driver use the graph listener instead of the python one.

### DIFF
--- a/dfusion_rust_core/src/database/graph_reader.rs
+++ b/dfusion_rust_core/src/database/graph_reader.rs
@@ -11,9 +11,15 @@ pub struct GraphReader {
 }
 
 impl GraphReader {
+    pub fn new(reader: Box<StoreReader>) -> Self {
+        GraphReader {
+            reader
+        }
+    }
+
     fn get_flux_of_slot(
         &self,
-        slot: u32,
+        slot: &U256,
         flux: &str,
     ) -> Result<Vec<models::PendingFlux>, DatabaseError> {
         let deposit_query = entity_query(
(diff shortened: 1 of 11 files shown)


listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -58,35 +57,21 @@ impl EventHandler for FluxTransitionHandler {
...
         info!(logger, "Received Flux AccountState Transition Event");
 
-        let account_query = util::entity_query(
-            "AccountState", EntityFilter::Equal("stateIndex".to_string(), state_index.to_value())
-        );
-        let mut account_state = AccountState::from(self.store
-            .find_one(account_query)?
-            .ok_or_else(|| failure::err_msg(format!("No state record found for index {}", &state_index)))?
-        );
+        let mut account_state = self.store
```
bh2smith: I like this! no need to specify query anymore!

dfusion_rust_core/src/database/graph_reader.rs
```diff
@@ -0,0 +1,113 @@
...
+        slot: &U256,
+    ) -> Result<Vec<models::PendingFlux>, DatabaseError> {
+        self.get_flux_of_slot(slot, "Withdraw")
+    }
+
+    fn get_orders_of_slot(
+        &self,
+        _: &U256,
+    ) -> Result<Vec<models::Order>, DatabaseError> {
+        unimplemented!()
```
bh2smith: I believe this Order collector has already been implemented... unless this function is meant to aggregate both orders and standing orders.... then I understand why this is empty for now.

fleupold: > I believe this Order collector has already been implemented...

I couldn't find where (it's not reading orders from db in the order listener). Can you link where I can find it?

> unless this function is meant to aggregate both orders and standing orders

It's not. It's supposed to do the semantically equivalent to its MongoDb counterpart (https://github.com/gnosis/dex-services/blob/master/driver/src/mongo_db.rs#L104-L110)

dfusion_rust_core/src/database/mod.rs
```diff
@@ -1,11 +1,13 @@
...
+mod graph_reader;
 
 use web3::types::{H256, U256};
 
 pub use error::*;
+pub use graph_reader::GraphReader;
 use super::models;
 
-pub trait DbInterface {
+pub trait DbInterface: Send + Sync + 'static {
```
josojo: What is the reason we need the Send etc here?

listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -105,22 +90,18 @@ impl EventHandler for FluxTransitionHandler {
...
+        let store = Arc::new(DbInterfaceMock::new());
 
         // Add previous account state and pending deposits into Store
         let existing_state = AccountState::new(H256::zero(), U256::zero(), vec![0, 0, 0, 0], 1);
-        let entity = existing_state.into();
-        store.apply_entity_operations(vec![EntityOperation::Set {
-            key: util::entity_key("AccountState", &entity),
-            data: entity
-        }], None).unwrap();
+        store.get_balances_for_state_index.given(U256::zero()).will_return(Ok(existing_state));
```
anxolin: nit: I'm usually more in favour in not chaining lot's of things in the same line. Even for tests. Otherwise you need to read all of it from left to right to understand.

I think in the book they also recommend to do only one thing in every line
```
store.get_balances_for_state_index
  .given(U256::zero())
  .will_return(Ok(existing_state)); 
```

listener/src/main.rs
```diff
@@ -210,12 +215,13 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
         SubgraphVersionSwitchingMode::Instant
     ));
 
+    let subgraph_name = SubgraphName::new("dfusion").unwrap();
```
anxolin: why here the subgraph is hardcoded and some lines above is loaded from a constant?

fleupold: Should have been the constant here. Good catch.

dfusion_rust_core/src/database/mod.rs
fleupold: DbInterface is a field on `FluxTransitionHandler`, which implements `EventHandler`, which itself is `Send + Sync + Debug`. Any field on such a datatype needs to be `Send + Sync` itself.

`EventHandler` is `Send + Sync` since it's a field on the RuntimeHost, which implements `pub trait RuntimeHost: Send + Sync + Debug`.

The static is actually unnecessary, good question/catch!

PR #219, opened 2019-08-14
[Easy] remove graph_node_reader

fleupold:
I ran into issues with clippy as it doesn't seem to allow me excluding the project when it is referenced as a local dependency (inside the same repo). Cf. the [travis failure](https://travis-ci.org/gnosis/dex-services/jobs/571308523#L205)

Since the graph_node_reader is a project that is mainly copied from https://github.com/graphprotocol/graph-node/tree/master/store I don't think addressing the clippy warnings is the right way to go (since we would have to redo them on every update).

I therefore decided to host the project in another [repo](https://github.com/fleupold/graph-node-reader) and reference it as an external git dependency in #217 

### Test plan

Nothing to test here, CI should still be passing.


anxolin:
Wouldn't it make sense to have it in Gnosis github if it's going to be part of the project?

fleupold:
> Wouldn't it make sense to have it in Gnosis github if it's going to be part of the project?

Yes, will move it

PR #220, opened 2019-08-16
172/settlement handler

### LINKED ISSUE #172: Auction Settlement handler [Rust]

bh2smith:
Closes #172 

Test Plan: Travis Build and E2E along with cargo test should suffice.

At this moment, however, there is no clean way of handling the `trivial solution` properly, since `from_bytes` can not accept an additional parameter about the number of orders. At this moment, it might make sense to emit the Trivial Solution directly from the contract as we would expect it to be parsed by the event listener.


dfusion_rust_core/src/database/graph_reader.rs
```diff
@@ -85,16 +88,60 @@ impl DbInterface for GraphReader {
...
+            };
+            let standing_order_option = self.reader
+                .find_one(standing_order_query)
+                .map_err(|e| DatabaseError::chain(ErrorKind::ConnectionError, "Could not execute query", e))?;
+
+            if let Some(standing_order) = standing_order_option {
+                let order_ids = standing_order.get("orders")
+                        .and_then(|orders| orders.clone().as_list())
+                        .ok_or_else(|| DatabaseError::new(ErrorKind::StateError, "No list orders found on standing order entity"))?;
+                    let relevant_order_query = entity_query("SellOrder", EntityFilter::In("id".to_string(), order_ids));
```
fleupold: Maybe we should capture here that we need to check that the orders are returned in the same sorting order as they were inserted. If not, we probably have to use some kind of other index to make sure they align with the order that the smart contract expects them to be in.

At the end of the day there should be a hash consistency check in the smart contract that ensures the orders are aligned correctly.

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -229,5 +247,68 @@ pub mod tests {
 
         assert_eq!(entity, state.clone().into());
         assert_eq!(state, AccountState::from(entity));
-  }
+    }
+
+    #[test]
+    fn test_apply_auction() {
+//        let one_eth = 1 * (10 as u128).pow(18);
```
fleupold: remove?

listener/src/event_handler/auction_settlement_handler.rs
```diff
@@ -0,0 +1,211 @@
...
+
+impl AuctionSettlementHandler {
+    pub fn new(store: Arc<DbInterface>) -> Self {
+        AuctionSettlementHandler {
+            store
+        }
+    }
+}
+
+impl Debug for AuctionSettlementHandler {
```
fleupold: Can you not `#[derive(Debug)]` that?

listener/src/event_handler/auction_settlement_handler.rs
```diff
@@ -0,0 +1,211 @@
...
+        _transaction: Arc<Transaction>,
+        log: Arc<Log>,
+    ) -> Result<Vec<EntityOperation>, Error> {
+        let mut event_data: Vec<u8> = log.data.0.clone();
+        info!(logger, "Parsing Auction Settlement event from {} bytes: {:?}", event_data.len(), event_data);
+        let auction_id = U256::pop_from_log_data(&mut event_data);
+        let state_index = U256::pop_from_log_data(&mut event_data).saturating_sub(U256::one());
+        let new_state_hash = H256::pop_from_log_data(&mut event_data);
+
+        // Strange info coming with packed bytes
```
fleupold: Can we make a more professional comment here? Maybe we can research what these values are by looking how ABI encoding works in detail.

listener/src/event_handler/auction_settlement_handler.rs
```diff
@@ -0,0 +1,211 @@
...
+
+#[cfg(test)]
+pub mod unit_test {
+    use super::*;
+    use dfusion_core::database::tests::DbInterfaceMock;
+    use dfusion_core::models::{AccountState, TOKENS, Order, BatchInformation, StandingOrder};
+    use web3::types::{Bytes, H256, U256};
+
+    #[test]
+    fn test_from_log() {
```
fleupold: Can we add tests for the unhappy cases as well?

test/e2e-tests-auction.sh
```diff
@@ -55,4 +55,20 @@ step_with_retry "Account 4 has now 4 of token 1" \
 "mongo dfusion2 --eval \"db.accounts.findOne({'stateHash': '$EXPECTED_HASH'}).balances[121]\" | grep -w 4000000000000000000"
 
 step_with_retry "Account 3 has now 52 of token 0" \
-"mongo dfusion2 --eval \"db.accounts.findOne({'stateHash': '$EXPECTED_HASH'}).balances[90]\" | grep -2 52000000000000000000"
+"mongo dfusion2 --eval \"db.accounts.findOne({'stateHash': '$EXPECTED_HASH'}).balances[90]\" | grep -w 52000000000000000000"
+
+step_with_retry "[theGraph] Account 4 has now 4 of token 1" \
```
fleupold: Should we add this to the standing order e2e test as well?

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -66,25 +68,40 @@ impl AccountState {
         self.state_hash = self.rolling_hash(self.state_index.low_u32());
         valid_withdraws
     }
+
+    pub fn apply_auction(&mut self, orders: &[Order], results: Solution) {
```
fleupold: This could have probably been its own PR.

dfusion_rust_core/src/database/graph_reader.rs
bh2smith: Yes I agree. Should we make this an issue somewhere in one or both of the projects?

listener/src/event_handler/auction_settlement_handler.rs
bh2smith: It appears I cannot. I would ask you the same for `FluxTransitionHandler`.

listener/src/event_handler/auction_settlement_handler.rs
bh2smith: Done!

test/e2e-tests-auction.sh
bh2smith: The standing order test doesn't really have anything to do with auction_settlement...

dfusion_rust_core/src/models/account_state.rs
bh2smith: Yes, I agree.

listener/src/event_handler/auction_settlement_handler.rs
bh2smith:
```suggestion
        // Meta-data coming from event attributes of type bytes (cf. Ethereum RLP Encoding)
```

PR #224, opened 2019-08-19
222/trivial auction

### LINKED ISSUE #222: Trivial Auction Handler
Handling the application of trivial auctions in a natural way requires a small refactoring of apply auction (i.e. looping over the executed amounts rather than orders) in `models::account_state.apply_auction`.

bh2smith:
Closes #222 

Slight refactoring of `apply_auction` which iterates over the executed amounts rather than the orders. Introducing a new e2e-test for trivial auctions.


dfusion_rust_core/src/models/account_state.rs
```diff
@@ -73,9 +73,11 @@ impl AccountState {
         let buy_amounts = results.executed_buy_amounts;
         let sell_amounts = results.executed_sell_amounts;
 
-        for (i, order) in orders.iter().enumerate() {
-            self.increment_balance(order.buy_token, order.account_id, buy_amounts[i]);
-            self.decrement_balance(order.sell_token, order.account_id, sell_amounts[i]);
+        for (i, execution_amounts) in buy_amounts.iter().zip(sell_amounts.iter()).enumerate() {
```
fleupold: nit: maybe easier to just iterate by index i and not zip onto the other amounts?

test/e2e-tests-trivial-auction.sh
```diff
@@ -0,0 +1,23 @@
...
+
+step "Setup" \
+"npx truffle exec scripts/setup_environment.js"
+
+EXPECTED_AUCTION=0
+
+step "Ensure no orders yet in auction slot 1" \
+"mongo dfusion2 --eval \"db.orders.find({'auctionId': ${EXPECTED_AUCTION}}).size()\" | grep -w 0"
+
+step "Place 2 orders that cannot be filled" \
```
fleupold: Is this really giving us the trivial solution with 0 bytes? I'd assume the trivial auction to be chosen if there is e.g. an error with the solver.

Otherwise, I feel like this should just be a subtest of the test auction test case.

dfusion_rust_core/src/models/account_state.rs
bh2smith: Ya, I could do that.

test/e2e-tests-trivial-auction.sh
bh2smith: This is surely giving the trivial solution (at least in the `NaiveSolver`). Why would no solution be an error with the solver?

I can put this into test auction if you'd prefer.

bh2smith: Note that trivial solution not only serves the purpose of applying an auction with no proposed solutions, but also is valid in the event that there is no possible order to match.

bh2smith: Another place where trivial solution is valid occurs when there is only one order in the batch.

dfusion_rust_core/src/models/account_state.rs
bh2smith: The reason I chose to do this is because I find there is a lack of symmetry if I loop over one of `buy_amounts.len()` or `sell_amounts.len()`. I am not fond of this lopsidedness.

test/e2e-tests-trivial-auction.sh
bh2smith: AHH @fleupold  I see what you mean. This gives the Trivial Solution that is currently provided by the Naive Solver. It isn't exactly relevant to the `0x0` trivial solution that is emitted by the contract. That trivial event is a problem for later... once the new driver code is ready.

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -73,7 +73,8 @@ impl AccountState {
         let buy_amounts = results.executed_buy_amounts;
         let sell_amounts = results.executed_sell_amounts;
 
-        for (i, order) in orders.iter().enumerate() {
+        for i in 0..buy_amounts.len() {
+            let order = &orders[i];
```
anxolin: It's not equivalent? If so, I would favour the previous approach

test/e2e-tests-trivial-auction.sh
anxolin: Not sure if I'm getting the comments, but If my understanding is correct, I would expect in this test, to test both:
- There's a solution with no matched orders if the constraints don't overlap
- Add a order that overlaps. My understanding is that Naive Solver, know how to do basic matching.

fleupold: Yes, @anxolin your understanding is correct. I'd argue that this test should be part of the general auction test (and not a separate e2e test file).

The thing I was referring to in my first comment was the "fallback" that is programmed into the [contract](https://github.com/gnosis/dex-contracts/blob/master/contracts/SnappAuction.sol#L294-L296), which will emit the trivial solution in case no-one has provided a solution for a batch in time (e.g. our driver is down).

I believe this one could be tested either in the general auction test script or in a separate one.

bh2smith: > The thing I was referring to in my first comment was the "fallback" that is programmed into the contract, which will emit the trivial solution in case no-one has provided a solution for a batch in time (e.g. our driver is down).

I only noticed later that this is what you were referring to. However that is not what this test is meant to handle. It is meant to deal with "apply_auction" on a trivial solution (as implemented in this codebase - and returned by naive solver).

> I believe this one could be tested either in the general auction test script or in a separate one.

This seems to say "leave it in a separate file as is, or put it into the existing auction test script"

dfusion_rust_core/src/models/account_state.rs
bh2smith: Actually it is not equivalent. The number of orders is potentially greater than the number of executed amounts. A perfect example of this is exactly the purpose of the PR (The Trivial Solution) where `orders.len() == 2` and `results.executed_buy_amounts.len() == 0`.

test/e2e-tests-trivial-auction.sh
fleupold: i think the test as it is right now should be in a e2e-apply-auction.sh. The test that checks the behavior if the solver is down and not providing solutions (trivial solution as smart contract level fallback) can be its own file.

bh2smith: Yes! So there is still some stuff to be done here. 

> i think the test as it is right now should be in a e2e-apply-auction.sh.

Does this mean our apply auction script should be updated to have a second auction running after the first test where the solver returns a trivial solution?

fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FJ336VCs1JC42zGRhjH%2Fgiphy.gif&token=yes" data-gitmeme-token="yes"></a>

bh2smith:
Closing since the contract no longer accepts trivial solutions!

Issue #178, opened 2019-07-16
Multi-Phase Auction - part II (Submit solution when appropriate)


bh2smith:
Driver should 
- detect if it is the winner of phase I and submit solution during the first half of the settlement period it it was (the winner).
- detect if the winner did not submit a full solution and be prepared to submit one in the second half of the settlement period.

Issue #177, opened 2019-07-16
Multi-Phase Auction - part I (Propose solution when appropriate)


bh2smith:
This is the second task to be executed along side [this](https://github.com/gnosis/dex-contracts/issues/150) issue from the contract code. It involves checking the contract for best solution proposal and submitting a proposal if what the driver has is better. Furthermore, driver will want to be later prepared to post this full solution (even if it doesn't make a proposal) during the second phase.

bh2smith:
closed by #225

PR #225, opened 2019-08-20
Two phase driver

fleupold:
This PR adjusts the driver to implement the two phase "bid + apply" auction model that the contracts are already implementing.

For that I'm updating the dex-contracts subrepo and refactoring the order processor into a stateful component that can save existing bids in memory.

Moving forward, we might want to persist the bids on disk so that we can recover from restarts, crashes between auctions.

Depends on https://github.com/gnosis/dex-contracts/pull/169 & https://github.com/gnosis/dex-contracts/pull/168

### Test Plan

Unit tests and e2e tests passing.


driver/src/util.rs
```diff
@@ -38,18 +38,31 @@ pub fn hash_consistency_check(
     Ok(())
 }
 
+pub enum ProcessingState {
```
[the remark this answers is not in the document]
fleupold: Yes, good point. We should add it when it becomes a possibility in the code.

driver/src/contract.rs
```diff
@@ -559,17 +597,31 @@ pub mod tests {
...
+            objective_value: U256,
+        ) -> Result<()> {
+            self.auction_solution_bid.called((
                 slot,
                 Val(prev_state),
                 Val(new_state),
                 Val(order_hash),
                 Val(standing_order_index),
-                Val(prices_and_volumes),
+                objective_value,
```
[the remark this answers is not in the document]
fleupold: Val is values of type Matcher (such as Any, SmallerThan, etc.) whereas passing the concrete value will always require an exact match.

Confer the last value in the type definition of auction_solution_bid is U256 and not Matcher<U256>.
```rs
pub auction_solution_bid: Mock<
            (
                U256,
                Matcher<H256>,
                Matcher<H256>,
                Matcher<H256>,
                Matcher<Vec<U128>>,
                U256,
            ),
            Result<()>,
        >,

```

driver/src/order_driver.rs
```diff
@@ -1,108 +1,170 @@
...
+
+use std::collections::HashMap;
+
+struct AuctionBid {
+    previous_state: H256,
+    new_state: H256,
+    solution: Solution,
+}
+
+pub struct OrderProcessor<'a, D: DbInterface, C: SnappContract> {
```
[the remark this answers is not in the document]
fleupold: Whenever we instantiate an order processor (e.g. using OrderProcessor::new) the lifetime of this object is implicitly defined by the lifetime of db, contract and price finder that we are passing along.

Since all of these are created in the same scope (cf. main.rs), the compiler can figure out that they anyways have the same lifetime and thus no explicit lifetime specification is needed. Let's chat offline to clarify more if needed.

driver/src/contract.rs
```diff
@@ -55,13 +55,20 @@ pub trait SnappContract {
         withdraw_hash: H256,
     ) -> Result<()>;
     fn apply_auction(
+        &self,
+        slot: U256,
+        prev_state: H256,
+        new_state: H256,
+        prices_and_volumes: Vec<u8>,
+    ) -> Result<()>;
+    fn auction_solution_bid(
```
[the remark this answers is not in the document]
anxolin: Maybe the code has changed since this comment and my review, but I see just the trait definition (64) and the implementation on 344. Nothing on 604

driver/src/contract.rs
```diff
@@ -327,13 +332,41 @@ impl SnappContract for SnappContractImpl {
...
             .call(
                 "applyAuction",
+                (slot, prev_state, new_state, prices_and_volumes),
+                account,
+                options,
+            )
+            .wait()
+            .map_err(DriverError::from)
+            .map(|_| ())
+    }
```
anxolin: It's not related to this PR. But some small comments about some logic here:
-  The gas limit is a bit hardcoded
-  `account_with_sufficient_balance` seams to find the first account with funds. Its fine, but I would find it ok to assume it's the first one. It seems like we foresee to set it up with a mnemonic/pks and different accounts with balances. 
- The "Not enough balance to send Txs" check only verifies that we have more than 0 ether. It should check we have enough for covering the `gas limit * gas price` we are setting
- Also, I see the API of this methods return a Result. Wouldn't it make sense to return a `Futute<Result<T>>`

driver/src/deposit_driver.rs
```diff
@@ -17,29 +19,30 @@ where
...
-            contract.apply_deposits(
-                slot,
-                state_root,
-                balances.state_hash,
-                contract_deposit_hash,
-            )?;
-            return Ok(true);
-        } else {
-            info!("Need to wait before processing deposit_slot {:?}", slot);
+        match batch_processing_state(slot, contract, &|i| {
```
anxolin: nit: Maybe is simpler to read if we do one thing in each line.  Like:

```rust
let processing_state = batch_processing_state(slot, contract, &|i| {
    contract.creation_timestamp_for_deposit_slot(i)
})?;

match processing_state {
   // ...
}
```

driver/src/util.rs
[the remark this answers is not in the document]
anxolin: One note about this, if we add the state Ben is suggesting in the future, in the matching of `withdraw_driver.rs:26`, you won't get a compiler error, cause it was handled the default to join the `AcceptsBids` and `AcceptsSolution`. 

It might be a good idea for this cases, match the specific cases, so if you add items to the enum, you are forced to handle the case. 

I think you can target multiple cases by using `AcceptsBids | AcceptsSolution => ...`

driver/src/util.rs
```diff
@@ -38,18 +38,31 @@ pub fn hash_consistency_check(
...
-    let slot_creation_block = creation_block(slot)?;
-    if slot_creation_block == U256::zero() {
-        return Ok(false);
+    let slot_creation_block_time = creation_block_time(slot)?;
+    if slot_creation_block_time == U256::zero() {
+        return Ok(ProcessingState::TooEarly);
+    }
+
+    let current_block_time = contract.get_current_block_timestamp()?;
+    if slot_creation_block_time + 360 < current_block_time {
```
anxolin: I would consider using a centralized set of constants for this (avoid magic numbers)

driver/src/deposit_driver.rs
josojo: sweet. I like it

driver/src/order_driver.rs
```diff
@@ -1,108 +1,170 @@
...
+                    if let Some(bid) = self.auction_bids.get(&slot) {
+                        self.contract.apply_auction(
+                            slot,
+                            bid.previous_state,
+                            bid.new_state,
+                            bid.solution.bytes(),
+                        )?;
+                        return Ok(ProcessResult::AuctionApplied(slot));
+                    }
+                    info!("Cannot find saved bid for auction slot {:?}", slot);
```
josojo: Should we return an error here?

driver/src/contract.rs
fleupold: Agree with 1-3.

For the last point the question is where to put the bound for asynchronous operations. In my view it's more intuitive for the driver to interact with a synchronous interface since it needs to wait on this interaction to complete anyways (and synchronous interfaces are easier to reason about). However, if we find ourselves issuing multiple transactions which could be run in parallel I'd be fine changing it to return non blocking futures.

fleupold:
Thanks for the good comments, will address shortly.

> What would it happen if you reboot?
@anxolin The worst case would be that we are the winning bidder of a previous auction that is now in the phase waiting for a solution submission. In this case we won't find the solution and thus likely get slashed.

driver/src/order_driver.rs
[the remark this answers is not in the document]
anxolin: This is some really advance stuff. I'll get there at some point when I get some time to resume the book!

driver/src/contract.rs
anxolin: Fair enough!

anxolin:
> I feel like the code is getting easier and easier to read. so sweet.

I'm sure! In addition I'm also sure you are getting more rustacean, hehe

anxolin:
> > @anxolin The worst case would be that....

Let's trust devops for these first versions. They are 99.99999% uptime guys

bh2smith:
Just for reference sake, this PR closes #177 and closes #178

PR #226, opened 2019-08-20
176 Part 1 - Point Driver at GraphDB

bh2smith:
The first part of removing the old listener is to remove all dependencies on it. This means that the driver should get all its DB query info from the new listener DB (the Graph).

**Test Plan:** existing tests.


driver/src/deposit_driver.rs
```diff
@@ -39,10 +43,10 @@ where
             )?;
             return Ok(true);
         } else {
-            info!("Need to wait before processing deposit_slot {:?}", slot);
+            debug!("Need to wait before processing deposit_slot {:?}", slot);
         }
     } else {
-        info!("All deposits are already processed");
+        info!("No pending deposit batches.");
```
bh2smith:
```suggestion
        debug!("No pending deposit batches.");
```

driver/src/mongo_db.rs
```diff
@@ -15,6 +15,7 @@ use web3::types::{H256, U256};
 pub struct MongoDB {
     pub client: Client,
 }
+
```
bh2smith:
```suggestion
```

driver/Cargo.toml
```diff
@@ -13,9 +13,11 @@ bson = "0.9.1"
 chrono = "0.4.6"
 ethabi = "6.1.0"
 dfusion_core = { path = "../dfusion_rust_core" }
+env_logger = "0.5.10"
```
fleupold: This doesn't look used.

driver/src/bin/main.rs
```diff
@@ -1,21 +1,31 @@
...
     simple_logger::init_with_level(log::Level::Info).unwrap();
-    let db_host = env::var("DB_HOST").expect("Specify DB_HOST env variable");
-    let db_port = env::var("DB_PORT").expect("Specify DB_PORT env variable");
-    let db_instance = MongoDB::new(db_host, db_port).unwrap();
+    let graph_logger = logger(false);
+    
+    let postgres_url = env::var("POSTGRES_URL").expect("Specify POSTGRES_URL variable");
+    let store_reader = GraphNodeReader::new(postgres_url, &graph_logger);
+    let db_instance = GraphReader::new(Box::new(store_reader));
     let contract = SnappContractImpl::new().unwrap();
```
anxolin: @bh2smith @fleupold  
Is it right to give the `GraphReader` to the run_driver_components?

It seams to me that the driver need to be more decoupled from the implementation. So it should send just a DbInterface, so we can provide any implementation we want:

image

anxolin: Same would apply to the other params. Shouldn't we pass traits, instead of implementations?

anxolin:
hehe:

There was a race condition :) 

image

Anyways, the PR was good, just added an existential question there

driver/src/bin/main.rs
fleupold: Yes it should. I believe this still dates back from our first days in rust and we had trouble getting it to compile when passing traits. But you are totally right. I'll submit a separate PR.

PR #227, opened 2019-08-21
176 Part 2 - Delete all python

### LINKED ISSUE #176: Phase out Django Event Listener
This task depends on having completed #175 and involves removing all the python code from this project completely. Note that there are other (configuration) components that might lie outside of the `event_listener` directory of the project!

bh2smith:
In summary: 

- Most of the changes here are deleting the entire `event_listener` directory (i.e. the directory containing all the python files).

- 

This is the final phase of removing the django-event-listener. 

Removing all mongoDB queries from the end to end tests. When reviewing, I would think it is important to pay special attention to the remaining E2E-tests. That is, we should ensure that our end to end tests are still covering all the cases we had hoped to. Ideally, if the individual Graph Listener PRs had covered the tests they were supposed to, we should have to worry about this, but it would not hurt (since this is the final phase of transition), to keep this in mind while reviewing!

Test Plan - e2e-testing also do a quick global search on the project for "python" and "mongo"

closes #176


driver/src/bin/main.rs
```diff
@@ -1,15 +1,13 @@
...
 use dfusion_core::database::GraphReader;
 
 use driver::contract::SnappContractImpl;
-use driver::price_finding::{
-    NaiveSolver,
-    LinearOptimisationPriceFinder,
-    PriceFinding
-};
+use driver::order_driver::OrderProcessor;
+use driver::price_finding::NaiveSolver;
```
bh2smith: @fleupold  is there any particular reason we are using 

```
use driver::order_driver::OrderProcessor;
use driver::price_finding::NaiveSolver;
use driver::price_finding::LinearOptimisationPriceFinder;
use driver::price_finding::PriceFinding;
```

over 
```
use driver::price_finding::{
    NaiveSolver,
    LinearOptimisationPriceFinder,
    PriceFinding
};
```

fleupold: No, and I actually prefer the latter. Must have been an oversight when resolving merge conflicts.

bh2smith: Cool, I'll come back to this someday and deal with it at some point!

PR #228, opened 2019-08-27
Annotate withdraws with a valid flag when processing State Transition

fleupold:
This is needed, in order to query the withdraws that were successfully applied (withdraws that would lead to a negative balance are ignored).

All valid withdraws are hashed into a merkle tree and submitted to the smart contract (cf. [here](https://github.com/gnosis/dex-contracts/blob/master/contracts/libraries/SnappBaseCore.sol#L232). Later, when people want to claim a withdraw they need to provide an inclusion proof of their withdraw request in that merkle root (cf. [here](https://github.com/gnosis/dex-contracts/blob/master/contracts/libraries/SnappBaseCore.sol#L255)).

In order to compute such proofs for our e2e tests we use [this script](https://github.com/gnosis/dex-contracts/blob/master/scripts/claim_withdraw.js#L58), that relies on being able to query only "valid" withdraws.

This is why we need to add this flag when the listener processes flux transitions.

### Test Plan

Run deposit/withdraw tests, then run following query:

```
query {
  withdraws {
    id
    accountId
    tokenId
    amount
    slot
    slotIndex
    valid
  } 
}
```

and observe the following repsonse:

```
{
  "data": {
    "withdraws": [
      {
        "accountId": 2,
        "amount": "18000000000000000000",
        "id": "0 - 0",
        "slot": "0",
        "slotIndex": 0,
        "tokenId": 2,
        "valid": true
      }
    ]
  }
}
```


listener/src/event_handler/flux_transition_handler.rs
```diff
@@ -73,16 +74,30 @@ impl EventHandler for FluxTransitionHandler {
                     .store
                     .get_withdraws_of_slot(&slot)
                     .map_err(|e| failure::err_msg(format!("{}", e)))?;
-                account_state.apply_withdraws(&withdraws);
+                let valid_withdraws = account_state.apply_withdraws(&withdraws);
+                updated_records = valid_withdraws
```
josojo: In future refactorings, I might be more confident about the correctness of this logic part, if we have a unit test for this logic

PR #231, opened 2019-08-29
Less caching for rust

fleupold:
This should hopefully reduce the amount of time our unit tests run on CI.


fleupold:
With only rust cargo subcommand caching (as suggested [here](https://levans.fr/rust_travis_cache.html)) the cache shrinks from 2.9GB to 29MB (99% reduction). However, compilation takes very long (500s for regular build + 400s for test build).

Will try the approach suggested here: https://travis-ci.org/rust-analyzer/rust-analyzer/jobs/463540847/config

fleupold:
It's now 1.4GB (50% reduction) and takes in total 7 min 5 sec (previously 20min)

PR #232, opened 2019-08-29
Using docker-compose kill to speed up restarts

fleupold:
Restarting the containers between tests takes ~60s at the moment (cf. https://travis-ci.org/gnosis/dex-services/jobs/578335529#L530)

With this change I was able to reduce the time between restarts to <15s. This should lead to a speedup of ~1.5mn for our overall travis time.

### TestPlan

Observe restart-container.sh timing in travis.


fleupold:
Restart took 13.5s instead of almost 60s

https://travis-ci.org/gnosis/dex-services/jobs/578352107#L537

PR #234, opened 2019-09-03
[Easy] misc clippy fixes

fleupold:
This PR fixes a bunch of clippy suggestions which are not land blocking (since they are minor). More specifically:

- trait objects without an explicit `dyn` are deprecated
- &vec![...] == &[...]
- complex types in mocks should be extracted (typedef)
- new methods without arguments should be implemented via Default trait
- don't use clone on copy types

### DIFF
--- a/driver/src/contract.rs
+++ b/driver/src/contract.rs
@@ -322,7 +322,10 @@ impl SnappContract for SnappContractImpl {
         new_state: H256,
         prices_and_volumes: Vec<u8>,
     ) -> Result<()> {
-        debug!("Applying Auction with result bytes: {:?}", &prices_and_volumes);
+        debug!(
+            "Applying Auction with result bytes: {:?}",
+            &prices_and_volumes
+        );
         let account = self
             .account_with_sufficient_balance()
             .ok_or("Not enough balance to send Txs")?;
@@ -398,6 +401,24 @@ pub mod tests {
     use mock_it::Matcher::*;
     use mock_it::Mock;
 
+    type ApplyDeopositArguments = (U256, Matcher<H256>, Matcher<H256>, Matcher<H256>);
+    type ApplyWithdrawArguments = (
+        U256,
+        Matcher<H256>,
+        Matcher<H256>,
+        Matcher<H256>,
+        Matcher<H256>,
+    );
+    type ApplyAuctionArguments = (U256, Matcher<H256>, Matcher<H256>, Matcher<Vec<u8>>);
+    type ApplySolutionArguments = (
+        U256,
+        Matcher<H256>,
+        Matcher<H256>,
+        Matcher<H256>,
+        Matcher<Vec<U128>>,
+        U256,
+    );
+
     #[derive(Clone)]
     pub struct SnappContractMock {
         pub get_current_block_timestamp: Mock<(), Result<U256>>,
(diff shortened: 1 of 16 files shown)


driver/src/order_driver.rs
```diff
@@ -1,18 +1,13 @@
...
 
 use dfusion_core::database::DbInterface;
 use dfusion_core::models::{
-    AccountState,
-    ConcatenatingHashable,
-    Order,
-    RollingHashable,
-    Serializable,
-    Solution,
+    AccountState, ConcatenatingHashable, Order, RollingHashable, Serializable, Solution,
```
bh2smith: I kinda prefer the look of the vertical/stacked imports here...

driver/src/order_driver.rs
```diff
@@ -85,8 +80,8 @@ impl<'a, D: DbInterface, C: SnappContract> OrderProcessor<'a, D, C> {
                     }
                     return Err(DriverError::new(
                         &format!("Cannot find saved bid for auction slot {:?}", slot),
-                        ErrorKind::StateError
-                    ))
+                        ErrorKind::StateError,
+                    ));
```
bh2smith: How is it even possible that we did not have a semi-colon here before?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -333,9 +331,9 @@ pub mod tests {
 
         let expected_solution = models::Solution {
             surplus: U256::from_dec_str("15854632034944469292777429010439194350").ok(),
-            prices: vec![14024052566155238000, 1526784674855762300],
-            executed_sell_amounts: vec![0, 318390084925498118944],
-            executed_buy_amounts: vec![0, 95042777139162480000],
+            prices: vec![14_024_052_566_155_238_000, 1_526_784_674_855_762_300],
```
bh2smith: Nice one!

driver/src/withdraw_driver.rs
```diff
@@ -15,10 +15,9 @@ where
     let withdraw_slot = contract.get_current_withdraw_slot()?;
 
     info!("Current top withdraw_slot is {:?}", withdraw_slot);
-    let slot = find_first_unapplied_slot(
-        withdraw_slot,
-        &|i| contract.has_withdraw_slot_been_applied(i),
-    )?;
+    let slot = find_first_unapplied_slot(withdraw_slot, &|i| {
```
bh2smith: I guess this looks better...

driver/src/contract.rs
```diff
@@ -322,7 +322,10 @@ impl SnappContract for SnappContractImpl {
         new_state: H256,
         prices_and_volumes: Vec<u8>,
     ) -> Result<()> {
-        debug!("Applying Auction with result bytes: {:?}", &prices_and_volumes);
```
bh2smith: I don't see that this was exceeding the line width (considering there are others added here that were longer). This seems like an unnecessary change.

driver/src/order_driver.rs
fleupold: I guess for return statements, you don't need to have a `;`

driver/src/contract.rs
fleupold: rustfmt breaks it for me whenever I save

PR #235, opened 2019-09-04
Refactoring account balances to be represented as a sparse HashMap

fleupold:
We currently represent balances inside the AccountState model as a continuous array with values in the form of 

```
acc1_token1, acc1_token2, ... acc1_tokenN, acc2_token1 ..., accM_tokenN
```

While this works well for a small number of accounts, this doesn't scale to the new dfusion model, where our address space is 256 bits. 

The number of participants in a batch is upper bound by the number of orders in a batch. We only need to pass the balances to the solver of accounts that are participating in the current auction. This suggests that we should be using a sparse HashMap rather than a list over the full address space to represent balances.

For backwards compatibility, I kept the interface that allows creating the AccountState object from a continuous list and also kept the way we store the state in the GraphDb (this will likely change as we plan to increase or address space to at least 24 bits there as well).

The new driver component will be able to expose another new method, passing in the sparse HashMap directly.

### Testplan

existing unit tests and e2e tests

### DIFF
--- a/dfusion_rust_core/src/models/account_state.rs
+++ b/dfusion_rust_core/src/models/account_state.rs
@@ -2,6 +2,7 @@ use byteorder::{BigEndian, WriteBytesExt};
 use graph::data::store::Entity;
 use serde_derive::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
+use std::collections::{hash_map::Entry, HashMap};
 use std::sync::Arc;
 use web3::types::{Log, H256, U256};
 
@@ -14,7 +15,7 @@ use super::util::*;
 pub struct AccountState {
     pub state_hash: H256,
     pub state_index: U256,
-    balances: Vec<u128>,
+    balances: HashMap<U256, HashMap<u8, u128>>, // UserId => (TokenId => balance)
     pub num_tokens: u8,
 }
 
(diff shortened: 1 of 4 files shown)


dfusion_rust_core/src/models/account_state.rs
```diff
@@ -23,39 +24,64 @@ impl AccountState {
...
+                .collect(),
             num_tokens,
         }
     }
-    fn balance_index(&self, token_id: u8, account_id: u16) -> usize {
-        self.num_tokens as usize * account_id as usize + token_id as usize
+    pub fn get_balance_vector(&self) -> Vec<u128> {
+        let mut i = U256::zero();
+        let mut result = vec![];
+        while let Some(account) = self.balances.get(&i) {
```
josojo: If you wanna introduce more syntax sugar, you might wanna implement the iterate trait for the balances and then simply iterate of it.
Not 100% sure, that this would work. But if it works, it would be more sugary.

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -92,14 +118,30 @@ impl AccountState {
...
     }
+
+    fn modify_balance<F>(&mut self, account_id: u16, token_id: u8, func: F)
+    where
+        F: FnOnce(&mut u128),
+    {
+        match self.balances.entry(U256::from(account_id)) {
+            Entry::Occupied(mut account) => match account.get_mut().entry(token_id) {
+                Entry::Occupied(mut balance) => func(balance.get_mut()),
+                Entry::Vacant(_) => panic!(
```
josojo: Why aren't vacant balances just to be assumed 0 and we apply the function on 0?

fleupold: Because at the moment we don't have to use this assumption and so I wanted the logic to be rather strict (have the behavior as tight as possible and as lose as necessary). If there is a good use case in the future where we insert into 0 balances that don't yet explicitly exist I think we can change this behaviour.

dfusion_rust_core/src/models/account_state.rs
fleupold: When iterating, I wouldn't be able to make sure that the ids are indeed adjacent. `{1: [0,0], 42, [0,0]}` would lead the same result as `{1: [0,0], 2, [0,0]}`, but shouldn't be a valid vector

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -14,48 +15,85 @@ use super::util::*;
...
         }
     }
-    fn balance_index(&self, token_id: u8, account_id: u16) -> usize {
-        self.num_tokens as usize * account_id as usize + token_id as usize
+
+    /// Returns the balances as a vector, under the assumptions
+    /// that self.balances contains consecutive addresses (0..num_accounts)
+    /// as keys, and each address has a balance for all (0..num_tokens) tokens.
+    ///
+    /// Panics, if this is not the case.
```
anxolin: nit: I think the panics have a special syntax

https://doc.rust-lang.org/1.30.0/book/first-edition/documentation.html#special-sections

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -14,48 +15,85 @@ use super::util::*;
 pub struct AccountState {
     pub state_hash: H256,
     pub state_index: U256,
-    balances: Vec<u128>,
+    balances: HashMap<U256, HashMap<u8, u128>>, // UserId => (TokenId => balance)
```
anxolin: This looks like a complex structure: Map of a Map
Is that better than modeling this with traits that model the `AccountDetails` and allows you to get the balance for each token. It would be this way just a `HashMap<U256, AccountDetails>`

dfusion_rust_core/src/models/account_state.rs
fleupold: # Panics above the panic sentence, good call.

dfusion_rust_core/src/models/account_state.rs
fleupold: Nice suggestion. Would be a bit of effort to expose this extra interface around it, but I agree that it'd be much cleaner.

PR #236, opened 2019-09-05
Fixing linear optimzer

fleupold:
Using the linear solver was broken (like everything that doesn't run as part of CI usually is), because the path from which the driver is run changed one up (when moving from rust projects to rust workspace).

I used this opportunity to adjust the repo to the new gitlab repo, add some explanation in the README and pin the repo against a version that we know works with the old prototype.

When running the stablecoin excange, we'll likely make the tag/rev an environment variable so it can be different for StableX and dƒusion.

### Test Plan

Follow the steps in the readme to make sure the linear optimiser is giving good auction results (I ran the auciton e2e test to try it out).

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ cargo test --lib
 
 For this to work, you will need to have read access to the [solver's gitlab repo](https://gitlab.gnosisdev.com/dfusion/batchauctions). You will also need to have SSH Key authentication for your gitlab account enabled (see [tutorial](https://docs.gitlab.com/ee/ssh/))
 
-In your top-level git folder, create a `.ssh/` folder and move or copy the private and public ssh key that are registered with your gitlab account in there.
+In your top-level git folder, create a `.ssh/` folder and move or copy the private and public ssh key (`id_rsa` & `id_rsa.pub`) registered with your gitlab account in there.
 Then, run
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -105,6 +105,19 @@ cd driver
 cargo test --lib
 ```
 
+## Running with linear optimization solver
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -105,6 +105,19 @@ cd driver
 cargo test --lib
 ```
 
+## Running with linear optimization solver
+
+For this to work, you will need to have read access to the [solver's gitlab repo](https://gitlab.gnosisdev.com/dfusion/batchauctions). You will also need to have SSH Key authentication for your gitlab account enabled (see [tutorial](https://docs.gitlab.com/ee/ssh/))
+
+In your top-level git folder, create a `.ssh/` folder and move or copy the private and public ssh key that are registered with your gitlab account in there.
```
josojo: Might wanna mention that their names have to be id_rsa and id_rsa.pub

fleupold: I think that's how they are always called, but sure.

bh2smith:
@josojo 
> where all solvers find the optimal solution

Not sure if we can use the proprietary solver in our CI since it would require an access key that isn't public.

josojo:
> @josojo
> 
> > where all solvers find the optimal solution
> 
> Not sure if we can use the proprietary solver in our CI since it would require an access key that isn't public.

Sure, but at least locally, I want to tests it easily

bh2smith:
Also, not that this is relevant atm, but I am also failing the deposit-test

```
✔ Setup 
✔ Deposit 18 of token 2 for user 2 
❌ Deposit was added to graph DB 
Command: source ../test/utils.sh && query_graphql     "query {         deposits(where: { accountId: 2}) {             amount         }     }" | grep 18000000000000000000
```

PR #237, opened 2019-09-05
Only passing the balances that are required by the solver to python

fleupold:
At the moment, we serialise the entire account state for the solver's input json, where really we only need the balances of the accounts that are trading.

This PR changes it to make it more sparse. Note, that we need to provide the buyBalance as well since otherwise, I'm running into the following python error:

```sh
[INFO: D_verify_SNARK.py:60 | checkAccountBalances()]  === ACCOUNT BALANCES ===
| Traceback (most recent call last):
|   File "./batchauctions/scripts/optimize_e2e.py", line 125, in <module>
|     args.solverTimelimit)
|   File "./batchauctions/scripts/optimize_e2e.py", line 93, in run
|     return verifySNARK(jsonFile_03, 1e13)
|   File "/app/batchauctions/scripts/verification/D_verify_SNARK.py", line 192, in run
|     c3 = checkAccountBalances(accounts, orders)
|   File "/app/batchauctions/scripts/verification/D_verify_SNARK.py", line 71, in checkAccountBalances
|     _acc[o['accountID']][o['buyToken']] += Decimal(o['execBuyAmount'])
| KeyError: 'token1'
```

cc @twalth3r I guess the feature of implicitly assuming 0 was added after April 16th (which is the version I'm currently pegged to)? Anyway, no big deal - we can do this logic in rust.

### Test Plan

Same as #236

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ cargo test --lib
 
 For this to work, you will need to have read access to the [solver's gitlab repo](https://gitlab.gnosisdev.com/dfusion/batchauctions). You will also need to have SSH Key authentication for your gitlab account enabled (see [tutorial](https://docs.gitlab.com/ee/ssh/))
 
-In your top-level git folder, create a `.ssh/` folder and move or copy the private and public ssh key that are registered with your gitlab account in there.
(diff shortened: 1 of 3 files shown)


twalth3r:
I see -  the crash occurs in the verification script (which doesn't do anything but checking constraints). So if everything else (i.e., solving and rounding) works fine, then this could verification could be commented out as a quick fix.

PR #238, opened 2019-09-06
Implement fee mechanism for naive solver

fleupold:
This PR adds an optional fee parameter to the NaiveSolver. The logic is changed as follows:
1) We make "space" for the fee to be taken. This is we change the limit price to be more permissive. To achieve this we either require more stuff to be bought or less stuff to be sold (both would make the deal "better" for the user).
2) We do the matching logic as before
3) We post-process the volumes taking away the fees (increasing sellVolume if feeToken is sold, reducing buyVolume if fee token is bought)

### Example

```
Order {
    account_id: 0,
    sell_token: 0,
    buy_token: 1,
    sell_amount: 20000,
    buy_amount: 9990,
},
Order {
    account_id: 0,
    sell_token: 1,
    buy_token: 0,
    sell_amount: 9990,
    buy_amount: 19960,
},
```

Adjusted orders after step 1)
```
Order {
    account_id: 0,
    sell_token: 0,
    buy_token: 1,
    sell_amount: 19980,
    buy_amount: 9990,
},
Order {
    account_id: 0,
    sell_token: 1,
    buy_token: 0,
    sell_amount: 9990,
    buy_amount: 19980 ,
},
```

Normal matching step 2) gives the following volumes:

```
Sell: [19980, 9990]
Buy [9990, 19980]
```

Adjusted volumes after step 3):

```
Sell: [20000, 9990]
Buy [9990, 19960]
```

Leading to a surplus of 40 units of token 0 (fee that goes to the solution finder)

### Test Plan

Added example as a unit test (comes from the [dex-smart contract repo](https://github.com/gnosis/dex-contracts/blob/master/test/stablecoin_converter.js#L38))

Also making sure current build (E2E tests) aren't broken

### DIFF
--- a/driver/src/price_finding/naive_solver.rs
+++ b/driver/src/price_finding/naive_solver.rs
@@ -4,7 +4,7 @@ use dfusion_core::models::{AccountState, Order, Solution, TOKENS};
 
 use crate::price_finding::error::PriceFindingError;
 
-use super::price_finder_interface::PriceFinding;
+use super::price_finder_interface::{Fee, PriceFinding};
 
 pub enum OrderPairType {
     LhsFullyFilled,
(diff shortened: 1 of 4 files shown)


driver/src/price_finding/naive_solver.rs
```diff
@@ -4,39 +4,52 @@ use dfusion_core::models::{AccountState, Order, Solution, TOKENS};
...
+    ) -> Option<OrderPairType>;
     fn opposite_tokens(&self, other: &Order) -> bool;
     fn have_price_overlap(&self, other: &Order) -> bool;
     fn surplus(&self, buy_price: u128, exec_buy_amount: u128, exec_sell_amount: u128) -> U256;
 }
 
 impl Matchable for Order {
-    fn attracts(&self, other: &Order) -> bool {
+    fn attracts(&self, other: &Order, fee: &Option<Fee>) -> bool {
+        // We can only match orders that touch the fee token
```
bh2smith: Is this desirable? How are we going to match orders that don't?

driver/src/price_finding/naive_solver.rs
```diff
@@ -145,6 +188,30 @@ impl PriceFinding for NaiveSolver {
     }
 }
 
+fn u128_to_u256(x: u128) -> U256 {
```
bh2smith: any reason for moving this? Do helper functions generally go below or above in Rust?

fleupold: I wanted to put them together, and since they are private I felt like they shouldn't be at the top of the file. Usually when I jump to a location I want to see the public structs and traits that I might use outside of this model, and not some model specific private helper method.

But not sure if this is general style.

driver/src/price_finding/naive_solver.rs
fleupold: Yes, otherwise we won't be able to produce a solution that is accepted by the smart contract. Therefore not matching (and thus potentially creating a more trivial but correct solution is better in my opinion).

PR #239, opened 2019-09-07
Refactor/Generalize smart contract interface

bh2smith:
Creating dedicated "contracts" module in driver which interfaces all relevant contracts for the protocol (currently _dFusion's SnappAuction_ and eventually _stableX's StableCoinConverter_).

This first PR introduces the module, moves existing contracts into it along with all relevant imports. Additionally, includes generalized ContractImpl that can be used when we introduce the StablConinConverter contract.

**Test Plan:**

Should see that no breaking changes were introduced.


driver/src/contracts/dfusion.rs
```diff
@@ -166,6 +166,19 @@ impl SnappContract for SnappContractImpl {
             .map_err(DriverError::from)
     }
 
+    fn calculate_order_hash(&self, slot: U256, standing_order_index: Vec<U128>) -> Result<H256> {
```
bh2smith: This was moved because of a warning that the trait functions were not implemented in the order they were defined.

driver/src/contracts/base_contract.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::fs;
+
+use web3::contract::Contract;
+use web3::futures::Future;
+use web3::types::Address;
+
+use crate::error::DriverError;
+
+type Result<T> = std::result::Result<T, DriverError>;
+// TODO - Generalize the ContractImpl struct out from each of the two contract handlers
```
fleupold: Is this still a to do?

driver/src/contracts/base_contract.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[allow(dead_code)] // event_loop needs to be retained to keep web3 connection open
+pub struct SmartContract {
```
fleupold: Maybe we should call this `ContractBase` or `BaseContract`? Smart Contract sounds a bit too powerful in my view.

driver/src/contracts/dfusion.rs
```diff
@@ -1,17 +1,30 @@
...
 
 type Result<T> = std::result::Result<T, DriverError>;
 
+pub struct SnappContractImpl {
+    base: SmartContract
+}
+
+impl SnappContractImpl {
+    pub fn new(base: SmartContract) -> Self {
+        // Should we assert that the contract is indeed a Snapp contract?
```
fleupold: I think if we wanted to, we should pass no arguments and read the abi file and contract address from env here.

fleupold:
I guess the title/description of the PR is not super applicable anymore

driver/src/contracts/base_contract.rs
bh2smith: Nope sorry!

driver/src/contracts/base_contract.rs
bh2smith: Yes! Good thinking!

driver/src/contracts/dfusion.rs
bh2smith: Ok. cool!

bh2smith:
> I guess the title/description of the PR is not super applicable anymore

Yes, Will come up with a new one.

driver/src/bin/main.rs
```diff
@@ -14,17 +15,22 @@ use graph::log::logger;
...
 fn main() {
-    // driver logger
     simple_logger::init_with_level(log::Level::Info).unwrap();
     let graph_logger = logger(false);
     let postgres_url = env::var("POSTGRES_URL").expect("Specify POSTGRES_URL variable");
     let store_reader = GraphNodeReader::new(postgres_url, &graph_logger);
     let db_instance = GraphReader::new(Box::new(store_reader));
-    let contract = SnappContractImpl::new().unwrap();
+
+    let contract_json = fs::read_to_string("dex-contracts/build/contracts/SnappAuction.json").unwrap();
```
josojo: This logic is used twice. Maybe you can refactor it.

driver/src/contracts/snapp_contract.rs
```diff
@@ -1,17 +1,34 @@
...
+pub struct SnappContractImpl {
+    base: BaseContract
+}
+
+impl SnappContractImpl {
+    pub fn new() -> Self {
+        let contract_json = fs::read_to_string("dex-contracts/build/contracts/SnappAuction.json").unwrap();
+        let address = env::var("SNAPP_CONTRACT_ADDRESS").unwrap();
+        SnappContractImpl {
+            base: BaseContract::new(address, contract_json).unwrap()
```
fleupold: Now, that this can fail, I'd suggest to return a Result<Self> from this method.

bh2smith: That is cool, although I have since changed the implementation and moved all the environment variables into the new (as suggested).

PR #240, opened 2019-09-10
StableX Contract Scaffolding

bh2smith:
Introducing the hollow/unimplemented functionality for the stableX contract interface to the Rust Driver. In summary, the driver will require three EVM interactions

- current_auction_index; query/read returning int
- auction_data; query/read returning Orders and Relevant Accounts
- submit_solution; write auction results/state transition based on solver output.


driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,138 @@
...
+        StableXContractImpl {
+            contract
+        }
+    }
+}
+
+pub trait StableXContract {
+    fn get_current_auction_index(&self) -> Result<U256>;
+    // TODO - auction_data should parse and return relevant orders and account balances.
+    fn get_auction_data(&self, _index: u32) -> Result<H256>;
```
fleupold: Let's define the interface the way it's intended to be in the end. This should return `Result<(Vec<Order>, AccountState)>`

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    fn submit_solution(
+        &self,
+        _batch_index: u32,
+        _owners: Vec<H256>,
```
fleupold: addresses are H160

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+pub trait StableXContract {
+    fn get_current_auction_index(&self) -> Result<U256>;
+    // TODO - auction_data should parse and return relevant orders and account balances.
+    fn get_auction_data(&self, _index: u32) -> Result<H256>;
+
+    fn submit_solution(
+        &self,
+        _batch_index: u32,
```
fleupold: I think we should either have both batch_index types as u32 or U256 (`get_current_auction_index ()`). Not sure which one makes more sense. @josojo what was your thought around this? I can see it being mixed in the contract as well (e.g. [here as U256](https://github.com/gnosis/dex-contracts/blob/master/contracts/StablecoinConverter.sol#L148) and [here as u32](https://github.com/gnosis/dex-contracts/blob/master/contracts/StablecoinConverter.sol#L162)).

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[allow(dead_code)] // event_loop needs to be retained to keep web3 connection open
+struct StableXContractImpl {
+    contract: BaseContract
+}
+
+#[allow(dead_code)] // event_loop needs to be retained to keep web3 connection open
+impl StableXContractImpl {
+    pub fn new(contract: BaseContract) -> Self {
+        // Should we assert that the contract is indeed a StableX contract?
```
fleupold: Let's use the same approach here as for the SnappBase

driver/src/contracts/stablex_contract.rs
bh2smith: Ok. I have never heard of H160, but thats cool.

bh2smith:
@fleupold - I would much rather use U128 or U256 for the `batchIndex`/`auctionId` because I am running into the following issue with u32:

```
self.base.contract
            .query("getCurrentStateIndex", (), None, Options::default(), None)
            .wait()
             ^^^^ the trait `web3::contract::tokens::Tokenizable` is not implemented for `u32`
```

which could easily be avoided.

driver/src/contracts/stablex_contract.rs
bh2smith: We could easily avoid the following error by using either of `U128` or `U256`

```
self.base.contract
            .query("getCurrentStateIndex", (), None, Options::default(), None)
            .wait()
             ^^^^ the trait `web3::contract::tokens::Tokenizable` is not implemented for `u32`
```

I have also noticed a place in the contract where `uint` is used instead of `uint32`... surprised this does cause any solidity errors.

fleupold: I'm fine using U256 everywhere. Can you adjust the smart contracts as well to be consistent?

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,152 @@
...
+        _volumes: Vec<U128>,
+        _prices: Vec<U128>,
+        _token_ids_for_price: Vec<u16>,
+    ) -> Result<()>;
+}
+
+impl StableXContract for StableXContractImpl {
+    fn get_current_auction_index(&self) -> Result<u32> {
+        unimplemented!();
+//        self.base.contract
```
fleupold: Can we keep this commit clean and remove the code that will be added later?

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,152 @@
...
+#[allow(dead_code)]
+struct StableXContractImpl {
+    base: BaseContract
+}
+
+#[allow(dead_code)]
+impl StableXContractImpl {
+    pub fn new() -> Result<Self> {
+        let contract_json = fs::read_to_string("dex-contracts/build/contracts/StablecoinConverter.json").unwrap();
+        let address = env::var("STABLEX_CONTRACT_ADDRESS").unwrap();
```
fleupold: Given that you return a result, you shouldn't need to use any unwraps here. In case any line fails, new should return the error flavor of Result.

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,152 @@
...
+use crate::error::DriverError;
+
+use super::base_contract::BaseContract;
+
+use std::env;
+use std::fs;
+
+type Result<T> = std::result::Result<T, DriverError>;
+
+#[allow(dead_code)]
```
fleupold: maybe a todo to remove this once base is used?

Alternatively, I'd also be fine to implement `get_current_auction_index` and `submit_solution` already, given they are very simple. Just `get_auction_data` will be more involved since it will require a good amount of parsing.

driver/src/contracts/stablex_contract.rs
bh2smith: Yes sure,

driver/src/contracts/stablex_contract.rs
bh2smith: Ok. I was not aware of that.

driver/src/contracts/stablex_contract.rs
bh2smith: Yes. Good thinking. Will do.

driver/src/contracts/stablex_contract.rs
bh2smith: I am pretty sure these need to be unwrapped in order for the next line to work.

```
BaseContract::new(address, contract_json).unwrap()
```

driver/src/contracts/snapp_contract.rs
```diff
@@ -20,11 +20,13 @@ pub struct SnappContractImpl {
 
 impl SnappContractImpl {
     pub fn new() -> Result<Self> {
-        let contract_json = fs::read_to_string("dex-contracts/build/contracts/SnappAuction.json").unwrap();
-        let address = env::var("SNAPP_CONTRACT_ADDRESS").unwrap();
-        Ok(SnappContractImpl {
-            base: BaseContract::new(address, contract_json).unwrap()
-        })
+        let contract_json = fs::read_to_string("dex-contracts/build/contracts/SnappAuction.json");
```
fleupold: nit: let's put the ? on this line so that the variables are already the correct success type.

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#[allow(dead_code)] // TODO - remove once used
+impl StableXContractImpl {
+    pub fn new() -> Result<Self> {
+        let contract_json = fs::read_to_string("dex-contracts/build/contracts/StablecoinConverter.json");
+        let address = env::var("STABLEX_CONTRACT_ADDRESS");
+        Ok(
+            StableXContractImpl {
+                base: BaseContract::new(address?, contract_json?)?
```
fleupold: ditto

driver/src/contracts/stablex_contract.rs
```diff
@@ -0,0 +1,162 @@
...
+            .wait()
+            .map_err(DriverError::from)
+    }
+
+    fn get_auction_data(&self, _index: U256) -> Result<(AccountState, Vec<Order>)> {
+        unimplemented!();
+    }
+
+    fn submit_solution(
+        &self,
```
fleupold: I wonder if it would be better to use `Solution` and Vec<Order> as an exposing abstraction level here and have the conversion logic into all these separate arrays happen inside the contract wrapper.

What do you think?

driver/src/contracts/snapp_contract.rs
bh2smith: Ahh! ok will do.

driver/src/contracts/stablex_contract.rs
bh2smith: I think I understand your reference to `Solution` as in passing a `Solution` type into this function, but I don't know what you mean by `Vec` here... unless there is additional information I am forgetting about. I would generally agree with this suggestion, but probably would want to make these changes in the following PR (since it is currently unimplemented and would not affect this meaningfully).

bh2smith: So yes, @fleupold, overall I can completely agree with these suggestions but would like to carry over this discussion to the next PR where submit solution will be implemented if that's alright with you!

bh2smith: Ok, I now see that this PR would make the most sense for these suggested changes, however, I am not entirely sure I fully understand the proposal. I suppose this would look something like 

```rs
    fn submit_solution(
        &self,
        batch_index: U256,
        owners: Vec<H160>,
        order_ids: Vec<U128>,
        solution: Solution,
        token_ids_for_price: Vec<U128>,
    ) -> Result<()>;
```
but in this case I have only reduced prices and volumes. I will have to look closer to see what else can be logically packed into a `Solution` type.

fleupold: I'd suggest for the method to look like:

```
fn submit_solution(
        &self,
        batch_index: U256,
        orders: Vec<Order>,
        solution: Solution,
    ) -> Result<()>;
```

Owners should be extractable from orders and token_ids should be extractable from the solution.

bh2smith: Ya cool! Thanks.. I had already determined we wouldnb't need `token_ids` but wasn't yet all the way there with passing orders and getting owners from them. Thanks!

PR #241, opened 2019-09-10
[Refactor] Make Accounts wide enough to hold addresses

fleupold:
This PR changes the data type of account_id from u16 to H160 (address type in solidity).

This is needed since in the StableX contract we don't compress the address space to 16 bits anymore.

### Test Plan

Travis

### DIFF
--- a/dfusion_rust_core/src/models/account_state.rs
+++ b/dfusion_rust_core/src/models/account_state.rs
@@ -4,7 +4,7 @@ use serde_derive::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::collections::{hash_map::Entry, HashMap};
 use std::sync::Arc;
-use web3::types::{Log, H256, U256};
+use web3::types::{Log, H160, H256, U256};
 
 use crate::models::{Order, PendingFlux, RollingHashable, Solution};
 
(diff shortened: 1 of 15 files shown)


anxolin:
I think the e2e is failing

fleupold:
> Why are we modeling the entities using this web3 types?

Rust doesn't have native types for hashes and integers that are wider than 128 bits. Therefore unless we wanted to deal with strings in rust, we'd need to use an external library to model these to begin with, so why not use the web3 types? I think a problem is that the data model is actually changing. Before, we thought an account would be a very slim number (u16) and now we changed our protocol to allow a much larger number of accounts.

bh2smith:
@fleupold 
> Rust doesn't have native types for hashes and integers that are wider than 128 bits.

This reminds me that I wanted to ask if there is any special reason we use `U128` rather than `u128`. My guess was that U128 has some more ethereum specific extensions than plain old `u128`

fleupold:
> Finding all these must have been tedious.

Not if you have a strong compiler like rust has. I imagine it being tedious in python ;-) 

> if there is any special reason we use U128 rather than u128

I believe u128 was added a bit later, so blockchain projects started with U128 (but I'm not sure about that). In any case, I think we should use the native types (e.g. u128) where possible.

PR #243, opened 2019-09-12
PoC submit solution

bh2smith:
This change inmplements the `submit_solution` part of the StableX contract interface. For the purpose of simplified unit testing, two helper functions `encode_prices_for_contract` and `encode_execution_for_contract` are introduced to convert auction result data from the `<Order, Solution>` into a collection of vectors that the smart contract recognizes. 

A few new unit tests involving all possible corner cases to auction result parser are introduced. 

Test Plan: `cargo test`

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -63,41 +63,26 @@ impl StableXContract for StableXContractImpl {
         orders: Vec<Order>,
         solution: Solution,
     ) -> Result<()> {
-        let account = self.base
+        let account = self
+            .base
             .account_with_sufficient_balance()
             .ok_or("Not enough balance to send Txs")?;
 
-        // Representing the solution's price vector more compactly as:
-        // sorted_touched_token_ids, non_zero_prices which are logically bound by index.
-        // Example solution.prices = [3, 0, 1] will be transformed into [0, 2], [3, 1]
-        let mut token_ids_for_price: Vec<U128> = vec![];
-        let mut prices: Vec<U128> = vec![];
+        let (owners, order_ids, volumes, prices, token_ids_for_price) =
+            parse_auction_results(orders, solution);
 
-        for (token_id, price) in solution.prices.into_iter().enumerate() {
-            if price > 0 {
-                token_ids_for_price.push(U128::from(token_id as usize));
-                prices.push(U128::from(price as usize));
-            }
-        }
-
-        let mut owners: Vec<H160> = vec![];
-        let mut order_ids: Vec<U128> = vec![];
-        let mut volumes: Vec<U128> = vec![];
-        let zipped_amounts = solution.executed_buy_amounts.into_iter().zip(solution.executed_sell_amounts.into_iter());
-        for (order_id, (buy_amount, sell_amount)) in zipped_amounts.enumerate() {
-            if buy_amount > 0 && sell_amount > 0 {
-                // order was touched!
-                owners.push(orders[order_id].account_id);
-                order_ids.push(U128::from(order_id));
-                // Currently all orders are sell orders, so volumes are sell_amounts.
-                volumes.push(U128::from(sell_amount as usize));
-            }
-        }
-
-        self.base.contract
+        self.base
+            .contract
             .call(
                 "submitSolution",
-                (batch_index, owners, order_ids, volumes, prices, token_ids_for_price),
+                (
+                    batch_index,
+                    owners,
+                    order_ids,
+                    volumes,
+                    prices,
+                    token_ids_for_price,
+                ),
                 account,
                 Options::default(),
             )
@@ -105,6 +90,41 @@ impl StableXContract for StableXContractImpl {
             .map_err(DriverError::from)
             .map(|_| ())
     }
+}
+
+fn parse_auction_results(
+    orders: Vec<Order>,
+    solution: Solution,
+) -> (Vec<H160>, Vec<U128>, Vec<U128>, Vec<U128>, Vec<U128>) {
+    // Representing the solution's price vector more compactly as:
+    // sorted_touched_token_ids, non_zero_prices which are logically bound by index.
+    // Example solution.prices = [3, 0, 1] will be transformed into [0, 2], [3, 1]
+    let mut ordered_token_ids: Vec<U128> = vec![];
+    let mut prices: Vec<U128> = vec![];
+    for (token_id, price) in solution.prices.into_iter().enumerate() {
+        if price > 0 {
+            ordered_token_ids.push(U128::from(token_id as usize));
+            prices.push(U128::from(price as usize));
+        }
+    }
+
+    let mut owners: Vec<H160> = vec![];
+    let mut order_ids: Vec<U128> = vec![];
+    let mut volumes: Vec<U128> = vec![];
+    let zipped_amounts = solution
+        .executed_buy_amounts
+        .into_iter()
+        .zip(solution.executed_sell_amounts.into_iter());
+    for (order_id, (buy_amount, sell_amount)) in zipped_amounts.enumerate() {
+        if buy_amount > 0 && sell_amount > 0 {
+            // order was touched!
+            owners.push(orders[order_id].account_id);
+            order_ids.push(U128::from(order_id));
+            // Currently all orders are sell orders, so volumes are sell_amounts.
+            volumes.push(U128::from(sell_amount as usize));
+        }
+    }
+    (owners, order_ids, volumes, prices, ordered_token_ids)
 }
 
 #[cfg(test)]
@@ -161,5 +181,59 @@ pub mod tests {
             self.submit_solution
                 .called((batch_index, Val(orders), Val(solution)))
         }
+    }
+
+    #[test]
+    fn test_parse_auction_results() {
+        let solution = Solution {
+            surplus: None,
+            prices: vec![3, 0, 1],
+            executed_sell_amounts: vec![1, 3],
+            executed_buy_amounts: vec![3, 1],
+        };
+
+        let address_1 = H160::from(1);
+        let address_2 = H160::from(0);
+
+        let order_1 = Order {
+            batch_information: None,
+            account_id: address_1,
+            sell_token: 0,
+            buy_token: 2,
+            sell_amount: 1,
+            buy_amount: 2,
+        };
+        let order_2 = Order {
+            batch_information: None,
+            account_id: address_2,
+            sell_token: 2,
+            buy_token: 0,
+            sell_amount: 3,
+            buy_amount: 4,
+        };
+
+        let zero = U128::from(0);
+        let one = U128::from(1);
+        let two = U128::from(2);
+        let three = U128::from(3);
+
+        let expected_owners = vec![address_1, address_2];
+        let expected_order_ids = vec![zero, one];
+        let expected_volumes = vec![one, three];
+        let expected_prices = vec![three, one];
+        let expected_token_ids = vec![zero, two];
+
+        let expected_results = (
+            expected_owners,
+            expected_order_ids,
+            expected_volumes,
+            expected_prices,
+            expected_token_ids,
+        );
+
+        assert_eq!(
+            parse_auction_results(vec![order_1, order_2], solution),
+            expected_results
+        );
     }
 }
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -118,11 +118,7 @@ fn parse_auction_results(
     let mut owners: Vec<H160> = vec![];
     let mut order_ids: Vec<U128> = vec![];
     let mut volumes: Vec<U128> = vec![];
-    let zipped_amounts = solution
-        .executed_buy_amounts
-        .into_iter()
-        .zip(solution.executed_sell_amounts.into_iter());
-    for (order_index, (_buy_amount, sell_amount)) in zipped_amounts.enumerate() {
+    for (order_index, sell_amount) in solution.executed_sell_amounts.into_iter().enumerate() {
         if sell_amount > 0 {
             // order was touched!
             // Note that above condition is only holds for sell orders.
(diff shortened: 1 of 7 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,74 @@ impl StableXContract for StableXContractImpl {
...
+    }
+
+    let mut owners: Vec<H160> = vec![];
+    let mut order_ids: Vec<U128> = vec![];
+    let mut volumes: Vec<U128> = vec![];
+    let zipped_amounts = solution
+        .executed_buy_amounts
+        .into_iter()
+        .zip(solution.executed_sell_amounts.into_iter());
+    for (order_id, (buy_amount, sell_amount)) in zipped_amounts.enumerate() {
```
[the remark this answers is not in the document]
fleupold: This information comes as part of the auction data. We might be able to use the BatchInformation struct on the Order model for storing the index?

driver/src/contracts/stablex_contract.rs
```diff
@@ -38,9 +38,9 @@ pub trait StableXContract {
     fn get_auction_data(&self, _index: U256) -> Result<(AccountState, Vec<Order>)>;
     fn submit_solution(
         &self,
-        _batch_index: U256,
-        _orders: Vec<Order>,
-        _solution: Solution,
+        batch_index: U256,
```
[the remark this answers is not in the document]
bh2smith: At this moment we do not. However, I have already noticed that this will necessarily need to be changes to `u32` anyways for the purpose of comparing with `order.validTo` and `order.ValidFrom` when filtering out the `EncodedAuctionElements`. Thanks for pointing this out.

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,74 @@ impl StableXContract for StableXContractImpl {
...
+fn parse_auction_results(
+    orders: Vec<Order>,
+    solution: Solution,
+) -> ContractRecognizedAuctionResults {
+    // Representing the solution's price vector more compactly as:
+    // sorted_touched_token_ids, non_zero_prices which are logically bound by index.
+    // Example solution.prices = [3, 0, 1] will be transformed into [0, 2], [3, 1]
+    let mut ordered_token_ids: Vec<U128> = vec![];
+    let mut prices: Vec<U128> = vec![];
+    for (token_id, price) in solution.prices.into_iter().enumerate() {
```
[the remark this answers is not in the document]
bh2smith: Yes, at some point (if we every reach 2^16 ~ 65K tokens) it will be imperative that we do not iterate (even once) over this list. Back when we were using 2^8 this would be no problem at all! Ha ha.

driver/src/contracts/stablex_contract.rs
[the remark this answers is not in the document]
bh2smith: Good points here indeed. I have also been thinking that our existing `Order` struct is slightly insufficient for the requirements of the StableX Order type. Observe the comparison, 

StableX
```js
struct Order {
    uint16 buyToken;
    uint16 sellToken;
    uint32 validFrom;  // order is valid from auction collection period: validFrom inclusively
    uint32 validUntil;  // order is valid till auction collection period: validUntil inclusively
    bool isSellOrder;
    uint128 priceNumerator;
    uint128 priceDenominator;
    uint128 remainingAmount; // remainingAmount can either be a sellAmount or buyAmount, depending on the flag isSellOrder
}
```

SnappAuction

```js
SellOrder(
    uint auctionId,
    uint16 slotIndex,
    uint16 accountId,
    uint8 buyToken,
    uint8 sellToken,
    uint96 buyAmount,
    uint96 sellAmount
)
```

There are actually very few values which coincide here. Even those that do have differing types (`u8` vs `u16`). Perhaps it would be in our best interest to introduce a new Order Type for the StableX to interact with? @josojo, @fleupold and/or  @anxolin any suggestions? At this point I see very little similarities. Should we try and make this a single (robust) Order type or is it really worth it?

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,14 +59,77 @@ impl StableXContract for StableXContractImpl {
...
+
+    let mut owners: Vec<H160> = vec![];
+    let mut order_ids: Vec<U128> = vec![];
+    let mut volumes: Vec<U128> = vec![];
+    let zipped_amounts = solution
+        .executed_buy_amounts
+        .into_iter()
+        .zip(solution.executed_sell_amounts.into_iter());
+    for (order_id, (buy_amount, sell_amount)) in zipped_amounts.enumerate() {
+        if buy_amount > 0 && sell_amount > 0 {
```
bh2smith: Question for @josojo , @fleupold  and @twalth3r. I would like to confirm that these are the necessary and sufficient condition for an order to be considered "touched". I can imagine a situation of infinities where merely one is non-zero that would say buy zero of A for non-zero of B (or vice-versa). I imagine the contract would struggle with division errors or other such things and just wanna make sure we are all on the same page. 

I ask this, because I have unit test for each of the four cases: `(buyAmount, sellAmount) \in {zero, non-zero}^2`

driver/src/contracts/stablex_contract.rs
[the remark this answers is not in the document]
josojo: >Perhaps it would be in our best interest to introduce a new Order Type for the StableX to interact with?

I guess so.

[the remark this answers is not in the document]
bh2smith: I am already working on another branch implementing `get_auction_info` which will explicitly deal with `Orders::from_bytes`. This will be a more relevant place to make the changes.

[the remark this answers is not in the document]
fleupold: I'm not sure if we need different models. isSellOrder will be relevant for the dfusion order as well (and since right now it's always true, I'd ignore it). `priceNumerator` `priceDenominator` and `remainingAmount` can be used to compute buyAmount and sellAmount.

The most shoehorning I see with using `slotIndex` vs `orderID` interchangeably on the same model (since in fact they are different things).

If we create a new model, we will also have to implement the price finder in a way that can handle both subtypes, which might open up a few other can of worms down the road. This is why I'd hope we can share the data model (as at the end of the day our orders didn't change that much)

driver/src/contracts/stablex_contract.rs
fleupold: I believe that if the sellAmount is 0, we will also not process it in the smart contract. However, if the sellAmount is positive, it shouldn't matter what the buyVolum is, the smart contract (only receiving the sellVolume) would process it.

Therefore, I think the condition might only be

`if sell_amount  > 0`

bh2smith: Good enough for me! Will readjust the unit test back to work for this and move on to the final issue regarding casting. Let us note that this condition only holds in the case that the order is a sell order.

```
U128::from(x as usize)
```
downcasts to u64 and will result in a loss of precision.

driver/src/contracts/stablex_contract.rs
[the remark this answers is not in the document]
bh2smith: Thats cool for now, then I will use slotIndex for orderId in this PR and deal with whatever else is necessary when I get to the `get_auction_data`. Cool with everyone else here?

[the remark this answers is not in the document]
bh2smith: We should be careful to note that `slot_index` (from and Order's BatchInformation) is a `u16` which could result in future issues. I will include a comment about this in the code.

driver/src/contracts/stablex_contract.rs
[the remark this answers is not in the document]
bh2smith: I will create a "task" for this in our backlog and perhaps include a `todo` here in the code.

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,86 @@ impl StableXContract for StableXContractImpl {
...
+        .zip(solution.executed_sell_amounts.into_iter());
+    for (order_index, (_buy_amount, sell_amount)) in zipped_amounts.enumerate() {
+        if sell_amount > 0 {
+            // order was touched!
+            // Note that above condition is only holds for sell orders.
+            owners.push(orders[order_index].account_id);
+            let order_batch_info = orders[order_index]
+                .batch_information
+                .as_ref()
+                .expect("Batch Information on StableX Order");
```
fleupold: nit: "StableX order needs to have batch_information"

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,86 @@ impl StableXContract for StableXContractImpl {
...
+        orders: Vec<Order>,
+        solution: Solution,
     ) -> Result<()> {
-        unimplemented!();
+        let account = self
+            .base
+            .account_with_sufficient_balance()
+            .ok_or("Not enough balance to send Txs")?;
+
+        let (owners, order_ids, volumes, prices, token_ids_for_price) =
```
fleupold: nit: would it make sense to have one method to get owners, orders, volumes and a separate one for prices and their indices (given they are disjunct in logic)?

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,86 @@ impl StableXContract for StableXContractImpl {
...
+    for (order_index, (_buy_amount, sell_amount)) in zipped_amounts.enumerate() {
+        if sell_amount > 0 {
+            // order was touched!
+            // Note that above condition is only holds for sell orders.
+            owners.push(orders[order_index].account_id);
+            let order_batch_info = orders[order_index]
+                .batch_information
+                .as_ref()
+                .expect("Batch Information on StableX Order");
+            // TODO - using slot_index (u16) for order_id (U128) is temporary and not sustainable.
```
fleupold: I think generally, it's quite hacky to reuse the BatchInformation (which represents a pair of auction and index within) here. Long term we might want to turn BatchInformation into an enum which can be either the slot/slot_index pair or the order Id of stable x

driver/src/contracts/stablex_contract.rs
bh2smith: Ya this is a weird typo. It used to say "Missing ..."

driver/src/contracts/stablex_contract.rs
bh2smith: Yea, sure! I like it!

driver/src/contracts/stablex_contract.rs
bh2smith: Yes. I would get behind this. We will also find that we need to rework `BatchInformation` to potentially include `valid_from` and `valid_till` especially for the new stableX order type. This issue will be revisited in the next PR (get_auction_data).

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,11 +59,51 @@ impl StableXContract for StableXContractImpl {
...
+            .account_with_sufficient_balance()
+            .ok_or("Not enough balance to send Txs")?;
+
+        // Representing the solution's price vector more compactly as:
+        // touched_token_ids, sorted_prices which are logically bound by index.
+        // Example solution.prices = [3, 0, 1] will be transformed into [2, 0], [1, 3]
+        let mut price_token_binding: Vec<(U128, U128)> = vec![];
+        for (token_id, price) in solution.prices.iter().enumerate() {
+            if *price > 0 {
+                price_token_binding.push((U128::from(token_id as usize), U128::from(*price as usize)))
```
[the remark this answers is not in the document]
bh2smith: Fixed this with `price.to_be_bytes()`!

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,86 @@ impl StableXContract for StableXContractImpl {
...
+    }
+
+    let mut owners: Vec<H160> = vec![];
+    let mut order_ids: Vec<U128> = vec![];
+    let mut volumes: Vec<U128> = vec![];
+    let zipped_amounts = solution
+        .executed_buy_amounts
+        .into_iter()
+        .zip(solution.executed_sell_amounts.into_iter());
+    for (order_index, (_buy_amount, sell_amount)) in zipped_amounts.enumerate() {
```
josojo: I have not yet understood why we do the zipping of amounts. Wouldn't it be simpler to enumerate just over the orders and then we get the sellvolume of an order by exectued_sell_amounts[enumeratorId]?

bh2smith: Ya, I guess this could be done that way. Although I prefer the _symmetry_ of iterating over them both at once. I suppose, overall, I have chosen to go with symmetry.

driver/src/contracts/stablex_contract.rs
bh2smith: Although, separating this into two functions will result in an overhaul of the unit tests (i.e. testing them each). Do is this a change that could be made in the future or would you suggest it be done here?

fleupold: As you feel. It's really optional anyways.

bh2smith: Ok, so I made these changes and they drastically reduced the complexity of the unit testing. Unfortunately, I would suspect that (eventually) we wouldn't require the price information to be "converted" since this is ultimately not scalable in the number of tokens. Anyway. Let me know what you think and I believe this is ready for merge!

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,84 @@ impl StableXContract for StableXContractImpl {
...
+        orders.len(),
+        executed_sell_amounts.len(),
+        "Received inconsistent auction result data."
+    );
+    // Note that buy_amounts do not play a role here since all orders are sell orders.
+    let mut owners: Vec<H160> = vec![];
+    let mut order_ids: Vec<U128> = vec![];
+    let mut volumes: Vec<U128> = vec![];
+    for (order_index, sell_amount) in executed_sell_amounts.into_iter().enumerate() {
+        if sell_amount > 0 {
```
anxolin: If you apply a filter function before iterating, you reduce the cyclomatic complexity

driver/src/contracts/stablex_contract.rs
```diff
@@ -59,12 +59,84 @@ impl StableXContract for StableXContractImpl {
...
+    }
+}
+
+fn encode_prices_for_contract(price_vector: Vec<u128>) -> (Vec<U128>, Vec<U128>) {
+    // Representing the solution's price vector more compactly as:
+    // sorted_touched_token_ids, non_zero_prices which are logically bound by index.
+    // Example solution.prices = [3, 0, 1] will be transformed into [0, 2], [3, 1]
+    let mut ordered_token_ids: Vec<U128> = vec![];
+    let mut prices: Vec<U128> = vec![];
+    for (token_id, price) in price_vector.into_iter().enumerate() {
```
anxolin: ubernit: Perfectly fine, but a more functional approach would be a --> filter + map + unzip

PR #244, opened 2019-09-12
[StableX] Driver component

fleupold:
This PR adds a new driver component that can handle the StableX logic (check for current batch index, get order data, run solver, push solution).

I moved all driver components into their own submodule and created some helper methods required for this PR (e.g. converting a PriceFindingError into a DriverError or creating accountState for a vector of orders).

I also simplified the way we mutate account balances and not requiring a value to be explicitly set before mutating it (if it's not set 0 will be assumed).

### TestPlan

Unit tests


driver/src/driver/stablex_driver.rs
```diff
@@ -0,0 +1,171 @@
...
+use crate::price_finding::PriceFinding;
+
+use std::collections::HashSet;
+
+use web3::types::U256;
+
+pub struct StableXDriver<'a, C: StableXContract> {
+    past_auctions: HashSet<U256>,
+    contract: &'a C,
+    price_finder: &'a mut dyn PriceFinding,
```
anxolin: Why PriceFinding needs to be mutable? 

If this is the signature:
image


I would understand it only gets a solution for the input data, but it doesn't need to mutate it's state, right?

driver/src/driver/stablex_driver.rs
```diff
@@ -0,0 +1,171 @@
+use crate::contracts::stablex_contract::StableXContract;
+use crate::error::DriverError;
+use crate::price_finding::PriceFinding;
+
+use std::collections::HashSet;
+
+use web3::types::U256;
+
+pub struct StableXDriver<'a, C: StableXContract> {
```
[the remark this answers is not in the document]
fleupold: No you are right, we can use `dyn StableXContract`. I only did it like this because the other components did it (which I believe was because in the beginning we didn't know better).

Good call.

driver/src/driver/stablex_driver.rs
fleupold: The reason is that for the solver, we need to pass in the previousPrices (it's used as a good starting point for the new price finding).

At the moment the prices are stored in the PriceFinding component and is updated/mutated on every invocation of `find_prices()`.

anxolin: Okay, and would it make sense to decouple PriceFinding and PriceStorage

PriceFinding = delegates to the solver, return the solution
PriceStorage = stores the solution, the driver persist the solution there. Also it reads the previous prices for handing them to the price finding to be used as an heuristic

fleupold: I think this could be a good idea (especially now that in the two-phased bid process of the dƒusion POC we store the old solutions and thus old prices in the driver anyways). Would you suggest that PriceFinder has a PriceStorage, or that it is passed into every call of `
`find_prices()`?

anxolin: I would just use the driver as the coordinator of all the job, so I would make him responsible of calling to PriceStorage in order to get the prices, then hand those price to the PriceFinder.

This way, unit testing would be simpler:
- PriceFinder: you can change directly the inputs to define the cases (no need to mock the PriceStorage)
- PriceStorage: could stores a solution, return a solution for a given batch. Easy to test too
- Driver: orchestrate. Not as important to test as the others, but if we feel it's imp, we can test that it does the actual delegations

bh2smith: Cool ideas going around here!

PR #247, opened 2019-09-13
[Refactor] Change token datatype from u8 to u16

fleupold:
Since stableX supports up to 2**16 tokens

### TestPlan

CI

### DIFF
--- a/dfusion_rust_core/src/models/account_state.rs
+++ b/dfusion_rust_core/src/models/account_state.rs
@@ -15,12 +15,12 @@ use super::util::*;
 pub struct AccountState {
     pub state_hash: H256,
     pub state_index: U256,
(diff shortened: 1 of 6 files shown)


bh2smith:
Doesn't this change now drop support for the Snapp Contracts? I am specifically concerned with `pop_from_log_data` which I assumed popped exactly the relevant amount of information from bytes. If we now use an extra byte for tokenID, shouldn't this break the Snapp Tests?

fleupold:
> If we now use an extra byte for tokenID, shouldn't this break the Snapp Tests?

The log data happens to be 32byte wide regardless of the underlying datatype (u8, u16, ... u256). Thus it doesn't matter how wide our target field is.

If you think it would be less confusing to parse the `u8::pop_from_log_data` and then convert them into u16, I can change that.

PR #248, opened 2019-09-13
Stablex main

fleupold:
This PR creates a new PR in the driver sub-porject - one to run the stablex processor

### TestPlan

`cargo run --bin stablex`

Will fail since the `STABLEX_CONTRACT_ADDRESS` is not yet exporterd. Even after exporting that it will fail since the contract is not yet fully implemented.


driver/src/util.rs
```diff
@@ -65,3 +67,14 @@ pub fn batch_processing_state(
     }
     Ok(ProcessingState::TooEarly)
 }
+
+pub fn create_price_finder() -> Box<dyn PriceFinding> {
+    let solver_env_var = env::var("LINEAR_OPTIMIZATION_SOLVER").unwrap_or_else(|_| "0".to_string());
```
anxolin: ubernit: Would it make sense to use `unwrap_or` instead?

driver/src/bin/stablex.rs
```diff
@@ -0,0 +1,23 @@
...
+use driver::driver::stablex_driver::StableXDriver;
+
+use std::thread;
+use std::time::Duration;
+
+fn main() {
+    simple_logger::init_with_level(log::Level::Info).unwrap();
+
+    let contract = StableXContractImpl::new().unwrap();
+    let mut price_finder = driver::util::create_price_finder();
```
anxolin: nit: the past two lines might be in a different level of coupling with the actual implementation. Line 14 instantiates a specific implementation, the 15 calls a factory method that instantiates the specific implementation.

driver/src/util.rs
fleupold: since we call `to_string()`, it's better to use a closure (cf. https://rust-lang.github.io/rust-clippy/master/#or_fun_call)

driver/src/bin/stablex.rs
fleupold: Do you mean, the two lines together have a different level of coupling than the rest or between each other there is a different level of coupling?

I can see the first but not really the second case, since both lines "wire" up the implementation that needs to be used (one is using a helper method since it is somewhat more involved).

Are you suggesting to pass the env var into the `create_price_finder()` method in order to not rely on static state?

driver/src/util.rs
anxolin: Aha, to avoid copy the &str into a String unless it's necessary

driver/src/bin/stablex.rs
anxolin: I mean that stablex.rs knows the implementation details in line 14, but in line 15 not.

It would make sense to me to also use a factory for line 14 and have stablex.rs depend only on an interface. This is nit/ubernit right now

PR #249, opened 2019-09-16
[PoC] Get (relevant) Auction Data from Smart Contract

bh2smith:
The driver's smart contract interface should implement the function `get_auction_data` which takes auction ID (or batch index) and returns the state of accounts along with all the relevant orders in the requested auction.

This implementation invokes the StableX contract's "getAuctionElementsFunction" which returns every order that has ever been placed in a byte encoded format. Thus, much of the work here involves deserializing the byte-encoded orders and filtering out those which are irrelevant for the current batch.


dfusion_rust_core/src/models/util.rs
```diff
@@ -22,6 +22,12 @@ impl PopFromLogData for u16 {
     }
 }
 
+impl PopFromLogData for u32 {
+    fn pop_from_log_data(bytes: &mut Vec<u8>) -> Self {
+        U256::pop_from_log_data(bytes).as_u32().try_into().unwrap()
```
fleupold: Doesn't `as_u32` return a u32 already? What die `try_into` and `unwrap` add here?

driver/src/contracts/stablex_contract.rs
```diff
@@ -44,6 +47,60 @@ pub trait StableXContract {
...
+        } else {
+            buy_amount = remaining_amount;
+            sell_amount = (price_denominator * remaining_amount) / price_numerator;
+        }
+
+        AuctionElement {
+            valid_from,
+            valid_until,
+            order: Order {
+                batch_information: None,
```
fleupold: We probably need to set the current batch index here?

driver/src/contracts/stablex_contract.rs
```diff
@@ -44,6 +47,60 @@ pub trait StableXContract {
...
+    order: Order,
+}
+
+impl AuctionElement {
+    fn from_bytes(bytes: &[u8; 113]) -> Self {
+        let mut data_vector = bytes.to_vec();
+        println!("{}", bytes.len());
+        let account_id = H160::pop_from_log_data(&mut data_vector);
+        // TODO - not sure what this is for.
+        let _sell_token_balance = U256::pop_from_log_data(&mut data_vector);
```
fleupold: It's the amount of tokens the user has currently in their account (for the `balance > 0` constraint).

driver/src/contracts/stablex_contract.rs
```diff
@@ -53,8 +110,48 @@ impl StableXContract for StableXContractImpl {
...
+        let auction_info: Vec<AuctionElement> = packed_auction_bytes
+            .chunks(AUCTION_ELEMENT_WIDTH)
+            .map(|chunk| {
+                let mut chunk_array = [0; AUCTION_ELEMENT_WIDTH];
+                chunk_array.copy_from_slice(chunk);
+                AuctionElement::from_bytes(&chunk_array)
+            })
+            .collect();
+
+        // TODO - can we use drain_filter here?
```
fleupold: You probably want `filter().map()`

dfusion_rust_core/src/models/util.rs
bh2smith: Good question, I will take a look into this!

driver/src/contracts/stablex_contract.rs
bh2smith: Yes we certainly do!

driver/src/contracts/stablex_contract.rs
bh2smith: Right, and the remaining amount is the amount left in their order I suppose.

dfusion_rust_core/src/models/util.rs
bh2smith:
```suggestion
        U256::pop_from_log_data(bytes).as_u32()```

File: driver/src/contracts/stablex_contract.rs [hunk 5]
```diff
@@ -42,6 +45,59 @@ pub trait StableXContract {
...
+        let valid_from = U256::from(u32::pop_from_log_data(&mut data_vector));
+        let valid_until = U256::from(u32::pop_from_log_data(&mut data_vector));
+        let is_sell_order = bool::pop_from_log_data(&mut data_vector);
+        let price_numerator = u128::pop_from_log_data(&mut data_vector);
+        let price_denominator = u128::pop_from_log_data(&mut data_vector);
+        let remaining_amount = u128::pop_from_log_data(&mut data_vector);
+
+        // Todo - Will likely have to compute this differently.
+        // TODO - put in own function fn [u128; 3] -> [u128; 2]
+        let buy_amount: u128;
```
fleupold: I like the idea of putting this into its own method. Otherwise it'd be slightly more rusty to do

```rust
let (buy_amount, sell_amount) = if is_sell_order {
  (remaining_amount, (price_numerator * remaining_amount) / price_denominator)
} else {
  ((price_denominator * remaining_amount) / price_numerator, remaining_amount)
}
```

driver/src/contracts/stablex_contract.rs
```diff
@@ -42,6 +45,59 @@ pub trait StableXContract {
...
+            buy_amount = remaining_amount;
+            sell_amount = (price_denominator * remaining_amount) / price_numerator;
+        }
+
+        AuctionElement {
+            valid_from,
+            valid_until,
+            sell_token_balance,
+            order: Order {
+                batch_information: None, // TODO - can't recover order_id from this information include it in the return of
```
fleupold: What would you suggest doing here? Should we add this to the data we return from the smart contract or count the number of items ourselves?

dfusion_rust_core/src/models/account_state.rs
```diff
@@ -20,6 +20,15 @@ pub struct AccountState {
 }
 
 impl AccountState {
+    pub fn default(state_index: U256) -> Self {
```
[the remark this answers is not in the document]
bh2smith: State index is relevant information here. I suppose we could get away without it (since the caller knows the batch_index they are calling with), but It would be like intentionally forgetting relevant information.

driver/src/contracts/stablex_contract.rs
```diff
@@ -143,17 +136,20 @@ impl StableXContract for StableXContractImpl {
...
+        let mut account_state = AccountState::default(index);
         let relevant_orders = auction_elements
             .into_iter() // using iter here gives "cannot move out of borrowed content"
             .filter(|x| x.in_auction(index))
-            .map(|e| e.order)
+            .map(|element| {
+                account_state.modify_balance(
+                    element.order.account_id,
+                    element.order.sell_token,
+                    |x| *x += element.sell_token_balance,
```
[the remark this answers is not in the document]
bh2smith: Good catch. I suppose we would need another function that simply "sets" the balance rather than adding to it. How does that sound?

dfusion_rust_core/src/models/util.rs
```diff
@@ -10,6 +10,12 @@ pub trait PopFromLogData {
     fn pop_from_log_data(bytes: &mut Vec<u8>) -> Self;
 }
 
+impl PopFromLogData for bool {
+    fn pop_from_log_data(bytes: &mut Vec<u8>) -> Self {
+        U256::pop_from_log_data(bytes).as_u32() != 0
```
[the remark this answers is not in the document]
bh2smith: Or maybe `!is_zero()`?

driver/src/contracts/stablex_contract.rs
```diff
@@ -51,8 +107,50 @@ impl StableXContract for StableXContractImpl {
...
+            "Each auction should be packed in 113 bytes"
+        );
+        let auction_elements: Vec<AuctionElement> = packed_auction_bytes
+            .chunks(AUCTION_ELEMENT_WIDTH)
+            .map(|chunk| {
+                let mut chunk_array = [0; AUCTION_ELEMENT_WIDTH];
+                chunk_array.copy_from_slice(chunk);
+                AuctionElement::from_bytes(&chunk_array)
+            })
+            .collect();
```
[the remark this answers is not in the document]
bh2smith: Ya, I believe that would be a really nice approach!

driver/src/contracts/stablex_contract.rs
bh2smith: I could also get behind the slightly more rusty idea (since I don't really want to have to name this function) and have been looking for a reason to use a `let if`.

driver/src/contracts/stablex_contract.rs
bh2smith: Yes, this will require inclusion of more information being returned from the `getAuctionElement` method. Will need to update the contracts before we can finish this.

bh2smith:
Sorry fellas (@josojo  and @fleupold), its has just come to my attention that this info is not "log data" so we cannot use `pop_from_log_data` to parse it. Instead we should treat it as encoded bytes and do something like we have done for parsing byte-encoded Orders 

https://github.com/gnosis/dex-services/blob/33d6c0e0a56fe45b578ef08cbbfeb3ec8f615287/dfusion_rust_core/src/models/order.rs#L31-L39

It appears the results are returned backwards (i.e. in little endian format), so I will have to read everything from right to left.

driver/src/contracts/stablex_contract.rs
josojo: I like the counting approach more...

dfusion_rust_core/src/models/auction_element.rs
```diff
@@ -22,30 +22,37 @@ impl AuctionElement {
...
-        let remaining = BigEndian::read_u128(&bytes[0..15]);
+        let is_sell_order = bytes[48] > 0;
+        let numerator = BigEndian::read_u128(&bytes[32..48]);
+        let denominator = BigEndian::read_u128(&bytes[16..32]);
+        let remaining = BigEndian::read_u128(&bytes[0..16]);
 
+        let mut other = 0;
         let (buy_amount, sell_amount) = if is_sell_order {
-            (remaining, (numerator * remaining) / denominator)
+            if denominator > 0 {
```
fleupold: @josojo Should it be valid to put these kind of offers into the smart contract to begin with?

If so I think there is sufficient logic here that this should become a helper method. Could be called something like `compute_buy_sell_amount`

dfusion_rust_core/src/models/auction_element.rs
```diff
@@ -0,0 +1,88 @@
...
+use byteorder::{BigEndian, ByteOrder};
+
+use std::collections::HashMap;
+
+use web3::types::{H160, U256};
+
+use crate::models::{BatchInformation, Order};
+
+#[derive(Debug, Default, PartialEq)]
+pub struct AuctionElement {
```
fleupold: Why are we putting this in the shared library? Right now this object is still pretty private to the smart contract and not passed around at all. Might be worth while keeping this an implementation detail.

dfusion_rust_core/src/models/auction_element.rs
```diff
@@ -0,0 +1,88 @@
...
+        let counter = order_count.entry(account_id).or_insert(0);
+        *counter += 1;
+
+        AuctionElement {
+            valid_from,
+            valid_until,
+            sell_token_balance,
+            order: Order {
+                batch_information: Some(BatchInformation {
+                    slot_index: *counter,
```
fleupold: The first order should be index 0, shouldn't it?

dfusion_rust_core/src/models/auction_element.rs
bh2smith: Yes. that would make sense. I suspect that these values could be zero (as they are adjusted by subtraction in the contract).

dfusion_rust_core/src/models/auction_element.rs
bh2smith: Ya, I suppose so.

dfusion_rust_core/src/models/auction_element.rs
josojo: > Should it be valid to put this kind of offers into the smart contract, to begin with?

Currently, it is a valid input. Should it be prohibited? Well, at least I can not think of any situation, where it would make sense. But I think, it does not hurt to have the case distinction in rust anyways.

Yeah, I will put it into a separate function

dfusion_rust_core/src/models/auction_element.rs
josojo: I am really not sure where a good place for the auction_element is.
I agree with Felix that in the short term, it is a private object, hence it does not belong into dfusion_core.
Then I thought to put it into stablex_contract.rs, as it is only used there, but then the whole file was soo big and messy to read. Hence, I put it all into stable_contract_auction_element.rs. 
I think, it is the best solution, but I am not sure. So I am very interested into your opinions.
@fleupold @bh2smith

driver/src/contracts/stablex_contract.rs
```diff
@@ -90,6 +106,34 @@ impl StableXContract for StableXContractImpl {
     }
 }
 
+fn parse_auction_data(packed_auction_bytes: Vec<u8>, index: U256) -> (AccountState, Vec<Order>) {
+    // extract packed auction info
+    assert_eq!(
+        packed_auction_bytes.len() % AUCTION_ELEMENT_WIDTH,
+        0,
+        "Each auction should be packed in 113 bytes"
```
fleupold: nit: This string should probably also use the constant

driver/src/contracts/stablex_contract.rs
```diff
@@ -90,6 +106,34 @@ impl StableXContract for StableXContractImpl {
...
+    );
+
+    let mut account_state = AccountState::default();
+    let mut order_count = HashMap::new();
+    let relevant_orders = packed_auction_bytes
```
fleupold: What a rust mastery!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FaLTFHi2aowjJe%2Fgiphy.gif&token=master" data-gitmeme-token="master"></a>

driver/src/contracts/stablex_contract_auction_element.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn in_auction(&self, index: U256) -> bool {
+        self.valid_from < index && index <= self.valid_until
+    }
+
+    pub fn from_bytes(order_count: &mut HashMap<H160, u16>, bytes: &[u8; 113]) -> Self {
+        let account_id = H160::from(&bytes[0..20]);
+
+        // these go together (since sell_token_balance is emitted as u256 and treated as u128
+        let sell_token_balance = BigEndian::read_u128(&bytes[36..52]);
+        let hopefully_null = BigEndian::read_u128(&bytes[20..36]);
```
fleupold: I'd call this `sell_token_balance_padding`

driver/src/contracts/stablex_contract_auction_element.rs
```diff
@@ -0,0 +1,224 @@
+use byteorder::{BigEndian, ByteOrder};
+
+use std::collections::HashMap;
+
+use web3::types::{H160, U256};
+
+use dfusion_core::models::{BatchInformation, Order};
+
+#[derive(Debug, Default, PartialEq)]
```
fleupold: Why do we need Default?

driver/src/contracts/stablex_contract_auction_element.rs
```diff
@@ -0,0 +1,224 @@
...
+pub struct AuctionElement {
```
fleupold: I think this is the right place for this struct, but I'd try to name it in accordance with the file. Maybe it should be a `StableXAuctionElement`? And the file could be called `stablex_auction_element`

driver/src/contracts/stablex_contract_auction_element.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub sell_token_balance: u128,
+    pub order: Order,
+}
+
+impl AuctionElement {
+    pub fn in_auction(&self, index: U256) -> bool {
+        self.valid_from < index && index <= self.valid_until
+    }
+
+    pub fn from_bytes(order_count: &mut HashMap<H160, u16>, bytes: &[u8; 113]) -> Self {
```
fleupold: I like the fixed sized array.

driver/src/contracts/stablex_contract_auction_element.rs
```diff
@@ -0,0 +1,224 @@
...
+pub struct AuctionElement {
+    valid_from: U256,
+    valid_until: U256,
+    pub sell_token_balance: u128,
+    pub order: Order,
+}
+
+impl AuctionElement {
+    pub fn in_auction(&self, index: U256) -> bool {
+        self.valid_from < index && index <= self.valid_until
```
fleupold: Just saw in the other PR, this should be <= on both sides.

dfusion_rust_core/src/models/order.rs
```diff
@@ -16,7 +16,7 @@ pub struct BatchInformation {
     pub slot_index: u16,
 }
 
-#[derive(Debug, Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
+#[derive(Debug, Clone, Default, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
```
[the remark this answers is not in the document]
josojo: Yep,

We want to have defaul for AuctionElement  and auctionElements contain orders

driver/src/contracts/stablex_contract.rs
josojo: Ben and I did this together, paircoding 👍 ! 

That is what happens after reviewing so many PRs from you 💃

driver/src/contracts/stablex_contract_auction_element.rs
josojo: As we want to provide Default for AuctionElement

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+fn compute_buy_sell_amounts(
+    numerator: u128,
+    denominator: u128,
+    amount: u128,
+    is_sell_order: bool,
+) -> (u128, u128) {
+    if denominator > 0 {
+        let other = (numerator * amount) / denominator;
```
fleupold: This overflows, please fix before merging.

driver/src/contracts/stablex_contract_auction_element.rs
bh2smith: Yes, needed for `AuctionElement::default()`

driver/src/contracts/stablex_contract_auction_element.rs
bh2smith: If so, then one of the unit tests will have to be changed. `in_auction_left` should have the left value shifted by one.

driver/src/contracts/stablex_auction_element.rs
bh2smith: Good catch, is there a rusty way to prevent overflow?

driver/src/contracts/stablex_contract.rs
```diff
@@ -270,6 +315,66 @@ pub mod tests {
         );
     }
 
+    #[test]
+    fn generic_parse_auction_data_test() {
```
bh2smith: Are the bytes here different than the already existing unit tests? I would think that it might be added value to include a different bytes vector here.

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -0,0 +1,227 @@
...
+        StableXAuctionElement::from_bytes(&mut order_count, &[1u8; 113]);
+    }
+
+    // Testing in_auction
+    #[test]
+    fn in_auction_left() {
+        let mut element = StableXAuctionElement::default();
+        element.valid_from = U256::from(2);
+        element.valid_until = U256::from(5);
+        assert_eq!(element.in_auction(U256::from(2)), true);
```
bh2smith: This will have to be changed when the `<=` is introduced.

dfusion_rust_core/src/models/order.rs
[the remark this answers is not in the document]
fleupold: `AuctionElement::default()` is only needed in unit tests. I'd argue that it's only used for convenience and I'd prefer to not allow production code to use these.

We can either expose it in unit tests only, but even there, I'd prefer writing out the extra few lines (we have a dummy `create_order_for_test method` somewhere

driver/src/contracts/stablex_contract_auction_element.rs
bh2smith: I thought you might.

driver/src/contracts/stablex_auction_element.rs
bh2smith: Furthermore, even without overflow, this formula is not correct.

Notice, from the  previous version of the code, that other is computed differently in different scenarios.

```rs
        let mut other = 0;
        let (buy_amount, sell_amount) = if is_sell_order {
            if denominator > 0 {
                other = (numerator * remaining) / denominator;
            }
            (remaining, other)
        } else {
            if numerator > 0 {
                other = (denominator * remaining) / numerator;
            }
            (other, remaining)
        };
```

josojo: Yeah,
I guess it depends on how we gonna store the price for a buyOrder, do we us /pi or 1/pi. I guess your definition is closer to the one Tom was using in his papers.

I think we would not implement any of this logic in rust, as long as we do not have any in the smart contract. 

Probably, I should have taken isSellOrder completely out of the interface of smartcontract ...

driver/src/contracts/stablex_auction_element.rs
josojo: It was already changed in my PR. Otherwise, tests would not have passed :). Maybe you are again looking at old prs?

driver/src/contracts/stablex_contract.rs
josojo: Hm... I like it the way it is. 
Here I am just testing that the bytes are correctly chunked into the elements.

Parsing the elements is tested in the other tests already sufficiently, in my opinion.

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -68,19 +68,33 @@ impl StableXAuctionElement {
...
+    // TODO - use u128_to_u256 for overflow handling (this is a conversion used in naive_solver)
+    //    assert!(remaining <= denominator);
+    let mut other = 0;
+    let (buy_amount, sell_amount) = if is_sell_order {
+        // 0 on sellAmount (remaining <= denominator) is nonsense, but solver can handle it.
+        // 0 on buyAmount (numerator) is a Market Sell Order.
+        if denominator > 0 {
+            let p = remaining * numerator;
+            let q = denominator;
+            // Recall that ceil(p / float(q)) == (p + q - 1) / q
```
josojo: Good call @bh2smith. Nice! I completely missed that the rounding errors could be a significant issue

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -68,19 +68,33 @@ impl StableXAuctionElement {
...
+            other = (p + q - 1) / q;
         }
+        (other, remaining)
     } else {
+        error!("Unable to to handle buy orders, returning unfulfillable amounts");
         (0, 0)
-    }
+        //        if numerator > 0 {
```
josojo: Nice, yes, let's do not yet implement it. Leaving as comments is the best way right now

driver/src/contracts/stablex_contract.rs
bh2smith: Cool with me! I can't recall if we have tested this function with more than one Encoded Order?

driver/src/contracts/stablex_auction_element.rs
bh2smith: That would be a cool fun thing to do! Although the semantic of buy and sell orders are completely different and not symmetric (since sell orders are unbounded and buy orders are bounded). This means we can't treat them in the same way.

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -0,0 +1,299 @@
...
+            let other_u256 = (top + bottom - 1) / bottom;
+
+            // Conversion never overflows because remaining <= denominator
+            let x = other_u256.0;
+            other = u128::from(x[0]) + u128::from(x[1]) * (2 as u128).pow(64);
+        }
+        (other, remaining)
+    } else {
+        error!("Unable to to handle buy orders, returning unfulfillable amounts");
+        (0 as u128, 0 as u128)
```
fleupold: nit: `0u128`

driver/src/contracts/stablex_auction_element.rs
fleupold: I feel leaving unused and unfinished code in the codebase is unclean and makes it harder to understand what's going on. Is it code that would be hard to rewrite? it's also in version control as part of this PR (a comment with a link to a commit that contains the code could suffice instead of leaving it here).

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -0,0 +1,299 @@
...
+    denominator: u128,
+    remaining: u128,
+    is_sell_order: bool,
+) -> (u128, u128) {
+    // TODO - use u128_to_u256 for overflow handling (this is a conversion used in naive_solver)
+    assert!(
+        remaining <= denominator,
+        "Smart contract should never allow this inequality"
+    );
+    let mut other = 0 as u128;
```
fleupold: Why is other defined outside of the if statement? Also isn't it the buy amount? Can we name it more appropriately.

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -0,0 +1,299 @@
...
+        error!("Unable to to handle buy orders, returning unfulfillable amounts");
+        (0 as u128, 0 as u128)
+        //        if numerator > 0 {
+        //            let p = denominator * remaining;
+        //            let q = numerator;
+        //            other = (p + q - 1) / q;
+        //            (remaining, other)
+        //        }
+    };
+    (buy_amount, sell_amount)
```
fleupold: nit: This line is unnecessary if you remove the `;` after the if statement and don't assign into the variable.

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -0,0 +1,313 @@
...
+        if denominator > 0 {
+            // up-casting to handle overflow
+            let top = u128_to_u256(remaining) * u128_to_u256(numerator);
+            let bottom = u128_to_u256(denominator);
+            // Recall that ceil(p / float(q)) == (p + q - 1) / q
+            let other_u256 = (top + bottom - 1) / bottom;
+
+            // Conversion never overflows because remaining <= denominator
+            let x = other_u256.0;
+            other = u128::from(x[0]) + u128::from(x[1]) * (2 as u128).pow(64);
```
fleupold: Here you could just use `u128::from_str(x.toString()).unwrap()` but I'd suggest creating a separate helper method next to `u128_to_u256`

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -0,0 +1,313 @@
...
+    let mut other = 0 as u128;
+    let (buy_amount, sell_amount) = if is_sell_order {
+        // 0 on sellAmount (remaining <= denominator) is nonsense, but solver can handle it.
+        // 0 on buyAmount (numerator) is a Market Sell Order.
+        if denominator > 0 {
+            // up-casting to handle overflow
+            let top = u128_to_u256(remaining) * u128_to_u256(numerator);
+            let bottom = u128_to_u256(denominator);
+            // Recall that ceil(p / float(q)) == (p + q - 1) / q
+            let other_u256 = (top + bottom - 1) / bottom;
```
fleupold: The naming here is quite terrible. You can reuse variable names. E.g. this could be `let buy_amount`

driver/src/contracts/stablex_auction_element.rs
fleupold: Why exactly do we need the ceil-rounding here? I don't think we do any of this in the smart contract.

driver/src/contracts/stablex_auction_element.rs
bh2smith:
```suggestion
        (0u128, 0u128)
```

driver/src/contracts/stablex_auction_element.rs
bh2smith: nice.

driver/src/contracts/stablex_auction_element.rs
bh2smith: was not aware I could reuse variable names.

bh2smith:
> Is it code that would be hard to rewrite?

No, it is not hard to duplicate I was going to do it, but @josojo insisted we don't bother untill we have buy Orders.

> it's also in version control as part of this PR (a comment with a link to a commit that contains the code could suffice instead of leaving it here).

Actually this is not true, since our merges are squashed and branches deleted.

In any case, I am fine to take it out since it is incorrect in its current state anyway.

driver/src/contracts/stablex_auction_element.rs
bh2smith: > Is it code that would be hard to rewrite?

It should not be hard to write, but perhaps we shouldn't bother until we have buy Orders.

> it's also in version control as part of this PR (a comment with a link to a commit that contains the code could suffice instead of leaving it here).

Actually this is not true, since our merges are squashed and branches deleted.

In any case, I am fine to take it out since it is incorrect in its current state anyway.

driver/src/contracts/stablex_auction_element.rs
bh2smith: It was put outside so that the else branch of the if statement could also treat it as the `sell_amount`. Of course, since this remains unimplemented, it does not need to be here at the moment.

driver/src/contracts/stablex_auction_element.rs
bh2smith:
```
91 |             let x = u128::from_str(x.toString()).unwrap();
   |                           ^^^^^^^^ function or associated item not found in `u128`
```

bh2smith: @fleupold  - do you think we should update our dependencies here. My PR to parity was closed because our dependencies are depreciated. 

https://github.com/paritytech/primitives/pull/68

driver/src/contracts/stablex_auction_element.rs
fleupold: > Actually this is not true, since our merges are squashed and branches deleted.

Merges are squashed into one commit on master the original commits remain. Branches are just labels for commits, so it doesn't matter if they are deleted.

E.g. this link points at a shell script of my rustfmt PR: https://github.com/gnosis/dex-services/pull/246/commits/7b78c74be4426062520078bca2de3e9cc7fd2be8#diff-b81e3e3f4c6ef7d2b9b3c06990d4cc17

The code got removed, the commit squash-merged and the branch deleted. Yet the code is still there.

driver/src/contracts/stablex_auction_element.rs
fleupold: I think we should look into it. However, I'm not sure how easy it is since the graph also depends on web3 types and I'm not sure if the versions are compatible.

fleupold: Regarding the from_str. It is implemented (https://doc.rust-lang.org/std/primitive.u128.html#impl-FromStr)

Maybe you have to import the trait `use std::str::FromStr`

driver/src/contracts/stablex_auction_element.rs
bh2smith: Interesting, thanks for that!

driver/src/contracts/stablex_auction_element.rs
bh2smith: Oh man! If that is really the case, I think we may recover from some other inconveniences in other places as well. I'll check this!

bh2smith: @josojo  and I were really stumped by this the other day!

bh2smith: All good now, that worked.

driver/src/contracts/stablex_auction_element.rs
fleupold: If you really want to do the rounding here (it makes at most 1 wei difference and I don't think we do it in the smart contract which should be the single source of truth), let's at least share the logic with the naive solver by moving this into a helper method (and making sure this comment has to only be in the codebase once).

bh2smith: Yes, just a moment!

bh2smith: I made a Branch with this ceiling quotient as a utility function. It still needs some work. If we merge this the way it is, I could continue and make the appropriate changes. Just wanna make sure this is not blocking anyone.

bh2smith: I made a Branch with this ceiling quotient as a utility function. It still needs some work. If we merge this the way it is, I could continue and make the appropriate changes. Just wanna make sure this is not blocking anyone.

driver/src/contracts/stablex_contract.rs
```diff
@@ -90,6 +106,35 @@ impl StableXContract for StableXContractImpl {
...
+    let mut account_state = AccountState::default();
+    let mut order_count = HashMap::new();
+    let relevant_orders = packed_auction_bytes
+        .chunks(AUCTION_ELEMENT_WIDTH)
+        .map(|chunk| {
+            let mut chunk_array = [0; AUCTION_ELEMENT_WIDTH];
+            chunk_array.copy_from_slice(chunk);
+            StableXAuctionElement::from_bytes(&mut order_count, &chunk_array)
+        })
+        .filter(|x| x.in_auction(index))
```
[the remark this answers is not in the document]
bh2smith: Yes, that makes sense... although the "remaining" is gone by this point. Will look into it.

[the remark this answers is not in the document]
bh2smith:
```suggestion
    .filter(|x| x.in_auction(index) && x.order.sell_amount > 0) 
```

PR #252, opened 2019-09-19
[NaiveSolver] bugfixes and more exhaustive unit tests

fleupold:
There were some issues with the NaiveSolver, which prevented the stableX e2e tests from working. Most importantly:

1) Price computation for `OrderPairType::BothFullyFilled` was inverted.
2) The rounding in `order_with_buffer_for_fee ` was done on the fee multiplier (.999 -> 1.0) instead of on the product of amount and fee leading to no reduction in the sell amount (and thus violating max. sell amounts when fee was added).

I interpreted this as a sign of poor unit tests and thus spent some time building more robust one:

First, I added a function that performs the validity checks of the smart contract (so that our tests can check that a solution will be accepted).

I then added unit tests for the three cases (left fully matched, right fully matched, both fully matched) that also work with the fee.

This should give us reasonable confidence that the smart contract accepts solutions from the NaiveSolver

### DIFF
--- a/driver/src/price_finding/naive_solver.rs
+++ b/driver/src/price_finding/naive_solver.rs
@@ -72,11 +72,8 @@ impl Matchable for Order {
     }
 
     fn surplus(&self, buy_price: u128, exec_buy_amount: u128, exec_sell_amount: u128) -> U256 {
-        // Note that: ceil(p / float(q)) == (p + q - 1) // q
-        let relative_buy = (u128_to_u256(self.buy_amount) * u128_to_u256(exec_sell_amount)
-            + u128_to_u256(self.sell_amount)
-            - 1)
-            / u128_to_u256(self.sell_amount);
+        let relative_buy = 
+            (u128_to_u256(self.buy_amount) * u128_to_u256(exec_sell_amount)).ceiled_div(u128_to_u256(self.sell_amount));
         (u128_to_u256(exec_buy_amount) - relative_buy) * u128_to_u256(buy_price)
     }
 }
--- a/driver/src/price_finding/naive_solver.rs
+++ b/driver/src/price_finding/naive_solver.rs
@@ -212,7 +212,7 @@ fn order_with_buffer_for_fee(order: &Order, fee: &Option<Fee>) -> Order {
 #[cfg(test)]
 pub mod tests {
     use super::*;
-    use crate::util::{u256_to_u128};
+    use crate::util::u256_to_u128;
     use dfusion_core::models::account_state::test_util::*;
     use std::collections::HashMap;
     use web3::types::{H160, H256};
--- a/driver/src/price_finding/naive_solver.rs
+++ b/driver/src/price_finding/naive_solver.rs
@@ -212,6 +212,7 @@ fn order_with_buffer_for_fee(order: &Order, fee: &Option<Fee>) -> Order {
 #[cfg(test)]
 pub mod tests {
     use super::*;
+    use crate::util::{u256_to_u128};
     use dfusion_core::models::account_state::test_util::*;
     use std::collections::HashMap;
     use web3::types::{H160, H256};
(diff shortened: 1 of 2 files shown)


driver/src/price_finding/naive_solver.rs
```diff
@@ -497,12 +472,110 @@ pub mod tests {
...
+            let sell_token_price = solution.prices[order.sell_token as usize];
+
+            let exec_sell_amount = solution.executed_sell_amounts[i];
+            let mut exec_buy_amount = if buy_token_price > 0 {
+                exec_sell_amount * sell_token_price / buy_token_price
+            } else {
+                0
+            };
+            if let Some(fee) = fee {
+                exec_buy_amount = (exec_buy_amount as f64 * (1.0 - fee.percentage)) as u128;
```
josojo: nit: you might wanna use the same fee calculation logic with the same rounding issues as in the contract.

driver/src/price_finding/naive_solver.rs
```diff
@@ -497,12 +472,110 @@ pub mod tests {
...
+    }
+
+    fn order_pair_both_fully_matched() -> Vec<Order> {
+        vec![
+            Order {
+                batch_information: None,
+                account_id: H160::from(1),
+                sell_token: 2,
+                buy_token: 1,
+                sell_amount: 10 * 10u128.pow(18),
```
bh2smith: Why the need to use 10^18 here?

driver/src/price_finding/naive_solver.rs
```diff
@@ -497,12 +472,110 @@ pub mod tests {
...
 
-        let mut solver = NaiveSolver::new(Some(Fee {
+        let fee = Some(Fee {
             token: 2,
             percentage: 0.001,
-        }));
+        });
+        let mut solver = NaiveSolver::new(fee.clone());
         let res = solver.find_prices(&orders, &state).unwrap();
         assert_eq!(res.executed_sell_amounts, [0, 0]);
```
bh2smith: Here, for example, we only check executed sell amounts. But we should also check buy amounts and prices.

driver/src/price_finding/naive_solver.rs
fleupold: They are still valid. However in reality we are always going to be trading amounts in the area of 10**18 and thus in order to get more confidence that our e2e tests would also work I increased the numbers to be in the same order of magnitude.

bh2smith: However, knowing that this is all working for smaller instances is still relevant for unit testing.

driver/src/util.rs
```diff
@@ -8,6 +8,24 @@ use web3::types::{H256, U256};
...
+
+impl CeiledDiv for u128 {
+    fn ceiled_div(&self, divisor: u128) -> u128 {
+        //ceil(p / float(q)) == (p + q - 1) / q
+        (self + divisor - 1) / divisor
+    }
+}
+
+impl CeiledDiv for U256 {
+    fn ceiled_div(&self, divisor: U256) -> U256 {
```
bh2smith: Gonna test these?

fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F3o6ZtqPYkSprj7nJUA%2Fgiphy.gif&token=caught_me" data-gitmeme-token="caught_me"></a>

bh2smith: I would recommend the tests in the branch I created earlier. Should be called ceiling quotient.

driver/src/price_finding/naive_solver.rs
```diff
@@ -67,7 +67,9 @@ impl Matchable for Order {
     }
 
     fn have_price_overlap(&self, other: &Order) -> bool {
-        u128_to_u256(self.buy_amount) * u128_to_u256(other.buy_amount)
```
bh2smith: This looks alright, but I find the long broken statement to be a little bit cosmetically unattractive. Maybe we can make an `all` and `any` macro at some point. I would love to learn how to make my own macros!

bh2smith:
All that remains is a `cargo fmt`

driver/src/price_finding/naive_solver.rs
```diff
@@ -601,16 +602,26 @@ pub mod tests {
             let buy_token_price = solution.prices[order.buy_token as usize];
             let sell_token_price = solution.prices[order.sell_token as usize];
 
-            let exec_sell_amount = solution.executed_sell_amounts[i];
-            let mut exec_buy_amount = if buy_token_price > 0 {
-                (exec_sell_amount * sell_token_price).ceiled_div(buy_token_price)
+            let exec_buy_amount = solution.executed_buy_amounts[i];
+            let exec_sell_amount = if sell_token_price > 0 {
```
bh2smith: New question... If we only submit buy amounts to the contract, do we really need to compute sell amounts? I know that the Snapp Contracts require this, but is that the only reason? Also, should we be looking at running the error shifting algorithm here along with the naive solver as well? I would be very excited to implement [Edmond's algorithm](https://en.wikipedia.org/wiki/Edmonds%27_algorithm) in Rust!

driver/src/util.rs
```diff
@@ -13,6 +14,10 @@ pub fn u128_to_u256(x: u128) -> U256 {
     U256::from_big_endian(&x.to_be_bytes())
 }
 
+pub fn u256_to_u128(x: U256) -> u128 {
```
bh2smith: I am pretty sure this already exists [here](https://github.com/gnosis/dex-services/pull/249/commits/e07b1c8a25ec3db3144a04fa7d85f3e01aad60c9),. Did you cherry pick it? Test it?

driver/src/price_finding/naive_solver.rs
```diff
@@ -212,109 +212,126 @@ fn order_with_buffer_for_fee(order: &Order, fee: &Option<Fee>) -> Order {
 #[cfg(test)]
 pub mod tests {
     use super::*;
+    use crate::util::{u256_to_u128};
```
bh2smith: This is where the Travis is failing!

```suggestion
    use crate::util::u256_to_u128;
```

driver/src/util.rs
fleupold: I cherry picked the code, forgot the tests.

driver/src/price_finding/naive_solver.rs
fleupold: I don't think the error shifting will be necessary since we only match two single orders, where one side is already the fee token. Thus we never incur any rounding errors and thus also don't need to shift any errors.

As for computing the sell amounts, this is still required for the original dƒusion system and thus should probably stay here for now.

driver/src/util.rs
bh2smith: Now worries, it will be there eventually.

PR #253, opened 2019-09-24
[PoC] Submit Buy Amounts to StableX Contracts

bh2smith:
As discussed, we submit executed buy amounts to the smart contract which uniquely determines the sell amounts according to equation (2) from our discussion [here](https://github.com/gnosis/dex-contracts/issues/173#issuecomment-526163117)

This change refactors those (previously submitted) sell amounts to buy amounts. 

**Test Plan:** Cargo.

**NOTE** should not be merged before contracts are updated or, at the very latest before e2e testing is complete.


driver/src/contracts/stablex_contract.rs
```diff
@@ -131,7 +130,7 @@ fn encode_execution_for_contract(
             order_ids.push(U128::from(order_batch_info.slot_index));
             // all orders are sell orders, so volumes are sell_amounts.
```
fleupold: These two comments seem outdated now.

PR #257, opened 2019-09-25
Stablex travis

fleupold:
Goal of this PR is to make the e2e tests from the previous PRs run continuously on travis

### Test Plan

See demo run in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,45 +1,43 @@
 if: (branch = master) OR (type = pull_request) OR (tag IS present)
-language: generic
-  dist: xenial
-  sudo: required
-  cache: 
-    directories:
-      - $HOME/.nvm/versions/node/v8.12.0/lib/node_modules
-      - dex-contracts/node_modules/
-  install:
-    - git submodule update --init
-    - npm install -C dex-contracts
-    - cd dex-contracts && npx truffle compile && cd -
-  after_failure:
-    - docker-compose logs
-matrix:
-  include:
-    - language: rust
-      rust: 1.36.0
-      cache: cargo
-      before_cache:
-        - find ./target/debug -type f -maxdepth 1 -delete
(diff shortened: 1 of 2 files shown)


fleupold:
Ready for review!

.travis.yml
```diff
@@ -17,23 +18,30 @@ matrix:
...
         directories:
-          - $HOME/.nvm/versions/node/v8.12.0/lib/node_modules
           - dex-contracts/node_modules/
       install:
-        - git submodule update --init
-        - npm install -C dex-contracts
-        - cd dex-contracts && npx truffle compile && cd -
-        - docker-compose up -d driver graph-listener | while read line ; do echo "$(date)| $line"; done; # prints timestamps for profiling
-        - cd dex-contracts && npx wait-port -t 30000 8545 && npx truffle migrate && cd -
+        - ./test/setup_contracts.sh
```
josojo: Why do you start ganache from a docker? 
Didn't we had that before and changed it, since running ganache "naitively" was faster?

test/setup_contracts.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+docker-compose up -d ganache-cli
+cd dex-contracts
+npm install --production
+npx wait-port -t 30000 8545
+npx truffle migrate
```
josojo: minor: You could use the migration flag --onlyMigrateStableX or --onlyMigrateDfusion to save some milliseconds

fleupold: This would affect the contract addresses though. Not sure if it's worth it.

.travis.yml
fleupold: This has been like this for a while. 
I don't fully remember why this was better, but for one we can get better logging in docker. I also don't think it's much faster but am happy to be convinced otherwise (not subject to this PR though).

bh2smith: I am in favour of having ganache as part of the container orchestration (since is it a part of the project). Leaving it out completely seems somewhat incomplete. Like the demo wouldn't run without it.

PR #268, opened 2019-10-01
Running client against Rinkeby

josojo:
This PR enables us to run our driver against rinkeby.

---
testplan:
Run
```
docker-compose build --build-arg use_solver=1 stablex-rinkeby
docker-compose up stablex-rinkeby
```
and see
```
stablex_1         |     Finished dev [unoptimized + debuginfo] target(s) in 0.25s
stablex_1         |      Running `target/debug/stablex`
stablex_1         | 2019-10-11 11:23:39,375 INFO  [driver::util] Using linear optimisation price finder
stablex_1         | 2019-10-11 11:23:43,778 INFO  [driver::driver::stablex_driver] Computed solution: Solution { surplus: Some(103390010673887838820432721250764750708), prices: [1000000000000000000, 1999997999985000000, 1999999996001000000, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], executed_buy_amounts: [522166492440842636996, 0, 0, 1042244320991726700000, 521643805343190302297, 0, 0, 0], executed_sell_amounts: [1045377317858726621541, 0, 0, 521643805343190302297, 522166492440842636996, 0, 0, 0] }
stablex_1         | 2019-10-11 11:23:44,022 INFO  [driver::contracts::base_contract] Sending tx from 0x740a…e96f with nonce 75
stablex_1         | 2019-10-11 11:23:44,231 INFO  [driver::driver::stablex_driver] Successfully applied solution to batch 5235975
```


docker-compose.yml
```diff
@@ -4,8 +4,6 @@ x-rust-service:
   build:
     context: .
     dockerfile: docker/rust/Dockerfile
-  depends_on:
```
fleupold: But for the development version we still need ganache, right?

common-rinkeby.env
```diff
@@ -0,0 +1,3 @@
+STABLEX_CONTRACT_ADDRESS=0x9046451F7cF124c1d7d1832F76F5e98a33D1610E
```
fleupold: #271 added a few more variables that are needed for rinkeby. We probably don't want to expose the PK but at least the NETWORK_ID

docker-compose.yml
josojo: yeah, if we want to run it locally, we would need to type docker-compose up ganache-cli, stablex.

I will do more research how we could do it. If you have another suggestion, let me know

README.md
```diff
@@ -133,3 +133,13 @@ or rebuild everything if you are desperate (will take longer, but might solve ot
...
 docker-compose build
 ```
+
+#### Different networks:
+
+In order to start StableX for the Rinkeby network, overwrite certain environmental variables by adding common-rinkeby.env in the docker-compose, like this:
+
+```
+env_file:
+    - common.env
```
fleupold: I was hoping we could find a way where a different command could invoke the rinkeby version of our services. Something like:

`docker-compose stablex --env_file common-rinkeby.env`

Since that doesn't seem to be possble maybe it'd be better to use a separate target (e.g. stablex rinkeby)? We can probably reuse the existing code by using << yaml inclusion similar to what we do for `&rust-service`?

docker-compose.yml
bh2smith: I am not sure I understand this comment. ganache is still directly below this line.

docker/truffle/Dockerfile
```diff
@@ -7,4 +7,5 @@ COPY docker/truffle/run.sh /app/run.sh
 
 # Dependencies will Install app dependencies
 RUN apk add --update --no-cache --virtual build-dependencies git python make g++ ca-certificates
-RUN npm install --production
\ No newline at end of file
+RUN npm install --production
+RUN npm run network-inject
```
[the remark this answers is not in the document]
josojo: Yes, they are. But anyway, we need to run them here in the docker setup to insert the newest addresses

PR #271, opened 2019-10-08
[Driver] Signing txs in rust

fleupold:
This PR introduces functionality to sign transactions by the driver in rust before submitting them to the ethereum client.

This is needed to deploy our system on rinkeby and mainnet since there we don't have "unlocked" clients (such as ganache for testing). Therefore our transactions are rejected as invalid.

Unfortunately our rust web3 library didn't have built in support for this (there is an open issue but not yet an implementation https://github.com/tomusdrw/rust-web3/issues/116).

With this change, we now provide the network id and private key in our env file and sign the raw transaction using the `ethereum-tx-sign` library. This will make the system work with both unlocked and locked accounts.

### Test Plan

In common.env set `NETWORK_ID` to 4, `PRIVATE_KEY` to your rinkeby eth funded account's pk, `STABLEX_CONTRACT_ADDRESS` to `0x9046451F7cF124c1d7d1832F76F5e98a33D1610E`

`docker-compose up`

Then use the truffle scripts to add tokens, deposits and orders (note that you will have to make a trade happen for the driver to actually send a transaction).

Alternatively, you can remove this if check 
https://github.com/gnosis/dex-services/blob/89448372eab3a860ecdf673607acbf54f05d4d0f/driver/src/driver/stablex_driver.rs#L42
to make it also submit trivial solutions.

See transaction going through successfully.


common.env
```diff
@@ -3,6 +3,8 @@ SNAPP_BASE_LIB_ADDRESS=0x9561C133DD8580860B6b7E504bC5Aa500f0f06a7
 STABLEX_CONTRACT_ADDRESS=0x9b1f7F645351AF3631a656421eD2e40f2802E6c0
 ETHEREUM_NODE_URL=http://ganache-cli:8545
 NETWORK_NAME=dev
+NETWORK_ID=42
```
bh2smith: Does the 42 represent ganache/localhost as a default network?

driver/Cargo.toml
```diff
@@ -12,7 +12,9 @@ path = "src/lib.rs"
 bson = "0.9.1"
 byteorder = "1.3.1"
 chrono = "0.4.6"
+clarity = "0.1.8"
 ethabi = "6.1.0"
+ethereum-tx-sign = { git = "https://github.com/synlestidae/ethereum-tx-sign.git" }
```
bh2smith: Is there any particular reason we are pegged to a specific repo here? Does this imply the package is not published in cargo?

driver/src/contracts/base_contract.rs
```diff
@@ -13,6 +21,10 @@ pub struct BaseContract {
     pub contract: Contract<web3::transports::Http>,
     pub web3: web3::Web3<web3::transports::Http>,
     event_loop: web3::transports::EventLoopHandle,
+    abi: ethabi::Contract,
+    network_id: u8,
+    public_key: H160,
+    private_key: H256,
```
bh2smith: Will this private key be stored in the driver's memory? Is this something one would need to be cautious about when deploying a driver?

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+            .eth()
+            .send_raw_transaction(Bytes::from(signed_tx))
+            .wait()
+            .map_err(DriverError::from)
     }
 }
+
+fn address_from_private_key(pk: &H256) -> Result<H160> {
+    let public_key = PrivateKey::from_slice(&pk.to_vec())
+        .and_then(|pk| pk.to_public_key())
```
bh2smith: I assume `to_public_key` is a function implemented for `H256`

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+    pub fn send_signed_transaction<P>(
+        &self,
+        function: &str,
+        params: P,
+        options: Options,
+    ) -> Result<H256>
+    where
+        P: Tokenize,
+    {
+        let nonce = if let Some(nonce) = options.nonce {
```
bh2smith: I am not sure I see what the `Some` is doing here. I read it like "If nonce is an attribute of options"... is this what it means?

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+                .wait()?
+        };
+
+        info!("Sending tx from {} with nonce {:x}", self.public_key, nonce);
+
+        let signed_tx = self
+            .abi
+            .function(function)
+            .and_then(|function| function.encode_input(&params.into_tokens()))
+            .map(move |data| {
```
bh2smith: Can you elaborate on the requirement for "move" here? I guess we are "forcing our closure to take ownership of its environment", but I don't really know

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+                let tx = RawTransaction {
+                    nonce,
+                    to: Some(self.contract.address()),
+                    value: options.value.unwrap_or_default(),
+                    gas_price: options
+                        .gas_price
+                        .unwrap_or_else(|| U256::from(1_000_000_000)),
+                    gas: options.gas.unwrap_or_else(|| U256::from(100_000)),
```
bh2smith: Would this be the place we would like to get an appropriate gas estimate for?

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+                    data,
+                };
+                tx.sign(&self.private_key, &4)
```
bh2smith: What is this &4? Are we sure this shouldn't be using the environment variable `NETWORK_ID`?

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+            })?;
+
+        self.web3
+            .eth()
+            .send_raw_transaction(Bytes::from(signed_tx))
```
bh2smith: is `send_raw` different, or preferred, to some sort of just `send`

driver/src/contracts/snapp_contract.rs
```diff
@@ -276,20 +276,12 @@ impl SnappContract for SnappContractImpl {
         new_state: H256,
         deposit_hash: H256,
     ) -> Result<()> {
-        let account = self
-            .base
-            .account_with_sufficient_balance()
```
bh2smith: So do we no longer ensure that the account has sufficient balance?

driver/src/contracts/snapp_contract.rs
```diff
@@ -302,24 +294,16 @@ impl SnappContract for SnappContractImpl {
         withdraw_hash: H256,
     ) -> Result<()> {
         // HERE WE NEED TO BE SURE THAT THE SENDING ACCOUNT IS THE OWNER
```
bh2smith:
```suggestion
        // SENDING ACCOUNT MUST BE CONTRACT OWNER
```

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -184,7 +184,7 @@ pub mod tests {
         ];
         let mut order_count = HashMap::new();
         StableXAuctionElement::from_bytes(&mut order_count, &bytes);
-        let mut bytes_modified = bytes.clone();
+        let mut bytes_modified = bytes;
```
bh2smith: interesting... did clippy point this out?

common.env
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FJ336VCs1JC42zGRhjH%2Fgiphy.gif&token=yes" data-gitmeme-token="yes"></a>

driver/src/contracts/base_contract.rs
fleupold: if options.nonce is Some (as in it's not None) then the local variable nonce will have the value of the option (unwrapped).

driver/src/contracts/base_contract.rs
fleupold: I wouldn't be too concerned about memory. More importantly it will be an environment variable and thus also stored on disk somewhere. But I believe this is how most services work these days. Alternatively we could allow the private key to be entered on startup. But not sure that's worth the hassle.

driver/src/contracts/base_contract.rs
fleupold: You are right that it's not needed since none of the variables used are referenced outside the closure later. Thus a mutable borrow is fine.

driver/src/contracts/base_contract.rs
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FWJKdDTR8gPD0Y%2Fgiphy-downsized-medium.gif&token=good_catch" data-gitmeme-token="good_catch"></a>

driver/src/contracts/base_contract.rs
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FA3dD2XWfJ7tV6%2Fgiphy.gif&token=indeed" data-gitmeme-token="indeed"></a>

driver/src/contracts/base_contract.rs
fleupold: It's implemented on PrivateKey: https://docs.rs/clarity/0.1.23/clarity/private_key/struct.PrivateKey.html#method.to_public_key

driver/src/contracts/stablex_auction_element.rs
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FNEvPzZ8bd1V4Y%2Fgiphy-downsized-medium.gif&token=yes" data-gitmeme-token="yes"></a>

driver/src/contracts/snapp_contract.rs
fleupold: The tx will fail if we don't have sufficient balance and it will say in the error message.

driver/src/contracts/base_contract.rs
fleupold: https://stackoverflow.com/a/50992399/11476590 applies to rust as well.

driver/Cargo.toml
fleupold: Good point. We need the chain_id parameter in the sign functions (added in version 1.0.0 here: https://github.com/synlestidae/ethereum-tx-sign/pull/3/files). The latest version on Cargo is 0.0.2.

But I will pin to a specific commit/tag and created an issue: https://github.com/synlestidae/ethereum-tx-sign/issues/11

driver/src/contracts/base_contract.rs
bh2smith: Yes... but I find it weird that you are using `Some(nonce)` where `nonce` doesn't appear to exist yet...

fleupold: It's a different variable (I could call it foo) and basically says whatever the value inside the option is (which is of type Some(value)) should be bound to that variable.

driver/src/contracts/base_contract.rs
```diff
@@ -1,10 +1,18 @@
+extern crate ethereum_tx_sign;
```
nlordell: If I am not mistaken `extern crate` statements should no longer be needed in edition 2018 which seems to be used by the driver.

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+        let abi_string = json.get("abi").ok_or("No ABI for contract")?.to_string();
+        let abi = abi_string.as_bytes();
 
         let decoded_address = hex::decode(&address[2..])?;
         let contract_address: Address = Address::from(&decoded_address[..]);
-        let contract = Contract::from_json(web3.eth(), contract_address, abi.as_bytes())?;
+        let contract = Contract::from_json(web3.eth(), contract_address, abi)?;
+
+        let network_id = env::var("NETWORK_ID")?.parse()?;
+        let private_key = H256::from_str(&env::var("PRIVATE_KEY")?).unwrap();
```
nlordell: Should this error not also be propagated in case it fails to parse the private key?

driver/src/contracts/base_contract.rs
nlordell: If I am not mistaken, `extern crate` statements are no longer needed in edition 2018 which I believe is being used.

nlordell: oops, looks like the comment got added twice?

driver/src/contracts/base_contract.rs
nlordell: As we require storage of private keys and other sensitive information, are there any future plans to do a security audit of the crate dependencies we use?

driver/src/contracts/base_contract.rs
nlordell: Its part of rust's pattern matching:
```rust
if let /* pattern */ {
    // pattern matched, you can use extracted bindings from the pattern here
} else {
    // pattern didn't match
}
```
 the full pattern syntax is here: https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html

driver/src/contracts/stablex_auction_element.rs
nlordell: @fleupold your memes are on point...

driver/src/contracts/base_contract.rs
nlordell: Also, for symmetry with the previous line, it is possibble to write:
```
let private_key = env::var("PRIVATE_KEY")?.parse()?;
```
and let type inference do its magic since `str::parse` is implemented for all `FromStr`.

driver/src/contracts/base_contract.rs
```diff
@@ -22,26 +34,81 @@ impl BaseContract {
...
+        let abi_string = json.get("abi").ok_or("No ABI for contract")?.to_string();
+        let abi = abi_string.as_bytes();
 
         let decoded_address = hex::decode(&address[2..])?;
         let contract_address: Address = Address::from(&decoded_address[..]);
-        let contract = Contract::from_json(web3.eth(), contract_address, abi.as_bytes())?;
+        let contract = Contract::from_json(web3.eth(), contract_address, abi)?;
+
+        let network_id = env::var("NETWORK_ID")?.parse()?;
+        let private_key = H256::from_str(&env::var("PRIVATE_KEY")?)?;
```
nlordell: Also, for symmetry with the previous line, it is possibble to write:
```
let private_key = env::var("PRIVATE_KEY")?.parse()?;
```
and let type inference do its magic since `str::parse` is implemented for all `FromStr`. Then the `use std::str::FromStr` would no longer be needed.

bh2smith: Noice!

driver/src/contracts/base_contract.rs
fleupold: Cool, let me remove all of those in a cleanup PR

driver/src/contracts/base_contract.rs
fleupold: @anxolin what did we do in this regard for the dutchX. Did we have an audit for the services or the dependencies they require? I think our general strategy is to use accounts, which don't have a ton of eth and fund them as we go.

tritone11:
Be careful that until https://github.com/synlestidae/ethereum-tx-sign/pull/13 is merged in ethereum-tx-sign, transactions signed by the library sent to the node have a chance to fail if they have leading zero bytes as they are not currently removed.

driver/src/contracts/base_contract.rs
anxolin: Hi, I missed this comment since back then I didn't have the Gmail tagging rules I have now. Hopefully I'll miss less comments now. Let me know if I'm being unresponsive anytime, sometimes it can take me a bit to answer but I'll try not to miss this ones.

For DutchX we started simple, as you suggest, with accounts with limited amounts. We used a mnemonic, and we were using one account per bot. We use to have dozens of bots.

Then we added another layer of security by using the safe. We wrote a module that was allowing ONLY OUR ADDRES to trade ONLY SOME TOKENS and ONLY IN DUTCHX CONTRACT

So the module didn't allow to withdraw money, only trade

Issue #279, opened 2019-10-30
[PoC] Driver to compute objective value of solutions


bh2smith:
Upon receiving solution from the solver, the driver should compute the objective value and determine (from smart contract) if the obtained solution is better than what is currently stored in the contract. This way, the driver can pre-determine if it worth submitting its solution.

This issue is related to the analogous [smart contract issue](https://github.com/gnosis/dex-contracts/issues/248) which (upon implementation) will require this proposed objective value for solution submission.

bh2smith:
For evaluations of objective Function we have already implemented these in Solidity and Javascript as follows:

**Total Utility** is evaluated [here](https://github.com/gnosis/dex-contracts/pull/236/files)

**Total Disregarded Utility** is evaluated here [here](https://github.com/gnosis/dex-contracts/pull/229/files)

nlordell:
So it looks like the driver uses reported utility values from the solver to compute an objective value. I don't know if this objective value matches our current objective value. Should it be ignored? Also there are quite a few unit tests that test computed objective values for things like the naive solver

bh2smith:
I suppose this issue should be closed too!

PR #281, opened 2019-10-30
Add objective value calculation to solution

nlordell:
This PR implements solution objective value calculation. This allows the driver to submit a solution with its objective value to the contract. The contract can then skip verification if the reported objective value is lower than the latest winning solution and save gas.

This PR is the continuation from PR #291 as this was too big and got difficult to review.

This PR addresses issue #279. Note that other issues arose from this implementation and are being tracked here: #285 #286 #290 

### Testplan

Unit tests were added with and will be run on CI to verify calculations


dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
         }
     }
+
+    /// Calculates the utility of an order based on a buy and sell price as well
+    /// as the executed buy amount.
+    ///
+    /// # Returns
+    ///
+    /// Returns `Some(U256)` utility result if the inputs were valid and it
+    /// it produced a valid positive utility.
```
bh2smith: "... were valid and it it ..."

double "it"!

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
+    pub fn utility(
+        &self,
+        buy_price: u128,
+        sell_price: u128,
+        exec_buy_amount: u128,
+    ) -> Option<U256> {
+        // first lets convert everything into U256s
```
bh2smith: nit:
```suggestion
        // first convert everything into U256s
```

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
+        let fee_denominator = U256::from(2);
+
+        // adj_sell_amount =
+        //   (
+        //     (
+        //       (exec_buy_amount * buy_price) / (fee_denominator - 1)
+        //     ) *
+        //     fee_denominator
+        //   ) /
+        //   sell_price
```
bh2smith: Some of the line breaks here seem excessive... are you working with a 15 character line width?

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
+        //
+        // utility =
+        //   (
+        //     (exec_buy_amount * buy_price - adj_sell_amount * buy_amount) *
```
bh2smith:
```suggestion
        //     (exec_buy_amount * sell_amount - adj_sell_amount * buy_amount) *
```

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
+        //     buy_price
+        //   ) /
+        //   sell_amount
+        let adj_sell_amount = exec_buy_amount
+            .checked_mul(buy_price)?
+            .checked_div(fee_denominator - 1)?
+            .checked_mul(fee_denominator)?
+            .checked_div(sell_price)?;
+        let utility = exec_buy_amount
+            .checked_mul(buy_price)?
```
bh2smith:
```suggestion
            .checked_mul(order.sell_amount)?
```

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
+            .checked_sub(adj_sell_amount.checked_mul(buy_amount)?)?
+            .checked_mul(buy_price)?
+            .checked_div(sell_amount)?;
+
+        Some(utility)
```
bh2smith: We must be cautious here that your utility is computed based on the formula in the contract's comment, but the contract actually makes one step of simplification canceling out `order.sell_amount` from the first term with the subtraction and moves the order of operations on the division of the "sell_amount". I believe this was done to reduce the amount of precision lost in the calculation and may have been missed because you currently have any extra copy of `buy_price` in the place where `order.sell_amount` should be!

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
+    ///
+    /// Returns `Some(U256)` disregarded utility result if the inputs were valid
+    /// and it it produced a valid positive utility.
+    pub fn disregarded_utility(
+        &self,
+        buy_price: u128,
+        sell_price: u128,
+        exec_sell_amount: u128,
+    ) -> Option<U256> {
+        // first lets convert everything into U256s
```
bh2smith: nit:
```suggestion
        // first convert everything into U256
```

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,90 @@ impl Order {
...
+        // disregarded_utility =
+        //   (
+        //     (sell_price * sell_amount - buy_price * buy_amount) *
+        //     (sell_amount - exec_sell_amount)
+        //   ) /
+        //   sell_amount
+
+        sell_price
+            .checked_mul(sell_amount)?
+            .checked_sub(buy_price.checked_mul(buy_amount)?)?
```
bh2smith: ubernit: I would like to swap the order of `buy_amount` and `buy_price` here because, although it appears there is a symmetry in the naming convention, the term itself is a result of cross multiplication. I was hoping to preserve some of the intuition in the calculation without too many mathe-magical tricks (like multiplicative commutativity of most number systems)

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +16,65 @@ pub struct Solution {
...
+    /// Returns `Some(U256)` for any solution that is positive valid solution.
+    /// Returns `None` for invalid solutions or orders. A solution might be
+    /// invalid if:
+    /// - The number of orders do not match the solution buy/sell amounts
+    /// - The solution values lead to invalid utility or disregarded utility
+    ///   results
+    /// - The orders reference tokens that are not priced by the solution
+    pub fn objective_value(&self, orders: &[Order]) -> Option<U256> {
+        if orders.len() != self.executed_buy_amounts.len()
+            || orders.len() != self.executed_sell_amounts.len()
```
bh2smith: There will likely be many more orders in an auction than can be touched. Is this condition prepared to handle that? Are we assuming that `objective_value` is only ever being passed a collection of touched orders?

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +16,65 @@ pub struct Solution {
...
+                // this order was not touched, so skip
+                continue;
+            }
+
+            u = u.checked_add(order.utility(buy_price, sell_price, exec_buy_amount)?)?;
+            du = du.checked_add(order.disregarded_utility(
+                buy_price,
+                sell_price,
+                exec_sell_amount,
+            )?)?;
```
bh2smith: These actually turn out to look super clean when all the ugly calculations are buried deep inside the `utility` and `disregarded utility` functions!

dfusion_rust_core/src/models/solution.rs
```diff
@@ -149,4 +198,37 @@ pub mod unit_test {
         };
         assert_eq!(parsed_solution, expected);
     }
+
+    #[test]
+    fn test_objective_value_is_zero_for_trivial_solution() {
```
bh2smith: cool test!

dfusion_rust_core/src/util.rs
```diff
@@ -0,0 +1,42 @@
+use std::str::FromStr;
+use web3::types::U256;
+
+pub fn u128_to_u256(x: u128) -> U256 {
+    U256::from_big_endian(&x.to_be_bytes())
+}
+
+pub fn u256_to_u128(x: U256) -> u128 {
+    u128::from_str(&x.to_string()).unwrap()
+}
```
bh2smith: Were these moved from somewhere else? I feel like I've seen them before.

dfusion_rust_core/src/util.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[test]
+    fn test_u128_to_u256() {
+        assert_eq!(
+            u128_to_u256(u128::max_value()),
+            U256::from_dec_str("340282366920938463463374607431768211455").unwrap(),
+            "failed on 128::max_value()"
+        );
+        assert_eq!(u128_to_u256(1u128), U256::from(1), "failed on 1u128");
+        assert_eq!(u128_to_u256(0u128), U256::from(0), "failed on 0u128");
+    }
```
bh2smith: I have definitely seen this all the code contained in this file before... was it renamed?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -291,14 +294,18 @@ pub mod tests {
...
             executed_buy_amounts: vec![0, 95_042_777_139_162_480_000],
         };
 
         let solution = deserialize_result(&json, 2).expect("Should not fail to parse");
         assert_eq!(solution, expected_solution);
+
+        // TODO(nlordell): we need orders in order to calcaulate this utility
+        let expected_objective_value =
+            U256::from_dec_str("15854632034944469292777429010439194350").ok();
+        let _ = expected_objective_value;
```
bh2smith: Would this be a place to use some sort of mock?

driver/src/util.rs
```diff
@@ -12,13 +11,7 @@ use crate::price_finding::{Fee, LinearOptimisationPriceFinder, NaiveSolver, Pric
 
 const BATCH_TIME_SECONDS: u32 = 3 * 60;
 
-pub fn u128_to_u256(x: u128) -> U256 {
-    U256::from_big_endian(&x.to_be_bytes())
-}
-
-pub fn u256_to_u128(x: U256) -> u128 {
-    u128::from_str(&x.to_string()).unwrap()
-}
```
bh2smith: I see now!

dfusion_rust_core/src/util.rs
nlordell: They were moved from `driver` crate to `dfusion_core` crate as it was needed in both now.

driver/src/price_finding/linear_optimization_price_finder.rs
nlordell: I think here we would update the `expected_objective_value` to be what is actually expected given the orders, solution and the new objective function and then test against that. I left a TODO so I don't forget to come back and fix this test.

dfusion_rust_core/src/models/order.rs
nlordell: Oops

dfusion_rust_core/src/models/order.rs
nlordell: Oops

dfusion_rust_core/src/models/order.rs
nlordell: I'll make it less line-y

dfusion_rust_core/src/util.rs
nlordell: See above, moved from `driver/src/util.rs` to `dfusion_rust_core/src/util.rs`

dfusion_rust_core/src/models/solution.rs
nlordell: From what I understand the solution `executed_buy_amounts` and `executed_sell_amounts` are created with the length of `num_orders` and are just 0 for untouched orders.

nlordell: If this isn't true anymore, then we would have to change the solution function so that `excuted_{buy,sell}_amount` is a `map(order_id -> u128)` instead of just a vector; or else I don't see a way to recover the values for a given order.

dfusion_rust_core/src/models/order.rs
nlordell: Easy fix

dfusion_rust_core/src/models/order.rs
nlordell: I think I just made a mistake copying the formula from the contract to my comment and then repeated it when implementing. I wasn't trying to be clever or anything.

nlordell:
Addressed some of your comments @bh2smith.

nlordell:
To imitate the contract objective function, we need to calculate the total fees which requires `i256` math. I have a minimum `I256` implementation with overflow checking.

dfusion_rust_core/src/util.rs
```diff
@@ -1,12 +1,25 @@
...
 pub fn u256_to_u128(x: U256) -> u128 {
-    u128::from_str(&x.to_string()).unwrap()
+    let mut bytes = [0u8; 32];
+    x.to_big_endian(&mut bytes[..]);
+
+    let hi = u128::from_be_bytes(bytes[..16].try_into().expect("correct slice length"));
+    let lo = u128::from_be_bytes(bytes[16..].try_into().expect("correct slice length"));
+
+    assert_eq!(hi, 0, "U256 to u128 overflow");
+    lo
```
bh2smith: Was there something wrong with relying on the errors that I would assume are emitted upon overflow from `u128::from_str`? We all thought the one liner was more elegant and relied on already tested code.

nlordell: It was a bit of nit-pick of mine to convert an int to a string and then back to an int when really just the LSBs of the larger int need to be read.

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +17,78 @@ pub struct Solution {
...
+            return None;
+        }
+
+        // note that we don't actually have to do the same checking here (such as
+        // account balances and such) as we do in the smart contract - we only
+        // care if the solution is valid given the provided orders
+
+        let mut u = U256::zero();
+        let mut du = U256::zero();
+        let mut fee = I256::zero();
```
josojo: Maybe we don't even need the I256.

If we would run two times through the orders, and first do all the feeToken balance adjustments for the cases where the feeToken is bought, and then, in a second loop, doing all the adjustments for the cases where the feeToken is sold, the fee should never be negative.

And later for calculating the objective value, we would just add and substract web3::U256 types with checks that we will never get negative. ( If we decided for this strategy, you check the smart contracts. They are already implementing this).

I think, in the near future, we don't need the I256, and hence, we could try to do it without. 
But for sure, for simplicity, future-readiness and readability of code etc, we could go with this variant.

nlordell:
@josojo I am going to split this PR as it is getting really big

dfusion_rust_core/src/models/solution.rs
nlordell: That's a great idea! We can leave the `I256` in a branch somewhere if we want it in the future.

nlordell:
@josojo implemented your idea of using two U256s instead of I256 - this makes this PR much more manageable

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,73 @@ impl Order {
...
     }
+
+    /// Calculates the utility of an order based on a buy and sell price as well
+    /// as the executed buy amount.
+    ///
+    /// # Returns
+    ///
+    /// Returns `Some(U256)` utility result if the inputs were valid and it
+    /// produced a valid positive utility.
+    pub fn utility(
```
fleupold: Can we add a unit test for `utility` and `disregarded_utility`? Those could also include cases where this returns None.

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,73 @@ impl Order {
...
+        let buy_price = u128_to_u256(buy_price);
+        let exec_buy_amount = u128_to_u256(exec_buy_amount);
+        let exec_sell_amount = u128_to_u256(exec_sell_amount);
+        let buy_amount = u128_to_u256(self.buy_amount);
+        let sell_amount = u128_to_u256(self.sell_amount);
+
+        // utility = (exec_buy_amount - (exec_sell_amount * buy_amount) / sell_amount)
+        //   * buy_price
+
+        let utility = exec_buy_amount
```
fleupold: Do we need checked operations here? Are there examples where this can overflow?

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +17,81 @@ pub struct Solution {
...
+            }
+
+            u = u.checked_add(order.utility(buy_price, exec_buy_amount, exec_sell_amount)?)?;
+            du = du.checked_add(order.disregarded_utility(
+                buy_price,
+                sell_price,
+                exec_sell_amount,
+            )?)?;
+
+            if order.buy_token == 0 {
```
fleupold: Is this correct? We would only count a fee if we are trading with the fee token (buying or selling it). While this is a restriction that the naive solver does, the "real" solver would also trade non-fee token with each other while subtracting a fee. I think we might have to compute the token conservation excess for the fee token and half that value.

dfusion_rust_core/src/models/solution.rs
```diff
@@ -149,4 +215,37 @@ pub mod unit_test {
...
+            Some(U256::zero())
+        );
+        assert_eq!(
+            Solution::trivial(5).objective_value(&orders),
+            Some(U256::zero())
+        );
+    }
+
+    #[test]
+    fn test_objective_value_for_non_trivial_solution() {}
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F4Zd5CCT47enl32Sx3P%2Fgiphy-downsized-medium.gif&token=where" data-gitmeme-token="where"></a>

driver/src/price_finding/naive_solver.rs
```diff
@@ -151,18 +154,29 @@ impl PriceFinding for NaiveSolver {
...
+                    .objective_value(
+                        prices[y.buy_token as usize],
+                        prices[x.buy_token as usize],
+                        exec_buy_amount[j],
+                        exec_sell_amount[j],
+                    )
+                    .unwrap();
+                let u = x_u + y_u;
+                let du = x_du + y_du;
+                if u < du {
```
fleupold: Can we add a unit test for that?

driver/src/price_finding/naive_solver.rs
```diff
@@ -110,19 +113,19 @@ impl PriceFinding for NaiveSolver {
...
         let mut exec_sell_amount: Vec<u128> = vec![0; orders.len()];
-        let mut total_objective_value = U256::zero();
 
         let mut found_flag = false;
 
         for (i, x) in orders.iter().enumerate() {
             for j in i + 1..orders.len() {
                 // Preprocess order to leave "space" for fee to be taken
-                let x = order_with_buffer_for_fee(&x, &self.fee);
+                let x = order_with_buffer_for_fee(x, &self.fee);
```
fleupold: Any particular reason for this change?

driver/src/price_finding/naive_solver.rs
```diff
@@ -234,7 +247,9 @@ pub mod tests {
         let solver = NaiveSolver::new(None);
         let res = solver.find_prices(&orders, &state).unwrap();
 
-        assert_eq!(Some(u128_to_u256(16 * 10u128.pow(36))), res.objective_value);
+        // -28_000_000_000_000_100_000_000_000_000_000_000_000
```
fleupold: That's a shame. @bh2smith can we come up with an example or a small tweak to the simple solver logic to avoid not getting a good solution for this very simple example?

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +17,81 @@ pub struct Solution {
...
+    /// - The number of orders do not match the solution buy/sell amounts
+    /// - The solution values lead to invalid utility or disregarded utility
+    ///   results
+    /// - The orders reference tokens that are not priced by the solution
+    ///
+    /// Note that just because this method returns some value does not mean that
+    /// the solution is actually valid WRT to the smart contract as it currently
+    /// does not verify that things like `execSellAmount * p[sellToken] * (1-phi)
+    /// == execBuyAmount * p[buyToken]` or token conservation hold.
+    pub fn objective_value(&self, orders: &[Order]) -> Option<U256> {
```
fleupold: Would it make sense to return a Result that can give more detail what about the solution is wrong?

dfusion_rust_core/src/models/order.rs
nlordell: In the works :smiley:

dfusion_rust_core/src/models/order.rs
nlordell: Yes we do need to check. Since we don't actually check if the solution is valid, things like having a higher `exec_sell_amount` than `sell_amount` can lead to overflows.

driver/src/price_finding/naive_solver.rs
bh2smith: Sorry, I am not sure I understand the request? What happened here?

bh2smith: Oh, now that I see the example yields negative objective and is no good! I will see what we can do here.

dfusion_rust_core/src/models/order.rs
nlordell: Well, we have a few options:
- assume that the solver always behaves, then we don't have to check anything and can just use regular math. This can leads to sub-optimal gas usage in that we submit solutions that won't actually succeed
- do some checking as with this math

dfusion_rust_core/src/models/solution.rs
nlordell: Its invisible...

dfusion_rust_core/src/models/solution.rs
nlordell: It depends on what we decide with how much checking we want to do. I think its a good idea and we can add it, maybe as a separate PR.

driver/src/price_finding/naive_solver.rs
nlordell: Nope, reverting. I think it happened when I was making changes to the naive solver and I left it in by accident.

dfusion_rust_core/src/models/solution.rs
nlordell: I'm not sure... I was hoping it was. Maybe I tried to do what `TokenConservation.sol` but only when it affects the token conservation for the fee token. I might have missed something though.

fleupold: Right now I don't think we accrue any fee in this logic if we are trading e.g. token1 and token2. However, in reality that also incurs a few (which just gets pushed around and eventually converted in token0)

dfusion_rust_core/src/models/order.rs
fleupold: Yes, if we check the solution, I'd suggest to do a separate method (something like https://github.com/gnosis/dex-services/blob/87a6826420f3a1410c0749ac2db56199da24a73f/driver/src/price_finding/naive_solver.rs#L604-L608).

For the purpose of utility calculation I'd suggest we assume a valid solution (since we don't check balances etc. anyways).

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +17,81 @@ pub struct Solution {
...
+            || orders.len() != self.executed_sell_amounts.len()
+        {
+            return None;
+        }
+
+        // note that we don't actually have to do the same checking here (such as
+        // account balances and such) as we do in the smart contract - we only
+        // care if the solution is valid given the provided orders
+
+        let mut u = U256::zero();
```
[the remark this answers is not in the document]
nlordell: I called them `total_utility` and `total_disregarded_utility` to be more descriptive

dfusion_rust_core/src/models/order.rs
nlordell: I removed some of the overflow checking where it isn't needed but kept it in some places where it could happen.

driver/src/price_finding/naive_solver.rs
nlordell: I've moved this to a separate issue #285, for this PR we can just ignore the objective value calculation in the naive solver.

driver/src/price_finding/naive_solver.rs
nlordell: I've moved this to a separate issue #285, for this PR we can just ignore the objective value calculation in the naive solver.

dfusion_rust_core/src/models/solution.rs
nlordell: I created issue #286 to address this.

dfusion_rust_core/src/models/solution.rs
nlordell: I am a little confused about how the fee token conservation works. From the contract, token conservation is calculated here:
```
tokenConservation.updateTokenConservation(
    order.buyToken,
    order.sellToken,
    tokenIdsForPrice,
    executedBuyAmount,
    executedSellAmount
);
```
and the calculation looks like this:
```
function updateTokenConservation(
    int[] memory self,
    uint16 buyToken,
    uint16 sellToken,
    uint16[] memory tokenIdsForPrice,
    uint128 buyAmount,
    uint128 sellAmount
) internal pure {
    uint buyTokenIndex = findPriceIndex(buyToken, tokenIdsForPrice);
    uint sellTokenIndex = findPriceIndex(sellToken, tokenIdsForPrice);
    self[buyTokenIndex] = self[buyTokenIndex].sub(int(buyAmount));
    self[sellTokenIndex] = self[sellTokenIndex].add(int(sellAmount));
} 
```

and from what I understood `{buy,sell}TokenIndex = 0` and `{buy,sell}Token = 0` for the fee token, that is why I wrote it the way I did.

I don't really understand how the calculation works.

nlordell: The other thing I can work from is: `exec_sell_amount * sell_price * (1 - phi) = exec_buy_amount * buy_price` so would the fee just be: `exec_buy_amount * buy_price - exec_sell_amount * sell_price` converted into token0?

dfusion_rust_core/src/models/order.rs
nlordell: Added a few unit tests

nlordell:
This PR ended up getting fairly large, however a lot of the added lines are unit test code. Let me know if you want me to split it in 2 to make the review easier. It could be something like:
- Add utility and disregarded utility calculation functions to order
- Calculate solution objective value

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,7 +17,7 @@ pub struct Solution {
 impl Solution {
     pub fn trivial(num_orders: usize) -> Self {
         Solution {
-            prices: vec![1; TOKENS as usize],
+            prices: vec![0; TOKENS as usize],
```
bh2smith: Funny, I was under the impression that this was changed a while back...

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +17,76 @@ pub struct Solution {
...
+            )?)?;
+            total_disregarded_utility = total_disregarded_utility.checked_add(
+                order.disregarded_utility(buy_price, sell_price, exec_sell_amount)?,
+            )?;
+
+            if order.buy_token == 0 {
+                fee_buy_amount = fee_buy_amount.checked_add(u128_to_u256(exec_buy_amount))?;
+            }
+            if order.sell_token == 0 {
+                fee_sell_amount = fee_buy_amount.checked_add(u128_to_u256(exec_buy_amount))?;
```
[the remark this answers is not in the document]
bh2smith: ha ha ha!

dfusion_rust_core/src/models/solution.rs
nlordell: It was, I unfixed it and then re-fixed it

dfusion_rust_core/src/models/order.rs
```diff
@@ -301,6 +368,143 @@ pub mod tests {
         );
     }
 
+    #[test]
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F1UYrFKWDkGmJGhVldV%2Fgiphy.gif&token=awesome_tests" data-gitmeme-token="awesome_tests"></a>

dfusion_rust_core/src/models/solution.rs
```diff
@@ -17,12 +17,79 @@ pub struct Solution {
...
+                // this order was not touched, so skip
+                continue;
+            }
+
+            total_utility = total_utility.checked_add(order.utility(
+                buy_price,
+                exec_buy_amount,
+                exec_sell_amount,
+            )?)?;
+            total_disregarded_utility = total_disregarded_utility.checked_add(
```
fleupold: Do we take the user's balance into account somehow? Disregarded utility should not be positive if the user had no remaining funds.

nlordell: No we currently don't account for balances at all. We could add it for sure, I created #290 for this.

nlordell:
Alright, I'll split this PR into 2 then.

nlordell:
I split this PR into 2, this is the second part of the calculation of the objective function, the first being #291.

I labeled this PR on hold until the former is merged.

nlordell:
Now that the order utility and disregarded utility calculations were merged, this PR should be ready to merge after review. Note that this still does not implement #290, #285 or #286

josojo:
@nlordell With the new idea of getting the objective value from solidity, this needs no longer to be reviewed, right? Or do you want to merge it anyways?

(The PR looks already so good. It hurts that we did not decide for this new solution earlier. Sorry!)

nlordell:
@josojo I think we can just close this as we have a new plan on how to deal with the objective value calculation.

PR #275, opened 2019-10-29
Adapting rust code to dex-contract PR with removed IsSellOrderFlag

josojo:
as title

---
testplan:
e2e tests & unit tests

PR #282, opened 2019-10-31
[Easy] resync submodule with master

fleupold:
We should never commit our submodule at a commit that is not included in the master branch of the original dex-contracts repo, since otherwise the two repos can diverge (otherwise the worst thing that can happen is that it is outdated).

#275 has committed `3c600d07c721df3b5f7c8192612af13a087c249b` which is not part of dex-contracts's master.

This PR makes it so that dex-contracts submodule track dex-contracts master branch again

### Testplan

CI


josojo:
Completely agreed! Sorry

PR #283, opened 2019-10-31
[Docker] fix truffle docker file

fleupold:
#280 excluded the git file in the truffle docker. However the command suggested doesn't actually copy the file structure as expected (it only copies all folders recursively but no top level files).

The reason things were still working are that we are always mounting the original dex-contracts repo as a volume in the docker container. Thus, removing the git file and most importantly doing any installation or injection are pointless.

This PR removes the unnecessary steps and makes sure we can actually run the migrations from within docker.

Note that it will still be required to run npm install locally before starting the container.

### Test Plan

```
cd dex-contract && rm -rf build && cd ..
docker-compose build truffle
docker-compose up truffle
```

See no errors

### DIFF
--- a/docker/truffle/Dockerfile
+++ b/docker/truffle/Dockerfile
@@ -2,11 +2,8 @@ FROM node:10.15-alpine
 
 # Create app directory
 WORKDIR /app/dex-contracts
-# Don't copy git artifacts as they contain backpointers to the parent repo
-COPY dex-contracts/[^.git]* ./
+COPY dex-contracts ./
 COPY docker/truffle/run.sh /app/run.sh
 
 # Dependencies will Install app dependencies
 RUN apk add --update --no-cache --virtual build-dependencies git python make g++ ca-certificates
-RUN npm install
-RUN npm run network-inject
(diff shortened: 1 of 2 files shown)


nlordell:
Everything worked like a charm.

docker/truffle/Dockerfile
```diff
@@ -2,11 +2,4 @@ FROM node:10.15-alpine
 
 # Create app directory
 WORKDIR /app/dex-contracts
-# Don't copy git artifacts as they contain backpointers to the parent repo
-COPY dex-contracts/[^.git]* ./
```
josojo: If you don't copy any dex-contracts files at all, how can it possibly work? Should then just "nothing migrate"?
Could you explain what you are doing? sorry, I don't see it right now.

bh2smith:
I am not sure what is the conflict with dex-contracts here...

docker/truffle/Dockerfile
fleupold: Check out 

https://github.com/gnosis/dex-services/blob/3268c7faca3f800b2c66bdf477ce854141239e77/docker-compose.yml#L36-L44

The contracts are included as a volume, thus they are taken from the local file system at runtime. When at runtime the run script is executed the most recent host contracts are used.

This means npm install/network-inject etc. can happen on the host, but you can still get your entire environment in a demo/test ready state by running `docker-compose up` (instead of `docker-compose up stablex` + doing migration in another shell)

fleupold:
> We might want to either stick with mounting or copying.

I agree, however I think we might need the build artefacts (at least very soon when we use ethcontract-rs) in order for the project to compile correctly. This happens at build time, whereas the mount is only available at runtime.

PR #287, opened 2019-11-05
[CI] build a stablex image on every master build

fleupold:
The plan is to build an image that uses the sophisticated solver on each travis build and push it to a private AWS docker repo.

Then, we can run a kubernetes pod which uses the image and drives the most recent rinkeby deployment.

### Test Plan

it builds: https://travis-ci.org/gnosis/dex-services/jobs/608409613


docker/deploy.sh
```diff
@@ -0,0 +1,20 @@
...
+
+sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
+
+# Get login token and execute login
+$(aws ecr get-login --no-include-email --region $AWS_REGION)
+echo -e $GITLAB_PRIVATE_KEY > .ssh/id_rsa
+chmod 0500 .ssh/id_rsa
+
+echo "Building latest image with solver...";
+docker-compose build --build-arg use_solver=1 stablex
```
josojo: maybe you want to build stablex-rinkeby?

docker/deploy.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -euo pipefail
+
+sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
+
+# Get login token and execute login
+$(aws ecr get-login --no-include-email --region $AWS_REGION)
+echo -e $GITLAB_PRIVATE_KEY > .ssh/id_rsa
```
josojo: Sry, my admin knowledge is soo bad:
What prevents me from downloading the docker image, run it and log in and simply read the .ssh/id_rsa file?

josojo: Ah, we never make the docker image public! I see.

docker/deploy.sh
fleupold: I didn't because the only difference is the .env file which doesn't get build into the image anyways (it only gets loaded when you run the image using docker-compose which on AWS we don't do).

In the long run, we should probably only have one var that indicates which network we want, so I thought we might be able to get rid of stablex-rinkeby and just pass this single parameter via the `-e` flag when running for rinkeby on our local machine.

docker/deploy.sh
fleupold: Right, we should probably delete the ssh file from the built image though just in case it ever leaks.

docker/deploy.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -euo pipefail
```
bh2smith: Cool. I just learned about this from a little article here:

https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

anxolin:
Good news! 👍 
congrats!!

Issue #290, opened 2019-11-07
[PoC] Account for balances in objective value calculation in Rust


nlordell:
We currently account for remaining balance calculation in the contract but not in the service. This can lead to different disregarded utility values in both places which would be nice to avoid.

nlordell:
We are now using the smart contract to validate and compute objective value so this is no longer needed.

PR #291, opened 2019-11-07
Add order utility and disregarded utility calculation

nlordell:
In order to ultimately calculate a solution's objective value we need to calculate the utility and disregarded utility per order. This PR implements those calculations.

This PR is the first part of the PR #281 as it was too big and got difficult to review.

### Testplan

Unit tests were added with and will be run on CI to verify calculations

### DIFF
--- a/dfusion_rust_core/src/models/order.rs
+++ b/dfusion_rust_core/src/models/order.rs
@@ -67,12 +67,6 @@ impl Order {
         let buy_amount = u128_to_u256(self.buy_amount);
         let sell_amount = u128_to_u256(self.sell_amount);
 
-        // essential_utility = (exec_buy_amount - (exec_sell_amount * buy_amount)
-        //   / sell_amount) * buy_price
-        // essential_utility = (((exec_sell_amount * buy_amount) % sell_amount)
-        //   * buy_price) / sell_amount
-        // utility = essential_utility - essential_utility
-
         let essential_utility = exec_buy_amount
             .checked_sub((exec_sell_amount * buy_amount).checked_div(sell_amount)?)?
             * buy_price;
@@ -101,10 +95,6 @@ impl Order {
         let exec_sell_amount = u128_to_u256(exec_sell_amount);
         let buy_amount = u128_to_u256(self.buy_amount);
         let sell_amount = u128_to_u256(self.sell_amount);
-
-        // limit_term = sell_price * sell_amount - buy_amount * buy_price
-        // leftover_sell_amount = sell_amount - exec_sell_amount
-        // disregarded_utility = (limit_term * leftover_sell_amount) / sell_amount
 
         let limit_term = (sell_price * sell_amount).checked_sub(buy_amount * buy_price)?;
         let leftover_sell_amount = sell_amount.checked_sub(exec_sell_amount)?;
@@ -383,7 +373,7 @@ pub mod tests {
             order.utility(buy_price, exec_buy_amount, exec_sell_amount),
             // u = ((xb * os - xs * ob) * bp) / os
             //   = ((10 * 100 - 90 * 10) * 9) / 100
-            //   = 100
+            //   = 9
             Some(9.into())
         );
     }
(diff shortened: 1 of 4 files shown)


dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,73 @@ impl Order {
...
+    }
+
+    /// Calculates the disregarded utility of an order based on a buy and sell
+    /// price as well as the executed sell amount.
+    ///
+    /// # Returns
+    ///
+    /// Returns disregarded utility of an order or None when there was an error
+    /// in the calculation such as overflow or divide by 0.
+    pub fn disregarded_utility(
```
josojo: I think, before we merge it, we should integrate here the accounts balance, similar as it is done in the smart contract

nlordell: I created a separate issue for this here #290, I referenced it in the solution objective value PR and planned on implementing it as a separate PR.

The other thing that is also missing here is update the disregarded utility calculation to include the fees, which I am waiting to see how it is done in the contracts so we do the same calculations, and get the same results.

I would rather have these in separate PRs as it would make the review a bit easier.

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,73 @@ impl Order {
...
+    ) -> Option<U256> {
+        let buy_price = u128_to_u256(buy_price);
+        let exec_buy_amount = u128_to_u256(exec_buy_amount);
+        let exec_sell_amount = u128_to_u256(exec_sell_amount);
+        let buy_amount = u128_to_u256(self.buy_amount);
+        let sell_amount = u128_to_u256(self.sell_amount);
+
+        // essential_utility = (exec_buy_amount - (exec_sell_amount * buy_amount)
+        //   / sell_amount) * buy_price
+        // essential_utility = (((exec_sell_amount * buy_amount) % sell_amount)
```
fleupold: These comments don't match up 

`utility = essential_utility - essential_utility`

==> utility = 0

Is there any rusty way to make sure those comments are up to date? E.g. Using a documentation test or such?

dfusion_rust_core/src/models/order.rs
```diff
@@ -301,6 +368,143 @@ pub mod tests {
...
+        };
+        let buy_price = 9;
+        let exec_buy_amount = 10;
+        let exec_sell_amount = 90;
+
+        assert_eq!(
+            order.utility(buy_price, exec_buy_amount, exec_sell_amount),
+            // u = ((xb * os - xs * ob) * bp) / os
+            //   = ((10 * 100 - 90 * 10) * 9) / 100
+            //   = 100
```
fleupold: but 100 != 9 😕

dfusion_rust_core/src/models/order.rs
```diff
@@ -47,6 +47,73 @@ impl Order {
...
+        sell_price: u128,
+        exec_sell_amount: u128,
+    ) -> Option<U256> {
+        let buy_price = u128_to_u256(buy_price);
+        let sell_price = u128_to_u256(sell_price);
+        let exec_sell_amount = u128_to_u256(exec_sell_amount);
+        let buy_amount = u128_to_u256(self.buy_amount);
+        let sell_amount = u128_to_u256(self.sell_amount);
+
+        // limit_term = sell_price * sell_amount - buy_amount * buy_price
```
fleupold: Do you think this comment is necessary?

nlordell: I had used them as a reference when I was writing the calculations. I guess these comments can be removed since they add no value: they are basically an almost verbatim copy of the code below them. They just confuse people when they are wrong like :point_up:.

dfusion_rust_core/src/models/order.rs
nlordell: `((10 * 100 - 90 * 10) * 9) / 100 != 100` either :/. Easy fix :)

nlordell:
@fleupold I addressed some of the issues you pointed out. Using doc tests sounds like a cool idea but I'm not sure exactly how we can leverage them here.

nlordell:
@bh2smith You Ok with me merging this? I plan on fixing the disregarded utility to account for fees and user balances in a separate PR.

bh2smith:
Yea boi! Merge away!

PR #292, opened 2019-11-07
Set contracts to 4 minute batch PR and check for breaking changes.

### LINKED ISSUE #278: [PoC] Solution submission only up to 4 minutes batch closing
Change timing logic in rust to know how long it can submit auction results to the contract. This issue is directly related to and should be done in parallel with a smart contract [issue](https://github.com/gnosis/dex-contracts/issues/241)

bh2smith:
This may not actually need to be merged (and shouldn't until the related branch has been merged to master).

This relates to [dex-contracts:pull275](https://github.com/gnosis/dex-contracts/pull/275) and will be used to make any repairs that the contract changes may impose

Closes #278

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
--- a/test/e2e-tests-stablex.sh
+++ b/test/e2e-tests-stablex.sh
@@ -23,5 +23,4 @@ step "Request withdraw" \
 step "Advance time to finalize withdraw" \
 "npx truffle exec scripts/wait_seconds.js 300"
 
-step "Claim withdraw" \
-"npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1"
+npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1
--- a/test/e2e-tests-stablex.sh
+++ b/test/e2e-tests-stablex.sh
@@ -23,4 +23,5 @@ step "Request withdraw" \
 step "Advance time to finalize withdraw" \
 "npx truffle exec scripts/wait_seconds.js 300"
 
-npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1
+step "Claim withdraw" \
+"npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1 | grep \"Success! Balance of token 1 before claim: 3000000000000000000000, after claim: 3999000000000000000000\""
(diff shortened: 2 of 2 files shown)


bh2smith:
My assumption is that the e2e tests won't fail because our test examples are solvable (to optimality) in much less than 4 minutes. However, we might find some problems with the five minute simulations running on real data.

bh2smith:
@josojo  - not sure if you have found the problem here yet, but I am running into the same as yours!

```
❌ Claim withdraw 
You can improve web3's peformance when running Node.js versions older than 10.5.0 by installing the (deprecated) scrypt package in your project
```

fleupold:
We currently limit the solver's running time to three minutes. This is controlled in the [solver config directly](https://gitlab.gnosisdev.com/dfusion/batchauctions/blob/master/scripts/e2e/C_solve_instance.py#L57) (we might want to make it controllable from the rust code eventually)

bh2smith:
> We currently limit the solver's running time to three minutes.

This is great to know, but I still can't tell what is wrong with this! It shouldn't be failing.

fleupold:
My comment was more targeted at https://github.com/gnosis/dex-services/pull/292#issuecomment-550984797

>  It shouldn't be failing.

Can you run the last command of the test manually and see what `npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1` returns?

bh2smith:
> Can you run the last command of the test manually 

When I tried this locally, everything worked fine.

fleupold:
In this case, I usually trigger a build in Travis to see if master is still green (here is a one from a minute ago https://travis-ci.org/gnosis/dex-services/builds/608807502?utm_medium=notification&utm_source=github_status)

fleupold:
Master is green. You might be able to adjust the travis script (to print the last command without grep) to get more insight into the error.

bh2smith:
> You might be able to adjust the travis script (to print the last command without grep) to get more insight into the error.

I don't see how this helps, it just didn't show me anything and passes cuz I wasn't checking anything...


Screenshot 2019-11-07 at 6 15 51 PM

fleupold:
Well you still need to run the command, but not as part of a `step` but just one off

turn 
```
step "Claim withdraw" \
"npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1 | grep \"Success! Balance of token 1 before claim: 3000000000000000000000, after claim: 3999000000000000000000\""
```

to

```
npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1
```

This should print the result and you might be able to see why the string "Success! Balance of token 1 before claim: 3000000000000000000000, after claim: 3999000000000000000000" is not in there.

bh2smith:
Looks like this should be passing... WTF?


Screenshot 2019-11-08 at 10 51 45 AM

bh2smith:
Thanks @josojo! I will get on this ASAP

bh2smith:
> in the stablex_driver.rs to check that the solution is only submitted in the first 4 minutes, but not in the last minute of the batch

So then @josojo, my assumption here would be that we update this little snippet 

https://github.com/gnosis/dex-services/blob/30d1db7a349857b18cc5f90466b0ecfcfbf3c9fe/driver/src/driver/stablex_driver.rs#L44-L51

to include a check on whether there is enough time left in the submission period to send the solution.

Also, we are having a real problem with the non=deterministic nature of this `claimWithdraw` e2e test. Do you have any idea whats up?

nlordell:
In #296 would the call to get the solution objective value would fail and not submit the solution if it is out of time? If so we don't need to add this check anymore.

Issue #285, opened 2019-11-04
New objective function for naive solver


nlordell:
As the objective function has changed, the solver's calculation of the objective function needs to be updated and perhaps the test cases need to be updated as well.

PR #296, opened 2019-11-11
Get solution objective value from contract before submitting

nlordell:
This PR implements retrieving objective value from the stableX contract before submitting the solution. This allows the driver to verify the solution before committing any gas to submit a solution with its objective value to the contract. The contract can then skip verification if the reported objective value is lower than the latest winning solution and save gas in case of multiple solutions racing to be submitted.

This PR addresses issue #279. Note that other issues arose from this implementation and are being tracked here: #285

This PR merge is pending on gnosis/dex-contracts#288

### Testplan

e2e tests


nlordell:
We still need e2e tests for this as well as updating the `dex-contracts` submodule to the new `submitSolution` function that returns the objective value

bh2smith:
I would like to point out that this PR depends on two contract changes:

1. Claimed Objective Value - https://github.com/gnosis/dex-contracts/pull/279
which introduced a new parameter into submit solution AND

2. Returned Objective Value - https://github.com/gnosis/dex-contracts/pull/288
which introduces a return value on `submitSolution`

bh2smith:
```
❌ Claim withdraw 
You can improve web3's peformance when running Node.js versions older than 10.5.0 by installing the (deprecated) scrypt package in your project
You can improve web3's peformance when running Node.js versions older than 10.5.0 by installing the (deprecated) scrypt package in your project
The command "./test/e2e-tests-stablex.sh" exited with 1.
```

@josojo - I recall you have dealt with this in the past.

nlordell:
The contract changes have been merged, so we should be ready to go :rocket:

driver/src/contracts/stablex_contract.rs
```diff
@@ -68,11 +75,43 @@ impl StableXContract for StableXContractImpl {
...
+    fn get_solution_objective_value(
+        &self,
+        batch_index: U256,
+        orders: Vec<Order>,
+        solution: Solution,
+    ) -> Result<U256> {
+        let (prices, token_ids_for_price) = encode_prices_for_contract(solution.prices);
+        let (owners, order_ids, volumes) =
+            encode_execution_for_contract(orders, solution.executed_buy_amounts);
+
```
josojo: Do 2 lines of code sharing already justify factoring it into an own function? Probably not, :)

driver/src/driver/stablex_driver.rs
```diff
@@ -41,8 +41,18 @@ impl<'a> StableXDriver<'a> {
             solution
         };
 
-        let submitted = if solution.executed_sell_amounts.iter().sum::<u128>() > 0 {
-            self.contract.submit_solution(batch, orders, solution)?;
+        let submitted = if solution.is_non_trivial() {
+            let objective_value = self.contract.get_solution_objective_value(
```
josojo: Maybe it is worth mentioning in a comment that this not just checks the conformity of the solution, but also that the following conditions are met:
// solution is better than the existing solution
// batchId still valid

driver/src/contracts/stablex_contract.rs
nlordell: I think not in this case. I can create a new function if you feel like it will make the code cleaner though.

nlordell: I think not in this case. I can create a new function if you feel like it will make the code cleaner though.

driver/src/driver/stablex_driver.rs
nlordell: Let me know if the comment I added makes sense

nlordell:
@josojo 
> Since you merged the networks.json, could you do me a favor and update the common-rinkeby.env file as well?

Is `STABLEX_CONTRACT_ADDRESS` the address of the `StablecoinConverter` contract?

bh2smith:
> Is STABLEX_CONTRACT_ADDRESS the address of the StablecoinConverter contract

Yes, but it is the local address you get when you run `truffle migrate` on a deterministic version of ganache (`ganache-cli -d`)

nlordell:
Closed with #300

PR #298, opened 2019-11-15
New env for auto deploying docker image

josojo:
This new env is required as we want to set a different tag for the docker image than the one specified in the $TRAVIS_BRANCH

--
testplan: praying


fleupold:
Why do we want this? Is there an advantage of having the image called develop instead of master? Now, when we start tagging "stable" release candidates we won't be able to deploy them under a different image name.

josojo:
@davidalbela can probably elaborate,

He told me that there is this convention naming for tags in staging clusters and the k8s-auto-deployment is prepared for that.

PR #297, opened 2019-11-15
Implement a new naive solver for StableX

nlordell:
The naive solver has not been updated for the changes to objective value calculation and fee token price restrictions in stableX. This PR splits the old naive solver into one for snappX and one for stableX for now to move forward with the MVP.

### Test plan

e2e and unit test

PR #300, opened 2019-11-15
Submit Solution With Computed Objective Value 2

nlordell:
This PR implements retrieving objective value from the stableX contract before submitting the solution. This allows the driver to verify the solution before committing any gas to submit a solution with its objective value to the contract. The contract can then skip verification if the reported objective value is lower than the latest winning solution and save gas in case of multiple solutions racing to be submitted.

Because of recent changes to the smart contract, changes to the naive solver were required. As an interim solution the naive solver for Snapp and StableX were split into two implementation. They intend to be merged at a later date.

This PR addresses issue #279.

This PR was originally two separate PRs #296 and #297  

### Testplan

e2e tests


josojo:
Green satisfaction.
But let's merge only, it is based on a master branch in dex-contracts

driver/src/driver/stablex_driver.rs
```diff
@@ -41,8 +41,22 @@ impl<'a> StableXDriver<'a> {
...
+        let submitted = if solution.is_non_trivial() {
+            // NOTE: in retrieving the objective value from the contract the
+            //   solution gets validated, ensured that it is better than the
+            //   latest submitted solution, and that solutions are still being
+            //   accepted for this batch ID.
+            let objective_value = self.contract.get_solution_objective_value(
+                batch,
+                orders.clone(),
+                solution.clone(),
+            )?;
```
fleupold: The print statement below is no longer accurate. You might not submit a non_trivial solution ("Not submitting trivial solution for batch {}")

driver/src/price_finding/stablex_naive_solver.rs
```diff
@@ -0,0 +1,242 @@
...
+/// to be as optimal.
+///
+/// The solver works by finding the first pair of orders which matching sell and
+/// buy tokens, where one of the traded tokens is the fee token, whose users have
+/// enough balance to fully fill, and that has a viable solution.
+///
+/// The solution is computed by taking the price of the larger order and and
+/// completely filling the smaller order. A large order is defined by having
+/// more non-fee token volume. Account balances are checked for each solution.
+pub struct StableXNaiveSolver;
```
fleupold: Why has this been merged, if https://github.com/gnosis/dex-services/pull/297#pullrequestreview-318346510 is still open?

driver/src/driver/stablex_driver.rs
nlordell: The `println!("Not submitting trivial solution...")` should only gets called if solution is trivial.

nlordell: The `println!("Not submitting trivial solution...")` still only gets called if solution is trivial.

driver/src/price_finding/stablex_naive_solver.rs
nlordell: We ended getting into an issue that we needed the new naive solver to merge the objective solution PR to make the e2e tests pass, but the naive solver needed updates to `dex-contracts` to include `close-auction` in order to reliably pass e2e tests (I believe master was failing, it was caused by the timing changes to the contracts that got merged in). And the `dex-contract` updates were done on latest so they needed the services to be updated to the new contract interface (which is what the objective value PR was).

Essentially we had a circular dependency of PRs. What we could have done would be to apply the `close_acution` script changes to the commit that master was based on, merge the naive solver, then merge the objective value PR. However, we both the naive solver PRs had been reviewed independently in #297 and #296 so we decided to just join them so that we could get things building again.

PR #302, opened 2019-11-18
[Easy] taking the `utilitlyPerOrder` out, as Tom's interface no longer support it

josojo:
[Easy] taking the `utilitlyPerOrder` out, as Tom's interface no longer support it

Without this fix, the integration of Tom's solver is no longer working. 

---
testplan:
```
docker-compose build --build-arg use_solver=1 stablex
docker-compose up stablex
cd dex-contracts
npx truffle migrate
```
and run the e2e test. It is expected to fail, though you want to check in the driver console that prices were submitted succesfully

### DIFF
--- a/driver/src/price_finding/linear_optimization_price_finder.rs
+++ b/driver/src/price_finding/linear_optimization_price_finder.rs
@@ -10,7 +10,7 @@ use std::collections::HashMap;
 use std::fs::File;
 use std::io::BufReader;
 use std::process::Command;
-use web3::types::{H160, U256};
+use web3::types::H160;
 
 const RESULT_FOLDER: &str = "./results/tmp/";
 type Prices = HashMap<String, String>;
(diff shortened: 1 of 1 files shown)


driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -291,7 +268,7 @@ pub mod tests {
         });
 
         let expected_solution = models::Solution {
-            objective_value: U256::from_dec_str("15854632034944469292777429010439194350").ok(),
+            objective_value: None,
```
fleupold: Would it make sense to take the objective value out of the solution struct completely, or do we use it somewhere?

josojo:
When the tesplan is run, unfortunately, the solver v3 only finds the trivial solution. Maybe @twalth3r  can you take a look. The instance is this one:
```
npx truffle exec scripts/stablex/get_auction_elements.js 
Using network 'development'.

[ { user: '0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1',
    sellTokenBalance: 3e+21,
    buyToken: 1,
    sellToken: 0,
    validFrom: 5247162,
    validUntil: 5247182,
    priceNumerator: 999000000000000000000,
    priceDenominator: 2e+21,
    remainingAmount: 2e+21 },
  { user: '0xffcf8fdee72ac11b5c542428b35eef5769c409f0',
    sellTokenBalance: 3e+21,
    buyToken: 0,
    sellToken: 1,
    validFrom: 5247162,
    validUntil: 5247182,
    priceNumerator: 1.996e+21,
    priceDenominator: 999000000000000000000,
    remainingAmount: 999000000000000000000 } ]
```

driver/src/price_finding/linear_optimization_price_finder.rs
josojo: It is also used in the whole snapp infrastructure. I don't want to touch this in this PR.

PR #303, opened 2019-11-20
Add Snapp Objective Value Calculation

nlordell:
Since the solver no longer calculates executed utility. Unfortunately this is required for computing the solution objective value which needs to be submitted with a solution for Snapp.

This PR is a first step to addressing this by calculating the Snapp objective value (the total executed utility of the orders) for submitting solutions.

This might help with #302 and with removing the `objective_value` field from the solution. This will also be nice for merging the two naive solvers together so it no longer needs to calculate this.

### Test Plan

Unit tests.

### DIFF
--- a/driver/src/snapp.rs
+++ b/driver/src/snapp.rs
@@ -7,6 +7,62 @@ use dfusion_core::models::{Order, Solution};
 use thiserror::Error;
 use web3::types::U256;
 
+/// Snapp specific extension trait for `Solution`. This extension trait provides
+/// an implementation to calculating the Snapp objective value which must be
+/// submitted with the solution.
+pub trait SnappSolution {
+    /// Returns the price for a token by ID or None if the token was not found.
+    fn get_token_price(&self, token_id: u16) -> Option<u128>;
+
+    /// Returns the objective value for submitting a solution to the Snapp smart
+    /// contract. The objective value is calculated as the total executed
+    /// utility of all orders.
+    ///
+    /// Note that this does not evaluate the validity of the solution and just
+    /// calculates the objective value.
+    ///
+    /// # Returns
+    ///
+    /// `Ok(U256)` when the solution's objective value is correctly calculated
+    /// and `Err` otherwise. Errors can occur when:
+    /// - The order book does not match the solution (number of orders)
+    /// - The order's token ids are not found
+    /// - An invalid order
+    /// - An order where the limit price was not respected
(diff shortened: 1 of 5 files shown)


driver/src/snapp.rs
```diff
@@ -0,0 +1,409 @@
...
+        // the utility is caculated in two parts, this is done to avoid overflows
+        let utility_with_error = exec_buy_amt
+            .checked_sub(
+                (exec_sell_amt * buy_amt)
+                    .checked_div(sell_amt)
+                    .ok_or(SnappObjectiveError::InvalidOrder)?,
+            )
+            .ok_or(SnappObjectiveError::LimitPriceNotRespected)?
+            * buy_price;
+        let utility_error = (((exec_sell_amt * buy_amt) % sell_amt) * buy_price) / sell_amt;
```
fleupold: Any reason we are not doing checked math here?

driver/src/snapp.rs
```diff
@@ -0,0 +1,409 @@
...
+        Ok(utility)
+    }
+}
+
+/// Represents and error that can occur during the calculation of the Snapp
+/// objective value.
+#[derive(Clone, Copy, Debug, Eq, Error, PartialEq)]
+pub enum SnappObjectiveError {
+    /// A token referenced by an order was not found in the solution.
+    #[error("token not found")]
```
fleupold: Nice, we should probably use this for the other error cases as well. Does it allow to chain errors, so that it will be easy to debug where a certain error came from (assuming multiple callsites could throw the same error type)?

driver/src/snapp.rs
nlordell: It can't overflow. `exec_sell_amt * buy_amt` is at most `u128::MAX * u128::MAX` which always fits into a `u256`. `x % sell_amt` is at most `u128::MAX - 1` when `sell_amt` is `u128::MAX` and so `(x % sell_amt) * buy_price` is at most `(u128::MAX - 1) * u128::MAX` which also always fits into a `u256`. Dividing by `sell_amt` has already been checked as well so it is guaranteed to not be `0`.

driver/src/snapp.rs
nlordell: It does. Although it does not do automatic boxing so your error types can get pretty big if you are not careful.

nlordell: I can look into it as a separate PR.

PR #304, opened 2019-11-20
Universal Naive Solver

### LINKED ISSUE #301: [Cleanup] Merge Naive Solver Implementations
Because of the changing StableX requirements, the naive solver was split into two. It will be nice to merge them back into one.

nlordell:
The solver was recently split in two because it was failing e2e tests for stablex. This PR merges the solver back together.

Closes #301. Depends on #303.

### Test Plan

Unit & e2e tests


nlordell:
Looks like the e2e test is failing. The solution the naive solver is generating is invalid unfortunately. I imagine it has something to do with rounding errors.

nlordell:
Ready for review!

driver/src/price_finding/naive_solver.rs
```diff
@@ -110,11 +105,8 @@ impl PriceFinding for SnappNaiveSolver {
         let mut prices: Vec<u128> = vec![0; TOKENS as usize];
         let mut exec_buy_amount: Vec<u128> = vec![0; orders.len()];
         let mut exec_sell_amount: Vec<u128> = vec![0; orders.len()];
-        let mut total_objective_value = U256::zero();
-
-        let mut found_flag = false;
 
-        for (i, x) in orders.iter().enumerate() {
+        'outer: for (i, x) in orders.iter().enumerate() {
```
bh2smith: `'outer: ` <---- What the heck is this? Some sort of space language?

driver/src/price_finding/naive_solver.rs
```diff
@@ -53,20 +49,26 @@ impl Matchable for Order {
         state: &AccountState,
         fee: &Option<Fee>,
     ) -> Option<OrderPairType> {
-        if self.sufficient_seller_funds(&state)
-            && other.sufficient_seller_funds(&state)
-            && self.attracts(other, fee)
+        if !self.sufficient_seller_funds(&state)
+            || !other.sufficient_seller_funds(&state)
+            || !self.attracts(other, fee)
+            || !fee
```
bh2smith: Why did these all become `||` from `&&`?

driver/src/price_finding/naive_solver.rs
```diff
@@ -151,45 +143,44 @@ impl PriceFinding for SnappNaiveSolver {
...
-                    prices[y.buy_token as usize],
-                    exec_buy_amount[j],
-                    exec_sell_amount[j],
-                );
-                total_objective_value = x_objective_value.checked_add(y_objective_value).unwrap();
-                break;
-            }
-            if found_flag {
-                break;
+                break 'outer;
```
bh2smith: spacey!

driver/src/price_finding/naive_solver.rs
```diff
@@ -218,33 +209,95 @@ fn order_with_buffer_for_fee(order: &Order, fee: &Option<Fee>) -> Order {
...
+        (((u128_to_u256(exec_sell_amt) * u128_to_u256(sell_price))
+            / u128_to_u256(fee_denominator))
+            * u128_to_u256(fee_denominator - 1))
+            / u128_to_u256(buy_price),
+    );
+
+    // we need to account for rounding errors here, since this function is
+    // essentially an inverse of `executed_sell_amount`; so find the maximum
+    // error that `v` can have and find a value that works
+    // TODO(nlordell): verify the maths
```
bh2smith:
```suggestion
    // TODO(nlordell): verify the math
```

driver/src/price_finding/naive_solver.rs
nlordell: `x && y` <=> `!x || !y`. I just wanted to remove the nested if statements as they were not needed.

driver/src/price_finding/naive_solver.rs
nlordell: It a loop label so you can break from nested loops: https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html

driver/src/price_finding/naive_solver.rs
nlordell: It was meant to be sarcastic. I should probably remove that comment. I was hoping one of you guys could take a look at the logic there and make sure it makes sense.

bh2smith: Will do

nlordell:
@bh2smith Sure! it works almost exactly as it did before. There are only differences when `fee` is `Some(...)`:
- must trade the fee token (this is required by stableX)
- all prices get adjusted for the price scaling hack
- the executed amounts get computed in a way that account for rounding errors so the solution is accepted by the contract.

Otherwise, there was no change in how it works. Some of the other code differences come from the fact that the solver no longer needs to calculate the objective value.

driver/src/price_finding/naive_solver.rs
nlordell: Removed the TODO :smile:

driver/src/price_finding/naive_solver.rs
```diff
@@ -180,32 +208,96 @@ fn order_with_buffer_for_fee(order: &Order, fee: &Option<Fee>) -> Order {
...
+            {
+                return Some($proposed_exec_buy_amt);
+            }
+        };
+    }
+
+    return_if_correct!(exec_buy_amt);
+
+    let maximum_error = (sell_price / buy_price) + 1;
+    for error in 1..=maximum_error {
```
fleupold: What's the bound of this loop? Do we have to be afraid to run it a billion times?

driver/src/price_finding/naive_solver.rs
```diff
@@ -180,32 +208,96 @@ fn order_with_buffer_for_fee(order: &Order, fee: &Option<Fee>) -> Order {
...
+        return_if_correct!(exec_buy_amt + error);
```
fleupold: Is there a better name for this? `return_if_sell_amount_match`?

driver/src/price_finding/naive_solver.rs
```diff
@@ -1,12 +1,13 @@
...
-
 use dfusion_core::models::{AccountState, Order, Solution, TOKENS};
 
 use crate::price_finding::error::PriceFindingError;
-use crate::util::{u128_to_u256, CeiledDiv};
+use crate::util::{checked_u256_to_u128, u128_to_u256, u256_to_u128, CeiledDiv};
 
 use super::price_finder_interface::{Fee, PriceFinding};
 
+const BASE_UNIT: u128 = 1_000_000_000_000_000_000u128;
```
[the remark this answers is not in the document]
nlordell: I'll do it as a separate PR and look into cleaning up `10**18` in other places as well.

driver/src/price_finding/naive_solver.rs
bh2smith: My understanding of the bound here was... `999_000` This can also be found (hardcoded) as a constant in the `dex-contracts/test/resources/math.js` as

```
const ERROR_EPSILON = new BN(999000)
```

nlordell: It potentially could for small buy prices and high sell prices. I guess we could put an upper bound on this. Also, let me try to formalize a bit what this represents so it feels a bit less magic. Essentially it is supposed to be the maximum error that the executed buy amount could have so that `x == executed_sell_amount(executed_buy_amount(x, buy_price, sell_price), buy_price, sell_price)`.

nlordell: We need this since the naive solver relies on fixing the executed buy amounts for orders that are selling the fee token and fixing the executed sell amounts for orders that are buying the fee token (since the rounding errors must be on the fee token).

nlordell: Hmm, now that I think about it, I might have made a mistake here.

nlordell: I think the `maximum_error == (fee_denominator / buy_price) + 1` which, because of price restrictions, is just 1.

It might also be `((fee_denominator / buy_price) + 1) + (((fee_denominator - 1) / sell_price) + 1) == 2` to account for errors in the `executed_sell_amount` calculation as well. Also I would expect the error to always be in the same direction so we might not have to check in both directions.

 I'm coming into the office in an hour or so, can someone help me look at this?

nlordell: @bh2smith That error epsilon is for something else I think. If I am not mistaken it is the maximum error for buy and sell order amounts so that they match an exact price. Something like:
```
order      price     order
    [              |              ]
-----                          -----
ERROR_EPSILON
```

nlordell: Turns out we needed to do a ceiled div. There are still values of executed sell amount that have no inverse in integer math. A future improvement would be to have the naive solver that deals with this by adjusting amounts to work.

Note that if the price of the non-fee token is not smaller than the base price, then the inverse exists and can be computed.

driver/src/price_finding/naive_solver.rs
```diff
@@ -244,32 +244,19 @@ fn executed_buy_amount(
...
-    return_if_sell_amount_matches!(exec_buy_amt);
-
-    let maximum_error = (sell_price / buy_price) + 1;
-    for error in 1..=maximum_error {
-        return_if_sell_amount_matches!(exec_buy_amt + error);
-        return_if_sell_amount_matches!(exec_buy_amt - error);
+    // essentially an inverse of `executed_sell_amount`; when the buy price is
+    // higher than the sell price, there are executed sell amounts that cannot
+    // be satisfied, check the executed buy amount correctly "round trips" to
+    // the specified exectued sell amount and return `None` if it doesn't
```
bh2smith: It appears the Macro has disappeared!

nlordell: It is no longer needed. It turns out that we can use a `ceil_div` and my error calculation was wrong.

nlordell:
@fleupold @bh2smith You guys OK with me merging this?

bh2smith:
Sure am! Approved yesterday.

nlordell:
OK, just wanted to make sure because of the changes to how the executed buy amount was calculated.
Also opened #309

PR #306, opened 2019-11-20
[WIP] Unified logger

nlordell:
We currently have multiple loggers for the graph and the rest of stablex and dfusion. This PR attempts to unify logging in both cases.

### Test Plan

It still logs


nlordell:
The graph version that we are using is on old slog version that is are not compatible with the log crate we are using. This will end up being a bit more involved.

nlordell:
This is superseded by #376

PR #307, opened 2019-11-20
[Easy] Clean Linear Solver JSON Deserialization Test Cases

nlordell:
#302 removed execUtility from the linear solver JSON but left it in some test cases, this just cleans it up.

### Test Plan

Unit tests still pass

### DIFF
--- a/driver/src/price_finding/linear_optimization_price_finder.rs
+++ b/driver/src/price_finding/linear_optimization_price_finder.rs
@@ -255,13 +255,11 @@ pub mod tests {
             "orders": [
                 {
                     "execSellAmount": "0",
-                    "execBuyAmount": "0",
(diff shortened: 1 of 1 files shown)


driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -370,8 +365,7 @@ pub mod tests {
             "orders": [
                 {
                     "execSellAmount": "0",
-                    "execBuyAmount": "0a",
```
bh2smith: What is the `a` for here?

fleupold: It's to make the buy volume not parseable as this test is checking that `serialize_result_fails_if_order_buy_volume_not_parseable `

PR #308, opened 2019-11-20
[Easy] Stricter Linting

nlordell:
This adds stricter linting by running `cargo fmt` on all workspace crates as well as running `cargo clippy` on all targets (including the `test` target, which we currently ignore) to check more code.

### Test plan

CI still passes

### DIFF
--- a/dfusion_rust_core/src/models/order.rs
+++ b/dfusion_rust_core/src/models/order.rs
@@ -91,11 +91,9 @@ impl From<Entity> for Order {
         let batch_information = entity
             .get("auctionId")
             .and(entity.get("slotIndex"))
-            .and_then(|_| {
-                Some(BatchInformation {
-                    slot: U256::from_entity(&entity, "auctionId"),
-                    slot_index: u16::from_entity(&entity, "slotIndex"),
-                })
+            .map(|_| BatchInformation {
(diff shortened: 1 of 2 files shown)


dfusion_rust_core/src/models/order.rs
```diff
@@ -91,11 +91,9 @@ impl From<Entity> for Order {
...
             .get("auctionId")
             .and(entity.get("slotIndex"))
-            .and_then(|_| {
-                Some(BatchInformation {
-                    slot: U256::from_entity(&entity, "auctionId"),
-                    slot_index: u16::from_entity(&entity, "slotIndex"),
-                })
+            .map(|_| BatchInformation {
+                slot: U256::from_entity(&entity, "auctionId"),
+                slot_index: u16::from_entity(&entity, "slotIndex"),
```
bh2smith: Can you tell me what was failing the linter before this change?

nlordell: `and_then(|_| Some(x))` <=> `map(|_| x)`. The latter is simpler and suggested by clippy.

nlordell: https://rust-lang.github.io/rust-clippy/master/index.html#option_and_then_some

PR #312, opened 2019-11-25
[CI] Add rinkeby e2e test to CI

fleupold:
This is a somewhat bold proposal which should be carefully discussed. My goal is to have an acceptance test that will only make travis green if the PR that is about to be merged will work with the currently deployed rinkeby version of the contracts (using Tom's solver and a simple trade example).

To do this I'm running the stablex-rinkeby image and am placing two limit orders (+deposits) on the deployed rinkeby contract. We then wait for `secondsRemainingInBatch` + 30 seconds so that the batch can close and the solver can propose a solution.

We then check that there are no errors in the logs.

Here are some problems with the current approach
1. There is no way to fast-forward rinkeby so in the worst case we will sleep 5 minutes.
2. The travis solver might be time racing with the solver that is already deployed on AWS making this test not very deterministic (however we'd still test that everything until submitSolution would be working e2e)
3. Just checking "there are no errors" might be too weak, but other checks might not work due to non determinism (e.g. other orders in the same batch might influence the result)
4. Orders are done with real rinkeby addresses, which would have to be funded constantly (1 OWL, DAI per commit). Maybe this could be circumvented by also requesting withdraws (however that would add to the time it takes to know whether the test passes).

If time is the only concern we could also create a job that only runs on master and tags a given commit as "stable" if it passes the long running test (so that our PRs can be merged quicker). We could then set up alerts if "stable" gets stuck and the frontend could depend on "stable".

I'd still much prefer this over a develop branch which we manually merge in fixed intervals as I'd like to have the frontend team work with and thus test our latest versions (otherwise no one would and so the merge from develop to master or release-candidate would be arbitrary and not guarantee any quality). 
But of course we want to make sure that our commits are not obviously broken.


bh2smith:
I fetched the error message from the recent build failure:

```
/home/travis/.travis/functions: line 109: ./test/e2e-tests-stablex-rinkeby.sh: Permission denied

The command "./test/e2e-tests-stablex-rinkeby.sh" exited with 126.
```

fleupold:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FxT5LMzIK1AdZJ4cYW4%2Fgiphy.gif&token=doh" data-gitmeme-token="doh"></a>

josojo:
Pretty cool that you are tackling this non-trivial task. It is important that we get more checks in place!

You mentioned some downsides of this approach. I agree with them. The biggest concern is that these tests could really slow down our development time. Debugging non-deterministic tests is awful. Especially, since tom's solver is not yet so robust.

Also, this might require merging small PRs together(e.g. with new deployments and potentially other updates) to get the e2e test working. That would make code review harder.

Overall, I think it would be best to have this e2e test only for "tagged releases", but not for every merge to master.

fleupold:
> Debugging non-deterministic tests is awful. Especially, since tom's solver is not yet so robust.

However, it might help us discover the edge cases and thus make it more robust.

> Overall, I think it would be best to have this e2e test only for "tagged releases", but not for every merge to master.

The problem with that is that then our development time might be faster but our release time will be slower, as for every release we will run the tests and might realize they failed. You then have to spend time (e.g. a good amount of last week) to try and find the changes that broke things. If you have it running on every commit you will know which commit broke it.

The passing travis run took 14 minutes (https://travis-ci.org/gnosis/dex-services/jobs/617099315?utm_medium=notification&utm_source=github_status), but I think there was something wrong with the sleep statement.

Our dfusion e2e tests take 20 minutes.

If time is our only concern, I'd still like to run it on every commit, but make the test not merge blocking and thus have a "sheriff" that makes sure our master branch is always passing the latest tests.

bh2smith:
> a "sheriff" that makes sure our master branch is always passing the latest tests

This is great!

.travis.yml
```diff
@@ -34,15 +34,29 @@ matrix:
...
+      cache: 
+        directories:
+          - dex-contracts/node_modules/
+      install:
+        - cd dex-contracts && npm ci && npx truffle compile && npm run networks-inject && cd ..
+        - echo -e $GITLAB_PRIVATE_KEY > .ssh/id_rsa && chmod 0500 .ssh/id_rsa
+        - docker-compose build --build-arg use_solver=1 stablex-rinkeby
+      script:
+        - docker-compose up -d stablex-rinkeby
+        - ./test/e2e-tests-stablex-rinkeby.sh
```
anxolin: Shouldn't this use `docker-compose run` so it uses docker? Otherwise it's kind of a mixed approach

docker/deploy.sh
```diff
@@ -6,12 +6,9 @@ sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install aw
...
 # Get login token and execute login
 $(aws ecr get-login --no-include-email --region $AWS_REGION)
-echo -e $GITLAB_PRIVATE_KEY > .ssh/id_rsa
-chmod 0500 .ssh/id_rsa
 
-echo "Building latest image with solver...";
-docker-compose build --build-arg use_solver=1 stablex
-docker tag stablex $REGISTRY_URI:$TRAVIS_BRANCH
+echo "Tagging latest image with solver...";
+docker tag stablex-rinkeby $REGISTRY_URI:$TRAVIS_BRANCH
```
anxolin: Why `stablex-rinkeby`? Would it make sense to make the environments as close as possible so their differents are abstracted away just with env var (like the provider url)

.travis.yml
bh2smith: Interesting comment!

fleupold: It is kind of a mixed approach. The system under test runs in its target environment (a docker container), however the interactions with this environment are coming from the travis machine directly. We could run the script in another (separate) docker container, but I don't really see the value.

docker/deploy.sh
fleupold: If you look inside the docker compose the only difference is the .env file. `docker-compose up` doesn't allow adding an env file which is why we added another target. But we can revisit this if you have a better suggestion.

fleupold:
> I would suggest to have this environment isolated, so the contract is only used by this scripts.

I don't fully understand your suggestion. If we want to make sure the currently deployed contract is compatible with the services, we need to run the e2e tests against that contract. However, if there was no new deployment on the commit it will be the same contract that the current AWS instance is pointing at.

josojo:
>If time is our only concern, I'd still like to run it on every commit, but make the test not merge blocking and thus have a "sheriff" that makes sure our master branch is always passing the latest tests.

That sounds like a good way to test it. If we are getting comfortable with it, I am happy to include it as merge requirement later.

test/e2e-tests-stablex-rinkeby.sh
```diff
@@ -0,0 +1,27 @@
...
+time_remaining_hex=`npx truffle exec scripts/stablex/invokeViewFunction.js --network rinkeby getSecondsRemainingInBatch | grep -Eo '[0-9]+'`
+time_to_sleep=`echo $((16#$time_remaining_hex + 30))`
+
+echo "Will sleep for $time_to_sleep seconds"
+
+step "Wait for batch to be closed and solved (+30 seconds) " \
+"sleep $time_to_sleep"
+
+step "Check there are no errors in the solver logs " \
+"! docker-compose logs | grep -e ERROR"
```
josojo: From time to time, we are getting this broken pipe error. It just happens. If we could allow for this error, it would be great

.travis.yml
anxolin: I don't find it so important, so I wouldn't argue too much for this if it's already working, but a e2e test is built with all the pieces in docker, all pieces but one.

Docker deals for example with auto containing this node version, and any library you might depend on, so you can make sure it runs in travis or in any machine no matter the version they have for their global packages.

anxolin:
> I don't fully understand your suggestion. If we want to make sure the currently deployed contract is compatible with the services, we need to run the e2e tests against that contract. However, if there was no new deployment on the commit it will be the same contract that the current AWS instance is pointing at.

What I mean, is a suggestion for what I understood of your point `2.`. Imo, this integration test should use an isolated version of the contracts, not connected to the web, not used by anyone else, so your test can be easily repeatable and doesn't fail (or pass when they shouldn't) because of two solvers are working on the same contract

docker/deploy.sh
anxolin: docker-compose run has `-e` so you could pass an env var. Would this help to abstract away the differences?

.travis.yml
fleupold: Travis itself is containerized and does a clean install of all dependencies on each run. Is that what you mean by "auto containing"? Otherwise please explain, as I would like to understand this better.

Maybe you can explain a concrete scenario which shows the potential weakness of this approach.

docker/deploy.sh
fleupold: Yes, unfortunately `docker-compose run` doesn't allow a daemon mode though, so it would be blocking (or would have to be called with & at the end). This could work however, since locally we don't usually run the daemon mode.

fleupold:
> What I mean, is a suggestion for what I understood of your point 2.. Imo, this integration test should use an isolated version of the contracts, not connected to the web, not used by anyone else, so your test can be easily repeatable and doesn't fail (or pass when they shouldn't) because of two solvers are working on the same contract

@anxolin We already have a e2e test case that runs on an isolated version (cf. https://travis-ci.org/gnosis/dex-services/jobs/617208146?utm_medium=notification&utm_source=github_status)

This test is really supposed to make sure we only merge and deploy versions of the service that will work with the deployed contract version on rinkeby.
Otherwise people depending on the rinkeby version (our alpha testers, which I'm hoping to include frontend devs in) would be screwed.

> If we are getting comfortable with it, I am happy to include it as merge requirement later.

@josojo How about the other way around? We try it like this and if it becomes flaky/takes too long we make it non-blocking. Having a job run on master and moving a stable branch upstream is not trivial. It's possible but would require more work.

FYI the test is likely not going to take more than 20 minutes in the worst case (which is already the time it takes to run the Snapp e2e tests) - https://travis-ci.org/gnosis/dex-services/builds/617208143?utm_medium=notification&utm_source=github_status

bh2smith:
> We try it like this and if it becomes flaky/takes too long we make it non-blocking. 

I am in favour of this! If I understand correctly, this would mean that we can merge approved branches before this test has finished

josojo:
>Having a job run on master and moving a stable branch upstream is not trivial. It's possible but would require more work.

This is actually a valid point.
 
> FYI the test is likely not going to take more than 20 minutes in the worst case (which is already the time it takes to run the Snapp e2e tests) - https://travis-ci.org/gnosis/dex-services/builds/617208143?utm_medium=notification&utm_source=github_status

You should not ignore the time to migrate the smart contracts to rinkeby and adding tokens.
And when I have done this, the test will also always pass, as there are no orders in the system and the solver will not submit anything at all. So the test is kinda worthless. 

If your e2e test also would put meaningful orders into the newly deployed Rinkeby contracts, it would be more valuable. 
---

Due to the fact that "Having a job run on master and moving a stable branch upstream is not trivial" I am actually in favor with implementing your suggestion to have this cool e2e test on rinkeby as a required check.
I guess in the early days I would have been opposed to this, but now, as I don't expect too many smart contract changes, I expect the test to pass most of the time and it will not slow down our development speed too much.

fleupold:
> If your e2e test also would put meaningful orders into the newly deployed Rinkeby contracts, it would be more valuable.

It does. Not necessarily long standing liquidity provision (that's not really part of a integration test) but it places two orders that can be matched.

> You should not ignore the time to migrate the smart contracts to rinkeby and adding tokens.

This has to happen with or without the test (since if we merge something into master without deploying after a breaking change the AWS instance would start crashing).


I'm happy to see increasing acceptance of this proposal. I agree that it is risky and might turn out creating more trouble than help, but let's try and learn from it?

Do you have any ideas how we can prevent having to fund the account that runs the tests too frequently? My biggest concern is rinkeby OWL @anxolin do we have unlimited supply of that?

anxolin:
> My biggest concern is rinkeby OWL @anxolin do we have unlimited supply of that?

I have 2M. If you run out of it, you need to mint more, it would take us a bit, but It's doable, anyways let's hope you don't need that much OWL

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FJpG2A9P3dPHXaTYrwu%2Fgiphy.gif&token=money" data-gitmeme-token="money"></a>

.travis.yml
```diff
@@ -34,15 +34,29 @@ matrix:
...
+      after_failure:
+        - docker-compose logs
+    - language: generic
+      name: "StableX e2e Tests (Rinkeby)"
+      cache: 
+        directories:
+          - dex-contracts/node_modules/
+      install:
+        - cd dex-contracts && npm ci && npx truffle compile && npm run networks-inject && cd ..
+        - echo -e $GITLAB_PRIVATE_KEY > .ssh/id_rsa && chmod 0500 .ssh/id_rsa
```
nlordell: This doesn't show up in logs anywhere right?

fleupold: No, just the command as far as I can tell: 

https://travis-ci.org/gnosis/dex-services/jobs/617208147#L535

test/e2e-tests-stablex-rinkeby.sh
josojo: Actually, this does not seem to be a challenge with the server on aws, it just used to be an issue on my local machine, so I think, it does not need to be addressed

PR #314, opened 2019-11-29
batchExchange rename

josojo:
This PR makes sure that the dex-service repo is up to date with the contracts:
It 
- looks for the BatchExchange contract instead of the of StablecoinConverter contract 
- it modifies the max_utility value, which was send as a parameter in the query submit_solution, as the new contracts multiply this new value by IMPROVEMENT_NOMIANTOR=1

----
testplan:
e2e tests


driver/src/contracts/stablex_contract.rs
```diff
@@ -23,8 +23,7 @@ pub struct StableXContractImpl {
 
 impl StableXContractImpl {
```
fleupold: Probably not worth changing this name. Looking forward to using ethcontracts-rs for all of this :-)

bh2smith:
Hello e2e-test my old friend

```
❌ Claim withdraw 
```

Last failing test!

josojo:
The No dust order solution improvements (#342) … broke dex-services, as we can no longer supply the U256::max_value() as the objective value, since this value will be multiplied by IMPROVEMENT_DENOMINATOR, and it would overflow.

I implemented a quick fix.
Though this fix uses an unwrap, which I don't want to use. But I could not figure out the error conversion:

I need to implement the conversion of error types. I want to implement:
```
impl From<std::option::NoneError>for DriverError {
    fn from(error: std::option::NoneError) -> Self {
        DriverError::new(error.description(), ErrorKind::NoneError)
    }
}
```
But that gives me the following error:
```
use of unstable library feature 'try_trait'

note: for more information, see https://github.com/rust-lang/rust/issues/42327rustc(E0658)
```

nlordell:
@josojo I think we can use `Option::ok_or_else` here to promote an `Option` to a `Result` and then use the standard `?` operator. https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

Hope that helps!

nlordell:
@josojo However, I would argue that you don't need to use `checked_add` here and can simply just use a `+`. If this checked add would fail then it would overflow on the contract and it would never accept any solutions. This would also remove the need to `unwrap`.

driver/src/contracts/stablex_contract.rs
```diff
@@ -84,15 +82,28 @@ impl StableXContract for StableXContractImpl {
...
+                "IMPROVEMENT_DENOMINATOR",
+                (),
+                None,
+                Options::default(),
+                None,
+            )
+            .wait()?;
+        // In the smart contracts the utility value will be multiplied by 1 + IMPROVEMENT_DENOMINATOR
+        // hence, the following reduction of the maximal possible utility value is required:
+        let max_utility_value = U256::max_value() / (U256::one() + improvement_denominator);
```
fleupold: I'm wondering if we should just calculate this value once and hardcode it as a constant. Feels like it won't change enough to justify a dynamic call and computation on every run. What do you think?

nlordell: Maybe memoize it?

fleupold: Why not just hardcode as a constant?

josojo: I am also in favor of hardcoding it for now. Just because we did it for other public constants for the smart contract as well. E.g. FeeDENOMINATOR.

But I am happy to change all of them to a memorized solution later, once we agreed on it.

driver/src/contracts/stablex_contract.rs
```diff
@@ -8,23 +8,26 @@ use web3::types::{H160, U128, U256};
...
 use super::stablex_auction_element::StableXAuctionElement;
+use crate::error::DriverError;
 
 type Result<T> = std::result::Result<T, DriverError>;
 
 pub const AUCTION_ELEMENT_WIDTH: usize = 112;
+lazy_static! {
+    // In the BatchExchange smart contract, the objective value will be multiplied by
+    // 1 + IMPROVEMENT_DENOMINATOR = 101. Hence, the maximal possible objective value is:
+    static ref MAX_OBJECTIVE_VALUE: U256 = U256::max_value() / (U256::from_dec_str("101").unwrap());
```
nlordell: I would use U256::from(101) here, so there is no need to parse a string or unwrap.

driver/src/lib.rs
```diff
@@ -5,6 +5,8 @@
 //   we successfully link.
 extern crate ethereum_tx_sign;
 extern crate web3;
+#[macro_use]
+extern crate lazy_static;
```
nlordell: This is not necessary, in Rust edition 2018 it is generally preferred to locally import macros with a use statement like this:
```
use lazy_static::lazy_static;
```

The crate documentation for `lazy_static` looks to still be in Rust edition 2015 :/

PR #315, opened 2019-11-30
Exclude fee price

bh2smith:
This simple fix should (at least temporarily) align with the corresponding contract changes. That is, the smart contract does not accept fee token price as part of the solution.

There are several alternatives, but this is a start and its success will determine how to continue.


driver/src/contracts/stablex_contract.rs
```diff
@@ -126,8 +126,8 @@ impl StableXContract for StableXContractImpl {
                     owners,
                     order_ids,
                     volumes,
-                    prices,
-                    token_ids_for_price,
+                    prices[1..],
+                    token_ids_for_price[1..],
                 ),
                 Options::with(|mut opt| {
                     // usual gas estimate is not working
```
bh2smith:
```suggestion
                    // usual gas estimation isn't working
```

nlordell:
Looks good, I suggest we merge the `dex-contract` changes first, then update the commit hash to point to the merged `master` commit.

driver/src/contracts/stablex_contract.rs
```diff
@@ -126,11 +126,11 @@ impl StableXContract for StableXContractImpl {
                     owners,
                     order_ids,
                     volumes,
-                    prices,
-                    token_ids_for_price,
+                    prices[1..],
```
fleupold: Should this logic maybe be part of `encode_prices_for_contract`? I was hoping to keep as much logic as possible outside of the `submit_solution` function as it's supposed to be dumb. In the other place it could be easily covered by a unit test.

bh2smith:
I don’t necessarily need you to commandeer, but given that I’m now officially away for two weeks, it wouldn’t hurt. Please notice that the tests have not run here yet and we have no idea if it will pass!

fleupold:
Nice, the newly added rinkeby e2e test caught that we need a new deployment for this: https://travis-ci.org/gnosis/dex-services/jobs/621113735?utm_medium=notification&utm_source=github_status

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2Fl0Iy8hquHU5npwMCs%2Fgiphy.gif&token=it_works" data-gitmeme-token="it_works"></a>

PR #354, opened 2019-12-12
Multi stage build

fleupold:
Goal of this PR is to avoid building our rust project on each travis step, and instead have a single build stage that creates a minified image only containing the binary.

Each subsequent E2E test would then download that image and not have to do any compilation.

Results:

||Total Time|Run Time|Link|
|---|---|---|---| 
|Before|57min|21min|https://travis-ci.org/gnosis/dex-services/builds/624108072
|After|42min|23min|https://travis-ci.org/gnosis/dex-services/builds/624341622

While the run time (the time we have to wait to see the green checkmark) didn't go down, we are using 15 minutes less total resource for each run (reducing the time we block other projects).

In order to reduce this time even further (I had a run with 16 minutes run time, 35 minutes total time) we need to reuse the locally built binaries and copy them into the docker image (instead of building them inside docker). For that to work all images need to have a xenial base, so it was starting to make this PR very large.

I will take a step back and create a xenial based solver image inside the dex-solver repo and then cycle back to this repo to simplify the Dockerfiles we have for our rust image (we should just need one that is based on the solver).


common-rinkeby.env
```diff
@@ -2,3 +2,5 @@ STABLEX_CONTRACT_ADDRESS=0x4eAfD2b1A2264228B87e35b77e50AA5b9bcC72dc
 ETHEREUM_NODE_URL=https://rinkeby.infura.io/v3/9408f47dedf04716a03ef994182cf150
 NETWORK_NAME=rinkeby
 NETWORK_ID=4
+PRIVATE_KEY=4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d
```
josojo: Is this line intended?

.travis.yml
```diff
@@ -19,13 +40,17 @@ matrix:
         - cargo fmt --all -- --check
     - language: generic
       name: "dƒusion e2e Tests"
-      cache: 
+      cache:
         directories:
-          - $HOME/.npm
```
josojo: Why is this Pr no longer valid?
https://github.com/gnosis/dex-services/pull/349

fleupold: Before, only saved the npm downloaded assets, because we used npm ci (which removes the cached `npm_modules` directory).

Now, we `npm install` and thus can cache the node modules. But it's a good point that in the very first stage (warm up npm cache) we should also be caching the downloaded files.

common-rinkeby.env
fleupold: Ah no, it seems to be not needed, since it is already defined in the common.env.

.travis.yml
nlordell: `$HOME/.npm` gets cached by default for projects with a `project-lock.json`

.travis.yml
```diff
@@ -1,8 +1,30 @@
...
+        - sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
+        - echo -e $GITLAB_PRIVATE_KEY > .ssh/id_rsa && chmod 0500 .ssh/id_rsa
+        - $(aws ecr get-login --no-include-email --region $AWS_REGION)
+      script:
+        # First build base image and run unit tests
+        - docker build --tag stablex --build-arg use_solver=1 -f docker/rust/Dockerfile .
+        # Then build and publish compact multi-stage image
+        - docker build --tag stablex-binary -f docker/rust/stablex_binary/Dockerfile .
+        - docker tag stablex-binary $REGISTRY_URI:$TRAVIS_COMMIT
+        - docker push $REGISTRY_URI:$TRAVIS_COMMIT
```
nlordell: This is a private registry right?

.travis.yml
nlordell: https://docs.travis-ci.com/user/caching/#npm-cache

.travis.yml
fleupold: yes

Issue #330, opened 2019-12-04
Use ethcontracts-rs instead of Smart Contract wrappers


fleupold:
Since @nlordell built this awesome library, we should be the first ones to use it internally.

PR #362, opened 2019-12-17
Using dex-solver image and having a single conditional base image

fleupold:
This PR builds on top of the previous work and dramatically cleans up the amount of build scripts we have in our repo. We now basically have two orthogonal choices
1. Which solver image to use (default a vanilla ubuntu without the dex-solver)
2. Which rust bas image to use (one that compiles everything or one that uses local binaries)

With this we can support both local development as well as cached travis builds and releases that will be deployed on AWS.

### Test Plan
I updated the README and made sure I can still run the system locally with both naive and linear solver. The rest is up to Travis.


fleupold:
https://travis-ci.org/gnosis/dex-services/builds/626676713

vs [current master](https://travis-ci.org/gnosis/dex-services/builds/626232855)

Improvements are not major (45min -> 41min total time, 24min ->21 min run time). However it dramatically simplifies the setup, thus I think it's an improvement.

Next step will be to try out sccache for faster rust builds (the long running e2e test could also be broken up into three small ones)

.travis.yml
```diff
@@ -1,18 +1,24 @@
...
+env:
+  - CARGO_INCREMENTAL=0
 matrix:
   include:
     - stage: WarmUp
       name: "Build"
-      language: generic
       install:
         - sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
-        - echo -e $GITLAB_PRIVATE_KEY > .ssh/id_rsa && chmod 0500 .ssh/id_rsa
```
josojo: This is so much more straight forward with the docker image build in dex-solver. Thumbs up

docker/rust/Dockerfile
```diff
@@ -1,18 +1,38 @@
...
+ONBUILD COPY dfusion_rust_core ./dfusion_rust_core
+ONBUILD COPY driver ./driver
+ONBUILD COPY listener ./listener
 
-COPY Cargo.* ./
-RUN cargo build
+ONBUILD COPY Cargo.* ./
+ONBUILD RUN cargo build
+
+# Build Option 1: Copy already compiled binary from host (used e.g. in travis)
```
josojo: option 2, I guess

docker/rust/Dockerfile
```diff
@@ -1,18 +1,32 @@
-FROM gnosispm/dfusion-rust-base:v4
+ARG SOLVER_BASE=ubuntu:bionic
```
josojo: This default value would be the right value for using the simple solver, right? Maybe a small comment would help. It took me some time to understand.

josojo:
Do we also need to update the docker/deploy.sh script to use these new deployment methods?

fleupold:
> Do we also need to update the docker/deploy.sh script to use these new deployment methods?

The deploy script already uses the binary built version. It's using another Docker image, because we have to copy the artifacts into the image. However #330 makes this mandatory even for the tests so I hope it will die soon.

josojo:
I have the following trouble:
With
```
docker-compose build --build-arg RUST_BASE=rust-binary stablex 
```
I get:
```
docker-compose down && docker-compose up stablex            
Stopping dex-services_stablex_1     ... done
Stopping dex-services_ganache-cli_1 ... done
Removing dex-services_stablex_1     ... done
Removing dex-services_ganache-cli_1 ... done
Removing network dex-services_default
Creating network "dex-services_default" with the default driver
Creating dex-services_ganache-cli_1 ... done
Creating dex-services_stablex_1     ... error

ERROR: for dex-services_stablex_1  Cannot start service stablex: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"cargo\": executable file not found in $PATH": unknown

ERROR: for stablex  Cannot start service stablex: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"cargo\": executable file not found in $PATH": unknown
```

fleupold:
If you use the binary build you have to also use the binary docker compose file (cf. travis file). 

‘’’
dicker-compose -f docker-compose.yml -f docker-compose-binary.yml up stablex 
‘’’

PR #363, opened 2019-12-18
Integrate `ethcontract-rs`

### LINKED ISSUE #330: Use ethcontracts-rs instead of Smart Contract wrappers

nlordell:
This PR integrates `ethcontract-rs` into `dex-services` for generating type safe contract bindings to the dFusion contracts.

Closes #330

### TODO

- [x] StableX
- [x] Snapp
- [x] Release `ethcontract-rs 0.2.0` and use it (with patches)

### Test Plan

CI


driver/build.rs
```diff
@@ -0,0 +1,19 @@
...
+use std::env;
+use std::path::Path;
+
+fn main() {
+    generate_contract("BatchExchange", "batch_exchange.rs");
+    generate_contract("SnappAuction", "snapp_auction.rs");
+}
+
+fn generate_contract(name: &str, out: &str) {
+    let artifact = format!("../dex-contracts/build/contracts/{}.json", name);
```
bh2smith: Does rust operate well with relative paths like this?

driver/src/contracts/snapp_contract.rs
```diff
@@ -109,163 +115,80 @@ impl SnappContract for SnappContractImpl {
...
-            .query(
-                "getCurrentWithdrawIndex",
-                (),
-                None,
-                Options::default(),
-                None,
-            )
-            .wait()
-            .map_err(DriverError::from)
+        Ok(self.instance.get_current_withdraw_index().call().wait()?)
```
bh2smith: This sequence of commands seems to be a common theme. Not that theres anything wrong with that.

driver/src/error.rs
```diff
@@ -100,6 +104,24 @@ impl From<PriceFindingError> for DriverError {
...
+        DriverError::new(&error.to_string(), ErrorKind::ContractDeployedError)
+    }
+}
+
+impl From<ethcontract::errors::ExecutionError> for DriverError {
+    fn from(error: ethcontract::errors::ExecutionError) -> Self {
+        DriverError::new(&error.to_string(), ErrorKind::ContractExecutionError)
+    }
+}
+
```
bh2smith: Nice to see some more refined Errors! Is this an added benefit of using `ethcontracts-rs`?

driver/src/contracts/stablex_contract.rs
```diff
@@ -437,7 +413,7 @@ pub mod tests {
 
         // Only contain non fee-token and non 0 prices
         let expected_prices = vec![1.into(), 2.into()];
-        let expected_token_ids = vec![2.into(), 3.into()];
+        let expected_token_ids = vec![2, 3];
```
bh2smith: Why does one need `into` and not the other?

driver/src/contracts/stablex_contract.rs
```diff
@@ -193,15 +173,15 @@ fn encode_prices_for_contract(price_vector: Vec<u128>) -> (Vec<U128>, Vec<U128>)
 fn encode_execution_for_contract(
     orders: Vec<Order>,
     executed_buy_amounts: Vec<u128>,
-) -> (Vec<H160>, Vec<U128>, Vec<U128>) {
+) -> (Vec<H160>, Vec<u64>, Vec<U128>) {
```
bh2smith: Why no longer using `U128`?

driver/src/contracts/stablex_contract.rs
```diff
@@ -193,15 +173,15 @@ fn encode_prices_for_contract(price_vector: Vec<u128>) -> (Vec<U128>, Vec<U128>)
...
         orders.len(),
         executed_buy_amounts.len(),
         "Received inconsistent auction result data."
     );
-    let mut owners: Vec<H160> = vec![];
-    let mut order_ids: Vec<U128> = vec![];
-    let mut volumes: Vec<U128> = vec![];
+    let mut owners = vec![];
+    let mut order_ids = vec![];
+    let mut volumes = vec![];
```
bh2smith: If the type declaration can be removed, what was it needed for in the first place?

driver/build.rs
nlordell: Build scripts are always run from the root of the crate. So for the driver crate that's `driver/`.

driver/src/contracts/snapp_contract.rs
nlordell: It is! Working as designed :stuck_out_tongue:

driver/src/error.rs
nlordell: Its more that `ethcontracs-rs` adds new errors that need to be accounted for.

driver/src/contracts/stablex_contract.rs
nlordell: It used to be the `U128` non-primitive type and now its the `u64` primitive type.

driver/src/contracts/stablex_contract.rs
nlordell: Because order**Indices** are `u16`s and not `u128`. After gnosis/ethcontract#8 it will be a `u16`.

driver/src/contracts/stablex_contract.rs
nlordell: It wasn't. I removed them since I needed to change it for one of them anyway, so next time the diff is even smaller :smile:

driver/build.rs
bh2smith: Cool I just remembered that there was a standard practice in python development that using relative paths for import statements was some sort of faux-pas.

nlordell: Here its needed unless you want to get needlessly fancy about it. Also its super deterministic seeing as it will always be relative to the same directory (unless the `driver` sub-crate moves).

driver/src/bin/stablex.rs
```diff
@@ -10,7 +10,7 @@ use std::time::Duration;
 fn main() {
     simple_logger::init_with_level(log::Level::Info).unwrap();
 
-    let contract = StableXContractImpl::new().unwrap();
+    let (contract, _event_loop) = BatchExchange::new().unwrap();
```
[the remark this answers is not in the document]
bh2smith: I will be excited to see what this means if we wind up doing it!

driver/src/contracts/mod.rs
```diff
@@ -1,4 +1,34 @@
...
+}
+
+fn method_defaults() -> Result<MethodDefaults, DriverError> {
+    let network_id = env::var("NETWORK_ID")?.parse()?;
+    let secret = {
+        let private_key: H256 = env::var("PRIVATE_KEY")?.parse()?;
+        SecretKey::from_raw(&private_key[..]).map_err(ethsign::Error::from)?
+    };
+    let account = Account::Offline(secret, Some(network_id));
+    let defaults = MethodDefaults {
```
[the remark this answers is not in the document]
nlordell: Yes, I kept that from the previous version. I'm going to do that in a separate PR though.

driver/src/bin/stablex.rs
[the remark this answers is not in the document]
nlordell: Hmm, my problem with that is that if we ever end up wanting to create and clean up multiple contract instances, then it might be problematic as they would never get cleaned up. I suggest we do `event_loop.into_remote()` for now, but then we have to keep in mind that these event loops are leaking every time we make a new contract instance.

.travis.yml
```diff
@@ -33,6 +34,8 @@ matrix:
         - rm -f  ./target/.rustc_info.json
       env:
         - CARGO_INCREMENTAL=0
+      install:
+        - ./test/setup_contracts.sh
```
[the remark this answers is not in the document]
nlordell: We need to deploy so we have dev network addresses in the artifact.

nlordell:
@fleupold Right now the contract interfaces are used for mocking.

.travis.yml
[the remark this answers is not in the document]
fleupold: Ah I thought those were only needed on runtime. But I guess they get hardcoded into the generated code at build time?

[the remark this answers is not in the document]
nlordell: Yes, they are part of the truffle artifact which is needed at build time. The artifact gets hardcoded into the generated code.

nlordell:
> Feature request for `ethcontract-rs`: Allow to create a mocked version of the contract.

gnosis/ethcontract-rs#83

driver/src/contracts/mod.rs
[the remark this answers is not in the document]
nlordell: Created #369

common.env
```diff
@@ -1,11 +1,8 @@
-SNAPP_CONTRACT_ADDRESS=0xe982E462b094850F12AF94d21D470e21bE9D0E9C
-SNAPP_BASE_LIB_ADDRESS=0x9561C133DD8580860B6b7E504bC5Aa500f0f06a7
-STABLEX_CONTRACT_ADDRESS=0x67B5656d60a809915323Bf2C40A8bEF15A152e3e
```
bh2smith: Where have these gone?

nlordell: They are no longer needed - its one of the bonuses of using `ethcontract`. It reads it from the truffle artifacts (`BatchExchange.json` and `SnappAuction.json`).

nlordell: This should mean that updating contract deployments should just require submodule updates (unless there are breaking changes of course).

PR #366, opened 2019-12-19
Use sccache

fleupold:
Hoping to speed up the build and unit test step

PR #367, opened 2019-12-19
parallelize all the builds

fleupold:
This will likely increase total time, but hopeful shorten the run time. Let's see the trade-off.


fleupold:
Should probably be rebased on top of #366

fleupold:
Not beneficial overall

PR #368, opened 2019-12-19
[Easy] Stable Rust

nlordell:
Build with stable Rust on CI, since Rust 1.40 was released. Also update README.

### Test Plan

CI


nlordell:
It was the new rustc and clippy (mostly clippy) version that spotted them.

PR #371, opened 2019-12-20
[Easy] Download awscli directly instead of via python

fleupold:
Micro-optimizing travis times by now.

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -4,8 +4,6 @@ set -euo pipefail
 
 image_name=$1
 
-sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
-
 # Get login token and execute login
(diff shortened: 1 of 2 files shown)


docker/deploy.sh
```diff
@@ -4,8 +4,6 @@ set -euo pipefail
 
 image_name=$1
 
-sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
```
nlordell: should this be `sudo apt-get update && sudo apt-get install awscli`?

fleupold: No, because it was already installed in the context before calling this script

fleupold:
Actually this doesn't work. Turns out downloading awscli via apt actually takes longer (and is an outdated version).

PR #372, opened 2019-12-23
Always log rinkeby e2e test logs

nlordell:
This change snuck in #363 but shouldn't have. This addresses that.

### Test Plan

CI -> verify logs are dumped at end of rinkeby e2e test.


nlordell:
Logs can be seen here: https://travis-ci.org/gnosis/dex-services/jobs/628644247#L424

PR #373, opened 2019-12-23
[Deploy] Allow multiple services to be restarted on master commit

fleupold:
Right now, we only restart the rinkeby AWS instance on master commits. This PR refactors the logic to allow for an array of autodeply URLs to be passed in (separated by semicolon).

This will allow us to restart both mainnet and rinkeby services on each master commit.

### Testplan
Printed the curl command locally. It remains to see bot services restart after this PR is merged in to master

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -2,6 +2,15 @@
 
 set -euo pipefail
 
+function autodeploy {
+    # Notifying webhook
+    curl -s --output /dev/null --write-out "%{http_code}" \
+    -H "Content-Type: application/json" \
+    -X POST \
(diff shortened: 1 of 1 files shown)


docker/deploy.sh
```diff
@@ -2,6 +2,15 @@
 
 set -euo pipefail
 
+function autodeploy {
+    # Notifying webhook
+    curl -s --output /dev/null --write-out "%{http_code}" \
+    -H "Content-Type: application/json" \
```
nlordell: nit: maybe indent params here so its easy to identify that they pertain to the `curl` call

PR #375, opened 2019-12-23
Logging Transport

nlordell:
Adds a logging transport implementation that logs JSON RPC requests and results to assist debugging.

### Test Plan

CI and take a look at the dumped rinkeby logs.

### DIFF
--- a/driver/src/contracts/mod.rs
+++ b/driver/src/contracts/mod.rs
@@ -3,16 +3,19 @@ pub mod stablex_auction_element;
 pub mod stablex_contract;
 
 use crate::error::DriverError;
+use crate::transport::LoggingTransport;
 use ethcontract::contract::MethodDefaults;
 use ethcontract::{ethsign, Account, SecretKey, H256};
+use log::Level;
 use std::env;
 use web3::api::Web3;
 use web3::transports::{EventLoopHandle, Http};
 
-fn web3_provider() -> Result<(Web3<Http>, EventLoopHandle), DriverError> {
+fn web3_provider() -> Result<(Web3<LoggingTransport<Http>>, EventLoopHandle), DriverError> {
     let url = env::var("ETHEREUM_NODE_URL")?;
     let (event_loop, http) = Http::new(&url)?;
-    let web3 = Web3::new(http);
+    let logging = LoggingTransport::new(http, Level::Debug);
+    let web3 = Web3::new(logging);
 
     Ok((web3, event_loop))
 }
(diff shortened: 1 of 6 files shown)


driver/src/contracts/mod.rs
```diff
@@ -3,16 +3,19 @@ pub mod stablex_auction_element;
...
+use crate::transport::LoggingTransport;
 use ethcontract::contract::MethodDefaults;
 use ethcontract::{ethsign, Account, SecretKey, H256};
+use log::Level;
 use std::env;
 use web3::api::Web3;
 use web3::transports::{EventLoopHandle, Http};
 
-fn web3_provider() -> Result<(Web3<Http>, EventLoopHandle), DriverError> {
+fn web3_provider() -> Result<(Web3<LoggingTransport<Http>>, EventLoopHandle), DriverError> {
```
fleupold: Do we need to be explicit about the type here? Couldn't it be an abstract Transport?

driver/src/contracts/mod.rs
```diff
@@ -3,16 +3,19 @@ pub mod stablex_auction_element;
...
     let url = env::var("ETHEREUM_NODE_URL")?;
     let (event_loop, http) = Http::new(&url)?;
-    let web3 = Web3::new(http);
+    let logging = LoggingTransport::new(http, Level::Info);
```
fleupold: I think this should be debug though. As all our other logging atm is info (and this is really something we only need when debugging issues)

driver/src/contracts/mod.rs
nlordell: We could make it `impl Transport`.

driver/src/contracts/mod.rs
nlordell: I had it like that originally but debug level is super verbose. Also, we currently have no way of enabling it without changing the default log level. I'm OK either way, but we either need to:
- make the default log level `Debug`
- keep this as `Info`

driver/src/contracts/mod.rs
nlordell: Not making this a concrete type makes things more complicated so I'll leave it in for now.

driver/src/contracts/mod.rs
fleupold: Maybe we can pass in the log level as an argument?

nlordell: OK, but lets do it as a separate PR and then we can use `env_logger` instead of `simple_logger` which reads log levels from `env` and even allows you to specify logging levels per module.

PR #376, opened 2019-12-23
`slog-envlogger`

### LINKED ISSUE #263: Unified Logger
The dfusion "service" is currently using two different loggers (`simple_logger` and `graph::log::logger`)
 both instantiated in the following snippet:

https://github.com/gnosis/dex-services/blob/3898d77c7247b457de785196a0437cc11424439e/driver/src/bin/dfusion.rs#L18-L19

Partial reasoning for this is that the graph logger is specific to `the graph` which we only introduced recently.

It would be nice if we used a single logger throughout the project and a _bonus_ if there were timestamps!

nlordell:
Switch `simple_logger` for `slog-envlogger` giving us more control over log levels (you can specify log level per module) that gets read from the environment. This has the added benefit that we unify the logger implementations being used for `graphprotocol` components and the `driver`. 

Additionally, the log level for the `LoggingTransport` was changed to `DEBUG`.

Closes #263

### Test Plan

CI and check output logs.


nlordell:
@bh2smith Do you mind re-reviewing? I took this opportunity to unify the loggers.

nlordell:
Logs are indeed showing up: https://travis-ci.org/gnosis/dex-services/jobs/628746405#L432

PR #378, opened 2019-12-23
Revert "[Deploy] Allow multiple services to be restarted on master commit"

fleupold:
Reverts gnosis/dex-services#373

No longer needed, due to the fact that we can encode two target services in a single URL.

https://dfusion.auto.gnosisdev.com/services/dfusion-mainnet-master,dfusion-rinkeby-master/restart

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -2,17 +2,6 @@
 
 set -euo pipefail
 
-function autodeploy {
-    # Notifying webhook
-    curl -s \
-      --output /dev/null \
-      --write-out "%{http_code}" \
-      -H "Content-Type: application/json" \
-      -X POST \
-      -d '{"token": "'$AUTODEPLOY_TOKEN'", "push_data": {"tag": "'$AUTODEPLOY_TAG'" }}' \
-      $1
-}
-
 image_name=$1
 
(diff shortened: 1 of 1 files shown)


docker/deploy.sh
```diff
@@ -29,10 +18,11 @@ docker push $REGISTRY_URI:$image_name
...
-    for url in "${AUTODEPLOY_URL_LIST[@]}"; do
-      autodeploy "$url"
-    done
+if [ "$image_name" == "master" ] && [ -n "$AUTODEPLOY_URL" ] && [ -n "$AUTODEPLOY_TOKEN" ]; then
+    # Notifying webhook
+    curl -s --output /dev/null --write-out "%{http_code}" \
+    -H "Content-Type: application/json" \
+    -X POST \
+    -d '{"token": "'$AUTODEPLOY_TOKEN'", "push_data": {"tag": "'$AUTODEPLOY_TAG'" }}' \
+    $AUTODEPLOY_URL
```
nlordell: Can we fix the indentation here or would you prefer to do that as a separate PR?

fleupold: I can fix this separately given that this is clean revert

PR #392, opened 2019-12-27
add mainnet compose config

fleupold:
This will allow running against mainnet contracts with local deployment

### Test Plan

```
docker-compose -f docker-compose.yml -f docker-compose.mainnet.yml up stablex
```


common-mainnet.env
```diff
@@ -0,0 +1,4 @@
+ETHEREUM_NODE_URL=https://mainnet.infura.io/v3/9408f47dedf04716a03ef994182cf150
```
nlordell: Shouldn't we be putting these Infura project IDs as Travis variables?

fleupold: It's not really important as it's not really a secret. 
If we did this it would require manual exports on the local machine. If Infura ever starts rate-limiting/charging for accounts we might have to switch to a secret ID.

bh2smith: > If Infura ever starts rate-limiting/charging for accounts we might have to switch to a secret ID.

I suppose they eventually will, after "to the moon"

PR #394, opened 2019-12-30
[Easy] Upgrade solver version

fleupold:
self explanatory. This line will upgrade the version of the solver that we use. @marcovc should we directly upgrade to the version that contains your fix as well?

### Test Plan
CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
       script:
         - cargo build
         # Build and publish compact image with compiled binary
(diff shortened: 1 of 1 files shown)


marcovc:
I was wondering if there is a way this PR could recur automatically whenever we tag a new version.

The last fix today was for an issue you raised: the fact that sometimes some keys are missing from the output json. If you already account for that in the solver client code, than I would say is not critical anymore.

fleupold:
We do have a fix for the missing key in place. We should probably re-release for https://github.com/gnosis/dex-solver/pull/43

As for doing it automatically, I believe we could depend on a `latest` tag and update that with every version tag. However, we would still require to make a new build of dex-services master so that the binary we publish to AWS gets updated. 

Such a build/PR could probably be triggered by the travis.yml inside the dex-solver repo. @nlordell was thinking about using something similar to automatically trigger a npm release whenever we tag a version of the dex-contracts.

PR #395, opened 2020-01-03
Use latest version of yarn and node10 for travis builds

fleupold:
We are seeing some flaky tests (e.g. [[1]](https://travis-ci.com/gnosis/dex-services/jobs/271235595) & [[2]](https://travis-ci.com/gnosis/dex-services/jobs/271380142)) which seem to stem from the fact that `~/.node-gyp/8.12.0/include/node/common.gypi` is empty where it should have a valid configuration inside. My hunch is that it is some kind of race condition with parallelising the build of the sha package and node-gyp.

This PR updates node to use the stable version we actually support (v10) and also the latest version of yarn. If this doesn't fix the issue, we likely have to enable verbose yarn logging to see in which order things are installed.

My hope is however that the issue is already fixed on the latest version.

### Test Plan

CI and observing if we keep getting random failures.


nlordell:
```
You need to run "nvm install 10" to install it before using it.
```

PR #398, opened 2020-01-03
Encode num_tokens Solution bytes

bh2smith:
At the moment, the project relies on a global constant `TOKENS` to decode Solution bytes  emitted from the `SnappAuction` contract. Rather than storing `num_tokens` in the database, or making and EVM query, this can simply be inferred as the length of the price vector in the `Solution` struct. This wouldn't necessarily be the number of tokens, but rather the max token ID represented by the length of the price vector.

**Test Plan** See that this change does not affect `dƒusion e2e Tests` and note that this change does not require any changes to the smart contracts (since `SnappAuction` does not interpret the bytes contained in `pricesAndVolumes`).

We may find that the unit tests fail becuase of hard-coded hash comparisons.


dfusion_rust_core/src/models/solution.rs
```diff
@@ -36,17 +37,24 @@ impl Serializable for Solution {
...
+        .flat_map(Serializable::bytes)
+        .collect()
     }
 }
 
 impl Deserializable for Solution {
     fn from_bytes(mut bytes: Vec<u8>) -> Self {
-        let volumes = bytes.split_off(TOKENS as usize * 12);
+        // First 12 bytes encode the length of price vector (i.e. num_tokens)
+        let len_prices = U128::from_big_endian(bytes.drain(0..12).collect::<Vec<u8>>().as_slice());
```
fleupold: Can we use 2 bytes instead of 12?

bh2smith: Yes sir! I had originally thought this would be a good idea, but didn't go with it immediately. Now that I have a second opinion! Will do.

dfusion_rust_core/src/models/solution.rs
```diff
@@ -29,24 +30,37 @@ impl Solution {
...
+        res.extend(prices_and_volumes);
+        return res;
     }
 }
 
 impl Deserializable for Solution {
     fn from_bytes(mut bytes: Vec<u8>) -> Self {
-        let volumes = bytes.split_off(TOKENS as usize * 12);
+        // First 2 bytes encode the length of price vector (i.e. num_tokens)
+        let len_prices = u16::from_be_bytes(
```
fleupold: nit: why not use split off here as well just like below? Also to avoid unwrapping, your could use a method like this one: https://github.com/gnosis/dex-services/blob/8473762420ceefac762799c823c659a212a1084d/dfusion_rust_core/src/models/util.rs#L207-L212

bh2smith: The link is dead, but I will check it out.

bh2smith: Ya, looks much cleaner than what I have here. I was just trying to "consume" the bytes instead of copy them (since they aren't needed anymore afterwards).

dfusion_rust_core/src/models/solution.rs
```diff
@@ -29,29 +30,36 @@ impl Solution {
...
-        [&self.prices, &alternating_buy_sell_amounts]
+
+        let prices_and_volumes: Vec<u8> = [&self.prices, &alternating_buy_sell_amounts]
             .iter()
             .flat_map(|list| list.iter())
             .flat_map(Serializable::bytes)
-            .collect()
+            .collect();
+        res.extend(prices_and_volumes);
+        return res;
```
bh2smith:
```suggestion
        res
```

dfusion_rust_core/src/models/solution.rs
```diff
@@ -29,29 +30,36 @@ impl Solution {
...
 }
 
 impl Deserializable for Solution {
     fn from_bytes(mut bytes: Vec<u8>) -> Self {
-        let volumes = bytes.split_off(TOKENS as usize * 12);
+        // First 2 bytes encode the length of price vector (i.e. num_tokens)
+        let len_prices = BigEndian::read_u16(&bytes[0..2]);
+        let volumes = bytes.split_off(2 + len_prices as usize * 12);
         let prices = bytes
+            .split_off(2)
```
nlordell: I would suggest `bytes[2..]` instead of `bytes.split_off(2)` here as the latter allocates a completely new `Vec` and since you don't re-use `bytes` it seems unnecessary.

bh2smith: Will include this great nit asap... but forgot here.

PR #400, opened 2020-01-03
remove gas limit for orders query

nlordell:
:point_up: 

We were running into a gas limit on `eth_call` requests because the number of orders was getting too large.

### Test Plan

See that the driver is no longer failing to get all orders on mainnet.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -68,11 +68,9 @@ impl StableXContract for BatchExchange {
     }
 
     fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)> {
-        let packed_auction_bytes = self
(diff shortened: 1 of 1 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -68,11 +68,9 @@ impl StableXContract for BatchExchange {
     }
 
     fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)> {
-        let packed_auction_bytes = self
-            .get_encoded_orders()
-            .gas(10_000_000.into())
-            .call()
-            .wait()?;
+        let mut orders_builder = self.get_encoded_orders();
+        orders_builder.m.tx.gas = None;
```
fleupold: Maybe a small comment here why we do this?

PR #402, opened 2020-01-06
[E2E] StableX e2e in native rust

fleupold:
Writing the e2e tests in rust instead of in shell + js will allow us drop the bloaty dex-services subrepo and replace it with the lightweight dex-services npm packages. 
It will also make it easier to do proper logic assertions (e.g. capturing the balance after a withdraw) without fancy grep and shell expressions.
Rust e2e tests also run much faster (this one went from 45s in shell/js to 12s in native rust).

This is the first of five e2e scenarios being converted. For now I don't run it in travis yet. This will be that last step of the PR series.

The orginal e2e test is this one: https://github.com/gnosis/dex-services/blob/master/test/e2e-tests-stablex-ganache.sh

### Test Plan

- Run the stablex system on ganace (`docker-compose up stablex`, then `cd dex-contracts && npx truffle migrate`)
- `cargo test test_stablex`


e2e/tests/common.rs
```diff
@@ -0,0 +1,137 @@
...
+            token
+                .approve(instance.address(), U256::exp10(18) * TOKEN_MINTED)
+                .from(Account::Local(*account, None))
+                .send()
+                .wait()
+                .expect("Cannot approve OWL for burning");
+        }
+    }
+
+    // token[0] is already addfed in constructor
```
bh2smith:
```suggestion
    // token[0] is already added in constructor
```

e2e/tests/stablex_test.rs
```diff
@@ -0,0 +1,89 @@
...
+        .send()
+        .wait()
+        .expect("Cannot place request withdraw");
+    common::close_auction(&web3, &instance);
+
+    let balance_before = tokens[1]
+        .balance_of(accounts[0])
+        .call()
+        .wait()
+        .expect("Cannot get balance before");
```
bh2smith: Just curious if it might be worth the effort to slightly reduce redundancy by making a couple helpers like `call_wait_expect` and/or `send_wait_expect`

e2e/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
bh2smith: Is this an empty file?

.travis.yml
```diff
@@ -60,7 +60,7 @@ matrix:
           popd
         - rustup component add clippy rustfmt
       script:
-        - cargo test --all
+        - cargo test --workspace --exclude e2e
```
nlordell: I would suggest adding a `default-members` that excludes the `e2e` crate to the `Cargo.toml` instead and then doing a `cargo test` (which will only run for the `default-members`).

e2e/src/lib.rs
nlordell: I think it is needed by cargo so its a crate. All the e2e tests seem to be implemented as integration tests. I might be wrong though.

e2e/tests/common.rs
```diff
@@ -0,0 +1,137 @@
+use ethcontract::web3::api::Web3;
```
nlordell: Just a suggestion, the common methods could be added to the `e2e` crate directly and then the tests would call into it. There is no real advantage to this, however, and just something that I thought would be "cleaner".

e2e/tests/common.rs
```diff
@@ -0,0 +1,137 @@
...
+            .wait()
+            .expect("Cannot add token");
+    }
+    (instance, accounts, tokens)
+}
+
+pub fn wait_for(web3: &Web3<Http>, seconds: u32) {
+    web3.transport()
+        .execute("evm_increaseTime", vec![seconds.into()]);
+    web3.transport().execute("evm_mine", vec![]);
```
nlordell: I finally now know how this is done! It was always black magic to me :star2:

e2e/tests/common.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+pub fn close_auction(web3: &Web3<Http>, instance: &BatchExchange) {
+    let seconds_remaining = instance
+        .get_seconds_remaining_in_batch()
+        .call()
+        .wait()
+        .expect("Cannot get seconds remaining in batch");
+    wait_for(web3, seconds_remaining.low_u32());
```
nlordell: I suggest using `as_u32` so it panics if it would overflow the u32 - it is a test after all and I think it would be a good idea to check all our assumptions.

e2e/tests/stablex_test.rs
fleupold: Agree. My feedback to @nlordell was that this might be something ethcontract-rs could provide. E.g. each future should have a default way to be sent (e.g. call for ViewMethodBuilders, deploy for DeployBuilders, etc) - https://github.com/gnosis/ethcontract-rs/issues/98

Regarding expect, I'm using custom messages to help me debug which call/tx failed. This is something that has been very recently added to ethcontracts but is not in the version we are depending on yet (https://github.com/gnosis/ethcontract-rs/pull/101).

e2e/src/lib.rs
fleupold: No, that's it

e2e/tests/common.rs
fleupold: I toke the suggestion of having a common submodule inside the tests from here: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

But in our case we have an entire project dedicated to tests so having it inside src/ can make sense

PR #404, opened 2020-01-06
Dynamic Tokens

bh2smith:
Currently the entire services repo relies on a global constant `TOKENS = 30`

https://github.com/gnosis/dex-services/blob/90d92089f4ea1a31c898beb9bea6a7c4ecb42587/dfusion_rust_core/src/models/mod.rs#L18-L19

which means these services would crash as soon as 31 tokens are listed and requested for trade on this exchange.

The changes here remove this constant using a dynamically allocated number of tokens based on the maximum token id appearing in the list of relevant orders for an auction batch.

Depends on #398 (which were the first phase of this improvement)


dfusion_rust_core/src/models/account_state.rs
```diff
@@ -214,26 +214,27 @@ pub mod test_util {
 pub mod tests {
     use super::*;
     use crate::models::order::BatchInformation;
-    use crate::models::TOKENS;
     use std::str::FromStr;
     use web3::types::{Bytes, H256};
 
+    const NUM_TOKENS: u16 = 10;
```
nlordell: Why 10? Is there a specific reason for that or is it just an arbitrary amount?

bh2smith: Just arbitrary for the unit tests, but I will probably remove this. Still need to make a helper function that maps u16 to bytes (i.e. Vec<u8>)

bh2smith: I have included a comment that this is not super relevant, just used for readability throughout the unit tests. Also set as a constant within the tests of some other files too. Hope this is alright,.

dfusion_rust_core/src/models/mod.rs
```diff
@@ -38,6 +37,7 @@ pub trait Serializable {
 }
 
 pub trait Deserializable {
+    // TODO: encode num_tokens in the first byte(s)
```
fleupold: This is a general trait, and should probably not have a comment about the particular `num_token` issue.

dfusion_rust_core/src/models/solution.rs
```diff
@@ -13,16 +13,15 @@ pub struct Solution {
 }
 
 impl Solution {
-    pub fn trivial(num_orders: usize) -> Self {
+    pub fn trivial(num_orders: usize, num_tokens: usize) -> Self {
```
fleupold: Shouldn't num_tokens inside the trivial solution be always 0?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -144,7 +143,12 @@ impl PriceFinding for LinearOptimisationPriceFinder {
         orders: &[models::Order],
         state: &models::AccountState,
     ) -> Result<models::Solution, PriceFindingError> {
-        let token_ids: Vec<String> = (0..models::TOKENS).map(token_id).collect();
+        let max_token_id = orders
```
fleupold: I think we can omit the tokens that don't have any orders. E.g. if we only have an order trading token0 for token7 the list can be [token0, token7] instead of [token0, token1, ...]

driver/src/price_finding/naive_solver.rs
```diff
@@ -101,8 +102,20 @@ impl PriceFinding for NaiveSolver {
...
+            .unwrap();
+        let num_tokens = if let Some(fee) = &self.fee {
+            cmp::max(fee.token, max_token_id) + 1
+        } else {
+            max_token_id + 1
+        };
+
         // Initialize trivial solution (default of zero indicates untouched token).
-        let mut prices: Vec<u128> = vec![0; TOKENS as usize];
+        let mut prices: Vec<u128> = vec![0; num_tokens as usize];
```
fleupold: Should this be a Map instead of a vector? Maybe this will imply changes to the Solution struct, which could be done in a separate PR on top?

dfusion_rust_core/src/models/solution.rs
bh2smith: At first, this is what I was hoping we could do and I would very much like it to be this way, however, this would require adapting solution evaluation for trivial solutions since we get missing token errors on trivial solutions with no price vector. 

Here:

https://github.com/gnosis/dex-services/blob/2a442f847dfe716b83a5784528910a0ad6abb60d/driver/src/snapp.rs#L40-L63


I was hoping you would point this out and encourage me to solve this problem in a better way. So I will go ahead and do that. It would require less overall changes anyway.


note that the above code fails with `TokenNotFoundError` on the following trivial solution

```
Solution {
    prices: []
    executed_buy_amounts: [0, 0], 
    executed_sell_amounts: [0, 0],
}
```

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Yes, I like this idea also and will work toward making this a reality! Looking forward to it!

driver/src/price_finding/naive_solver.rs
bh2smith: I would very much agree with this. Should the separate PR come first? The chicken or the egg? I would think the change to a map should come first.

dfusion_rust_core/src/models/mod.rs
bh2smith: Completely agree. I think this was a merge issue... whoops.

bh2smith:
Hey! So, I would say that this "phase" of the change is ready for merge. I have created a new issue introducing a further improvement to the `Solution` struct #412 to be addressed next. How does all this sound to you @fleupold?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -94,10 +93,10 @@ fn deserialize_result(
                 .unwrap_or_else(|| (token.to_owned(), "0".to_string()))
         })
         .collect::<Prices>();
-    let prices = (0..num_tokens)
+    let prices = (0..price_map.len())
```
fleupold: I'm not a huge fan of the unwrap(). Can we comment why this is not going to panic (price_map can have at most uint16 prices) or maybe even allow tokenId() to be called with usize as argument?

driver/src/price_finding/naive_solver.rs
```diff
@@ -101,8 +102,14 @@ impl PriceFinding for NaiveSolver {
         orders: &[Order],
         state: &AccountState,
     ) -> Result<Solution, PriceFindingError> {
+        // Fetch max_token_id to determine num_tokens
+        let max_token_id = orders
+            .iter()
+            .map(|o| cmp::max(o.buy_token, o.sell_token))
+            .max()
+            .unwrap();
```
fleupold: why not unwrap_or(0) here?

driver/src/price_finding/naive_solver.rs
```diff
@@ -268,6 +275,7 @@ pub mod tests {
         let state = create_account_state_with_balance_for(&orders);
 
         let solver = NaiveSolver::new(None);
+        println!("{:?}", orders);
```
fleupold: please remove before merging.

driver/src/price_finding/linear_optimization_price_finder.rs
fleupold: Did this happen?

bh2smith: I was saving this for part 3 which is essentially the content of #412

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: I would be in favour of allowing `token_id()` be called on `usize`

driver/src/price_finding/naive_solver.rs
bh2smith: Yes absolutely. I am actually surprised this is passing. Will fix this and add a test.

PR #410, opened 2020-01-07
Update README.md

fleupold:
YK was trying to work through our README and was having some issues. Looking at it I feel like it deserved a major revamp putting more focus on the StableX component as this is likely what most external developers are interested in.

Please suggest changes where things are not clear...


README.md
```diff
@@ -2,15 +2,75 @@
...
 ## Intro
 
-This repository contains the backend logic for the dfusion exchange based on the specification, see [here](github.com/gnosis/dex-research)
+This repository contains the backend logic for the dfusion exchange based on [this specification](github.com/gnosis/dex-research).
 
-## Architecture
+It contains two sub-projects that both implement the market mechanism described above in different ways. An fully on-chain solution with instant finality but limited scalability (referred to as "StableX") and a preliminary version that intends to achieves scalability by offloading computation and data-storage off-chain using an [optimistic roll-up](https://medium.com/plasma-group/ethereum-smart-contracts-in-l2-optimistic-rollup-2c1cef2ec537) approach. The latter is in early development stage and not yet ready for use.
 
-<p align="center">
-  <img src="documentation/architecture.png" alt="dex-services architecture" width="500">
```
bh2smith: I was also wondering if we might remove this diagram soon. No worries! It will return (eventually)

README.md
```diff
@@ -2,15 +2,75 @@
...
-</p>
+## Getting Started
+
+### Requirements
+
+- Rust (stable)
+- NodeJS 11.0, starting with version 12 some deprecated APIs were removed that cause `scrypt`, `keccak`, `secp256k1`, and `sha3` packages to fail to build
```
bh2smith:
```suggestion
- NodeJS <=11.0, starting with version 12 some deprecated APIs were removed that cause `scrypt`, `keccak`, `secp256k1`, and `sha3` packages to fail to build
```

README.md
```diff
@@ -39,30 +99,15 @@ Thus, our data layer has to rely only on the data emitted by the EVM. It cannot
...
-Clone the repository, its submodule, and run the container
-```bash
-git clone git@github.com:gnosis/dex-services.git
-cd dex-services
-git submodule update --init
-docker-compose up -d ganache-cli
-(cd dex-contracts && yarn && npx truffle migrate && npm run networks-inject)
-docker-compose up
+```
+docker-compose down && docker-compose up driver graph-listener truffle
```
bh2smith: What is wrong with just 

```
docker-compose down && docker-compose up
```

Does this have something to do with postgres?

README.md
```diff
@@ -39,30 +99,15 @@ Thus, our data layer has to rely only on the data emitted by the EVM. It cannot
...
 ```
 
 This will start:
 ganache-cli, the local ethereum chain
+a truffle image compiling and deploying your smart contracts to the local chain
 postgres, the database storing the data of the snapp
 graph-listener, a listener pulling data from the ganache-cli and inserting it into postgres
 driver, a service calculating the new states and push these into the smart contract
```
bh2smith: Can we itemize or enumerate this list... I found the transaction from one line to the next confusing before I realized it was bullet points.

README.md
fleupold: You will then have to `(cd dex-contracts && truffle migrate)` which the truffle box does for you

README.md
```diff
@@ -2,15 +2,75 @@
...
+It contains two sub-projects that both implement the market mechanism described above in different ways. An fully on-chain solution with instant finality but limited scalability (referred to as "StableX") and a preliminary version that intends to achieves scalability by offloading computation and data-storage off-chain using an [optimistic roll-up](https://medium.com/plasma-group/ethereum-smart-contracts-in-l2-optimistic-rollup-2c1cef2ec537) approach. The latter is in early development stage and not yet ready for use.
 
-<p align="center">
-  <img src="documentation/architecture.png" alt="dex-services architecture" width="500">
-</p>
+## Getting Started
+
+### Requirements
+
+- Rust (stable)
```
anxolin: why do we need Rust, and Docker?

Is not docker enough?

README.md
```diff
@@ -2,15 +2,75 @@
...
+
+### Installation
+
+Clone the repository, its submodule, and run the container
+```bash
+git clone git@github.com:gnosis/dex-services.git
+cd dex-services
+git submodule update --init
+docker-compose up -d ganache-cli
+(cd dex-contracts && yarn && npx truffle compile && npm run networks-inject && npx truffle migrate)
```
anxolin: For me is fine, and if it adds too much complication we can forget about it. But is a bit strange we require docker, but for this parts we use something else.

I think docker allows to map the source code of the contracts to the container, and also to connect to the ganache launched in line 26.

README.md
```diff
@@ -2,15 +2,75 @@
...
+```
+
+## StableX
+
+The StableX system only consists of a simple service that queries the relevant auction information (orders and balances) directly from the blockchain. It then tries to find and submit a valid solution as soon as the order collection phase for a given auction ends.
```
anxolin: Maybe point out that anyone can run a solver, and how they can implement one (point to the submitSolution function in the contract) and possible the naive implementation

README.md
```diff
@@ -2,15 +2,75 @@
...
+
+```bash
+docker-compose up stablex
+```
+
+In order to setup the environment and make a first deposit/order you can run:
+
+```
+cd dex-contracts
+npx truffle exec scripts/stablex/setup_environment.js
```
anxolin: I thinks it's more clear if you explain a bit what this lines do using a comment. i.e:

Actually, I'm not sure someone would realize this is for running a fake thing in a local ganache with fake data.

```bash
# Basic Test Setup: Lists 3 fake tokens, and creates a deposit for 3 fakes users
npx truffle exec scripts/stablex/setup_environment.js
```

README.md
fleupold: technically yes, you might want to run some things such as unit tests locally, which is why I left it in there.

README.md
fleupold: So you are suggesting to change the docker entry-point to run the truffle commands as a first step in the container before running the rust binary. That would work and could save the extra command.

However, generally it shouldn't be the responsibility of the service to compile or migrate any contracts. It's normal mode of operation is expecting that those have already been deployed (which happens here outside of the container). It would also require the node to have node installed.
Maybe we can add this for the ganache build and not use it for rinkeby/mainnet. I will take a look and change the README if that procedure changes.

README.md
fleupold: Naive solver is pointed to in the next paragraph. Will also point to submitSolution

bh2smith:
I was working with Nick earlier this morning and used the following commands to run rinkeby locally.

```
cargo run --bin stablex
```

Not sure if this was included in the doc somewhere... there were also several environment variables that needed to be configured (like `PK`) beforehand.

README.md
nlordell: I think this should be changed to `(cd dex-contracts && yarn && yarn prepack && npx truffle migrate)`. That way if the build process slightly changes, we won't have to remember to update the README.

anxolin: In reality I was thinking more of running a `docker-compose run`

Anyways, it's not important for now, at least for me. Let's see the feedback we get from YK and others

PR #413, opened 2020-01-08
[E2E] StableX Rinkeby test in rust

fleupold:
Converting the e2e test that actually runs an entire auction on Rinkeby. It requires waiting for up to 5 minutes (depending on when the batch will finish), but other than the existing javascript test we  actually wait on the confirmations for deposits/and order placements making it more reliable than the one we have today.

It's not yet being used on travis, only locally

### Test Plan

run 
```
docker-compose down && docker-compose -f docker-compose.yml -f docker-compose.rinkeby.yml up stablex
```

and then
```
cargo test -p e2e rinkeby -- --nocapture
```

and see the spectacle unfold...

### DIFF
--- a/e2e/tests/stablex_test.rs
+++ b/e2e/tests/stablex_test.rs
@@ -1,12 +1,21 @@
 use ethcontract::web3::api::Web3;
-use ethcontract::web3::transports::Http;
+use ethcontract::web3::futures::Future as F;
+use ethcontract::web3::transports::{Http, WebSocket};
 use ethcontract::web3::types::U256;
-use ethcontract::Account;
+use ethcontract::{ethsign, Account, SecretKey, H256};
+
+use futures::future::join_all;
 
 use e2e::common::{close_auction, setup, wait_for_condition, FutureWaitExt};
 
+use std::env;
+use std::process::Command;
+use std::time::Duration;
+
+ethcontract::contract!("dex-contracts/build/contracts/IERC20.json");
+
 #[test]
-fn test_stablex() {
+fn test_with_ganache() {
     let (eloop, http) = Http::new("http://localhost:8545").expect("transport failed");
     eloop.into_remote();
     let web3 = Web3::new(http);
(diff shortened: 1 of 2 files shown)


e2e/tests/stablex_test.rs
```diff
@@ -111,3 +120,124 @@ fn test_stablex() {
...
+        .send_and_confirm(Duration::from_secs(1), 1);
+
+    // Place orders
+    let first_order = instance
+        .place_order(0, 7, batch + 2, 1_000_000.into(), 10_000_000.into())
+        .nonce(nonce + 4)
+        .send_and_confirm(Duration::from_secs(1), 1);
+    let second_order = instance
+        .place_order(7, 0, batch + 1, 1_000_000.into(), 10_000_000.into())
+        .nonce(nonce + 5)
```
nlordell: Clever!

e2e/tests/stablex_test.rs
```diff
@@ -111,3 +120,124 @@ fn test_stablex() {
...
+        .send_and_confirm(Duration::from_secs(1), 1);
+
+    // Wait for all transactions to be confirmed
+    println!("Waiting for transactions to be confirmed");
+    let results = join_all(vec![
```
nlordell: I like `try_join!` for this, it also turns it into a `Result<(T1, T2, T3, ...), Error>` for you.

e2e/tests/stablex_test.rs
```diff
@@ -111,3 +120,124 @@ fn test_stablex() {
...
+    println!("Sleeping {} seconds...", sleep_time);
+    std::thread::sleep(Duration::from_secs(sleep_time));
+
+    // Make sure there was no error
+    let output = Command::new("docker-compose")
+        .arg("logs")
+        .output()
+        .expect("failed to execute process");
+    let logs = String::from_utf8(output.stdout).expect("failed to read logs");
+    assert!(!logs.to_lowercase().contains("error"));
```
nlordell: Super robust error checking :stuck_out_tongue:

e2e/tests/stablex_test.rs
fleupold: The problem with try_join is that it itself awaits the futures and thus can only be used in an async function. There is try_join_all which does the same as try_join but returns a Future instead of directly awaiting it.

However, I'd prefer knowing which tx failed and thus unwrapping per item helps with that. I agree, that once we log the function name in the Future Error this may no longer be needed.

e2e/tests/stablex_test.rs
fleupold: We print `docker-compose logs` on test failure, thus I'd expect the developer to identify what was the problem.

e2e/tests/stablex_test.rs
```diff
@@ -111,3 +120,124 @@ fn test_stablex() {
...
+        .gas(1_000_000.into())
+        .gas_price(8_000_000_000u64.into())
+        .from(account.clone())
+        .send_and_confirm(Duration::from_secs(1), 1);
+    let second_approve = IERC20::at(&web3, token_b)
+        .approve(instance.address(), 1_000_000.into())
+        .nonce(nonce + 1)
+        .gas(1_000_000.into())
+        .gas_price(8_000_000_000u64.into())
+        .from(account.clone())
```
fleupold:
```suggestion
        .from(account)
```

e2e/tests/stablex_test.rs
```diff
@@ -111,3 +120,124 @@ fn test_stablex() {
...
+        .nonce(nonce + 1)
+        .gas(1_000_000.into())
+        .gas_price(8_000_000_000u64.into())
+        .from(account)
+        .send_and_confirm(Duration::from_secs(1), 1);
+
+    // Deposit Funds
+    let first_deposit = instance
+        .deposit(token_a, 1_000_000.into())
+        .nonce(nonce + 2)
```
bh2smith: Is speficying `nonce` something that is necessary for the user to do manually with ethcontracts?

fleupold:
> I have just started to wonder about the difference between expect and unwrap_or

https://learning-rust.github.io/docs/e4.unwrap_and_expect.html explains it and all the other flavours nicely.

Expect doesn't give you a fallback (in case the Result is of type error) whereas unwrap_or does

e2e/tests/stablex_test.rs
fleupold: We need it here because we want to send multiple txs without waiting for the previous one to be confirmed (save a lot of time). 
Thus we need to manually increment the nonce for each tx. By default ethcontract-rs checks "how many confirmed txs has this account sent in the past" and then sets nonce as this value + 1.

bh2smith: VERY COOL!

PR #414, opened 2020-01-08
[Easy] check that Cargo lockfile is in sync

### LINKED ISSUE #411: Build should fail if Cargo.lock is not in sync with Cargo.toml
This will help avoid PRs like #408

fleupold:
Closes #411 

Adding a command to the travis config that will fail if the lockfile was updated while building the project.

### Test Plan

Change a version in the Cargo.lock, run tests. Then run

```
 git diff --name-only | grep -q Cargo.lock
```

And see that exit code is not success.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
         - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack)
         - ./test/setup_contracts.sh
       script:
-        - cargo build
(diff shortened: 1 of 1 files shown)


fleupold:
>Any reason you prefer this to cargo build --locked?

Ah cool, I tried `cargo update --locked` and a few commands like this but didn't see the wood for the trees. Thanks!

Moements:
_ __ _

PR #420, opened 2020-01-09
Use HashMap for Solution.prices

### LINKED ISSUE #412: Update Solution Struct
The current solution struct 

https://github.com/gnosis/dex-services/blob/2a442f847dfe716b83a5784528910a0ad6abb60d/dfusion_rust_core/src/models/solution.rs#L8-L13

Has a vector of prices, which is unnecessary if we know beforehand that some tokenIds will have a price of zero (i.e. those tokens which are not traded). Thus, it would be more reasonable to have the 
 `prices` attribute as a `HashMap<u16, u128>`. We may also want to include an attribute `max_non_zero_token`

We may also want to consider (simultaneously) adapting the `Fee` struct since the fee token is always 0.

https://github.com/gnosis/dex-services/blob/2a442f847dfe716b83a5784528910a0ad6abb60d/driver/src/price_finding/price_finder_interface.rs#L6-L10

bh2smith:
As described in #412, it will be more efficient to encode prices for solution data in a `HashMap` with inferred values of zero for non-existing elements. This PR introduces the new Solution structure and adjusts all components of the project which touch it.

Since many files were touched I have briefly summarized the important changes:

1. Models:

- Solution 
    - Prices became `HashMap<u16, u128>`
    - Serializable and Deserializable (to and from bytes) both had to be reworked.

2. Driver:

- StableX contract
    - encode_prices_for_contract takes the hashmap of prices and returns the two list  representation of the non-zero, non-fee token prices.

3. Price Finding:

- CombinatorialSolver

    - Introduce (and test) `serialize_tokens` - a helper method for `LinearOptimisationPriceFinder.find_prices` which returns ordered collection of unique tokens occurring in the order batch

- NaiveSolver
    - Update find_prices to adhere to the new price map.

TODO:

- [X] Test `serialize_token` on unordered orders.
- [X] Handle Parse Errors without unwrap in `LinearOptimisationPriceFinder.deserialize_result`
- [X] Include some commentary (perhaps the sentences included here in the PR summary).

Closes #412


dfusion_rust_core/src/models/solution2.rs
```diff
@@ -0,0 +1,204 @@
...
+impl Solution2 {
+    pub fn trivial(num_orders: usize) -> Self {
+        Solution2 {
+            prices: HashMap::new(),
+            executed_buy_amounts: vec![0; num_orders],
+            executed_sell_amounts: vec![0; num_orders],
+        }
+    }
+
+    // TODO - Why does this have to be a reference?
```
bh2smith: As in "Why does my IDE insist the return value be a reference here?"

nlordell: Because the HashMap owns all the keys, so unless you remove all of the keys, you have to return references to the keys instead. You should be able to return `.max().copied()` here as `u16` implements `Copy`.

fleupold: Or even just *(self.prices.keys().max())?

bh2smith: Is one better than the other? I guess I wouldn't have to dereference everywhere.

bh2smith: > Or even just *(self.prices.keys().max())?

Yes, I could do this instead of having a function, but I find the function is more readable... I could also include `*(self.prices.keys().max())` in the function... Essentially this is just dereferencing before returning.

bh2smith: Sadly, 

```
type `std::option::Option<&u16>` cannot be dereferenced
```

nlordell: The other option is `self.prices.keys().max().map(|max| *max)`, but that is pretty much what `.copied()` ends up doing.

nlordell: `.copied()` and `.cloned()` are both `Option` methods for turning an `Option<&T>` into an `Option<T>`. They both work the exact same way except they don't work for the same types. Specifically `Option::copied()` only works for `T: Copy`, so types that implement the `Copy` trait. `Option::cloned()` works for `T: Clone` which is a less restrictive trait. Note that `Copy: Clone` so all types that implement `Copy` also implement `Clone`. `Copy` is just used as a marker to indicate that a type can be cloned, and cloning is super cheap so you shouldn't feel about about using it everywhere.

bh2smith:
The previous two builds are failing on a line of code that doesn't exist in the project at all. 

I will record the main point of the message here:

```
error[E0308]: mismatched types
   --> driver/src/driver/stablex_driver.rs:343:21
    |
343 |             prices: vec![1, 2],
    |                     ^^^^^^^^^^ expected struct `std::collections::HashMap`, found struct `std::vec::Vec`
    |
    = note: expected type `std::collections::HashMap<u16, u128>`
               found type `std::vec::Vec<{integer}>`
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
error: aborting due to previous error
For more information about this error, try `rustc --explain E0308`.
error: could not compile `driver`.
warning: build failed, waiting for other jobs to finish...
error: build failed
The command "cargo test" exited with 101.
```

My only option now is to add a comment exactly at that line in the code claiming to be unbuildable.

driver/src/contracts/stablex_contract.rs
```diff
@@ -161,19 +161,21 @@ fn parse_auction_data(packed_auction_bytes: Vec<u8>, index: U256) -> (AccountSta
...
-    (prices, ordered_token_ids)
+fn encode_prices_for_contract(price_map: HashMap<u16, u128>) -> (Vec<U128>, Vec<u64>) {
+    // Representing the solution's price vector as:
+    // sorted_touched_token_ids, non_zero_prices (excluding price at token with id 0)
+    let mut token_ids: Vec<u16> = price_map
+        .keys()
+        .copied()
+        .filter(|t| *t > 0 && price_map[t] > 0)
+        .collect();
+    token_ids.sort();
```
bh2smith:
```suggestion
    token_ids.sort_unstable();
```

dfusion_rust_core/src/models/solution.rs
```diff
@@ -3,24 +3,30 @@ use super::*;
 use log::info;
 
 use byteorder::{BigEndian, ByteOrder};
+use std::collections::HashMap;
 use std::iter::once;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Solution {
-    pub prices: Vec<u128>,
+    // token_id => price
```
nlordell: nit: I would use `///` so this adds a doc-comment instead of just a regular comment here as this documentation for the struct field.

dfusion_rust_core/src/models/solution.rs
```diff
@@ -3,24 +3,30 @@ use super::*;
...
         Solution {
-            prices: vec![],
+            prices: HashMap::new(),
             executed_buy_amounts: vec![0; num_orders],
             executed_sell_amounts: vec![0; num_orders],
         }
     }
 
+    pub fn max_token(&self) -> Option<u16> {
+        self.prices.keys().max().copied()
```
nlordell: Gooooooorgeous

dfusion_rust_core/src/models/solution.rs
```diff
@@ -53,10 +66,16 @@ impl Deserializable for Solution {
         // First 2 bytes encode the length of price vector (i.e. num_tokens)
         let len_prices = BigEndian::read_u16(&bytes[0..2]);
         let volumes = bytes.split_off(2 + len_prices as usize * 12);
-        let prices = bytes[2..]
+        let price_vector: Vec<u128> = bytes[2..]
             .chunks_exact(12)
             .map(|chunk| util::read_amount(&util::get_amount_from_slice(chunk)))
             .collect();
```
nlordell: The `collect()` here should be unnecessary, especially since it being turned into an iterator right after with `price_vector.iter()`.

driver/src/price_finding/naive_solver.rs
```diff
@@ -707,19 +694,19 @@ pub mod tests {
             return Ok(());
         }
 
-        if let Some(fee_token) = fee.as_ref().map(|fee| fee.token as usize) {
-            if solution.prices[fee_token] != BASE_PRICE {
+        if let Some(fee_token) = fee.as_ref().map(|fee| fee.token) {
+            if *solution.prices.get(&fee_token).unwrap_or(&BASE_PRICE) != BASE_PRICE {
```
nlordell: Shouldn't this fail if the solution doesn't specify the price of the fee token?

dfusion_rust_core/src/models/solution.rs
```diff
@@ -3,24 +3,30 @@ use super::*;
...
+            prices: HashMap::new(),
             executed_buy_amounts: vec![0; num_orders],
             executed_sell_amounts: vec![0; num_orders],
         }
     }
 
+    pub fn max_token(&self) -> Option<u16> {
+        self.prices.keys().max().copied()
+    }
+
```
nlordell: I would add another method to `Solution` for getting the price for a token ID `pub fn price(token_id: u16) -> u128 { /*...*/ }` as I see that it is used a lot.

dfusion_rust_core/src/models/util.rs
```diff
@@ -220,3 +221,7 @@ pub fn read_amount(bytes: &[u8; 12]) -> u128 {
 
     BigEndian::read_u128(bytes.as_slice())
 }
+
+pub fn map_from_list(arr: &[(u16, u128)]) -> HashMap<u16, u128> {
```
nlordell: ubernit: use `slice` in the function name instead of `list` as you are mapping from a slice and not a list. Also, I would suggest that the function name reflects that this is not for any map, but for price map only, so something like `prices_from_slice` would be better IMO.

If you want to get fancy, you can define a generic `hash_map!` macro that works like the `vec!` one. It would be a great Rust exercise ;)

driver/src/price_finding/error.rs
```diff
@@ -36,6 +36,12 @@ impl From<&str> for PriceFindingError {
         PriceFindingError::new(error, ErrorKind::Unknown)
     }
 }
+
+impl From<(u16, u128)> for PriceFindingError {
+    fn from(_x: (u16, u128)) -> Self {
+        PriceFindingError::new("Some shit.", ErrorKind::Unknown)
+    }
+}
```
nlordell: nit: missing new-line between `impl` blocks.

driver/src/price_finding/error.rs
```diff
@@ -36,6 +36,12 @@ impl From<&str> for PriceFindingError {
         PriceFindingError::new(error, ErrorKind::Unknown)
     }
 }
+
+impl From<(u16, u128)> for PriceFindingError {
+    fn from(_x: (u16, u128)) -> Self {
+        PriceFindingError::new("Some shit.", ErrorKind::Unknown)
```
nlordell: Is this accidentally left around? Also the conversion from a price pair to a `PriceFindingError` does not seem obvious to me.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -86,25 +102,23 @@ fn deserialize_result(json: &serde_json::Value) -> Result<models::Solution, Pric
...
-                    PriceFindingError::new(
-                        &format!("Token {} not found in price map", t),
-                        ErrorKind::JsonError,
-                    )
-                })
-                .and_then(|price| price.parse::<u128>().map_err(PriceFindingError::from))
-        })
-        .collect::<Result<Vec<u128>, PriceFindingError>>()?;
+            (
+                token[5..].parse::<u16>().map_err(|_| {
```
nlordell: This may panic if `token.len() < 5`.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -271,6 +302,42 @@ pub mod tests {
...
+        let err = deserialize_result(&json).expect_err("Should fail to parse");
+        assert_eq!(err.description(), "Price value not a string");
+
+        let json = json!({
+            "prices": {
+                "tokenX": "1",
+            },
+        });
+        let err = deserialize_result(&json).expect_err("Should fail to parse");
+        assert_eq!(err.description(), "Failed to parse token id");
```
nlordell: It would be great to include the `ParseIntError` in the `PriceFindingError` so that you know why the parsing failed. Here its because `X` is an unexpected character, the next case is because of an overflow.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -341,7 +408,7 @@ pub mod tests {
     }
 
     #[test]
-    fn serialize_result_fails_if_order_buy_volume_not_parseable() {
+    fn serialize_result_fails_if_order_buy_volume_not_parsable() {
```
nlordell: Seems unrelated to the PR. Hehe, I've noticed that incorrect spelling in function names and comments really gets to you :stuck_out_tongue:

driver/src/price_finding/naive_solver.rs
```diff
@@ -761,12 +748,12 @@ pub mod tests {
             *token_conservation.entry(order.sell_token).or_insert(0) -= exec_sell_amount as i128;
         }
 
-        for j in 0..solution.prices.len() {
-            let balance = token_conservation.entry(j as u16).or_insert(0);
-            if *balance != 0 && (fee.is_none() || j as u16 != fee.as_ref().unwrap().token) {
+        for t in solution.prices.keys() {
```
nlordell:
```suggestion
        for token_id in solution.prices.keys() {
```

e2e/tests/stablex_test.rs
```diff
@@ -136,7 +136,7 @@ fn test_rinkeby() {
         let private_key: H256 = env::var("PK")
             .expect("PK env var not set")
             .parse()
-            .expect("PK not parseable");
+            .expect("PK not parsable");
```
nlordell: Seems unrelated to the PR.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Oh yes! Right thanks.

dfusion_rust_core/src/models/solution.rs
bh2smith: I have tried this in the past and gotten strange compile errors, but will try again.

driver/src/price_finding/naive_solver.rs
bh2smith: Ya, it seems wrong to me as well, but don't forget that we wrote this line together.

dfusion_rust_core/src/models/util.rs
bh2smith: Great suggestion. Would the util file still be the right location to define such a marco?

driver/src/price_finding/error.rs
bh2smith: Actually this isn't supposed to be included! Whoops!

driver/src/price_finding/error.rs
bh2smith: Ya, this was just something I was trying that didn't pan out and is not used anywhere.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Is there some sweet trick I can do to recover the `ParseIntError`? Like a string format?

driver/src/price_finding/naive_solver.rs
nlordell: Sorry! I think it maybe should unwrap to 0 if there is no price.

driver/src/price_finding/linear_optimization_price_finder.rs
nlordell: It looks like PriceFindingError already implements `impl From<std::num::ParseIntError> for PriceFindingError` so you should be able to use `?` or `.map_err(PriceFindingError::from)`.

bh2smith: Right, but then I wouldn't get the detailed string...

bh2smith: I was trying to avoid using `PriceFindingError::from` since it returns `ErroKind::unknown`

dfusion_rust_core/src/models/solution.rs
nlordell: Yeah, we could probably get rid of the `get_token_price` method on the SnappSolution.

bh2smith: Will do! What to do with all the nice doc strings?

driver/src/price_finding/linear_optimization_price_finder.rs
nlordell: Really? https://github.com/gnosis/dex-services/blob/master/driver/src/price_finding/error.rs#L29 The code here suggests its an `ErrorKind::ParseIntError`.

nlordell: If you want a JSON error you can do:
```
.map_err(|err| PriceFindingError::new(err.to_string(), ErrorKind::JsonError))
```

bh2smith: No no, I want parse Int error, but I wanted to be more specific about exactly what was trying to be parsed, so I had a string...

dfusion_rust_core/src/models/solution.rs
nlordell: You still need the `snapp_objective_value` method I think. You can move the `get_token_price` doc string for `Solution::price` :D.

dfusion_rust_core/src/models/util.rs
nlordell: It would be. May I suggest the following syntax (although, I suggest leaving this for separate PR):
```
let map: HashMap<String, i32> = hash_map! {
  "foo": 32,
  "bar": 44,
};
let prices: HashMap<u16, u128> = hash_map! {
    0: 1 * BASE_AMOUNT,
    1: 2 * BASE_AMOUNT,
};
```

driver/src/price_finding/linear_optimization_price_finder.rs
nlordell: Then may I suggest:
```
.map_err(|err| PriceFindingError::new(
    format!("error doing something: {}", err),
    ErrorKind::JsonError.
))
```

bh2smith: I believe I have already done this...

nlordell: But both errors seem to have the same message: `assert_eq!(err.description(), "Failed to parse token id");`

dfusion_rust_core/src/models/util.rs
```diff
@@ -220,3 +221,7 @@ pub fn read_amount(bytes: &[u8; 12]) -> u128 {
 
     BigEndian::read_u128(bytes.as_slice())
 }
+
+pub fn map_from_slice(arr: &[(u16, u128)]) -> HashMap<u16, u128> {
```
fleupold: nit: can this be generics?

dfusion_rust_core/src/models/solution.rs
fleupold: I still find it weird that max_tokens is optional. Shouldn't it return 0 if the map is empy?

driver/src/contracts/stablex_contract.rs
```diff
@@ -161,19 +161,21 @@ fn parse_auction_data(packed_auction_bytes: Vec<u8>, index: U256) -> (AccountSta
...
+    // sorted_touched_token_ids, non_zero_prices (excluding price at token with id 0)
+    let mut token_ids: Vec<u16> = price_map
+        .keys()
+        .copied()
+        .filter(|t| *t > 0 && price_map[t] > 0)
+        .collect();
+    token_ids.sort_unstable();
+    let prices = token_ids
+        .iter()
+        .filter(|token_id| price_map[*token_id] > 0)
```
fleupold: Haven't you already checked that above?

driver/src/contracts/stablex_contract.rs
```diff
@@ -161,19 +161,21 @@ fn parse_auction_data(packed_auction_bytes: Vec<u8>, index: U256) -> (AccountSta
...
-    let mut prices = vec![];
-    let mut ordered_token_ids = vec![];
-    for (token_id, price) in price_vector.into_iter().enumerate().skip(1) {
-        if price > 0 {
-            prices.push(U128::from(price));
-            ordered_token_ids.push(token_id as u64);
-        }
-    }
-    (prices, ordered_token_ids)
+fn encode_prices_for_contract(price_map: HashMap<u16, u128>) -> (Vec<U128>, Vec<u64>) {
```
fleupold: nit: price_map should be taken by ref?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -36,14 +37,29 @@ impl LinearOptimisationPriceFinder {
...
     format!("token{}", token)
 }
 
 fn account_id(account: H160) -> String {
     format!("{:x}", account)
 }
 
+fn serialize_tokens(orders: &[models::Order]) -> Vec<String> {
+    // Get collection of all token ids appearing in orders
+    let mut token_ids = orders.iter().map(|o| o.buy_token).collect::<Vec<u16>>();
```
fleupold: nit: why not use a HashSet to start with? HashSets should also implement the Extend trait (https://doc.rust-lang.org/src/std/collections/hash/set.rs.html#938-946)

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -36,14 +37,29 @@ impl LinearOptimisationPriceFinder {
...
+    token_ids.extend(orders.iter().map(|o| o.sell_token).collect::<Vec<u16>>());
+
+    // Remove duplicate tokens by casting to HashSet and convert back to Vec
+    let unique_token_ids: HashSet<u16> = HashSet::from_iter(token_ids.iter().cloned());
+    let mut token_vec = unique_token_ids.into_iter().collect::<Vec<u16>>();
+
+    // unstable sort has no performance loss since elements are unique
+    token_vec.sort_unstable();
```
fleupold: Is this needed? Afaik the solver doesn't care about the order.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -72,39 +88,55 @@ fn serialize_balances(state: &models::AccountState, orders: &[models::Order]) ->
...
+        "sellToken": token_id(order.sell_token),
+        "buyToken": token_id(order.buy_token),
         "sellAmount": order.sell_amount.to_string(),
         "buyAmount": order.buy_amount.to_string(),
         "ID": id //TODO this should not be needed
     })
 }
 
+fn parse_token(key: &str) -> Result<u16, PriceFindingError> {
+    if key.len() < 6 {
```
fleupold: This seems a bit hacky. Can we use `starts_with` or some other more explicit way to signal we expect the name to start with "token"

dfusion_rust_core/src/models/solution.rs
bh2smith: But that would suggest token id 0 is valid...

driver/src/contracts/stablex_contract.rs
bh2smith: Yes, I suppose I have. I actually wrote this line before the one above. I guess this was left over. Good catch.

driver/src/contracts/stablex_contract.rs
bh2smith: Sure that sounds pretty good. How do you know when to use ref and when not?

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Believe me I tried (for days)... but I will try again.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Yes! Great! I was looking for something like this.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Yes, agreed!

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Very nice. Worked like a charm!

driver/src/contracts/stablex_contract.rs
nlordell: That’s a hard one to explain. It’s to do with the semantics of the function. Here, your function just needs to read the data to produce the output so you should use a borrow `&`. If a function needs to modify its inputs (for example, imagine a function that appends to a `String` parameter) then you need a mutable borrow `&mut`. Now, if the function needs to take the input in a way that the caller can never use it again (some examples are if you want to express that the function inputs become invalid after the function, or you are moving the input into a struct or something) then you to transfer ownership, AKA move the value into the function.

Think of them as different ways to express your functions intentions on what it will do with the data:
- `&data` just look at the data
- `&mut data` modify the data
- `data` take the data away

The reason for these it allows you to write some pretty efficient code, and choosing which one of these is usually about avoiding having multiple copies of the same data.

nlordell: That’s a hard one to explain. It’s to do with the semantics of the function. Here, your function just needs to read the data to produce the output so you should use a borrow `&`. If a function needs to modify its inputs (for example, imagine a function that appends to a `String` parameter) then you need a mutable borrow `&mut`. Now, if the function needs to take the input in a way that the caller can never use it again (some examples are if you want to express that the function inputs become invalid after the function, or you are moving the input into a struct or something) then you to transfer ownership, AKA move the value into the function.

Think of them as different ways to express your functions intentions on what it will do with the data:
- `&data` just look at the data
- `&mut data` modify the data
- `data` take the data away

The reason for these it allows you to write some pretty efficient code, and choosing which one of these is usually about avoiding having multiple copies of the same data.

nlordell: That’s a hard one to explain. It’s to do with the semantics of the function. Here, your function just needs to read the data to produce the output so you should use a borrow `&`. If a function needs to modify its inputs (for example, imagine a function that appends to a `String` parameter) then you need a mutable borrow `&mut`. Now, if the function needs to take the input in a way that the caller can never use it again (some examples are if you want to express that the function inputs become invalid after the function, or you are moving the input into a struct or something) then you to transfer ownership, AKA move the value into the function.

Think of them as different ways to express your functions intentions on what it will do with the data:
- `&data` just look at the data
- `&mut data` modify the data
- `data` take the data away

The reason for these it allows you to write some pretty efficient code, and choosing which one of these is usually about avoiding having multiple copies of the same data.

dfusion_rust_core/src/models/util.rs
bh2smith: Ya sure! Great thinking! What about the suggestion to make a macro?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -72,39 +85,51 @@ fn serialize_balances(state: &models::AccountState, orders: &[models::Order]) ->
...
         "sellAmount": order.sell_amount.to_string(),
         "buyAmount": order.buy_amount.to_string(),
         "ID": id //TODO this should not be needed
     })
 }
 
+fn parse_token(key: &str) -> Result<u16, PriceFindingError> {
+    if key.starts_with("token") {
+        return key[5..].parse::<u16>().map_err(|_| {
+            PriceFindingError::new("Failed to parse token id", ErrorKind::ParseIntError)
```
nlordell:
```suggestion
            PriceFindingError::new(format!("Failed to parse token id: {}", err), ErrorKind::ParseIntError)
```
where `err` is the error being mapped (this also requires `.map_err(|_| {` -> `.map_err(|err| {`).

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -72,39 +85,51 @@ fn serialize_balances(state: &models::AccountState, orders: &[models::Order]) ->
...
+    ))
+}
+
+fn parse_price_value(value: &serde_json::Value) -> Result<u128, PriceFindingError> {
+    value
+        .as_str()
+        .ok_or_else(|| PriceFindingError::from("Price value not a string"))?
+        .parse::<u128>()
+        .map_err(|_| {
+            PriceFindingError::new("Failed to parse price string", ErrorKind::ParseIntError)
```
nlordell:
```suggestion
            PriceFindingError::new(format!("Failed to parse price string", err), ErrorKind::ParseIntError)
```

dfusion_rust_core/src/models/solution.rs
nlordell: I think `Some(0)` and `None` are slightly different here:
- `Some(0)` produces a price vector with 1 price.
- `None` produces an empty price vector.

nlordell:
Looks like the unit test needs to be updated with the new error message:

```
---- price_finding::linear_optimization_price_finder::tests::test_failed_deserialize_result stdout ----

thread 'price_finding::linear_optimization_price_finder::tests::test_failed_deserialize_result' panicked at 'assertion failed: `(left == right)`

  left: `"Token keys expected to start with \"token\""`,

 right: `"Insufficient key length 4 (expected at least 6)"`', driver/src/price_finding/linear_optimization_price_finder.rs:334:9
```

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Unfortunately, `err` doesn't exist here.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: oh!

dfusion_rust_core/src/models/util.rs
bh2smith: Done!

bh2smith:
Ok, @fleupold  and @nlordell  - I am beginning to see why it would make sense to return `Option<u128>` for `Solution.price`

Although, now that we have the HashMap, we have kinda lost some  information. when the key is not contained in the Map, we assume zero, but this could be equivalent to `TokenNotFound`. Not sure what is the best way to approach this. Before we had three cases. 
```
indexError -> TokenNotFound
notIndexError -> Price is either zero or non-zero.
```

Not we have only non-zero prices. It seems like `TokenNotFound` doesn't really exist anymore unless we want to consider absence of price as `TokenNotFound`

fleupold:
The smart contract requires a price to be supplied for each token that is traded and furthermore check that this price is at leas 10**4. Thus not finding a price for a token in the map is invalid and should cause a TokenNotFound error in my opinion.

dfusion_rust_core/src/models/solution.rs
```diff
@@ -3,24 +3,35 @@ use super::*;
...
-            prices: vec![],
+            prices: HashMap::new(),
             executed_buy_amounts: vec![0; num_orders],
             executed_sell_amounts: vec![0; num_orders],
         }
     }
 
+    /// Returns the price for a token by ID or 0 if the token was not found.
+    pub fn price(&self, token_id: u16) -> Option<u128> {
+        self.prices.get(&token_id).cloned()
```
nlordell: ubernit: I would suggest to use `copied()` for consistency with the `max_token` function and to give signal that the `Option<&T> -> Option<T>` is computationally inexpensive.

dfusion_rust_core/src/models/util.rs
```diff
@@ -220,3 +222,20 @@ pub fn read_amount(bytes: &[u8; 12]) -> u128 {
 
     BigEndian::read_u128(bytes.as_slice())
 }
+
+pub fn map_from_slice<T: Copy + Eq + Hash, U: Copy>(arr: &[(T, U)]) -> HashMap<T, U> {
```
nlordell: Boss.

driver/src/contracts/stablex_contract.rs
```diff
@@ -161,19 +161,20 @@ fn parse_auction_data(packed_auction_bytes: Vec<u8>, index: U256) -> (AccountSta
...
-            prices.push(U128::from(price));
-            ordered_token_ids.push(token_id as u64);
-        }
-    }
-    (prices, ordered_token_ids)
+fn encode_prices_for_contract(price_map: &HashMap<u16, u128>) -> (Vec<U128>, Vec<u64>) {
+    // Representing the solution's price vector as:
+    // sorted_touched_token_ids, non_zero_prices (excluding price at token with id 0)
+    let mut token_ids: Vec<u16> = price_map
+        .keys()
```
nlordell: for fun: You could write this to use `iter()` here, then you would have an iterator of `(u16, u128)` and you wouldn't have to do a `HashMap` access to get the prices. The `Vec` can be sorted with `sort_unstable_by_key`.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -36,14 +36,27 @@ impl LinearOptimisationPriceFinder {
...
 }
 
+fn serialize_tokens(orders: &[models::Order]) -> Vec<String> {
+    // Get collection of all token ids appearing in orders
+    let mut token_ids = orders.iter().map(|o| o.buy_token).collect::<HashSet<u16>>();
+    token_ids.extend(
+        orders
+            .iter()
+            .map(|o| o.sell_token)
+            .collect::<HashSet<u16>>(),
```
nlordell: nit: This extra collect is not needed, you should be able to call `extend` directly on an `Iterator`.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -271,6 +320,59 @@ pub mod tests {
...
+
+        let json = json!({
+            "prices": {
+                "token9999999999": "1",
+            },
+        });
+        let err = deserialize_result(&json).expect_err("Should fail to parse");
+        assert_eq!(
+            err.description(),
+            "Failed to parse token id: number too large to fit in target type"
```
nlordell: :+1:

PR #421, opened 2020-01-09
Do not retry solving the instance if  failed

### LINKED ISSUE #417: Do not retry solving the instance if `price_finder.find_prices` failed
If the solver was unable to solve an instance this is usually not a temporary issues. Thus it doesn't make sense to solve the same instance again.

Instead we should wait until the next auction where potentially the orders/balances have changed and thus the solver might produce a different result.

This could most easily be achieved by registering the auction in the [past auction hash set](https://github.com/gnosis/dex-services/blob/d594a6ca7d26d2bab81ca2b3c87f026aa2418165/driver/src/driver/stablex_driver.rs#L14-L15) when the solver fails [[1]](https://github.com/gnosis/dex-services/blob/d594a6ca7d26d2bab81ca2b3c87f026aa2418165/driver/src/driver/stablex_driver.rs#L39).

fleupold:
Closes #417 

The solver is running deterministically. Therefore, there is no point in calling it again with the same problem. Thus if it failed to produce a result, we may as well consider this batch as unsolvable.

I'd like to do the same for smart contract rejections. However, right now we can still have transient failures (e.g. 503 error from the full node) and thus it is not safe to rely on it. 

With an upgrade to ethcontracts we might be able to check for revert errors (and we should retry transient failures).

### Test Plan
Added a unit test.

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -36,9 +36,16 @@ impl<'a> StableXDriver<'a> {
             info!("No orders in batch {}", batch);
             Solution::trivial(0)
         } else {
-            let solution = self.price_finder.find_prices(&orders, &account_state)?;
-            info!("Computed solution: {:?}", &solution);
-            solution
+            match self.price_finder.find_prices(&orders, &account_state) {
+                Ok(solution) => {
+                    info!("Computed solution: {:?}", &solution);
+                    solution
+                }
+                Err(err) => {
+                    self.past_auctions.insert(batch);
+                    return Err(err.into());
+                }
+            }
         };
 
         let submitted = if solution.is_non_trivial() {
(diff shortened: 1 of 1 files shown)


driver/src/driver/stablex_driver.rs
```diff
@@ -271,4 +310,49 @@ mod tests {
                 .was_called_with((batch - 1, Any, Any, Any))
         ));
     }
+
+    #[test]
+    fn test_does_not_submit_solution_for_which_validation_failed() {
```
nlordell: A sneaky unrelated test case :P I'm totally cool with including this here though since it is vaguely related.

driver/src/driver/stablex_driver.rs
```diff
@@ -271,4 +310,49 @@ mod tests {
...
+            .given((orders, state))
+            .will_return(Ok(solution));
+
+        let mut driver = StableXDriver::new(&contract, &mut pf);
+        assert!(driver.run().is_err());
+        assert!(!mock_it::verify(
+            contract
+                .submit_solution
+                .was_called_with((batch - 1, Any, Any, Any))
+        ));
```
bh2smith: I was thinking there should be some other assertions here... We assert there was an error, and that submit solution was called, but we don't check that it `does_not_submit_solution_for_which_validation_failed`. Am I missing something here?

fleupold: You are missing the ! which negates the statement and this making sure submit_solution was not called.

PR #430, opened 2020-01-10
Updates solver call

### LINKED ISSUE #429: Fix deprecated solver invocation
The current way of calling the solver was deprecated by gnosis/dex-solver#65 . This should be updated here as well.

marcovc:
Closes #429.

This should be merged after the solver is bumped to a version >= 0.5.4. This should make travis happy.

@fleupold Please make sure there are no other places calling the solver. Thanks!

### DIFF
--- a/driver/src/price_finding/linear_optimization_price_finder.rs
+++ b/driver/src/price_finding/linear_optimization_price_finder.rs
@@ -185,7 +185,8 @@ fn write_input(input_file: &str, input: &serde_json::Value) -> std::io::Result<(
 
 fn run_solver(input_file: &str) -> Result<(), PriceFindingError> {
     let output = Command::new("python")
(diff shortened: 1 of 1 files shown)


driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -185,7 +185,7 @@ fn write_input(input_file: &str, input: &serde_json::Value) -> std::io::Result<(
 
 fn run_solver(input_file: &str) -> Result<(), PriceFindingError> {
     let output = Command::new("python")
-        .arg("./batchauctions/scripts/e2e/_run.py")
+        .arg(&["-m", "batchauctions.scripts.e2e._run"])
```
fleupold: either this should be `args` or two .arg calls (one with -m and the other one with the module). Cf. https://doc.rust-lang.org/std/process/struct.Command.html

PR #436, opened 2020-01-13
Refactor error handling

e00E:


e00E:
We talked about the error handling. This is how I suggest doing it. Splitting the code into smaller functions makes it simpler and we no longer store all intermediate errors.

e00E:
`.collect::<Result<PriceMap, PriceFindingError>>()?` does what we want it to: It either returns the first error or the full map in case there was no error.

e00E:
Feel free to copy paste this btw, I just made a PR so Github shows a diff.

e00E:
Having smaller functions also allows you test them individually, if you want to.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -96,28 +96,35 @@ fn serialize_order(order: &models::Order, id: &str) -> serde_json::Value {
...
+        .parse::<u16>()
+        .map_err(|_| PriceFindingError::new("Failed to parse token id", ErrorKind::JsonError))
+}
+
+fn parse_price_value(value: &serde_json::Value) -> Result<u128, PriceFindingError> {
+    value
+        .as_str()
+        .ok_or(PriceFindingError::from("Price value not a string"))?
+        .parse::<u128>()
+        .map_err(|_| PriceFindingError::new("Failed to parse price string", ErrorKind::JsonError))
```
nlordell: Maybe include the `ParseIntError` here so we have more details about the failure.

bh2smith: How does one "include" `ParseIntError`? Do you mean to simply use the `ParseIntError` instead or like format the string with `|err| format!("{}: Failed to parse price string", err.type)`?

bh2smith: I like this suggestion and will address it in the main branch! For now I will simply merge these great suggestions and continue from here! Thanks @e00E

PR #437, opened 2020-01-13
Mention postgresql dependency

e00E:
The `graph-store-postgres` crate requires libpq to be installed or compilation will fail.


bh2smith:
Feel free to merge!

PR #439, opened 2020-01-15
Patch dynamic prices

bh2smith:
Interpret null values in solution JSON as zeroes and update end to end tests to ensure these scenarios are included.


e2e/tests/stablex_test.rs
```diff
@@ -35,6 +35,12 @@ fn test_with_ganache() {
         .wait()
         .expect("Cannot get second token id");
 
+    let third_token_id = instance
+        .token_address_to_id_map(tokens[1].address())
+        .call()
+        .wait()
+        .expect("Cannot get second token id");
```
bh2smith:
```suggestion
        .expect("Cannot get third token id");
```

e2e/tests/stablex_test.rs
```diff
@@ -80,6 +86,21 @@ fn test_with_ganache() {
...
+            first_token_id,
+            third_token_id,
+            batch + 20,
+            0.into(),
+            0.into(),
+        )
+        .from(Account::Local(accounts[1], None))
+        .send()
+        .wait()
+        .expect("Cannot place first order");
```
bh2smith:
```suggestion
        .expect("Cannot place third order");
```

e2e/tests/stablex_test.rs
```diff
@@ -164,6 +185,13 @@ fn test_rinkeby() {
         .call()
         .wait()
         .expect("Cannot get second Token address");
+
+    let token_c = instance
+        .token_id_to_address_map(3)
+        .call()
+        .wait()
+        .expect("Cannot get second Token address");
```
bh2smith:
```suggestion
        .expect("Cannot get third Token address");
```

bh2smith:
@e00E  - DIdn't you have a great suggestion about how we could be using a `solutionJSON` struct that derives `Serialize` and `Deserialize` for which we could have significantly less room for error when parsing?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -106,7 +106,7 @@ fn parse_token(key: &str) -> Result<u16, PriceFindingError> {
 fn parse_price_value(value: &serde_json::Value) -> Result<u128, PriceFindingError> {
     value
         .as_str()
-        .ok_or_else(|| PriceFindingError::from("Price value not a string"))?
+        .unwrap_or("0")
```
nlordell: I think this is not what we want here. I think we should handle the different JSON value types:
```rust
match value {
    Value::String(value) => { /* parse same as before */ }
    Value::Number(value) => { /* here value is a serde_json::Number which can be converted to u128 */ }
    Value::Null => 0,
    other => return PriceFindingError::new(format!("Expected numerical value but got '{}'", other), ErrorKind::JsonError),
}
```

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -103,17 +104,29 @@ fn parse_token(key: &str) -> Result<u16, PriceFindingError> {
...
 }
 
-fn parse_price_value(value: &serde_json::Value) -> Result<u128, PriceFindingError> {
-    value
-        .as_str()
-        .ok_or_else(|| PriceFindingError::from("Price value not a string"))?
-        .parse::<u128>()
-        .map_err(|err| {
+fn parse_price_value(value: &Value) -> Result<u128, PriceFindingError> {
+    match value {
```
fleupold: nice! This is much cleaner than the version before!

e2e/tests/stablex_test.rs
```diff
@@ -80,6 +86,21 @@ fn test_with_ganache() {
         .send()
         .wait()
         .expect("Cannot place first order");
+
+    // This order will not be matched
```
fleupold: I wouldn't add this here. This is a very specific scenario that is being tested and we want to keep the overall e2e test as general as possible (and not add every regression). If this one had been properly in place it should have actually already caught this issue as rinkeby had a bunch of orders in it that would have produced a bad output json.

PR #440, opened 2020-01-15
Dynamic price vector (revisited)

bh2smith:
This addresses the reverted PR #420 which introduced a bug that is fixed in a patch PR #439.


nlordell:
Could you add #420 in the description so we know the original reverted PR?

bh2smith:
> this will only be merged once the patch PR #439 is merged into this branch.

Yes, although I am unaware of anyway to block such a merge in github.

PR #441, opened 2020-01-15
These hooks were outdated

bh2smith:
Will attempt to give these a try before requesting review.

**Test Plan:**

Run commands
```sh
cp hooks/pre-* .git/hooks/
chmod +x .git/hooks/pre-commit
chmod +x .git/hooks/pre-push
```

Then try to commit and push some bad (unlinted, unformated code).

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -8,8 +8,7 @@ PASS=true
 if [[ ("$STAGED_RS_FILES" != "") ]]; then
     printf "\nValidating Rust:\n"
 
-    cd driver
-    cargo clippy -- -D warnings
+    cargo clippy --all --all-targets -- -D warnings
 
     if [[ "$?" == 0 ]]; then
       printf "\t\033[32mcargo clippy Passed: $FILE\033[0m"
@@ -17,12 +16,24 @@ if [[ ("$STAGED_RS_FILES" != "") ]]; then
       printf "\t\033[41mcargo clippy Failed: $FILE\033[0m"
       PASS=false
     fi
-    cd ..
+
+    cargo fmt --all -- --check
+
+    if [[ "$?" == 0 ]]; then
+      printf "\t\033[32mcargo fmt Passed: $FILE\033[0m"
+    else
+      printf "\t\033[41mcargo fmt Failed: $FILE\033[0m"
+      PASS=false
+    fi
+
 
     printf "\nRust validation completed!\n"
 else
     printf "\nNo rust files touched, skipping validation\n"
 fi
+
+
+
 
 if ! $PASS; then
   printf "\033[41mCOMMIT FAILED:\033[0m Your commit contains files that should pass Linting but do not. Please fix the errors and try again.\n"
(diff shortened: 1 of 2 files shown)


hooks/pre-commit
```diff
@@ -8,22 +8,33 @@ PASS=true
...
+
 
     printf "\nRust validation completed!\n"
 else
     printf "\nNo rust files touched, skipping validation\n"
 fi
 
+
+
+
```
nlordell: nit: useless whitespace.

hooks/pre-commit
```diff
@@ -8,22 +8,33 @@ PASS=true
...
+    cargo fmt --all -- --check
+
+    if [[ "$?" == 0 ]]; then
+      printf "\t\033[32mcargo fmt Passed:\033[0m"
+    else
+      printf "\t\033[41mcargo fmt Failed: \033[0m Try running cargo fmt"
+      PASS=false
+    fi
+
```
nlordell: nit: extra newline.

hooks/pre-commit
```diff
@@ -8,22 +8,33 @@ PASS=true
...
-    cd driver
-    cargo clippy -- -D warnings
+    cargo clippy --all --all-targets -- -D warnings
 
     if [[ "$?" == 0 ]]; then
-      printf "\t\033[32mcargo clippy Passed: $FILE\033[0m"
+      printf "\t\033[32mcargo clippy Passed:\033[0m"
     else
-      printf "\t\033[41mcargo clippy Failed: $FILE\033[0m"
+      printf "\t\033[41mcargo clippy Failed: \033[0m Try running cargo clippy"
```
nlordell: ubernit: can you quote the command so that its clear what you are supposed to run. Also I would add the `--all --all-targets` so that its more clear what the exact command that failed is.

hooks/pre-commit
```diff
@@ -8,22 +8,33 @@ PASS=true
...
       PASS=false
     fi
-    cd ..
+
+    cargo fmt --all -- --check
+
+    if [[ "$?" == 0 ]]; then
+      printf "\t\033[32mcargo fmt Passed:\033[0m"
+    else
+      printf "\t\033[41mcargo fmt Failed: \033[0m Try running cargo fmt"
```
nlordell: ubernit: the command should be quoted.

hooks/pre-push
```diff
@@ -6,7 +6,6 @@ PASS=true
 
 if [[ ("$CHANGED_RS_FILES" != "") ]]; then
     echo 'Running cargo test'
-    cd driver
     cargo test
```
nlordell: should this maybe be `cargo test --all`?

bh2smith: In that case maybe we should specify what to test here. Why would all pick up the e2e tests and not without?

nlordell: `cargo test` runs only tests for the default crate members (see [here](https://github.com/gnosis/dex-services/blob/master/Cargo.toml#L8)). I think we want to run `cargo test` for the driver, listener and dfusion rust core unit tests.

PR #442, opened 2020-01-15
Enable rust e2e test for rinkeby

fleupold:
The fact that https://github.com/gnosis/dex-services/pull/420 made it into master is mainly caused by the fact that our rinkeby e2e tests were not correctly functioning.

We were matching case sensitively on "ERROR" which in a recent logging refactoring changed into "ERRO".

Instead of fixing up the legacy e2e test (which is turning out to have other problems such as not waiting for confirmations etc.) I'm replacing it with the new rust based e2e test.

I tried to see if there is a more generic way to filter errors (e.g. by looking at STDERR but that doesn't seem to be possible with docker-compose logs).

### Test Plan

- rebase this PR on top of 2a3385b6e3d9d4d713e456efe0385ac0c5413aeb
- run `docker-compose -f docker-compose.yml -f docker-compose.rin
keby.yml up stablex`
- run `cargo test -p e2e rinkeby -- --nocapture`

See e2e test failing.


nlordell:
Baller!

.travis.yml
```diff
@@ -128,7 +131,7 @@ matrix:
         - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack)
       script:
         - docker-compose -f docker-compose.yml -f docker-compose.rinkeby.yml -f docker-compose.binary.yml up -d stablex
-        - ./test/e2e-tests-stablex-rinkeby.sh
+        - cargo test -p e2e rinkeby -- --nocapture
```
bh2smith: What is the -p

Does this have something to do with private key?

nlordell: `-p` for `--package`, as in the crate to test. All the e2e tests are part of the `e2e` crate.

fleupold:
> I don't understand how one can checkout a different commit and then run the test plan.

You are right that you need to rebase this PR on top of the old commit. Updated the test plan

fleupold:
The problem is that we cannot merge this until rinkeby stops producing errors (currently there are a lot of utility overflow messages). We might have to "filter out" known errors for the time being.

.travis.yml
bh2smith: Given that the test function itself is called `test_rinkeby` and we are calling `rinkeby`, is this cargo test doing some sort of string matching?

fleupold: Yes, it's pattern matching the string with the test names, cf. https://doc.rust-lang.org/rust-by-example/cargo/test.html

Issue #284, opened 2019-11-01
Use `serde` to generate JSON serialization/deserialization code.


nlordell:
The JSON serialization in `linear_optimization_price_finder.rs` is a good candidate for this.

e00E:
@bh2smith This is exactly the thing I was suggesting to you in our conversation just now.

e00E:
I could probably do this too, if you have something else you would rather work on.

bh2smith:
> I could probably do this too, if you have something else you would rather work on.

That seems reasonable @e00E!

PR #443, opened 2020-01-16
Use serde to (de)serialize the solver input and output json

e00E:
This is less error prone and less repetitive than managing the json by
hand.

We do not change which fields are handled.

We remove code handling both strings and numbers in parse_price_value
because these fields are always strings.

We remove some assertions of a specific error descriptions in favor of
only asserting the error kind. Error descriptions are human readable.
The code should not rely on specific strings.

### DIFF
--- a/driver/src/price_finding/linear_optimization_price_finder.rs
+++ b/driver/src/price_finding/linear_optimization_price_finder.rs
@@ -5,11 +5,9 @@ use dfusion_core::models;
 
 use chrono::Utc;
 use log::{debug, error};
-use serde::Deserialize;
-use serde_json::{json, Value};
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
-use std::io::BufReader;
+use std::io::{BufReader, BufWriter, Read, Write};
 use std::process::Command;
 use web3::types::H160;
 
@@ -17,11 +15,70 @@ const RESULT_FOLDER: &str = "./results/tmp/";
 
 type PriceMap = HashMap<u16, u128>;
 
+mod solver_output {
+    use serde::Deserialize;
+    use std::collections::HashMap;
+    use std::vec::Vec;
+
+    #[derive(Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Order {
+        pub exec_sell_amount: Option<String>,
+        pub exec_buy_amount: Option<String>,
+    }
+
+    #[derive(Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Output {
+        pub orders: Vec<Order>,
+        pub prices: HashMap<String, Option<String>>,
+    }
+}
+
+mod solver_input {
+    use serde::Serialize;
+    use std::collections::HashMap;
+    use std::vec::Vec;
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Order {
+        #[serde(rename = "accountID")]
+        pub account_id: String,
+        pub sell_token: String,
+        pub buy_token: String,
+        pub sell_amount: String,
+        pub buy_amount: String,
+        #[serde(rename = "ID")]
+        pub id: String,
+    }
+
+    #[derive(Serialize)]
+    pub struct Fee {
+        pub token: String,
+        pub ratio: f64,
+    }
+
+    pub type Accounts = HashMap<String, HashMap<String, String>>;
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Input {
+        pub tokens: Vec<String>,
+        pub ref_token: String,
+        pub accounts: Accounts,
+        pub orders: Vec<Order>,
+        // TODO vk: is this needed even though its never set to anything?
+        pub prices_prev: HashMap<String, String>,
+        pub fee: Option<Fee>,
+    }
+}
+
 pub struct LinearOptimisationPriceFinder {
     // default IO methods can be replaced for unit testing
-    write_input: fn(&str, &Value) -> std::io::Result<()>,
+    write_input: fn(&str, &str) -> std::io::Result<()>,
     run_solver: fn(&str) -> Result<(), PriceFindingError>,
-    read_output: fn() -> std::io::Result<Value>,
+    read_output: fn() -> std::io::Result<String>,
     fee: Option<Fee>,
 }
 
@@ -53,8 +110,11 @@ fn serialize_tokens(orders: &[models::Order]) -> Vec<String> {
     token_ids.into_iter().map(token_id).collect::<Vec<String>>()
 }
 
-fn serialize_balances(state: &models::AccountState, orders: &[models::Order]) -> Value {
-    let mut accounts: HashMap<String, HashMap<String, String>> = HashMap::new();
+fn serialize_balances(
+    state: &models::AccountState,
+    orders: &[models::Order],
+) -> solver_input::Accounts {
+    let mut accounts = HashMap::new();
     for order in orders {
         let modify_token_balance = |token_balance: &mut HashMap<String, String>| {
             let sell_balance = state
@@ -75,17 +135,24 @@ fn serialize_balances(state: &models::AccountState, orders: &[models::Order]) ->
                 token_balance
             });
     }
-    json!(accounts)
+    accounts
 }
 
-fn serialize_order(order: &models::Order, id: &str) -> Value {
-    json!({
-        "accountID": account_id(order.account_id),
-        "sellToken": token_id(order.sell_token),
-        "buyToken": token_id(order.buy_token),
-        "sellAmount": order.sell_amount.to_string(),
-        "buyAmount": order.buy_amount.to_string(),
-        "ID": id //TODO this should not be needed
+fn serialize_order(order: &models::Order, id: &str) -> solver_input::Order {
+    solver_input::Order {
+        account_id: account_id(order.account_id),
+        sell_token: token_id(order.sell_token),
+        buy_token: token_id(order.buy_token),
+        sell_amount: order.sell_amount.to_string(),
+        buy_amount: order.buy_amount.to_string(),
+        id: id.to_owned(), //TODO this should not be needed
+    }
+}
+
+fn serialize_fee(fee: &Option<Fee>) -> Option<solver_input::Fee> {
+    fee.as_ref().map(|fee| solver_input::Fee {
+        token: token_id(fee.token),
+        ratio: fee.ratio,
     })
 }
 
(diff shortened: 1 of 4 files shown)


e00E:
Would have been nice if this also turned out to be less lines of code. Despite being longer its still a positive to use serde imo.

e00E:
#284

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -5,23 +5,80 @@ use dfusion_core::models;
...
+    pub type Accounts = HashMap<String, HashMap<String, String>>;
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Input {
+        pub tokens: Vec<String>,
+        pub ref_token: String,
+        pub accounts: Accounts,
+        pub orders: Vec<Order>,
+        // TODO vk: is this needed even though its never set to anything?
```
bh2smith: This would be a question for `dex-solver` which would make use of them to estimate prices if they were provided. @twalth3r might know if this is still needed.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -75,17 +135,24 @@ fn serialize_balances(state: &models::AccountState, orders: &[models::Order]) ->
...
+}
+
+fn serialize_order(order: &models::Order, id: &str) -> solver_input::Order {
+    solver_input::Order {
+        account_id: account_id(order.account_id),
+        sell_token: token_id(order.sell_token),
+        buy_token: token_id(order.buy_token),
+        sell_amount: order.sell_amount.to_string(),
+        buy_amount: order.buy_amount.to_string(),
+        id: id.to_owned(), //TODO this should not be needed
```
bh2smith: Also a good question for @twalth3r: Is "ID" used anywhere in the solver for any reason?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -104,68 +171,33 @@ fn parse_token(key: &str) -> Result<u16, PriceFindingError> {
...
-            })
-        }
-        Value::Null => Ok(0u128),
-        other => Err(PriceFindingError::new(
-            format!("Expected numerical string but got '{}'", other).as_ref(),
-            ErrorKind::JsonError,
-        )),
-    }
+fn parse_price(price: &Option<String>) -> Result<u128, PriceFindingError> {
+    price.as_ref().map_or(Ok(0), |price| {
```
bh2smith: Actually I was hoping we could actually leave zero prices out of the price map and adjust the logic in a couple other places to infer zero as a default when the prices are not included.

bh2smith: We would then no longer need the "TokenNotFound" error in objective function evaluation.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -238,32 +263,30 @@ fn run_solver(input_file: &str) -> Result<(), PriceFindingError> {
...
     fn test_parse_prices() {
-        let large_number: Value =
-            serde_json::from_str("340282366920938463463374607431768211457").unwrap();
-        let err = parse_price_value(&large_number).expect_err("Should fail");
-        assert_eq!(
-            err.description(),
-            "Failed to deserialize JSON Number: invalid number"
-        );
+        // 2**128 should not fit into a u128 (max value is 2**128-1)
+        let err = parse_price(&Some("340282366920938463463374607431768211457".to_owned()))
```
bh2smith: Right, 2**128 is already too big. I was using 2*128 + 1, but we could decrement this string by 1 and make the last character a "6"

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -313,7 +336,7 @@ pub mod tests {
         let json = json!({
             "prices": {
                 "token0": "14024052566155238000",
-                "token1": 1_526_784_674_855_762_300u128,
+                "token1": "170141183460469231731687303715884105728", // greater than max value of u64
```
bh2smith: Good idea!

driver/src/price_finding/linear_optimization_price_finder.rs
e00E: At this point I wanted to do everything as before - just with serde. Lets change the  behavior for missing prices in a separate PR.

driver/src/price_finding/linear_optimization_price_finder.rs
twalth3r: We currently don't use previous prices, because we currently don't get any price info from the services. Not sure if this will always stay the case though..

driver/src/price_finding/linear_optimization_price_finder.rs
twalth3r: The solver accesses orders through an ID. Mainly because it does stuff with them, e.g., aggregate/modify them, and needs a way to keep track of them throughout the process.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Yes, that is totally fine, but maybe we can create a new issue out of this.

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: Right, so in this case, we should leave it possible as a TODO - for implementation in the future.

fleupold: @twalth3r can we completely omit the field when it's empty or does the solver expect it to be there (even when empty)

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -5,23 +5,80 @@ use dfusion_core::models;
...
+        // TODO vk: is this needed even though its never set to anything?
+        pub prices_prev: HashMap<String, String>,
+        pub fee: Option<Fee>,
+    }
+}
+
 pub struct LinearOptimisationPriceFinder {
     // default IO methods can be replaced for unit testing
-    write_input: fn(&str, &Value) -> std::io::Result<()>,
+    write_input: fn(&str, &str) -> std::io::Result<()>,
```
fleupold: nit: would it make sense to use the strict types here and in read_output or does this make the unit tests harder?

driver/src/price_finding/linear_optimization_price_finder.rs
bh2smith: I would also advocate for using the id field for the actual order Id stored on the EVM. This is an index of the individual users orders and wouldn't necessarily be unique.

fleupold: > This is an index of the individual users orders and wouldn't necessarily be unique

This is why you can't use it. But you could make it unique by concatenating with address.

driver/src/price_finding/linear_optimization_price_finder.rs
twalth3r: Currently it is expected, but it would necessary to change 3 lines to make it optional - should I do this?

driver/src/price_finding/linear_optimization_price_finder.rs
twalth3r: The solver uses the ID if it is there, otherwise generates a hash from the order data.

driver/src/price_finding/linear_optimization_price_finder.rs
e00E: These functions are not really needed at the moment. The only time we use this is to check whether the `Fee` struct is correctly serialized (in which case `str` makes sense) and we test the individual `serialize_...` functions already.
I'll keep it like this for now and we can change that later if we want to.

driver/src/price_finding/linear_optimization_price_finder.rs
e00E: Does not matter. The reason I added the comment is that I was wondering what the number is. No reason to specifically test the lowest erroring number.

driver/src/price_finding/linear_optimization_price_finder.rs
fleupold: Yes please.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -5,23 +5,80 @@ use dfusion_core::models;
...
-use std::io::BufReader;
+use std::io::{BufReader, BufWriter, Read, Write};
 use std::process::Command;
 use web3::types::H160;
 
 const RESULT_FOLDER: &str = "./results/tmp/";
 
 type PriceMap = HashMap<u16, u128>;
 
+mod solver_output {
```
nlordell: Love the inline modules!

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -5,23 +5,80 @@ use dfusion_core::models;
...
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Order {
+        #[serde(rename = "accountID")]
+        pub account_id: String,
+        pub sell_token: String,
+        pub buy_token: String,
+        pub sell_amount: String,
+        pub buy_amount: String,
```
nlordell: Would love for these to have non-`String` types. Maybe a `TokenId(u16)` for token IDs (that implements `(De)Serialize`).

driver/src/price_finding/linear_optimization_price_finder.rs
twalth3r: Change has been merged: https://github.com/gnosis/dex-solver/pull/91

driver/src/price_finding/linear_optimization_price_finder.rs
e00E: I have tried this. It works but I like keeping these types what they really are in the json without extra converting logic because that makes it more obvious how the json looks. We can still revisit this in the future.

e00E:
Looks this fails because the `pricesPrev` change in dex-solver is not yet applied. Do I need to do something so that it uses the newest solver?

bh2smith:
> Looks this fails because the pricesPrev change in dex-solver is not yet applied. Do I need to do something so that it uses the newest solver?

I believe this might involve a version bump in our dependencies. @twalth3r  might know if there is a more recent release of the solver to use... Otherwise, I would save this change for a separate PR (if you want to get this thing landed).

twalth3r:
@bh2smith  @e00E  -- I tagged the current master of the solver (which includes the pricesPrev update) to v0.5.6, so dependencies here can be updated accordingly!

fleupold:
Here is a pointer to the code where the version needs to be bumped: https://github.com/gnosis/dex-services/blob/master/.travis.yml#L43

Change v0.5.5 -> v0.5.6

PR #444, opened 2020-01-16
[Prometheus] Implement a basic metrics server

fleupold:
This PR implements a basic Server to serve the Prometheus metrics we are going to collect to the central Prometheus server that is in charge of pulling them from all hosts.

Probably wouldn't technically require an entire thread of its own but since everything else is running on the main thread seams ok.

In the next PR I will actually add metrics

### Testplan

run stableX and see metrics being served under http://localhost:9586/

### DIFF
--- a/driver/src/metrics/metrics_server.rs
+++ b/driver/src/metrics/metrics_server.rs
@@ -1,0 +1,42 @@
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+
+use hyper::service::{make_service_fn, service_fn};
+use hyper::{Body, Error, Response, Server};
+use prometheus::{Encoder, Registry, TextEncoder};
+
+pub struct MetricsServer {
+    registry: Arc<Registry>,
+}
+
+impl MetricsServer {
+    pub fn new(registry: Arc<Registry>) -> Self {
+        Self { registry }
+    }
+    pub fn serve(&self, port: u16) -> Pin<Box<dyn Future<Output = Result<(), Error>> + Send>> {
+        let addr = ([0, 0, 0, 0], port).into();
+        let registry = self.registry.clone();
+        let service = make_service_fn(move |_| {
+            let registry = registry.clone();
+            async move {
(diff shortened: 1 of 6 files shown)


driver/src/metrics/metrics_server.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl MetricsServer {
+    pub fn new(registry: Arc<Registry>) -> Self {
+        Self { registry }
+    }
+    pub fn serve(&self, port: u16) -> Pin<Box<dyn Future<Output = Result<(), Error>> + Send>> {
+        let addr = ([0, 0, 0, 0], port).into();
+        let registry = self.registry.clone();
+        let service = make_service_fn(move |_| {
+            let registry = registry.clone();
```
fleupold: Is there any way to get around all these registry clones?

fleupold:
> Looks clean and nice, but I must admit I don't know what this "MetricServer" means or does for the project.

Nothing right now. But it will expose certain metrics we collect (e.g. the times it takes to process a batch, number of errors, suceesses etc.) to a central db which we can then query and alert on. This should allow us to get more tailored alerts and not be overloaded by all these logging statements at the moment.

driver/src/metrics/metrics_server.rs
e00E: I have tried this a bit. Afaict we would need to implement hyper's `Service` trait manually but its not worth the complexity for something where a small change in performance does not matter.

e00E:
We could avoid the whole async, tokio, hyper because we don't actually need async.
After looking at rust http servers I like https://crates.io/crates/rouille . It is easier to use, has fewer dependencies and does enough for our use case.

A minimal example:

```rust
fn main() {
    rouille::start_server("localhost:8000", move |_| {
        rouille::Response::text("hello world")
    })
}
```

To use the buffer from TexteEncoder: https://docs.rs/rouille/3.0.0/rouille/struct.Response.html#method.from_data

Issue #419, opened 2020-01-09
Add Prometheus Metrics


fleupold:
In order to get better alerting and metrics, we should use proper metrics logging (e.g. prometheus) directly in the dex-services.

Some metrics we could be logging
- number of successful/failed price_finding cycles
- objective value reported
- number of orders in a batch
- number of different tokens in a batch
- number of touched orders in a solution
- prices set per auction
- time between beginning of run loop to submitting the solution (and how far into the 4 minute solution finding period that is)

[This package](https://github.com/tikv/rust-prometheus) looks like a good rust implementation of a prometheus client.

@luarx has apparently experience with Prometheus on the dev ops side and could be our point of contact.

fleupold:
My suggestion would be to implement a StableXMetrics trait with the following interface:

- auction_processing_started(Result<batch, error>)
- auction_orders_fetched(Result<orders, error>)
- auction_solution_computed(Result<solution, error>)
- auction_solution_verified(Result<objective_value, error>)
- auction_solution_submitted(Result<tx_hash, error>)
- auction_skipped(reason) // e.g. trivial vs. too_late

Implementations of that trait would be responsible for doing all logging (e.g. to console and into prometheus metrics).

This would allow us to gather the metrics mentioned above. In particular it would allow us to measure the exact time passed between all sub-steps, which can be used for performance optimization of the flow.

Regarding the [types of the metrics](https://prometheus.io/docs/concepts/metric_types/#gauge):
- Number of cycles should probably be a counter
- Information about the batches (number of orders/tokens) could be a gauge as they shouldn't change dramatically between auctions
- Information about the solution (number of touched orders/number of prices/objective value) could be a histogram as it's likely to be very different from auction to auction
- timings should probably be histograms

@e00E what do you think?

e00E:
> My suggestion would be to implement a StableXMetrics trait with the following interface:

It is hard to visualize how this would work out in the code, its something we have to try out. A source of error could be that the functions have to be called in the correct order and that data from the previous call is needed: to calculate how long `solution_verified` took you need to have remembered the timestamp of `solution_computed`. It might not be a problem in practice.

Keep in mind that Prometheus is not logging, its metrics that are usually `counter`s that are accumulated. For example, in a webserver with Prometheus you cannot retrieve the handling time per individual request.
[...]

PR #447, opened 2020-01-17
Recording Metrics for StableX

fleupold:
Associated with #419 

This PR actually adds the metrics we are interested in. It hooks into the stablex driver to report relevant events such as starting auction processing (wrt. the time the batch was "released" for settlement), order fetching, price finding, etc.

### Test Plan

Run service, open http://localhost:9586/ and see metrics such as

```
# HELP failures number of auctions failed
# TYPE failures counter
failures{stage="start"} 18
# HELP orders number of orders in a batch
# TYPE orders counter
orders{type="orders"} 0
# HELP success number of auctions successfully processed
# TYPE success counter
success{type="skipped"} 1
# HELP tokens number of distinct tokens in a batch
# TYPE tokens counter
tokens{type="orders"} 0
# HELP users number of distinct users in a batch
# TYPE users counter
users{type="orders"} 0
```

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -1,5 +1,6 @@
 use crate::contracts::stablex_contract::StableXContract;
 use crate::error::DriverError;
+use crate::metrics::StableXMetrics;
 use crate::price_finding::PriceFinding;
 
 use dfusion_core::models::Solution;
@@ -14,35 +15,56 @@ pub struct StableXDriver<'a> {
     past_auctions: HashSet<U256>,
     contract: &'a dyn StableXContract,
     price_finder: &'a mut dyn PriceFinding,
+    metrics: StableXMetrics,
 }
 
 impl<'a> StableXDriver<'a> {
-    pub fn new(contract: &'a dyn StableXContract, price_finder: &'a mut dyn PriceFinding) -> Self {
+    pub fn new(
+        contract: &'a dyn StableXContract,
+        price_finder: &'a mut dyn PriceFinding,
+        metrics: StableXMetrics,
+    ) -> Self {
         StableXDriver {
             past_auctions: HashSet::new(),
             contract,
             price_finder,
+            metrics,
         }
     }
 
     pub fn run(&mut self) -> Result<bool, DriverError> {
         // Try to process previous batch auction
-        let batch = self.contract.get_current_auction_index()? - 1;
-        if self.past_auctions.contains(&batch) {
+        let batch_to_solve_result = self
+            .contract
+            .get_current_auction_index()
+            .map(|batch_collecting_orders| batch_collecting_orders - 1);
+        self.metrics
+            .auction_processing_started(&batch_to_solve_result);
+        let batch_to_solve = batch_to_solve_result?;
+
+        if self.past_auctions.contains(&batch_to_solve) {
             return Ok(false);
         }
-        let (account_state, orders) = self.contract.get_auction_data(batch)?;
+
+        let get_auction_data_result = self.contract.get_auction_data(batch_to_solve);
+        self.metrics
+            .auction_orders_fetched(batch_to_solve, &get_auction_data_result);
+        let (account_state, orders) = get_auction_data_result?;
+
         let solution = if orders.is_empty() {
-            info!("No orders in batch {}", batch);
+            info!("No orders in batch {}", batch_to_solve);
             Solution::trivial(0)
         } else {
-            match self.price_finder.find_prices(&orders, &account_state) {
+            let price_finder_result = self.price_finder.find_prices(&orders, &account_state);
+            self.metrics
+                .auction_solution_computed(batch_to_solve, &orders, &price_finder_result);
+            match price_finder_result {
                 Ok(solution) => {
                     info!("Computed solution: {:?}", &solution);
                     solution
                 }
                 Err(err) => {
-                    self.past_auctions.insert(batch);
+                    self.past_auctions.insert(batch_to_solve);
                     return Err(err.into());
                 }
             }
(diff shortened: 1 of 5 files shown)


driver/src/metrics/stablex_metrics.rs
```diff
@@ -0,0 +1,195 @@
...
+
+fn users_from_orders(orders: &Vec<Order>) -> i64 {
+    orders
+        .iter()
+        .map(|order| order.account_id)
+        .collect::<HashSet<_>>()
+        .len()
+        .try_into()
+        .unwrap_or(std::i64::MAX)
+}
```
bh2smith: I would think that these belong in the `models/orders.rs` and that tokens from orders can be used in the place where we construct the JSON instance to be passed to the solver.

e00E: This calculates the number of tokens.
In `linear_optimization_price_finder.rs` it needs the full `HashSet` and turns it into a `Vec<String>`.
The function is short and both places aren't doing the same thing so I would keep it separate.

e00E:
There is a conceptual question about how the duration metrics should work in

* `auction_processed_started`
* `auction_orders_fetched`
* `auction_solution_computed`
* ...

Currently these metrics measure the time it took from the start of the auction to the specific step, they are cumulative.  So we would get something like `5, 10, 15, ...` (seconds). Alternatively they could measure the time the step took so we would get `5, 5, 5, ...`.

To implement the alternative version, stablex_driver would calculate the durations and pass them to stablex_metrics whereas currently stablex_metrics calculates the durations based on the start time of the batch.

This might technically be equivalent in our special case because auctions takes significantly longer than the prometheus sample interval. But the alternative feels cleaner and more natural to me.

One bug I see with the current approach is that if one step for whatever reason takes really long, then the following steps will be counted as if they belonged to the next bucket. So what really is `5 290 310 315` turns into `5 290 10 15`.

driver/src/metrics/stablex_metrics.rs
```diff
@@ -0,0 +1,195 @@
...
+            .with_label_values(label)
+            .observe(time_elapsed_since_batch_start(batch));
+        self.successes.with_label_values(label).inc();
+    }
+}
+
+fn time_elapsed_since_batch_start(batch: U256) -> f64 {
+    let now = Utc::now().timestamp() as u64;
+    // A new batch is created every 5 minutes and becomes solvable one batch later
+    let batch_start = (batch.low_u64() + 1) * 300;
```
e00E: Is a batch a U256 that represents the unix time stamp in seconds or do the first 192 bits carry any meaning (are sometimes not 0)?

If its the first case then we could U256::as_u64 which panics if the value is too large.

fleupold: A batchID is defined as unixtimestamp_sec // 300 (5minutes), we really only need a few significant bits. In the [new contract](https://github.com/gnosis/dex-contracts/blob/2dfb10ea610e12109460c499c548361c6beb7ce1/contracts/EpochTokenLocker.sol#L145-L146) the data type is actually a u32

fleupold:
> There is a conceptual question about how the duration metrics should work in

I agree with your analysis.

I thought that the cumulative numbers are more interesting to e.g. alert on when we get close to the 4 minute limit until which we have to submit our solution (otherwise it's too late).
I'm not familiar with Prometheus but I was hoping we could still get e.g. the average time spent in each phase by subtracting the average values from the measurement points around them (e.g. *avg(solution finding) = avg(time until solution) - avg(time until orders fetched)*).

I also preferred not passing metric context state around (if we did, I think it should also be encapsulated in the StableXMetics implementation itself as the driver should only know that it is informing some metrics class about events but not what about the events the class is measuring).

fleupold:
I played around a bit with a local Prometheus server and running the driver against Rinkeby. 

The fact that we only get 1 datapoint roughly but not exactly every 5 minutes makes using int counters for orders/tokens and users quite weird. I cannot get a plot that smoothly shows me the amount of orders in the system, as when I set a range to less than 5 minutes I have some data-points with value 0 (when there was no solution within the last 5 minutes), and if I set a range of more than 5 minutes, I'm sometimes double counting auctions getting values ~2x the desired one (when there were two solutions in slightly less than 5 minutes). 

I think using a gauge for these metrics and just setting the target value every 5 minutes is better.

As for the timings, I'd also suggest to use a gauge as it allow the finest granularity (rather than buckets) and we can still compute `quantile_over_time` server side. Scalability shouldn't be an issue for our tiny dataset.

fleupold:
Here are some example queries on my local prometheus server:

This is the time from start processing to submitting a solution (the spike comes from waking up my computer from sleep in the middle of a batch):
Screen Shot 2020-01-22 at 21 40 35

This is the number of orders in the system (there was not a lot of activity)
Screen Shot 2020-01-22 at 21 40 58

This is the error rate (there was one incomplete message due to putting laptop to sleep): 
Screen Shot 2020-01-22 at 21 47 15

It can be smoothed to calculate the error rate over that last 10 minutes for example

driver/src/driver/stablex_driver.rs
```diff
@@ -14,35 +15,61 @@ pub struct StableXDriver<'a> {
     past_auctions: HashSet<U256>,
     contract: &'a dyn StableXContract,
     price_finder: &'a mut dyn PriceFinding,
+    metrics: StableXMetrics,
 }
 
 impl<'a> StableXDriver<'a> {
-    pub fn new(contract: &'a dyn StableXContract, price_finder: &'a mut dyn PriceFinding) -> Self {
+    pub fn new(
+        contract: &'a dyn StableXContract,
```
bh2smith: I never know when to use these `&'a` things. IS this something clippy tells you, or do you just know.

driver/src/driver/stablex_driver.rs
```diff
@@ -14,35 +15,61 @@ pub struct StableXDriver<'a> {
...
+            self.metrics
+                .auction_solution_computed(batch_to_solve, &orders, &price_finder_result);
+            match price_finder_result {
                 Ok(solution) => {
                     info!("Computed solution: {:?}", &solution);
                     solution
                 }
                 Err(err) => {
-                    self.past_auctions.insert(batch);
+                    self.past_auctions.insert(batch_to_solve);
```
bh2smith: Somewhat unrelated; Is this "past_auctions" something that is stored in memory? If so, what happens if our services go down between batches? Will it try to solve an already finished batch and fail on restart?

bh2smith:
The test plan seems a bit vague. Have not yet successfully gotten this to work. Let me retrace my attempt:

```sh
# Terminal 1:
docker-compose build stablex
docker-compose up stablex

# Terminal 2:
cd dex-contracts && truffle migrate
./test/e2e-tests-stablex-ganache.sh
```

**Errors**

stablex container failed with 
```
 thread 'main' panicked at 'attempt to subtract with overflow', driver/src/metrics/stablex_metrics.rs:167:5
```

Also, was not able to open http://localhost:9586/

The unit test failed as well: 

```
❌ Claim withdraw 
```

I must be running the service incorrectly.

driver/src/driver/stablex_driver.rs
fleupold: It will try to resolve it but likely not come up with a better solution and thus not resubmit the solution.

driver/src/driver/stablex_driver.rs
fleupold: It's the lifetime of the reference you are passing. You need to make sure that the person "owning" the object doesn't get rid of it before the person "borrowing" it goes out of scope (in other languages you might have a dangling pointer otherwise, which the rust compiler doesn't allow you to have).
It can be omitted if the compiler can automatically infer and thus verify the lifetimes are ok.

https://doc.rust-lang.org/1.9.0/book/lifetimes.html has a more detailed explanation.

Issue #331, opened 2019-12-04
Replace bash e2e tests with rust implementations


fleupold:
Our e2e tests are pushing the limits of bash programming and are hard for others to maintain. E.g. using the return value of a call in a next step of the test requires complicated and error prone grep magic.

Long term, we should move these tests to a real programming language (rust or js). We could potentially use the ethcontract-rs crate to make the interactions with the smart contract very simple.

bh2smith:
The following checklist will help track the progress:

- [X] StableX - Rinkeby
- [X] StableX - Ganache
- [] Snapp - Deposit & WIthdraw
- [] Snapp - Standing Orders
- [] Snapp - Auction

PR #452, opened 2020-01-21
[Snapp] deposit-withdraw e2e (rust)

bh2smith:
This marks first of 3 end to end refactorings; 

In brief, this PR 

1. introduces the `setup_snapp` scaffolding for end to end tests
2. Includes the rust implementation of deposit-withdraw

This PR partially fulfils #331


e2e/src/stablex.rs
```diff
@@ -50,7 +50,7 @@ pub fn setup_stablex(
     for token in &tokens {
         instance
             .add_token(token.address())
-            .gas(MAX_GAS.into())
+            .gas(MAX_GAS.into()) // Why do we specify gas here and nowhere else? wasn't this already set as a default?
```
bh2smith: @fleupold - I think this could be removed and it wouldn't change anything.

e2e/tests/stablex_test.rs
```diff
@@ -120,7 +120,7 @@ fn test_with_ganache() {
     assert_eq!(balance_after - balance_before, 999_000.into())
 }
 
-ethcontract::contract!("dex-contracts/build/contracts/BatchExchange.json");
+ethcontract::contract!("dex-contracts/build/contracts/BatchExchange.json"); // Isn't this already available in lib.rs?
```
bh2smith: @fleupold - I think this was left over from a time when this file structure was different (and could be removed).

e2e/src/snapp.rs
```diff
@@ -0,0 +1,79 @@
...
+        .call()
+        .wait()
+        .expect("Could not recover current state root")
+}
+
+//pub fn await_db_update(db: &dyn DbInterface, current_state: &H256) {
+//    wait_for_condition(|| {
+//        db.get_balances_for_state_root(current_state).is_ok()
+//    }).expect("No state change detected");
+//}
```
nlordell: nit: commented dead code.

e2e/src/stablex.rs
nlordell: I'm not sure, but I think this was added because the automatic gas estimation was not working correctly for contract deployment with Ganache. I'm not super confident that this is the reason though.

I think that if you remove it and we get the green check, than we should take it out!

e2e/tests/test_snapp_deposit_withdraw.rs
```diff
@@ -0,0 +1,152 @@
...
+fn test_deposit_and_withdraw() {
+    let (eloop, http) = Http::new("http://localhost:8545").expect("transport failed");
+    eloop.into_remote();
+    let web3 = Web3::new(http);
+    let (instance, accounts, tokens, db) = setup_snapp(&web3, 3, 3);
+
+    // Test environment values
+    let deposit_amount = 18_000_000_000_000_000_000u128;
+
+    let user_address = &accounts[2];
```
nlordell: nit: `&` is probably unnecessary here since `H160` implements `Copy`. This will also make it so you don't have to `*user_address` everywhere, and can just use `user_address`.

e2e/tests/test_snapp_deposit_withdraw.rs
```diff
@@ -0,0 +1,152 @@
...
+    let user_id = instance
+        .public_key_to_account_map(*user_address)
+        .call()
+        .wait()
+        .expect("Could not recover account id");
+    let db_account_id = H160::from_low_u64_be(user_id);
```
nlordell: This part is little confusing, it seams that we are using addresses as account indices for Snapp?

**edit**: After talking offline to @bh2smith it looks like this is because we are sharing `AccountState` between StableX and Snapp and they keep track of accounts differently, and this was the interim solution.

**edit 2**: There is also a TODO futher down in the code.

e2e/tests/test_snapp_deposit_withdraw.rs
```diff
@@ -0,0 +1,152 @@
...
+    instance
+        .request_withdrawal(token_id.into(), U128::from(deposit_amount))
+        .from(Account::Local(*user_address, None))
+        .send()
+        .wait()
+        .expect("Failed to request withdraw");
+
+    wait_for(&web3, 181);
+    let expected_withdraw_hash =
+        H256::from_str("7b738197bfe79b6d394499b0cac0186cdc2f65ae2239f2e9e3c698709c80cb67").unwrap();
```
nlordell: Should this maybe be calculated somehow eventually?

e2e/src/stablex.rs
bh2smith: Notice that this is set at the beginning of the file (which is not included in the diff)

Here: https://github.com/gnosis/dex-services/blob/0d2a4ed486ba852f3a29d2e682ab3d5f1210008d/e2e/src/stablex.rs#L23

e2e/tests/test_snapp_deposit_withdraw.rs
bh2smith: I originally tried using copy here, but "it" wouldn't allow me to... Actually turns out my IDE has been lying to me!

e2e/tests/test_snapp_deposit_withdraw.rs
bh2smith: I moved the TODO up to here.

e2e/tests/test_snapp_deposit_withdraw.rs
bh2smith: It is calculated by the driver. We are expecting this to be the state has and this is (thus) meant to be a hardcoded value.

.travis.yml
```diff
@@ -72,8 +72,46 @@ matrix:
...
+        - docker pull $REGISTRY_URI:$TRAVIS_COMMIT
+        - docker tag $REGISTRY_URI:$TRAVIS_COMMIT stablex-binary
+        - (cd dex-contracts && yarn --frozen-lockfile)
+        - ./test/setup_contracts.sh
+      script:
+        - docker-compose -f docker-compose.yml -f docker-compose.binary.yml up -d driver graph-listener
+        - cargo test -p e2e --test test_snapp_deposit_withdraw -- --nocapture
+      after_script:
+        - docker-compose logs
+      deploy:
```
nlordell: Just realized that this `deploy` is probably no needed for dFusion e2e tests.

e2e/src/stablex.rs
fleupold: yes, feel free to remove.

e2e/src/common.rs
```diff
@@ -57,7 +57,7 @@ pub fn create_accounts_with_funded_tokens(
     let accounts: Vec<H160> =
         web3.eth().accounts().wait().expect("get accounts failed")[..num_users].to_vec();
 
-    let tokens: Vec<IERC20> = (1..num_tokens)
+    let tokens: Vec<IERC20> = (0..num_tokens)
```
fleupold: Good catch! The 0th token used to be the OWL token. That's why I started at 1. But I think it got moved during refactorings. Your change looks right.

e2e/tests/stablex_test.rs
fleupold: 👍

e2e/tests/snapp_tests.rs
```diff
@@ -0,0 +1,147 @@
+use e2e::common::{wait_for, wait_for_condition, FutureWaitExt};
+use e2e::snapp::{await_state_transition, setup_snapp};
+use ethcontract::web3::api::Web3;
+use ethcontract::web3::transports::Http;
+use ethcontract::web3::types::{H160, H256, U128, U256};
+use ethcontract::Account;
+use std::str::FromStr;
+
+#[test]
+fn snapp_deposit_withdraw() {
```
fleupold: Maybe we can use the snapshotting function to reset the ganache server back to the state before the test.

.travis.yml
```diff
@@ -89,20 +118,27 @@ matrix:
...
         - docker-compose -f docker-compose.yml -f docker-compose.binary.yml up -d driver graph-listener
-        - ./test/e2e-tests-deposit-withdraw.sh
-        - ./test/restart-containers.sh
         - ./test/e2e-tests-auction.sh
         - ./test/restart-containers.sh
         - ./test/e2e-tests-standing-order.sh
       after_failure:
         - docker-compose logs
-    - language: generic
+    - language: rust
```
fleupold: We should clean up the travis file in the end. All this stuff can become the default.

e2e/src/snapp.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub fn setup_snapp(
+    web3: &Web3<Http>,
+    num_tokens: usize,
+    num_users: usize,
+) -> (SnappAuction, Vec<H160>, Vec<IERC20>, Box<dyn DbInterface>) {
+    let graph_logger = logger(false);
+    let postgres_url = "postgresql://dfusion:let-me-in@localhost/dfusion";
+    let store_reader = GraphNodeReader::new(postgres_url.parse().unwrap(), &graph_logger);
+    let db_instance = GraphReader::new(Box::new(store_reader));
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2Frt96ekQ9jrWg%2Fgiphy.gif&token=fine" data-gitmeme-token="fine"></a>

e2e/tests/snapp_tests.rs
bh2smith: I remember this... Can check if it still works!

e2e/tests/stablex_test.rs
bh2smith: I wasn't able to remove this without the compiler complaining that `BatchExchange` does not exist.

e2e/tests/snapp_tests.rs
bh2smith: Actually, I am not sure this will be sufficient here because we also need to remove all contents from the database (i.e. restart the service anyway).

e2e/tests/stablex_test.rs
nlordell: You probably have to `use crate::BatchExchange` or something similar.

bh2smith: is this a "crate"?

nlordell: `e2e` is a crate (a Rust word for library), its where the testing utilities are. `crate` is a special keyword that means the local crate that the code is part of.

My initial suggesting was wrong, as I thought that `BatchExchange` was in the same crate as this line of code, but that is not the case.

PR #455, opened 2020-01-23
[Easy] Remove ethabi crate as it is no longer used

nlordell:
:point_up: 

### Test Plan

CI build.


nlordell:
> Seems great. Just wondering how we handle "ABIError" now.

It doesn't happen anymore. `ethcontract` handles ABI encoding and decoding, so ABI errors are bubbling up as `ethcontract::errors::MethodError` errors.

PR #456, opened 2020-01-24
[Easy] e2e StableX Ganache Test

bh2smith:
Travis was still using the old bash ganache test. This PR points Travis CI at the rust implementation of the same test and removes the related bash script. 

**Test Plan:** See that "StableX e2e Tests (Ganache)" is still passing.


bh2smith:
> As we move the e2e tests to rust, I wonder if it makes sense running them in one job rather than in three separate ones like we do now.

Yes, I would agree here. So then we will go with 

1. e2e-rinkeby
2. e2e-ganache: This will include StableX-ganache and Snapp tests

PR #459, opened 2020-01-27
Introduce New Trait to Make Builders Waitable.

nlordell:
This allow `ethcontract` builders to produce default futures and `.wait()` and `.expect()` them directly.

### Test Plan

CI


nlordell:
Yeah it should work without any adapting, but we would probably want to adapt the new e2e test to use the `wait_and_expect`.

PR #464, opened 2020-01-29
[Trivial] Bump solver to v0.5.8

nlordell:
:point_up:


nlordell:
The power of Vim

fleupold:
>The power of Vim

Do you have a shell alias that checks out a new branch, replaces the line and creates a PR?

nlordell:
No, but now I think I might write something to do something like that...

PR #465, opened 2020-01-29
Decoupling the StableX contract and driver

fleupold:
This PR adds a layer of abstraction between the StableX driver and StableX contract. In particular, it separates the potential interactions in reading (OrderBookReading trait) and writing (SolutionSubmitting trait).

This allows us to implement more sophisticated logic for the two (e.g. an order book reader based on the graph instead of the contracts, a paginated reader, a reader that is using a blacklist of users or tokens, etc.) and decouples the driver from the smart contract.

In the next step I will implement a FilterinOrderBookReader that reads a blacklist from an environment variable.

Most of the added lines is boilerplate code. In particular I will look into replacing our mocking framework into something less verbose.

### Test Plan

Unit test.

### DIFF
--- a/driver/src/orderbook/mod.rs
+++ b/driver/src/orderbook/mod.rs
@@ -8,11 +8,11 @@ use web3::types::U256;
 type Result<T> = std::result::Result<T, DriverError>;
 
 pub trait StableXOrderBookReading {
-    /// Retunrs the index of the auction that is currently being solved
+    /// Returns the index of the auction that is currently being solved
     /// or an error in case it cannot get this information.
     fn get_auction_index(&self) -> Result<U256>;
 
-    /// Returns the current state of the order book, inclducing account balances
+    /// Returns the current state of the order book, including account balances
     /// and open orders or an error in case it cannot get this information.
     ///
     /// # Arguments
--- a/driver/src/orderbook/mod.rs
+++ b/driver/src/orderbook/mod.rs
@@ -1,0 +1,82 @@
+use crate::contracts::stablex_contract::StableXContract;
+use crate::error::DriverError;
+
+use dfusion_core::models::{AccountState, Order};
+
+use web3::types::U256;
+
+type Result<T> = std::result::Result<T, DriverError>;
+
+pub trait StableXOrderBookReading {
+    /// Retunrs the index of the auction that is currently being solved
+    /// or an error in case it cannot get this information.
+    fn get_auction_index(&self) -> Result<U256>;
+
+    /// Returns the current state of the order book, inclducing account balances
+    /// and open orders or an error in case it cannot get this information.
+    ///
+    /// # Arguments
+    /// * `index` - the auction index for which returned orders should be valid
+    fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)>;
+}
+
+/// A simple implementation of the order book trait that forwards calls
+/// directly to the smart contract.
+pub struct StableXOrderBookReader<'a> {
+    contract: &'a dyn StableXContract,
+}
+
+impl<'a> StableXOrderBookReader<'a> {
+    pub fn new(contract: &'a dyn StableXContract) -> Self {
+        Self { contract }
+    }
+}
+
+impl<'a> StableXOrderBookReading for StableXOrderBookReader<'a> {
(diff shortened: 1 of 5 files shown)


driver/src/orderbook/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::error::DriverError;
+
+use dfusion_core::models::{AccountState, Order};
+
+use web3::types::U256;
+
+type Result<T> = std::result::Result<T, DriverError>;
+
+pub trait StableXOrderBookReading {
+    /// Retunrs the index of the auction that is currently being solved
```
bh2smith:
```suggestion
    /// Returns the index of the auction that is currently being solved
```

driver/src/orderbook/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// or an error in case it cannot get this information.
+    fn get_auction_index(&self) -> Result<U256>;
+
+    /// Returns the current state of the order book, inclducing account balances
```
bh2smith:
```suggestion
    /// Returns the current state of the order book, including account balances
```

driver/src/solution_submission/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+type Result<T> = std::result::Result<T, DriverError>;
+
+pub trait StableXSolutionSubmitting {
+    /// Return the objective value for the given solution in the given
+    /// batch or an error.
+    ///
+    /// # Arguments
+    /// - `batch_index` - the auction for which this solutions should be evaluated
+    /// - `orders` - the list of orders for which this solution is applicable
+    /// - `solution` - the solution to be evaluated
```
bh2smith: Is this a JS doc string adaptation?

driver/src/orderbook/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// and open orders or an error in case it cannot get this information.
+    ///
+    /// # Arguments
+    /// * `index` - the auction index for which returned orders should be valid
```
nlordell: ubernit: the notation is a bit inconsistent, `* name - description` here and `- name - description` in other places.

driver/src/solution_submission/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
+    fn get_solution_objective_value(
+        &self,
+        batch_index: U256,
+        orders: Vec<Order>,
+        solution: Solution,
+    ) -> Result<U256> {
+        self.contract
+            .get_solution_objective_value(batch_index, orders, solution)
```
nlordell: Maybe the `get_solution_objective_value` implementation should move here, and call the contract's `submit_solution` directly?

driver/src/solution_submission/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    fn submit_solution(
+        &self,
+        batch_index: U256,
+        orders: Vec<Order>,
+        solution: Solution,
+        claimed_objective_value: U256,
+    ) -> Result<()> {
+        self.contract
+            .submit_solution(batch_index, orders, solution, claimed_objective_value)
```
nlordell: Same here.

driver/src/solution_submission/mod.rs
fleupold: Yes, for this PR I wanted to keep the factoring as is. Theoretically we could also minify the trait and just expose the submit solution method and have the fact that it's first verified and then sent hidden to the driver.

driver/src/solution_submission/mod.rs
fleupold: I tried to follow https://doc.rust-lang.org/rust-by-example/meta/doc.html

driver/src/solution_submission/mod.rs
nlordell: Sounds like a great plan

bh2smith:
Maybe merge master into here to fetch the changes you fix with the wait statement earlier today! It appear the build is never gonna finish.

bh2smith:
How did you trigger build without changing any files?!

fleupold:
> How did you trigger build without changing any files?!

`git commit --allow-empty`

PR #458, opened 2020-01-27
[Snapp-e2e] refactor auction test

bh2smith:
Rewriting `e2e-test-auction.sh` in native rust and removing bash script test.

Test Plan:

Make sure that your builds are up to date. From within the project source run:
```
docker-compose down && docker-compose up driver graph-listener truffle
# Wait for the driver to recognize the deployed contract
cargo test -p e2e snapp_auction -- --nocapture
```

You should expect to see the following logs

```sh
Acquired contract instance 0x9561…06a7
Depositing sufficient funds for trades
    deposit(tokenId=2, amount=300000000000000000000, { from: 0x90f8…c9c1 })
    deposit(tokenId=1, amount=300000000000000000000, { from: 0xffcf…09f0 })
    deposit(tokenId=2, amount=300000000000000000000, { from: 0x22d4…e32b })
    deposit(tokenId=1, amount=300000000000000000000, { from: 0xe11b…882d })
    deposit(tokenId=0, amount=300000000000000000000, { from: 0xd03e…b117 })
    deposit(tokenId=0, amount=300000000000000000000, { from: 0x95ce…6ebc })
Post Deposit State 0x601b0806ca63ec70e8ee746926b30e4d764463b205e2bd6fb605ea46b3c1ce00
Placing 6 orders in current auction
    placeOrder(buyToken=1, sellToken=2, buyAmount=12000000000000000000, sellAmount=12000000000000000000, { from: 0x90f8…c9c1 })
    placeOrder(buyToken=2, sellToken=1, buyAmount=2200000000000000000, sellAmount=2000000000000000000, { from: 0xffcf…09f0 })
    placeOrder(buyToken=0, sellToken=2, buyAmount=150000000000000000000, sellAmount=10000000000000000000, { from: 0x22d4…e32b })
    placeOrder(buyToken=0, sellToken=1, buyAmount=180000000000000000000, sellAmount=15000000000000000000, { from: 0xe11b…882d })
    placeOrder(buyToken=1, sellToken=0, buyAmount=4000000000000000000, sellAmount=52000000000000000000, { from: 0xd03e…b117 })
    placeOrder(buyToken=2, sellToken=0, buyAmount=20000000000000000000, sellAmount=280000000000000000000, { from: 0x95ce…6ebc })
Awaiting order inclusion in DB
Advancing time and waiting for bid in auction
Advancing time for auction settlement and awaiting state transition
[...]

PR #468, opened 2020-01-29
[Snapp-e2e] Refactor standing order test

### LINKED ISSUE #331: Replace bash e2e tests with rust implementations

bh2smith:
This test is the final of a three part refactoring (cf. #452 and #458). Closes #331 

**Changes Includes**

- [X] Introduce new test `snapp_standing_order`
- [X] Introduce `AuctionBid`Model for cleaner assertions
- [X] Add helper method `await_and_fetch_auction_bid`
- [X] Delete unneeded bash scripts (`utils` and the test itself)
- [X] Move `restart-containers` and `setup-contracts` into `./e2e`
- [X] Delete legacy e2e-test in travis script
- [X] Include README in `./e2e`

**Test Plan:**

Make sure that your builds are up to date. From within the project source run:

```sh
docker-compose down && docker-compose up driver graph-listener truffle
# Wait for the driver to recognize the deployed contract
cargo test -p e2e snapp_standing -- --nocapture
```

The above test execution should result in the follow logs:

```
Acquired contract instance 0x9561…06a7
Depositing balances for test
Placing (matching) sell and standing sell orders
Ensure standing order recorded in DB
Standing Order:
Order {
    batch_information: None,
    account_id: 0x0000000000000000000000000000000000000000,
    buy_token: 1,
    sell_token: 2,
    buy_amount: 1000000000000000000,
    sell_amount: 1000000000000000000,
}
Auction 0 settled - 0x429f2bea7176595c4f7a5ca2b64fe124e3fa59ef5caa573523a41d88c3e41e2e
Ensure standing order account traded
Placing another matching sell order for standing order
Auction 1 settled - 0x429f2bea7176595c4f7a5ca2b64fe124e3fa59ef5caa573523a41d88c3e41e2e
Ensure standing order is still traded
Update standing order and await DB inclusion
Updated Standing Order:
Order {
    batch_information: None,
    account_id: 0x0000000000000000000000000000000000000000,
    buy_token: 1,
    sell_token: 2,
    buy_amount: 1000000000000000000,
    sell_amount: 2000000000000000000,
}
Canceling standing order in same batch (only cancellation is processed)
Verify with DB that standing order has been deleted
Deleted Standing Order:
Order {
    batch_information: None,
    account_id: 0x0000000000000000000000000000000000000000,
    buy_token: 0,
    sell_token: 0,
    buy_amount: 0,
    sell_amount: 0,
}
Place, yet, another matching sell order for standing order
Waiting for Auction 2 to clear
Ensure standing order was no longer traded
test snapp_standing_order ... ok
```


e2e/src/auction_bid.rs
```diff
@@ -0,0 +1,87 @@
...
+    order_hash: H256,
+    num_orders: u64,
+    creation_time_stamp: U256,
+    pub solver: H160,
+    objective_value: U256,
+    pub tentative_state: H256,
+    solution_hash: H256,
+}
+
+type PreAuctionBid = (
```
e00E: Maybe a doc comment explaining why this type is useful. When I first read this I was confused about it until I saw `AuctionBid::from(&instance.auctions(auction_index).wait_and_expect(".."))`.

bh2smith: Great idea. Will do I am still not entirely satisfied with this `AuctionBid` model yet and may even consider removing it completely.

e2e/README.md
```diff
@@ -0,0 +1,39 @@
+# Guide to Running Tests
+
+It is important to note that one must ensure that all dfusion related builds are recent (i.e. driver, graph-listener and their dependents).
```
nlordell: What do you mean "related builds are recent", are these the docker images? Maybe include instructions on how to do that?

e2e/README.md
```diff
@@ -0,0 +1,39 @@
+# Guide to Running Tests
```
nlordell: These instructions are Snapp specific, maybe incuded a sub-section so that is clear. We can add the StableX specific steps in another PR I think.

e2e/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+It is important to note that one must ensure that all dfusion related builds are recent (i.e. driver, graph-listener and their dependents). 
+
+From within the project source run:
+
+```shell script
```
nlordell: ubernit:
```suggestion
```sh
```

e2e/README.md
```diff
@@ -0,0 +1,39 @@
...
+cargo test -p e2e snapp_<testname> -- --nocapture
+```
+The nocapture used here will display log statements while the test is running to help identify where the test is at.
+
+where `<testname>` can be any of "deposit_withdraw", "auction" or "standing_order".
+
+In order to run multiple tests some containers must be restarted and the database must be removed.
+For example, the following sequence of commands (in separate/alternating terminals) will run all the tests. 
+Notice that, although the travis file uses the bash script for restarting containers, this will not work locally.
```
nlordell: I'm surprised by this, `restart-containers.sh` works for me it just runs them in detached mode.

e2e/README.md
```diff
@@ -0,0 +1,39 @@
+# Guide to Running Tests
+
+It is important to note that one must ensure that all dfusion related builds are recent (i.e. driver, graph-listener and their dependents). 
+
+From within the project source run:
+
+```shell script
+docker-compose down && docker-compose up driver graph-listener truffle
```
nlordell: we should probably use the same testing procedure as the Travis file, as that is what we test:
```
docker-compose down
./scripts/setup-contracts.sh
docker-compose up -d driver graph-listener
```

Doing it this way also makes it so there are WAY less errors in both the driver and listener.

e2e/src/auction_bid.rs
```diff
@@ -0,0 +1,87 @@
...
+    U256,
+    H160,
+    U256,
+    [u8; 32],
+    [u8; 32],
+    U256,
+    U256,
+);
+
+impl AuctionBid {
```
nlordell: nit: this should probably be `impl From<&'_ PreAuctionBid> for AcutionBid` or even `impl From<PreAuctionBid> for AcutionBid`

e2e/README.md
e00E: +1. I am unable to run the tests successfully, probably doing something wrong related to this.

e2e/src/auction_bid.rs
```diff
@@ -0,0 +1,87 @@
+use ethcontract::web3::types::{H160, H256, U256};
+
+#[derive(Debug)]
+pub struct AuctionBid {
```
fleupold: You could just have a

`pub struct AuctionBid(([u8; 32], ...))`

and then:

```rs
impl AuctionBid {
  pub fn order_hash(&self) -> H256 {
    H256::from_slice(self.0)
  }
  ...
```

That should give you a lightweight named view on top of this monster tuple.

e2e/src/auction_bid.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_auction_bid_new() {
```
fleupold: It's a bit weird to test your test helper structs. Usually a sign that the tests are over-engineered.

e2e/tests/snapp_tests.rs
```diff
@@ -266,3 +255,192 @@ fn snapp_auction() {
...
+
+    wait_for(&web3, 181);
+    await_and_fetch_auction_bid(&instance, U256::zero());
+    wait_for(&web3, 181);
+    let post_auction_state = await_state_transition(&instance, &post_deposit_state);
+    let post_auction_state_hash = H256::from_slice(&post_auction_state);
+    println!("Auction 0 settled - {:?}", post_auction_state_hash);
+
+    println!("Ensure standing order account traded");
+    // TODO - let state = await_and_fetch_new_account_state(&db, &post_auction_state_hash);
```
fleupold: That sounds like a nice helper. Can we also improve the error message in the case where we don't observe the state update telling us the last seen state (helps debugging to now where in the test we failed).

e2e/src/snapp.rs
```diff
@@ -66,3 +68,20 @@ pub fn await_state_transition(instance: &SnappAuction, current_state: &[u8]) ->
...
+    wait_for_condition(|| {
+        instance
+            .auctions(auction_index)
+            .wait_and_expect("No auction bid detected on smart contract")
+            .3
+            != H160::from_str("0000000000000000000000000000000000000000").unwrap()
+    })
+    .expect("Did not detect bid placement in auction");
+
+    AuctionBid::from(
```
fleupold: Nit: You are racing a bit here, since the final query might see a different result than the query inside the `wait_for_condition` closure (what if there is another state change). I guess capturing the result inside the closure mutable will get you in trouble with the borrow checker?

nlordell:
> Next, we need to get our travis build times down. It's taking forever at the moment.

Noticed that yesterday too : #474

e2e/src/snapp.rs
nlordell: You can make the closure mutate captured references, but then `wait_for_condition` needs to accept an `FnMut` instead of `Fn`.

e2e/README.md
bh2smith: Right yes, this would require following the main project readme guide from cloning to building. I can include a reference to this, but not sure it makes sense to include word for word twice. 

@e00E - Your issue is probably something to do with the `dex-contracts` submodule setup IN particular, you might require a couple of the following commands from the main readme. 

```
git submodule update --init
cd dex-contracts && yarn && yarn prepack && cd ..
```

You will notice that I have not include any migrations, or manually running `ganache` because these are included in the truffle container. From the test plan.

e2e/README.md
bh2smith: Ya... My IDE auto completes to this for some stupid reason that I will have to investigate.

e2e/README.md
bh2smith: I am also surprised... since the last line of restart-containers uses that one compose file that I cannot use! Right, yes, this is a MacOS vs Linux thing I believe and this is why it works on yours.

e2e/README.md
bh2smith: ~~I am sorry, but this is simply not possible for those users who are on macOS. Unless it would make sense to split the guide into OS-specific instructions (or at least include a phrase in the readme about why the test guide differs from travis. I believe this has somethign to do with the fact that Docker does not run natively in darwin, but instead inside its own container. Then there is some problem running binaries etc... @fleupold might be able to elaborate on this.~~

Oh! You are proposing that we use the Guide instructions in Travis instead of what is currently there! Really really sweet idea. I think its also much easier anyway. However, if I recall, the reason for the special (and more complicated) Travis execution was for build speed improvement. Is this right @fleupold?

e2e/src/auction_bid.rs
bh2smith: Yes, I agree and was considering simply removing the AuctionBid model completely since I only really use it to abstract away the parsing of a smart contract read. For now I will remove the test and  use the other helper function you suggested to clean thing up a bit.

e2e/tests/snapp_tests.rs
bh2smith: Yes, I assume you mean better error handling with personalized messages. as opposed to generic fixed ones. I am excited to make this change!

As for your first point, that I should implement `await_and_fetch_new_account_state`, I would have already done this but struggled with passing of the DB into the helper method. Clippy told me at first that it needed to be a boxed reference, but then wouldn't let me. Any suggestions you might have regarding the DB as a parameter of this function would be helpful.

e2e/src/snapp.rs
bh2smith: Oh my... Looks like I'm in the deep-end now! Im gonna save this part for last.

e2e/README.md
bh2smith: I would argue (at least locally) that the three lines proposed here are equivalent to 

```
docker-compose down && docker-compose up driver graph-listener truffle
```

since `truffle` container is essentially the same as `./scripts/setup-contracts.sh`

I wonder if we could make this work...

e2e/src/auction_bid.rs
bh2smith: Yes, I like this, however, clippy will insist that I declare a type for the tuple consisting of `AuctionBidData`. Would love if it were possible to declare functions for types instead of structs. Will check and see if this is possible.

e2e/src/auction_bid.rs
bh2smith: Renamed this type to `AuctionBidData`

e2e/tests/snapp_tests.rs
bh2smith: @fleupold 

I had already tried to make this helper function a reality, but am still running into some issues that I could use some help with:

Without Boxed:

```
error[E0277]: the size for values of type `(dyn dfusion_core::database::DbInterface + 'static)` cannot be known at compilation time
  --> e2e/src/snapp.rs:91:5
   |
91 |     db: dyn DbInterface,
   |     ^^ doesn't have a size known at compile-time
```

As a reference:

```
error[E0277]: the trait bound `std::boxed::Box<dyn dfusion_core::database::DbInterface>: dfusion_core::database::DbInterface` is not satisfied
   --> e2e/tests/snapp_tests.rs:328:51
    |
328 |     let state = await_and_fetch_new_account_state(&db, post_auction_state_hash);
    |                                                   ^^^ the trait `dfusion_core::database::DbInterface` is not implemented for `std::boxed::Box<dyn dfusion_core::database::DbInterface>`
    |
    = note: required for the cast to the object type `dyn dfusion_core::database::DbInterface`
```

It seems I am not able to pass around Boxes...

The only thing else I can think of is to move the db initialization to the beginning of each test and pass it to the helper as a box...

e2e/src/snapp.rs
bh2smith: Managed to make this happen with ease (and the help of @fleupold). Thanks for the suggestions fellas!

e2e/README.md
bh2smith: I think I will save this suggestion for a separate PR. Particularly I will address this in #474 when reworking the travis file (if this is ok with you).

nlordell: Sure, my only concern was that if the contracts were not deployed on a testnet before building, then you get build errors for the driver and listener.

nlordell: Sure, my only concern was that if the contracts were not deployed on a testnet before building, then you get build errors for the driver and listener.

PR #471, opened 2020-01-30
Replacing mock-it with mockall

fleupold:
If I remember correctly, when we first evaluated the available mocking frameworks in rust (using [this](https://asomers.github.io/mock_shootout/) list), there was no convenient one available for stable rust. We therefore chose [mock-it](https://github.com/nathanielsimard/mock-it), which was on of the few libraries for stable.

One drawback with mock-it is that it requires writing our own mock-implementations of all the traits we are depending on. This creates a lot of repetitive boilerplate code.

A recent re-evaluation of the landscape made me discover [mockall](https://github.com/asomers/mockall) which seems to be able to auto derive the mock implementations using macros and checks almost all green boxes  on the comparison matrix.

This PR replaces mock-it with mockall, kill a bunch of boilerplate code and rewrites the tests in the new framework.

### Test Plan

`cargo tests`


dfusion_rust_core/Cargo.toml
```diff
@@ -11,7 +11,7 @@ graph = { git = "https://github.com/graphprotocol/graph-node", tag = "v0.17.0" }
 graph-node-reader = { git = "https://github.com/gnosis/graph-node-reader", tag = "v0.17.0" }
 lazy_static = "1.4.0"
 log = "0.4.8"
-mock-it = "0.3.0"
+mockall = "0.6.0"
```
e00E: In the mockall repository they recommend to have it as a dev dependency only:

```toml
[dev-dependencies]
mockall = "0.6.0"
```

Can we do this?

fleupold: The `#[automock]` macro is also compiled in non-dev mode and causes compilation errors if mockall is inside dev dependencies. I wasn't yet able to conditionally compile the mock macro for the test config.

Do you have an idea how to do it?

fleupold: Ok, found the explanation: https://github.com/asomers/mockall/issues/77

e00E: Maybe

```
#[cfg_attr(test, automock)]
trait MyTrait {
    fn foo(&self, x: u32) -> u32;
}
```

or the Readme is not updated, because I noticed that the example uses it as a normal dependency, too.

dfusion_rust_core/Cargo.toml
```diff
@@ -11,8 +11,8 @@ graph = { git = "https://github.com/graphprotocol/graph-node", tag = "v0.17.0" }
...
 serde_derive = "1.0"
 serde = "1.0"
 sha2 = "0.8.1"
-web3 = "0.8.0"
+web3 = "0.8.0"
```
nlordell: uber^2-nit: my newline :sob:

dfusion_rust_core/src/database/mod.rs
```diff
@@ -7,6 +7,9 @@ use super::models;
 pub use error::*;
 pub use graph_reader::GraphReader;
 
+use mockall::automock;
+
+#[automock]
```
nlordell:
```suggestion
#[cfg_attr(test, automock)]
```

fleupold: Unfortunately this causes compilation errors, as we are using the MockDbInterface in unit tests of a different crate (the driver crate).

I wasn't able to tell the driver crate that when it's compiled in test mode, to also compile its dependencies in test mode. Thus I had to make this class available outside of dev mode.

Do you have a workaround for that?

dfusion_rust_core/Cargo.toml
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F26uf44OiZfrA3pPNK%2Fgiphy.gif&token=give_it_back" data-gitmeme-token="give_it_back"></a>

dfusion_rust_core/src/database/mod.rs
e00E: A solution like in the answer here https://stackoverflow.com/questions/41700543/can-we-share-test-utilites-between-crates *Use a testing-feature across all crates* might work. But it doesn't feel good to have to add an extra feature for this and always specify it when running tests. That answer is 3 years old so maybe there is a better way now.
I don't like having to add a feature for this and having to remember to use it when testing.

Alternatively https://docs.rs/mockall/0.6.0/mockall/#external-traits could work as it allows creating mocks for things not in the same crate.
I don't like this solution because it would lead to a lot of boiler plate.

Based on this the best way imo is to have the `automock` unconditional and not just as a dev-dependency in the cases where this is needed because simplicity is more important.

driver/src/contracts/snapp_contract.rs
```diff
@@ -311,249 +316,3 @@ where
         Ok(())
     }
 }
-
-#[cfg(test)]
-pub mod tests {
```
bh2smith: Really nice to see that all this is no longer necessary. You are becoming quite the "rustacean"!

driver/src/contracts/stablex_contract.rs
```diff
@@ -207,82 +212,9 @@ fn encode_execution_for_contract(
...
-                    "Unexpected call to get_auction_data",
-                    ErrorKind::Unknown,
-                ))),
-                get_solution_objective_value: Mock::new(Err(DriverError::new(
-                    "Unexpected call to get_solution_objective_value",
-                    ErrorKind::Unknown,
-                ))),
-                submit_solution: Mock::new(Err(DriverError::new(
-                    "Unexpected call to submit_solution",
-                    ErrorKind::Unknown,
```
bh2smith: I wonder if some of our previously handcrafted error strings and expectations in test will have to be adjusted to whatever comes for free with the automock.

driver/src/driver/deposit_driver.rs
```diff
@@ -296,45 +284,42 @@ mod tests {
...
-        db.get_balances_for_state_root
-            .given(state_hash)
-            .will_return(Ok(state));
+        let mut db = MockDbInterface::new();
+        db.expect_get_deposits_of_slot()
+            .with(eq(U256::one()))
+            .return_const(Ok(deposits));
+        db.expect_get_balances_for_state_root()
+            .with(eq(state_hash))
+            .return_const(Ok(state));
```
bh2smith: what is the `eq` for?

driver/src/contracts/snapp_contract.rs
```diff
@@ -58,6 +62,7 @@ where
     }
 }
 
+#[cfg_attr(test, automock)]
```
bh2smith: So, if I understand correctly, this is what generated the`MockSnappContract` that is being used in our tests now. It somehow also magically comes along with a `::default()`. This is really great.

driver/src/contracts/stablex_contract.rs
fleupold: I had to add some specific `return_const(Err..)` as not mocking a call defaults to panicking.

driver/src/driver/deposit_driver.rs
fleupold: It's a matcher that will consider this method as being called when the argument is equal to state_hash

driver/src/contracts/snapp_contract.rs
fleupold: yes, since it has a new() method without args it can also implement default

driver/src/driver/deposit_driver.rs
bh2smith: Interesting. So then if we call this function with any other state_root, the mock would treat that as some kind of Error. Makes sense thanks.

fleupold: Yes, there are other matchers such as always (they match anything) and you can even provide a matching function that can inspect the argument and return a boolean if it should match or not.

dfusion_rust_core/src/database/mod.rs
nlordell: I think a testing feature is probably the way to go here.

nlordell: Here is a gist on how this could be done without much hassle. Note that gists don't allow directories, so just `s/_/\//`:
https://gist.github.com/nlordell/12a6c46ff26d30f034b061b052bc9c2f

nlordell: Guess there is no good way around this :(

Issue #432, opened 2020-01-10
Allow dynamic blacklisting of addresses/tokens


fleupold:
In case of malicious activities (bad orders being placed, fake tokens being used) we should have the opportunity to manually blacklist orders e.g. by certain addresses or involving certain tokens.
We would filter out those orders before passing the to the solver. This is mainly a "quick" defense mechanism in case we see bad actors in the system.

This task captures to build a mechanism that allows to specify such filters either at runtime (without restarting the service) or at instantiation time (requiring a restart with a different environment).

fleupold:
Should probably also allow to ignore a single order of a user (if that's not too hard)

PR #476, opened 2020-01-31
Implementing a blacklist for the order book

fleupold:
Implements #432 

This PR adds another implementation of the orderbook that allows filtering specific orders based on a config string currently passed in via an environment variable.

It mainly allows for three types of filters
- malicious/fake tokens (ignore all orders that have a certain token ID in buy or sell token)
- malicious users (ignore all orders from users with certain addresses)
- malicious orders (ignore specific pairs of user + order id)

The driver is now wired to an orderbook filter, which itself uses a regular on chain order book reader as a source of truth. The source of the filter could theoretically be changed to a off-chain orderbook reader or a paginated reader in the long term.

### Test Plan

Added unit tests.

### DIFF
--- a/driver/src/orderbook/filtered_orderbook.rs
+++ b/driver/src/orderbook/filtered_orderbook.rs
@@ -4,13 +4,13 @@ use serde::Deserialize;
 use std::collections::{HashMap, HashSet};
 use web3::types::{H160, U256};
 
-/// Data structure to specify what type orders to filter
+/// Data structure to specify what type of orders to filter
 #[derive(Debug, Default, Deserialize, PartialEq, Eq)]
 pub struct OrderbookFilter {
     /// The token ids that should be filtered/
     tokens: HashSet<u16>,
 
-    /// User addresses and which of their orders to filter
+    /// User addresses mapped to which of their orders to filter
     users: HashMap<H160, UserOrderFilter>,
 }
 
--- a/driver/src/orderbook/filtered_orderbook.rs
+++ b/driver/src/orderbook/filtered_orderbook.rs
@@ -8,9 +8,11 @@ use web3::types::{H160, U256};
 #[derive(Debug, Default, Deserialize, PartialEq, Eq)]
 pub struct OrderbookFilter {
     /// The token ids that should be filtered/
+    #[serde(default)]
     tokens: HashSet<u16>,
 
     /// User addresses mapped to which of their orders to filter
+    #[serde(default)]
     users: HashMap<H160, UserOrderFilter>,
 }
 
--- a/driver/src/orderbook/filtered_orderbook.rs
+++ b/driver/src/orderbook/filtered_orderbook.rs
@@ -1,0 +1,160 @@
+use super::*;
+
+use serde::Deserialize;
+use std::collections::{HashMap, HashSet};
+use web3::types::{H160, U256};
+
+/// Data structure to specify what type orders to filter
+#[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+pub struct OrderbookFilter {
+    /// The token ids that should be filtered/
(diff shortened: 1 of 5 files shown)


driver/src/orderbook/filtered_orderbook.rs
```diff
@@ -0,0 +1,160 @@
...
+    use dfusion_core::models::order::test_util::create_order_for_test;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_filter_deserialization() {
+        let json = "{
+            \"tokens\": [1,2], 
+            \"users\": {
+                \"0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0A\": {\"OrderIds\": [0,1]}
+                \"0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0B\": \"All\",
```
nlordell: Fancy!

fleupold:
> is there a reason why you chose to use an environment variable instead of a file?

Not in particular. I though it is a configuration that shouldn't necessarily live in the codebase and thus would be something specific to our kubernetes setup. There it seemed easier to use an environment variable. We might have to re-evaluate if we want to change those values without having to restart the service.

driver/src/orderbook/filtered_orderbook.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use dfusion_core::models::order::test_util::create_order_for_test;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_filter_deserialization() {
+        let json = "{
```
nlordell: One minor suggestion I just though of, you can use a raw string literal here, which would make you not need any escape sequences:
```rust
let json = r#"{
  "foo": "bar"
}"#;
```

bh2smith:
> We might have to re-evaluate if we want to change those values without having to restart the service.

I would certainly be in favour of having a way of dynamically updating these configurations without restart.

Also, with regards to the ENV variables vs file discussion, I believe we could have the best of both worlds. Why not have a file `.driverconfig`, `.dfusionrc' that contains a bunch of environment (default) variables and have the driver `source` this file whenever it detects a change. way we could dynamically adjust our configuration, continue using environment variables and have stuff all in a file. I would argue that having a runtime configuration for the project is important for other developers to understand and setup up their own instance.

driver/src/orderbook/filtered_orderbook.rs
```diff
@@ -0,0 +1,160 @@
...
+    use super::*;
+    use dfusion_core::models::order::test_util::create_order_for_test;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_filter_deserialization() {
+        let json = r#"{
+            "tokens": [1,2], 
+            "users": {
+                "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0A": {"OrderIds": [0,1]},
```
bh2smith: Not sure how much it matters, but should we maybe use an account that isn't attached to any transactions.

josojo:
Great PR!

Could you add a short section in the readme for setting up the filters?

driver/src/orderbook/filtered_orderbook.rs
fleupold: What do you mean by this?

fleupold:
@bh2smith 

> Why not have a file .driverconfig ...

We already have `common.env`, `common-rinkeby.env`, etc. 


@josojo 

> Could you add a short section in the readme for setting up the filters?

Will do

driver/src/orderbook/filtered_orderbook.rs
bh2smith: Oh... Never mind. I had assumed that this was an address that had mainnet transactions.

Issue #475, opened 2020-01-31
Driver out of sync with batches.


nlordell:
As the batches are now more complicated and the solver takes the whole 3 minutes, I suspect we might have to rework how the driver is trying to solve things.

Specifically, right now we can get into a situation where the driver can run the solver multiple times per batch, and since the solver runs for about 3 minutes, the second solver will miss the solution deadline and cause solver to start late for the next batch:

Here are the logs for one such example: https://logs.gnosisdev.com/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:'2020-01-29T19:45:50.000Z',mode:absolute,to:'2020-01-29T22:59:59.999Z'))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'7ea8a550-5471-11e9-98b4-3b011d27ba56',key:kubernetes.container_name,negate:!f,params:(query:dfusion-mainnet-master,type:phrase),type:phrase,value:dfusion-mainnet-master),query:(match:(kubernetes.container_name:(query:dfusion-mainnet-master,type:phrase))))),index:'7ea8a550-5471-11e9-98b4-3b011d27ba56',interval:auto,query:(language:lucene,query:'%22driver::%22'),sort:!('@timestamp',asc))

Here the driver:
- **`19:45:50.982`** checks the batch ID
- **`19:45:51.159`** generates the solver input and starts the solver
- **`19:48:53.091`** solver finishes with a solution
- **`19:48:53.098`** solution is checked
- **`19:48:53.178`** check fails (because of a subtraction underflow - might be related to the negative objective value issues that we are seeing)
- **`19:48:58.182`** batch ID is queried for the **same batch**
- **`19:48:58.345`** the solver is started again for the **same batch**
- **`19:52:00.436`** solver finishes late, as it was started late
- **`19:52:05.743`** solver is started for the next batch, but 2 minutes late
- **`19:55:08.060`** solver finishes within the 3 minute window, but is late for the batch because it was started late.

I hope this illustrates the problem!

fleupold:
I guess #421 should be extended to also not retry smart contract failures? It would help if we can inspect the error and make sure we can retry intermittent ones and not retry permanent ones like a subtraction overflow.

nlordell:
Sounds like a good solution to me!

nlordell:
[...]

PR #477, opened 2020-01-31
Do not run solver if previous solution validation and submission failed.

nlordell:
This was causing the driver to get "out of sync" with batches. More detailed description of the issue is found here: #475 

This PR is a temporary solution to the issue, in the long run a solution where we have separate threads for each batch that gets cancelled once the batch submission time is up might be a more robust solution.

### Test Plan

New unit test.

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -111,7 +111,6 @@ impl<'a> StableXDriver<'a> {
             self.metrics.auction_skipped(batch_to_solve);
             false
         };
-        self.past_auctions.insert(batch_to_solve);
         Ok(submitted)
     }
 }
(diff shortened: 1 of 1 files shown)


driver/src/driver/stablex_driver.rs
```diff
@@ -57,23 +57,20 @@ impl<'a> StableXDriver<'a> {
             .auction_orders_fetched(batch_to_solve, &get_auction_data_result);
         let (account_state, orders) = get_auction_data_result?;
 
+        self.past_auctions.insert(batch_to_solve);
```
fleupold: Can we remove the duplicate insert in line 114?

nlordell: Oops! missed that.

bh2smith:
Great investigation skillz on this one! I like the proposed real solution as well.

PR #480, opened 2020-02-03
Making the solver configurable

josojo:
Closes: 435

Makes the solver configurable for the different solvers mip,nlp and naive.

This PR will be used for the following deployments prepared here:
https://gitlab.gnosisdev.com/devops/gnosis-staging/merge_requests/162/diffs

This PR and the https://gitlab.gnosisdev.com/devops/gnosis-staging/merge_requests/162/diffs
need to be executed simultaneously.

---

testplan(does not yet work):
Add to the commen.env:
```
OPTIMIZATION_MODEL=MIP
```

Build the solver:
```
docker-compose build --build-arg SOLVER_BASE=163030813197.dkr.ecr.us-east-1.amazonaws.com/dex-solver
:latest
```
Run an e2e tests so see the new solver at works(test will fail due to a different solution)


driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -233,9 +234,12 @@ fn write_input(input_file: &str, input: &str) -> std::io::Result<()> {
 }
 
 fn run_solver(input_file: &str) -> Result<(), PriceFindingError> {
+    let optimization_config: String =
```
fleupold: I'd prefer if this can be passed in from above (as a solver config option or the like). Reading the env var directly in this component makes it hard to unit test (env vars are considered global state).

josojo:
@fleupold 
We discussed this in the solver team and we had the impression that such a free configuration would be good. There might be many more parameters in the future, such as Minimal_Economic_Viable_orderamoutn,  time_to_finish are just two possibilities and it would be good to quickly incorporate them.

But then, of course, I can also see your point. 
Currently, I favor the less "brittle" version. Hence, I updated the code.

Also, you will find in the code this "clowny" two-way transition: "env-string"->OptimizationModel->"paramter-string-for-python-code". For both transitions, I need the match-statements. I don't know how to get around it. :( Maybe you a have hint.

driver/src/util.rs
```diff
@@ -104,9 +106,23 @@ pub fn batch_processing_state(
 
 pub fn create_price_finder(fee: Option<Fee>) -> Box<dyn PriceFinding> {
     let solver_env_var = env::var("LINEAR_OPTIMIZATION_SOLVER").unwrap_or_default();
+    let optimization_model_string: String =
```
fleupold: How about having one variable here and using naive, mip, and nlp as flags, rather than two variables (e.g. the combination LINEAR_OPTIMIZATION_SOLVER=0, OPTIMIZATION_MODEL=mip seems invalid.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -15,6 +15,12 @@ const RESULT_FOLDER: &str = "./results/tmp/";
 
 type PriceMap = HashMap<u16, u128>;
 
+#[derive(Clone, Copy)]
+pub enum OptimizationModel {
```
fleupold: Nice, I really like this struct

driver/src/util.rs
```diff
@@ -104,9 +106,23 @@ pub fn batch_processing_state(
...
+        env::var("OPTIMIZATION_MODEL").unwrap_or_else(|_| String::from("mip"));
+    let optimization_model;
+    match optimization_model_string.as_str() {
```
fleupold: This logic could be implemented on the struct directly. E.g. [`tryFrom<String>`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) or [`From<String>`](https://doc.rust-lang.org/std/convert/trait.From.html). 

Should the logic be case insensitive?

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -232,11 +240,20 @@ fn write_input(input_file: &str, input: &str) -> std::io::Result<()> {
     Ok(())
 }
 
-fn run_solver(input_file: &str) -> Result<(), PriceFindingError> {
+fn run_solver(
+    input_file: &str,
+    optimization_model: OptimizationModel,
+) -> Result<(), PriceFindingError> {
+    let optimization_model_str;
+    match optimization_model {
```
fleupold: This logic could also be a method on OptimizationModel. I'd suggest to have a `to_args() -> &[String]` on the object and implement it for each variant. This way, if we create more sophisticated flavours we just have to implement this method and it's colocated with the data definition.

In any case
```rust
let result = match optimization_model {
  OptimizationModel::MIP => &"mip"
  OptimizationModel::NLP => &"nlp"
}
```

is a bit more idiomatic.

driver/src/price_finding/linear_optimization_price_finder.rs
```diff
@@ -70,22 +76,24 @@ mod solver_input {
     }
 }
 
-pub struct LinearOptimisationPriceFinder {
+pub struct OptimisationPriceFinder {
```
fleupold: 👍

driver/src/util.rs
josojo: I tried to implement it via the From trait. But I could not get it working :(

If I uncomment the the line 110 in utils.rs, then I am getting the following error:
```
no variant or associated item named `From` found for type `price_finding::price_finder_interface::OptimizationModel` in the current scope
   --> driver/src/util.rs:111:28
    |
111 |         OptimizationModel::From(optimization_model_string.clone());
    |                            ^^^^
    |                            |
    |                            variant or associated item not found in `price_finding::price_finder_interface::OptimizationModel`
    |                            help: there is a method with a similar name (notice the capitalization): `from`
    | 
   ::: driver/src/price_finding/price_finder_interface.rs:26:1
    |
26  | pub enum OptimizationModel {
    | -------------------------- variant or associated item `From` not found here
```

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -20,6 +22,38 @@ impl Default for Fee {
...
+            "mip" => OptimizationModel::MIP,
+            "nlp" => OptimizationModel::NLP,
+            // the naive solver is the standard solver.
+            _ => OptimizationModel::NAIVE,
+        }
+    }
+}
+
+impl OptimizationModel {
+    pub fn to_args(self) -> Result<&'static str, PriceFindingError> {
```
fleupold: nit: naming should be try_to_args since it returns a Result. I'm wondering if it would even be fine to panic here and not return an error, since this would be an invalid logical state if we end up with the naive option here.

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -20,6 +22,38 @@ impl Default for Fee {
...
+        match self {
+            OptimizationModel::MIP => Ok(&"mip"),
+            OptimizationModel::NLP => Ok(&"nlp"),
+            // to_args is currently only used for the optimizations models mip and nlp
+            _ => Err(PriceFindingError::new(
```
fleupold: Small but important note:

It's usually better to spell out all options (when they are finite). This way the compiler will yell at you whenever you create a new option and force you to think about how that new option should be handled in all callsites that are important. If you use the default fallback, the compiler will not warn you and you might end up with weird run time errors later.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -232,11 +234,16 @@ fn write_input(input_file: &str, input: &str) -> std::io::Result<()> {
...
+fn run_solver(
+    input_file: &str,
+    optimization_model: OptimizationModel,
+) -> Result<(), PriceFindingError> {
+    let optimization_model_str = optimization_model.to_args()?;
     let output = Command::new("python")
         .args(&["-m", "batchauctions.scripts.e2e._run"])
         .arg(RESULT_FOLDER)
         .args(&["--jsonFile", input_file])
+        .args(&["--optModel", optimization_model_str])
```
fleupold: Going forward `to_args` could return the full arguments it needs in case the configuration cannot just be done with a single `--optModel` parameter

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -20,6 +22,38 @@ impl Default for Fee {
...
 
+#[derive(Clone, Copy, PartialEq)]
+pub enum OptimizationModel {
+    NAIVE,
+    MIP,
+    NLP,
+}
+
+impl From<String> for OptimizationModel {
+    fn from(optimization_model_string: String) -> OptimizationModel {
```
fleupold: ubernit: could be a &str to avoid unnecessary clone below.

PR #484, opened 2020-02-04
433 batched orders

### LINKED ISSUE #433: Allow batched querying of order book
Our [rinkeby deployment](https://rinkeby.etherscan.io/address/0x66b41d8177d6f7cd7316d9a06ecff5302249ee7c) of the contract seems to have exceeded the maximum capacity of order that we can query with a single view function call (cf. invoking the get_auction_elements script results in `Error: Returned values aren't valid, did it run Out of Gas?` even when running it with an extremely high gas limit).

We might locally be able to work around this issue by increasing the gas limit for view function calls to something very large, but in the long run we should make use of the batched query API that the smart contracts provide.

Unfortunately, this requires a new deployment of the smart contracts which isn't scheduled until in two weeks.

e00E:
Fixes #433 
# Summary
Currently, when reading the order book we are retrieving all orders ever placed in a single call to the smart contract. This can fail if there are too many orders even with high gas limits.
This PR changes the reading method to read orders in batches of configurable size. This is already support by the smart contract.
# Test Plan
New unit tests in `batched_auction_data_reader` that extend the previous unit tests for order reading.


driver/src/contracts/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,255 @@
+use super::stablex_auction_element::{StableXAuctionElement, AUCTION_ELEMENT_WIDTH};
+use dfusion_core::models::{AccountState, BatchInformation, Order};
+use std::collections::HashMap;
+use web3::types::{H160, U256};
+
+/// Handles reading of auction data that has been encoded with the smart
```
fleupold: Nice documentation throughout this PR!

driver/src/contracts/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /// Applies one batch of data.
+    ///
+    /// A batch can come from `getEncodedUsersPaginated` or `getEncodedOrders`.
+    /// In the latter case there is only one batch.
+    ///
+    /// Returns the number of orders in the data.
+    ///
+    /// Panics if `packed_auction_bytes` is not a multiple of
```
fleupold: ubernit: Panics if length of ...

driver/src/contracts/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,255 @@
...
+        let last_order_user = self
+            .orders
+            .last()
+            .expect("have at least one order")
+            .account_id;
+        if self.pagination.previous_page_user == last_order_user {
+            self.pagination.previous_page_user_offset += number_of_added_orders;
+        } else {
+            self.pagination.previous_page_user = last_order_user;
+            self.pagination.previous_page_user_offset = number_of_added_orders;
```
fleupold: Is this correct? I believe this should be the number of orders of this particular user. Consider a page like: `user1, user2, user2`

The offset should be 2 but I believe here it would be 1. Maybe we can reuse user_order_counts logic here?

driver/src/contracts/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,255 @@
...
+
+/// Handles reading of auction data that has been encoded with the smart
+/// contract's `encodeAuctionElement` function.
+pub struct BatchedAuctionDataReader {
+    index: U256,
+    pub account_state: AccountState,
+    /// All orders in the order they were received.
+    pub orders: Vec<Order>,
+    /// The total number of orders per user.
+    pub user_order_counts: HashMap<H160, u16>,
```
fleupold: Maybe it would make sense to only store a user_order Map<H160, Order> on the struct and turn it into a Vec<Order> on demand? Rather than storing two fields? This way we'd have less implicitly semantically coupled fields.

driver/src/contracts/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    fn apply_auction_elements_to_account_state<'a, Iter>(&mut self, auction_elements: Iter)
+    where
+        Iter: Iterator<Item = &'a StableXAuctionElement>,
+    {
+        for element in auction_elements.into_iter() {
+            self.account_state.modify_balance(
+                element.order.account_id,
+                element.order.sell_token,
+                |x| *x = element.sell_token_balance,
```
fleupold: I know this wasn't there before, but I'm thinking we should maybe panic here if the value is set and not the same as the new balance. That seems like an invalid state. What do you think?

driver/src/contracts/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,255 @@
...
+            packed_auction_bytes.len() % AUCTION_ELEMENT_WIDTH,
+            0,
+            "Each auction should be packed in {} bytes",
+            AUCTION_ELEMENT_WIDTH
+        );
+
+        // Workaround for borrow checker that would complain that the map
+        // closure borrows self as mutable while at the same time the filter
+        // closure borrows self as immutable because of using `self.index`.
+        let index = self.index;
```
fleupold: Could you capture by move in the filter condition instead?

driver/src/contracts/stablex_contract.rs
```diff
@@ -79,8 +78,27 @@ impl StableXContract for BatchExchange {
         //   request to run into the gas limit
         orders_builder.m.tx.gas = None;
         let packed_auction_bytes = orders_builder.call().wait()?;
-        let auction_data = parse_auction_data(packed_auction_bytes, index);
-        Ok(auction_data)
+        let mut reader = BatchedAuctionDataReader::new(index);
```
fleupold: I really like this abstraction you built there.

I am thinking it should live in the OrderbookReader directly instead of the contract. The contract was meant to be a thin layer providing an API for interacting with the smart contracts directly. Since we are using ethcontracts now, we might even get rid of it completely. I'd like for any logic (parsing, etc.) to live in a layer on top, that can be unit tested.

Can we change the contract API to return the raw bytes rather than a parsed result and do the parsing in the OrderBook?

driver/src/contracts/batched_auction_data_reader.rs
e00E: Agree that it would be nicer from that perspective but after trying this I think the code is overall better the way it is now because there is no extra work to produce the final order vector and parsing the orders is a bit simpler.

driver/src/contracts/batched_auction_data_reader.rs
e00E: Making the filter closure `move` does not work if that is what you mean.

fleupold: Right, because we would move self and not just the index.

driver/src/contracts/batched_auction_data_reader.rs
e00E: In the paginated version it really could change if  the state of the smart contract changes between different retrievals. Not sure how exceptional that is and if it can be handled  gracefully.

driver/src/contracts/batched_auction_data_reader.rs
nlordell: I think if we want to keep it this way, then `orders` and `user_order_counts` can't be public, or else that would allow code outside the module to mutate these fields and invalidate the `BatchedAuctionDataReader`. The issue with that is you are attaching implicit requirements to your data and no way of enforcing it with the compiler.

driver/src/contracts/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// The total number of orders per user.
+    pub user_order_counts: HashMap<H160, u16>,
+    /// Used when reading data from the smart contract in batches.
+    pub pagination: Pagination,
+}
+
+/// Data for the next call to the smart contract's `getEncodedUsersPaginated`
+/// function (which should be named `getEncodedOrdersPaginated`).
+pub struct Pagination {
+    pub previous_page_user: H160,
```
nlordell: You added doc comments everywhere else, maybe forgot these two properties?

driver/src/contracts/batched_auction_data_reader.rs
nlordell: Maybe the reader can read for a specific block? Calls can be done on a specific block number which will guarantee the data will not change during retrieval. Because new orders are only valid for the next batch, it would be OK if we miss them.

driver/src/contracts/batched_auction_data_reader.rs
nlordell: You could user `filter_map` which would only mutably borrow `self` once.

driver/src/contracts/batched_auction_data_reader.rs
e00E: Should be fine with just `user_order_counts` private. Mutating `orders` doesn't change correctness.

driver/src/contracts/batched_auction_data_reader.rs
e00E: filter_map doesn't work because it applies the filter before the map but we actually perform a side effect in the filter updating `user_order_counts`.

driver/src/contracts/batched_auction_data_reader.rs
e00E: Good idea. I have implemented this.

driver/src/contracts/batched_auction_data_reader.rs
e00E: This would work but I find filter and map clearer.

driver/src/contracts/batched_auction_data_reader.rs
e00E: So there is still the question whether we should check and panic. On the one hand if the output of the smart contract cannot be trusted then we cannot expect anything to work correctly here. On the other hand it is a simple check...

driver/src/contracts/batched_auction_data_reader.rs
fleupold: If I add/delete orders from the vec without updating the user_order_counts that's invalid isn't it? If using a Map<H160, Order> is too ugly maybe we can make the fields private and read only?

driver/src/contracts/batched_auction_data_reader.rs
fleupold: Technically, the sell_token_balance can only change when also the auctionId changed (every 5 minutes). 
Since we query those orders at the beginning of a batch this shouldn't really happen or otherwise the batch we are looking to process can no longer be solved anyways. However, it can happen if we start the service just before a batch is about to end (we should probably not attempt to process batches if they are already 4 minutes or so in).

Thus I think it's fine to panic as it can be considered invalid state.

README.md
```diff
@@ -206,6 +206,8 @@ The following environment variables can be used to configure the behavior of the
 }
 ```
 
+- *AUCTION_DATA_BATCH_SIZE*: the batch size with which to read orders from the smart contract
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FRgfGmnVvt8Pfy%2Fgiphy.gif&token=nice" data-gitmeme-token="nice"></a>

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,288 @@
+use crate::contracts::stablex_auction_element::{StableXAuctionElement, AUCTION_ELEMENT_WIDTH};
+use dfusion_core::models::{AccountState, BatchInformation, Order};
+use std::collections::HashMap;
+use web3::types::{H160, U256};
+
+/// Handles reading of auction data that has been encoded with the smart
+/// contract's `encodeAuctionElement` function.
+pub struct BatchedAuctionDataReader {
+    /// The account state resulting from unfiltered handled orders.
+    pub account_state: AccountState,
```
fleupold: Shouldn't all of these fields be read only? We don't want anyone to mutate the reader's state.

driver/src/contracts/stablex_contract.rs
```diff
@@ -72,15 +79,21 @@ impl StableXContract for BatchExchange {
...
+    fn get_auction_data_batched(
+        &self,
+        block: u64,
+        page_size: u64,
+        previous_page_user: H160,
+        previous_page_user_offset: u64,
+    ) -> Result<Vec<u8>> {
+        let mut orders_builder = self.get_encoded_users_paginated(
+            previous_page_user,
+            previous_page_user_offset as u64,
```
fleupold: nit: this is already u64, right?

driver/src/contracts/batched_auction_data_reader.rs
e00E: It doesn't change how parsing new orders works and it doesn't change the pagination logic is what I meant, but it doesn't make sense to change the orders. Really I mostly wanted to avoid writing more getters for it but I'll do that now.

nlordell: The issue is that it doesn’t make sense now and in this context to change the orders vector, it may make sense to someone unfamiliar with the code base some at a different time and place. Having these implicit requirements is an easy way to introduce bugs in the future.

driver/src/contracts/batched_auction_data_reader.rs
nlordell: If we always query with the same block number, then all queries should agree since the batch index will stay the same. Or am I misunderstanding how that works?

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,288 @@
...
+                let order_counter = self
+                    .user_order_counts
+                    .entry(result.order.account_id)
+                    .or_insert(0);
+                result.order.batch_information = Some(BatchInformation {
+                    slot_index: *order_counter as u16,
+                    slot: U256::from(0),
+                });
+                *order_counter += 1;
+                result
```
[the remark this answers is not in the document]
e00E: It would work but I prefer first map then filter because this is not really the purpose of filter_map.

e00E:
I should add another test that makes sure the block number logic works correctly.

driver/src/contracts/batched_auction_data_reader.rs
fleupold: You are right, my comment was assuming that we wouldn't pin the block number, but that's even better.

driver/src/orderbook/batched_auction_data_reader.rs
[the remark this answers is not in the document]
fleupold: I really don't feel strongly here, but what is the purpose of filter_map if not doing a map and a filter operation at the same time? 
Do you mean we should only use it if we first want to filter and then do a map?

driver/src/contracts/stablex_contract.rs
```diff
@@ -72,15 +77,21 @@ impl StableXContract for BatchExchange {
...
+            previous_page_user,
+            previous_page_user_offset,
+            page_size,
+        );
         orders_builder.m.tx.gas = None;
-        let packed_auction_bytes = orders_builder.call().wait()?;
-        let auction_data = parse_auction_data(packed_auction_bytes, index);
-        Ok(auction_data)
+        orders_builder.block = Some(web3::types::BlockNumber::Number(block));
+        orders_builder.call().wait().map_err(From::from)
```
nlordell: Any reason not to use `?` here?

driver/src/models/order.rs
```diff
@@ -1,10 +1,15 @@
...
+// TODO: refactor the struct to be an enum that can be either the stableX
+// version (slot index) or the Snapp version (slot_index + slot).
 #[derive(Debug, Clone, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
 #[serde(rename_all = "camelCase")]
 pub struct BatchInformation {
+    // only applies to the Snapp version
     pub slot: U256,
+    /// the order index of a StableX order (the n-th order a user has ever
+    /// placed in the system)
     pub slot_index: u16,
```
nlordell: This should probably be called `order_id` then right? Just curious, we can change it in a separate PR (or get rid of the `BatchInformation` struct altogether).

e00E: I ... am not sure. I took the comment from an explanation Felix gave me on slack. With the removal of the snapp as you say we can probably remove BatchInformation.

e2e/tests/batched_auction_data_reader_test.rs
```diff
@@ -0,0 +1,61 @@
+use driver::orderbook::{PaginatedStableXOrderBookReader, StableXOrderBookReading};
+use e2e::common::FutureBuilderExt;
+use e2e::stablex::setup_stablex;
+use ethcontract::web3::api::Web3;
+use ethcontract::web3::transports::Http;
+use ethcontract::Account;
+
+#[test]
+fn test_with_ganache() {
```
nlordell: To me, this could be a unit test with mocked contract responses. Alternatively, you could set `AUCTION_DATA_BATCH_SIZE=5` and setup an auction with many orders and check that it is solved.

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /// The pagination data used when reading data from the smart contract in
+    /// batches.
+    pub fn pagination(&self) -> &Pagination {
+        &self.pagination
+    }
+
+    /// Signal that no more data will be read and return the result.
+    pub fn done(self) -> Done {
```
fleupold: What's the point of defining the Done struct? Shouldn't we just return a tuple? I'd also prefer if this method was called `get_auction_data` or something a little bit more explicit. There is nothing enforcing the no more data will be read.

driver/src/models/order.rs
fleupold: Yes, we should refactor it into a single order_id now (but not in this PR).

driver/src/orderbook/mod.rs
```diff
@@ -27,26 +29,47 @@ pub trait StableXOrderBookReading {
...
         self.contract
             .get_current_auction_index()
             .map(|batch_collecting_orders| batch_collecting_orders - 1)
     }
 
     fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)> {
-        self.contract.get_auction_data(index)
+        let block = self.web3.eth().block_number().wait()?.as_u64();
+        let mut reader = BatchedAuctionDataReader::new(index);
+        loop {
```
fleupold: ubernit: Would a while loop be more elegant here?

driver/src/orderbook/batched_auction_data_reader.rs
e00E: It takes `self` by value which consumes self which means it cannot be used anymore after the call.
Struct vs Tuple is a minor difference. With structs you can name the values with tuples you have only types. I prefer structs.

driver/src/orderbook/mod.rs
e00E: Don't think so. What I really want is a do-while loop but rust does not have them.

driver/src/orderbook/batched_auction_data_reader.rs
fleupold: Nice, I didn’t spot the consumption of self (maybe this can be pointed out in the name or doc string). 

I still think the method name is too generic and would argue that the definition of the Done struct for the purpose of packing the return type into a single value adds unnecessary indirection and cognitive overhead than just returning account state and orders without defining the extra type. If you feel strongly about keeping the struct I’d ask to give it a more descriptive name.

driver/src/orderbook/mod.rs
fleupold: The loop body would be empty since the condition would do the side-effect (we’d be checking apply_batch’s return value. Thus it shouldn’t matter if it’s while or do-while, shouldn’t it?

But maybe having the side effect in the condition is a bit ugly.

e00E:
I'll leave out the unit test for `StableXOrderBookReading::get_auction_data` because that function is small and any test would consist mostly of repetitions from the unit tests of `BatchedAuctionDataReader`.

driver/src/orderbook/batched_auction_data_reader.rs
nlordell: I like the `done` method, its like finalizing a builder. I do agree with @fleupold that maybe it can return a tuple instead.

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -45,14 +46,12 @@ impl StableXAuctionElement {
...
         let (buy_amount, sell_amount) = compute_buy_sell_amounts(numerator, denominator, remaining);
-        let order_counter = order_count.entry(account_id).or_insert(0);
-        *order_counter += 1;
         StableXAuctionElement {
             valid_from,
             valid_until,
             sell_token_balance,
             order: Order {
-                id: *order_counter - 1,
+                id: 0,
```
nlordell: This seams wrong to me. Should be `*order_counter - 1`.

driver/src/orderbook/mod.rs
```diff
@@ -27,26 +29,46 @@ pub trait StableXOrderBookReading {
...
-/// A simple implementation of the order book trait that forwards calls
-/// directly to the smart contract.
-pub struct StableXOrderBookReader<'a> {
+/// Implements the StableXOrderBookReading trait by using the underlying
+/// contract in a paginated way.
+/// This avoid hitting gas limits when the total amount of orders is large.
+pub struct PaginatedStableXOrderBookReader<'a, 'b> {
     contract: &'a dyn StableXContract,
+    page_size: u64,
+    web3: &'b Web3,
```
nlordell: I don't think this needs a separate lifetime. I may be wrong though.

driver/src/contracts/stablex_auction_element.rs
e00E: I have considered this but I would like the deserialize function to just deserialize what it can from the input bytes.

bh2smith:
I would have liked to to review this PR before it was merged. Please also include me as a reviewer on such changes.

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,317 @@
...
+                    } else {
+                        assert_eq!(
+                            *x,
+                            element.sell_token_balance,
+                            "got order which sets user {}'s sell token {} \
+                            balance to {} but sell_token_balance has already \
+                            been set to {}",
+                            element.order.account_id,
+                            element.order.sell_token,
+                            element.sell_token_balance,
```
bh2smith: What is this and why are we modifying account balance? Only the smart contract should be modifying account balances.

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,317 @@
...
+                let mut result = StableXAuctionElement::from_bytes(&chunk_array);
+                let order_counter = self
+                    .user_order_counts
+                    .entry(result.order.account_id)
+                    .or_insert(0);
+                result.order.id = *order_counter as u16;
+                *order_counter += 1;
+                result
+            })
+            .filter(|x| x.in_auction(index) && x.order.sell_amount > 0)
```
bh2smith: I think technically, since the contract can only have a minimum executed amounts as 10^4, this inequality could be the same.

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,317 @@
...
+        account_state.modify_balance(H160::from_low_u64_be(1), 258, |x| *x = 5);
+
+        assert_eq!(reader.account_state, account_state);
+        assert_eq!(reader.orders, [ORDER_1.clone(), ORDER_2.clone()]);
+        assert_eq!(reader.pagination.previous_page_user, ORDER_1.account_id);
+        assert_eq!(reader.pagination.previous_page_user_offset, 2);
+    }
+
+    #[test]
+    fn batched_auction_data_reader_multiple_batches() {
```
bh2smith: What are these Batches? Is this the same as auction index? Are these batches somehow related to "pages"?

driver/src/orderbook/batched_auction_data_reader.rs
fleupold: Here is the point where you filter this reducing the amount of orders taken from the page.

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// In the latter case there is only one batch.
+    ///
+    /// Returns the number of orders in the data.
+    ///
+    /// Panics if length of `packed_auction_bytes` is not a multiple of
+    /// `AUCTION_ELEMENT_WIDTH`.
+    pub fn apply_batch(&mut self, packed_auction_bytes: &[u8]) -> usize {
+        let previous_order_count = self.orders.len();
+        self.apply_auction_data(&packed_auction_bytes);
+        let number_of_added_orders = self.orders.len() - previous_order_count;
```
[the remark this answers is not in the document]
bh2smith: Great find! That’s definitely it.

driver/src/orderbook/batched_auction_data_reader.rs
e00E: Its modifying account balances in the `AccountState` struct in order to be able return it once the order reading is done. We have been doing the same thing since before this PR, the code just moved to a different file.
I don't know why we need `AccountState` in the grand scheme things, if that is what you're asking.

PR #485, opened 2020-02-04
Improved CI

### LINKED ISSUE #474: Investigate slow CI times.
It seams that `cargo build` is taking MUCH longer than it used to. Here is a build from about a month ago where `cargo build` took around 150 seconds:

https://travis-ci.com/gnosis/dex-services/jobs/272188154#L603

The last build on master took around **550 seconds**:

https://travis-ci.com/gnosis/dex-services/jobs/281975494#L627

bh2smith:
Attempting to consolidate local tests into a single Travis "stage".

Closes #474

Test Plan: Check that all expected tests are included in travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
     - RUSTC_WRAPPER=$SCCACHE
 matrix:
   include:
-    - stage: "Warm Up"
+    - stage: "Local and Rinkeby in Parallel"
       name: "Build and Local Testing"
       cache:
         directories:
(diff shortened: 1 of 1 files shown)


bh2smith:
> We should also check if a more minimal caching might improve overall travis time, as I believe we are spending >10 minutes down- and uploading the cache.

I will have to figure out what is even being cached... Perhaps none of this is necessary.

bh2smith:
Closing in favour of #488. Which will undoubtedly reduce CI times.

fleupold:
Can we still merge this PR before we implement #488. Regardless of whether or not we run the snapp tests we might want to consolidate the pipeline into a single job.

bh2smith:
> Can we still merge this PR before we implement #488. Regardless of whether or not we run the snapp tests we might want to consolidate the pipeline into a single job.

I suppose I could reopen this and continue working on it. Just note that it wasn't passing and will still require some work.

fleupold:
>Just note that it wasn't passing and will still require some work.

I can take a look at it

fleupold:
Running everything in a single job takes 28 minutes, which is not significantly less than before but it now only uses one job (thus allowing 4 PRs to be built in parallel). Will also try removing the cache and see if that helps.

fleupold:
It's slightly fast without a cache (26minutes vs 28minutes) but a significantly simpler setup. Therefore, I'd like to kill the cache as well.

Can I have a review please?

bh2smith:
> Can I have a review please?

Only you can approve my PR.

PR #489, opened 2020-02-05
[Easy] Lift environment variables from contract helpers to main

fleupold:
Environment variables are considered global state and as such should not be used deep down inside our components. 
Instead, we should read them in the main method and panic if they are not set (currently we return an error which get swallowed by a DriverError in an upper layer).

I didn't yet touch the LINEAR_OPTIMIZATION_SOLVER variable as this would lead to merge conflicts with #480 

### Test Plan

CI


josojo:
I merged #480 just right now.

I can do the update required for LINEAR_OPTIMIZATION_SOLVER
 Once we successfully deployed the multi-solver setup manually. Would love to take this PR for testing the auto-deployment.

PR #496, opened 2020-02-06
Merge `dfusion_core` crate with driver.

nlordell:
:point_up: 

### Test Plan

CI


nlordell:
> `driver/src/models/util.rs` seems to be added as an empty file?

Oops! nice catch!

PR #497, opened 2020-02-06
Sort tokenIds and filter out 0 balances

### LINKED ISSUE #473: Sort token IDs and remove 0 balances in account state.
Request by @twalth3r, this will make the JSON instances easier to compare.

fleupold:
Closes #473 

Requested by @twalth3r .
I'm making this an implementation detail of the serialisation logic as it doesn't seem as if it has to be enforced on a higher level.

### Test Plan

Added a unit test. 

Also checked an instance files after running the mainnet solver by running:

```
docker-compose down && docker-compose -f docker-compose.yml -f docker-compose.mainnet.yml run stablex bash
cargo run --bin stablex
# After one batch is processed
cat instances/*
```

### DIFF
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -36,8 +36,8 @@ mod solver_output {
 }
 
 mod solver_input {
-    use serde::Serialize;
-    use std::collections::HashMap;
+    use serde::{Serialize, Serializer};
+    use std::collections::{BTreeMap, HashMap};
     use std::vec::Vec;
 
     #[derive(Serialize)]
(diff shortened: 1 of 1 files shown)


nlordell:
Build is failing because of a clippy warning:
```
error: writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices.
  --> driver/src/price_finding/optimization_price_finder.rs:74:33
   |
74 |     fn ordered_tokens<S>(value: &Vec<String>, serializer: S) -> Result<S::Ok, S::Error>
   |                                 ^^^^^^^^^^^^
   |
   = note: `-D clippy::ptr-arg` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
help: change this to
   |
74 |     fn ordered_tokens<S>(value: &[String], serializer: S) -> Result<S::Ok, S::Error>
   |                                 ^^^^^^^^^
help: change `value.clone()` to
   |
78 |         let mut sorted = value.to_owned();
   |
```

twalth3r:
>This is great, but it would be worth noting that dex-solver is not ordering tokens by ID, but rather by their string order. That is, for example, in a solution instance with "prices", the tokens are ordered as follows:
```
"token1",
"token10",
"token2",
etc...
```

Would it be an option to rename "token1" -> "token01" etc?

nlordell:
> This is great, but it would be worth noting that dex-solver is not ordering tokens by ID, but rather by their string order.

It look like this is also ordering by string order 

> Would it be an option to rename "token1" - "token01" etc?

The number of 0s would have to be enough to fit the maximum token ID, I don’t remember off the top of my head what that is right now.

bh2smith:
I think this would involve some hackey string manipulation in a few places throughout our  code base since all tokens are encoded by their integer IDs. I would not be in favour of doing this since it introduces unnecessary “fiddling”

PR #503, opened 2020-02-07
Convert `driver` crate into a `bin`

nlordell:
This PR turns `driver` crate into a `bin` crate instead of a `lib` crate. It also corrects all the clippy warnings that surfaced because of it.

Documention, CI and Dockerfiles were also updated.

### Test Plan

CI + :pray:


e00E:
For the batched auction reader I am writing a test in the `e2e` folder/crate that uses types from driver. If this is a bin does that mean I can no longer do that? Or maybe it means I should move that test into the driver crate (but it still needs ganache to run).

driver/src/models/account_state.rs
```diff
@@ -68,19 +30,60 @@ impl AccountState {
...
+                state_index: U256::zero(),
+                state_hash: H256::zero(),
+                balances: HashMap::new(),
+                num_tokens: std::u16::MAX,
+            };
+            for order in orders {
+                state.increment_balance(order.sell_token, order.account_id, order.sell_amount);
+            }
+            state
+        }
```
bh2smith: I like this! This is nice.

nlordell:
> For the batched auction reader I am writing a test in the `e2e` folder/crate that uses types from driver. If this is a bin does that mean I can no longer do that? Or maybe it means I should move that test into the driver crate (but it still needs ganache to run).

Is there value in making this a e2e test instead of a unit test with mocked contract responses?

I think a valid e2e test would be to set the `AUCTION_DATA_BATCH_SIZE` to a low number and then run an auction with a large number of orders (relative to `AUCTION_DATA_BATCH_SIZE`) and then assert that a solution was found.

e00E:
I think you are right. Don't let that stop this PR.

PR #508, opened 2020-02-09
Use estimated gas amount and price

### LINKED ISSUE #369: Remove Fixed Default Gas And Gas Price
Instead use `estimate_gas` and `estimate_gas_price` (or gas station) to get the values to use. `ethcontract` will automatically estimate the gas amount when sending transactions.

fleupold:
Closes #369 

With the update to ethcontract-rs 0.3.0 we can use the estimated gas amount and price (with a factor to ensure we are getting fast execution speed).

This should prevent us from drastically overpaying tx fees by using a fixed gas price of 20GWEI (the avg. solution submission costs between 5-10$ atm). We will now be using 3x the avg. gas price. At the time of testing this was about 3x6GWEI

### Test Plan

E2E tests, check submission times as soon as it's in staging.

Get average gas price (in hex) from full node:

```
curl -H "Content-type:application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_gasPrice","params":[],"id":7398897897897}' https://node.mainnet.gnosisdev.com
```

### DIFF
--- a/driver/src/contracts/mod.rs
+++ b/driver/src/contracts/mod.rs
@@ -28,8 +28,8 @@ fn method_defaults(network_id: u64) -> Result<MethodDefaults, DriverError> {
     let account = Account::Offline(key, Some(network_id));
     let defaults = MethodDefaults {
         from: Some(account),
-        gas: Some(100_000.into()),
-        gas_price: Some(1_000_000_000.into()),
+        gas: None,
+        gas_price: None,
     };
 
     Ok(defaults)
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -128,7 +128,7 @@ impl StableXContract for BatchExchange {
             prices,
             token_ids_for_price,
         )
-        .gas_price(GasPrice::high())
+        .gas_price(GasPrice::Scaled(3.0))
         .send()
         .wait()?;
 
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -3,6 +3,7 @@
 use std::collections::HashMap;
 use std::env;
 
+use ethcontract::transaction::GasPrice;
 use lazy_static::lazy_static;
 #[cfg(test)]
 use mockall::automock;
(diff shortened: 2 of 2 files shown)


bh2smith:
Could we use a smaller factor like 2? Or why not just use 1.1 with the fast gas price estimate. I have a feeling this would, on average, cost less and still yield successful submissions.

driver/src/contracts/stablex_contract.rs
```diff
@@ -128,8 +128,7 @@ impl StableXContract for BatchExchange {
             prices,
             token_ids_for_price,
         )
-        .gas(5_000_000.into())
-        .gas_price(20_000_000_000u64.into())
+        .gas_price(GasPrice::Scaled(3.0))
```
bh2smith: Safe to assume this is only possible because of the recent ethcontract update?

bh2smith: I don’t really see any reference to the word estimate here. I suppose then that current gas price means the current average

fleupold:
> Could we use a smaller factor like 2? Or why not just use 1.1 with the fast gas price estimate. I have a feeling this would, on average, cost less and still yield successful submissions.

The “fast” option in ethcontracts uses a factor of 6 times the average gas price. The more sophisticated version of asking for a “low”/“safe”/“fast”  isn’t supported by full nodes and requires more sophisticated non-standardised  queries. Cc @nlordell

driver/src/contracts/mod.rs
```diff
@@ -28,8 +28,8 @@ fn method_defaults(network_id: u64) -> Result<MethodDefaults, DriverError> {
     let account = Account::Offline(key, Some(network_id));
     let defaults = MethodDefaults {
         from: Some(account),
-        gas: Some(100_000.into()),
-        gas_price: Some(1_000_000_000.into()),
+        gas: None,
+        gas_price: None,
```
nlordell: nit:
```
    let defaults = MethodDefaults {
        from: Some(account),
        ..Default::default()
    };
```

driver/src/contracts/stablex_contract.rs
nlordell: So `GasPrice::Scaled` will scale the average gas price by a factor of 3.0. Eventually, we can integrate something like ETH Gas Station as well and just use the "fast" gas price.

PR #516, opened 2020-02-10
Bump default page size

bh2smith:
This is a better temporary fix that is easy to review, which (hopefully) buys some time while we fix the broken pagination orderbook query.


fleupold:
I think we should set this inside the Kubernetes deployment (where we define all env vars like private key, etc.). 

Here I think we should leave it a value that incurs noses can handle.

PR #519, opened 2020-02-11
Batched orders again

e00E:
This reverts the revert of #484 . There is a second commit after the revert commit which fixes the problem and adds another unit test.

I will attempt to run this locally against mainnet to verify that is picking up all orders.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ The following environment variables can be used to configure the behavior of the
 }
 ```
 
-- *AUCTION_DATA_BATCH_SIZE*: the batch size with which to read orders from the smart contract
+- *AUCTION_DATA_PAGE_SIZE*: the page size with which to read orders from the smart contract
 
 blacklists all orders that contain token 1 & 2, all orders of _0x...B_ and orderId 0 & 1 or _0x...A_
 
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ The following environment variables can be used to configure the behavior of the
 }
 ```
 
+- *AUCTION_DATA_BATCH_SIZE*: the batch size with which to read orders from the smart contract
+
 blacklists all orders that contain token 1 & 2, all orders of _0x...B_ and orderId 0 & 1 or _0x...A_
 
 ## Troubleshooting
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -7,18 +7,14 @@ use ethcontract::transaction::GasPrice;
 use lazy_static::lazy_static;
 #[cfg(test)]
 use mockall::automock;
-use web3::transports::EventLoopHandle;
 use web3::types::{H160, U128, U256};
 
 use crate::contracts;
-use crate::contracts::stablex_auction_element::StableXAuctionElement;
 use crate::error::DriverError;
-use crate::models::{AccountState, Order, Solution};
+use crate::models::{Order, Solution};
 use crate::util::FutureWaitExt;
 
 type Result<T> = std::result::Result<T, DriverError>;
-
-pub const AUCTION_ELEMENT_WIDTH: usize = 112;
 
 lazy_static! {
     // In the BatchExchange smart contract, the objective value will be multiplied by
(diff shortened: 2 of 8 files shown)


e00E:
@bh2smith Since you didn't get to review the original PR feel free to add anything else that I should address.

README.md
```diff
@@ -144,6 +144,8 @@ The following environment variables can be used to configure the behavior of the
 }
 ```
 
+- *AUCTION_DATA_BATCH_SIZE*: the batch size with which to read orders from the smart contract
```
bh2smith:
```suggestion
- *AUCTION_DATA_PAGE_SIZE*: the batch size with which to read orders from the smart contract
```

driver/src/contracts/stablex_contract.rs
```diff
@@ -51,7 +46,17 @@ impl BatchExchange {
 #[cfg_attr(test, automock)]
 pub trait StableXContract {
     fn get_current_auction_index(&self) -> Result<U256>;
-    fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)>;
+    /// Retrieve one batch of auction data.
```
bh2smith: page?

driver/src/contracts/stablex_contract.rs
```diff
@@ -51,7 +46,17 @@ impl BatchExchange {
...
+    /// `block` is needed because the state of the smart contract could change
+    /// between blocks which would make the returned auction data inconsistent
+    /// between calls.
+    fn get_auction_data_batched(
```
bh2smith: I might suggest calling this `get_paginated_auction_data` or get `get_auction_data_paginated`

driver/src/orderbook/mod.rs
```diff
@@ -27,26 +29,46 @@ pub trait StableXOrderBookReading {
...
             .map(|batch_collecting_orders| batch_collecting_orders - 1)
     }
 
     fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)> {
-        self.contract.get_auction_data(index)
+        let block = self.web3.eth().block_number().wait()?.as_u64();
+        let mut reader = BatchedAuctionDataReader::new(index);
+        loop {
+            let number_of_added_orders =
+                reader.apply_batch(&self.contract.get_auction_data_batched(
```
bh2smith: Is this `apply_batch` doing something like `append_page`?

nlordell: I think this helps with the confusion of "batches" vs "pages" since the word "batch" has a very specific meaning in the exchange protocol.

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub fn get_auction_data(self) -> (AccountState, Vec<Order>) {
+        (self.account_state, self.orders)
+    }
+
+    /// Applies one batch of data.
+    ///
+    /// A batch can come from `getEncodedUsersPaginated` or `getEncodedOrders`.
+    /// In the latter case there is only one batch.
+    ///
+    /// Returns the number of orders in the data, before filtering for `index`.
```
fleupold: nit:  Returns the number of orders in the packed_acution_bytes

I'd also leave out the second half sentence (or change it to before applying *any* filtering) as we are technically filtering by amount as well.

driver/src/orderbook/batched_auction_data_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+            .expect("user has order but no order count");
+        number_of_orders
+    }
+
+    fn apply_auction_data(&mut self, auction_elements: Vec<StableXAuctionElement>) {
+        self.apply_auction_elements_to_account_state(auction_elements.iter());
+        // Workaround for borrow checker that would complain that `extend`
+        // borrows self as mutable while at the same time the filter
+        // closure borrows self as immutable because of using `self.index`.
+        let index = self.index;
```
fleupold: Do we still need this workaround now that we are not borrowing self mutably in the same method anymore?

e00E: It is still needed.

fleupold: I see. It says why in the comment.

e00E:
I have ran this locally against mainnet where it read a total of 1031 orders and returned 387. This seems correct and consistent with the non paged orderbook reading so I am going to merge this now.

PR #522, opened 2020-02-11
[Trivial] prefixing prometheus metrics

fleupold:
Requested by @luarx 

This is needed as metric names need to be globally unique in our entire network. A general name like success is not really serving this purpose.

### Test Plan

Run locally and see new test names showing up on `http://localhost:9586/`

### DIFF
--- a/driver/src/metrics/stablex_metrics.rs
+++ b/driver/src/metrics/stablex_metrics.rs
@@ -21,7 +21,7 @@ pub struct StableXMetrics {
 impl StableXMetrics {
     pub fn new(registry: Arc<Registry>) -> Self {
         let processing_time_opts = Opts::new(
-            "processing_times",
+            "dfusion_service_processing_times",
             "timings between different processing stages",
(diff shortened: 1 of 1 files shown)


driver/src/metrics/stablex_metrics.rs
```diff
@@ -21,31 +21,40 @@ pub struct StableXMetrics {
 impl StableXMetrics {
     pub fn new(registry: Arc<Registry>) -> Self {
         let processing_time_opts = Opts::new(
-            "processing_times",
```
nlordell: Should these be prefixed with `$NETWORK` as well so we can distinguish rinkeby from mainnet or are we not going to record prometheus metrics for rinkeby?

fleupold: I believe filtering by network/solver strategy can be done with the host label that the event emits. I think the global prefix is just so that we don't confuse the general source of the metrics.

PR #523, opened 2020-02-11
[Easy] sync src code between local and docker filesystem

fleupold:
This allows for changes to the local filesystem to be directly reflected in the docker container without having to rebuild the entire container. It makes testing changes in a docker setup much faster.

### Test Plan

- docker-compose up stablex
- Make a change inside the driver code
- stop and restart the container
- see change is reflected without having had to docker-compose build.

### DIFF
--- a/docker-compose.binary.yml
+++ b/docker-compose.binary.yml
@@ -1,7 +1,7 @@
 version: "3.6"
 services:
   stablex:
-    command: ./driver
+    command: ./stablex
     build:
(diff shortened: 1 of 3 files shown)


fleupold:
Heads up @e00E this will likely break the staging pods again since I had to rename the binary one more time (to allow mounting the volume under the same path also in the release docker).

Let's see if our new alerting system catches it.

PR #526, opened 2020-02-12
[Easy] Token representation

bh2smith:
In relation to solver issue https://github.com/gnosis/dex-solver/pull/129#issuecomment-585201068

### DIFF
--- a/driver/src/orderbook/paginated_auction_data_reader.rs
+++ b/driver/src/orderbook/paginated_auction_data_reader.rs
@@ -115,8 +115,8 @@ impl PaginatedAuctionDataReader {
                             *x,
                             element.sell_token_balance,
                             "got order which sets user {}'s sell token {} \
-                            balance to {} but sell_token_balance has already \
-                            been set to {}",
+                             balance to {} but sell_token_balance has already \
+                             been set to {}",
                             element.order.account_id,
                             element.order.sell_token,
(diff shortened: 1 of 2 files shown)


bh2smith:
> Looks fine from our side but don't we have to wait for this format to be accepted by the solver first? Otherwise the format wouldn't be compatible. Or does the driver already accept both formats?

Very good question. I suppose we should include a solver version bump with this PR. @twalth3r can you let me know when this is ready!?

twalth3r:
The solver PR is merged and a new version tagged: https://github.com/gnosis/dex-solver/releases/tag/v0.5.10

bh2smith:
It definitely makes sense to bump the solver version but it appears to be passing with the current version anyway. Any ideas @fleupold? I have the version bump ready to go, but perhaps it should be in a separate PR if not necessary here.

e00E:
Maybe we don't run a test in CI with the real solver? We probably should add that if we don't. Or the solver was forwards compatible in this case.

fleupold:
We do run CI with the "real" solver. I don't think this change would cause any crash, it just leads to bad formatting in the debug output and potentially bad price estimations (for which we don't specifically test). @marcovc described the impact here: https://github.com/gnosis/dex-solver/pull/129#issuecomment-584244543

PR #529, opened 2020-02-12
[WIP] Intial Price Estimation

nlordell:
This PR implements some initial price estimation using Kraken as an oracle. It is being implemented in a way that allows for multiple sources to be added later on.

For HTTP client library `isahc` was used. The reasons are that it:
- uses statically linked `libcurl` under to hood, so it should be fairly stable
- has both sync and async (which may be useful for price oracles where price pairs need to be fetched individually)
- async implementation uses an implicit runtime and does not rely on more complex setups like `tokio`.

### Test Plan

New unit tests, and integrate with current integration tests.


nlordell:
Closing and re-opening as draft PR.

PR #530, opened 2020-02-12
[WIP] Kraken price estimation

nlordell:
This PR implements some initial price estimation using Kraken as an oracle. It is being implemented in a way that allows for multiple sources to be added later on.

For HTTP client library `isahc` was used. The reasons are that it:
- uses statically linked `libcurl` under to hood, so it should be fairly stable
- has both sync and async (which may be useful for price oracles where price pairs need to be fetched individually)
- async implementation uses an implicit runtime and does not rely on more complex setups like `tokio`.

### Test Plan

New unit tests, and integrate with current integration tests.


driver/src/price_estimation/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// Cache of token information that is used for
+    tokens: Vec<Token>,
+    /// Cached count of read tokens.
+    read_token_count: u16,
+}
+
+lazy_static! {
+    /// Tokens that are pegged to USD.
+    static ref USD_TOKENS: HashSet<String> = HashSet::from_iter(
+        ["USDT", "TUSD", "USDC", "PAX", "GUSD", "DAI", "sUSD"]
```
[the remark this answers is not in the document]
nlordell: These are just used as a fallback price in case Kraken does not list prices for them. "PAX" for example is not listed on Kraken so it gets the default `10**18` value.

Note that I tried to set this up in a way that the price oracle can have multiple `PriceSource`s, where Kraken is one of them. So hopefully, as more price sources are included, the fallback prices will not have to be used.

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,248 @@
...
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct Asset {
+    pub altname: String,
+}
+
+impl Asset {
+    /// Create a new asset from an alternate name.
+    #[cfg(test)]
+    pub(crate) fn new(altname: &str) -> Asset {
```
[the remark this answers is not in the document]
nlordell: No, its what Kraken calls the alternate name of a token. So US currency is called "ZUSD" but has an "altname" of "USD". The idea is that it allows you to use those name interchangeably for API calls.

nlordell:
> just left some random comments - I will review it later in more detail - and I wanna let you know that I am having the following issue with mac to build it:
https://gist.github.com/josojo/fed57604399d3afca44bec50ef1b668f

Hmm, odd.. I'll have to take a look.

driver/src/price_estimation/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+            .call()
+            .wait()
+            .unwrap();
+        let erc20 = ERC20Detailed::at(&self.contract.web3(), address);
+        let symbol = {
+            // NOTE: We check if the symbol is part of the overrides map, and
+            //   use the overridden value if it is. This allows ERC20 tokens
+            //   like WETH to be treated by ETH, since exchanges generally only
+            //   track ETH prices and not WETH.
+            let symbol = erc20.symbol().call().wait().unwrap();
```
fleupold: I wonder if it is save to rely on the ERC20 symbol. I can create my own contract with an arbitrary symbol. Personally I would feel more comfortable using e.g. our trusted [token list](https://github.com/gnosis/dex-js/blob/master/src/tokenList.json) to get the symbols for tokens we now are real.

But then, all you can really do is get wrong estimate for your fake token, not sure if that is a concern.

nlordell: I figured that we would eventually extend token filtering to this here.

fleupold: I could imaging that we might not want to fully blacklist a token but still not necessarily do a price estimation for them.

nlordell: We can implement a price estimation token filter analogous to the orderbook filter. This would allow us to quickly filter tokens without needing to rebuild. Or, for price estimation, we can go the opposite direction and have a whitelist (instead of a blacklist).

nlordell:
@josojo It hopefully should be fixed now that master is merged.

nlordell:
This PR was split into multiple smaller PRs and can now be closed:
#562
#563
#571

PR #531, opened 2020-02-12
Upgrade to crates.io ethcontract

nlordell:
This PR upgrades to use the most recent version of the crates.io `ethcontract` crate. This means we will no longer have to maintain the special compatibility branch for `dex-services`. Because of this update, the version of some of the shared dependencies have changed, and the generated contract bindings use the correct primitive integer types (no longer coerced to `u64`, `U128` and `U256`). This PR:
- Updates code to use re-exported `web3` and `jsonrpc_core` crates, reducing the dependency update burden in `dex-services`
- Updates code to use the correct contract primitive integer types. Specifically, there are no more `u16` to `u64` and `u128` to `U128` conversions.

The main benefit of this update is that:
- Dependabot will automatically create PRs when ethcontract updates again
- No longer required to managed dependencies for web3 related things as that is all part of `ethcontract`
- Hopefully fixed some error message decoding for Parity nodes so we can create better filters for mainnet client

### Test Plan

CI and :pray:

### DIFF
--- a/driver/src/orderbook/mod.rs
+++ b/driver/src/orderbook/mod.rs
@@ -61,13 +61,20 @@ impl<'a> StableXOrderBookReading for PaginatedStableXOrderBookReader<'a> {
         let mut reader = PaginatedAuctionDataReader::new(index);
         loop {
             let number_of_orders: u16 = reader
-                .apply_page(&self.contract.get_auction_data_paginated(
-                    block,
-                    self.page_size,
-                    reader.pagination().previous_page_user,
-                    reader.pagination().previous_page_user_offset.try_into()?,
-                )?)
-                .try_into()?;
+                .apply_page(
+                    &self.contract.get_auction_data_paginated(
+                        block,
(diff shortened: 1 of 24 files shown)


driver/src/orderbook/mod.rs
```diff
@@ -59,13 +60,15 @@ impl<'a> StableXOrderBookReading for PaginatedStableXOrderBookReader<'a> {
...
-            )?);
-            if (number_of_orders as u64) < self.page_size {
+            let number_of_orders: u16 = reader
+                .apply_page(&self.contract.get_auction_data_paginated(
+                    block,
+                    self.page_size,
+                    reader.pagination().previous_page_user,
+                    reader.pagination().previous_page_user_offset.try_into()?,
+                )?)
+                .try_into()?;
```
fleupold: nit: should we expect here since apply_page should never return more requested?

nlordell: Good point!

PR #533, opened 2020-02-13
Get gas price from gnosis safe relay

e00E:
Related to #369 .
Test Plan: CI and unit test for new functionality.

The way it is implemented is
* try to retrieve the gas prices from the gnosis safe relay
* if it worked then use the `fast` price
* if it did not work (gnosis safe relay is down) fall back on the current behavior which is to use `GasPrice::Scaled(3.0)`

I am not sure if this is exactly what you had in mind @fleupold . Do you still want to scale the fast price by 1.2?


nlordell:
Maybe we should use the same HTTP client library. Any reason why you chose `reqwest`? My issue with it is that is needs a `tokio` runtime for the `async` methods. While we don't use `async` at all now, one of the exchanges we want to estimate prices from require 1 request per token and `async` would probably be useful in that scenario.

e00E:
Which http client library do you mean by "the same"? I didn't see one in Cargo.toml but I might have missed it.

I chose reqwest because it is the most popular and I used it before.

nlordell:
Sorry, my PR is still a draft PR. I chose `isahc` since it also has sync and async implementation but implicitly drives the future and doesn’t require `tokio`. It uses cURL under the hood so it should be robust.

I mean, this is all up for debate! I’m just saying between this and #530 we should settle on a single HTTP client.

driver/src/contracts/stablex_contract.rs
```diff
@@ -139,7 +140,11 @@ impl StableXContract for BatchExchange {
             prices,
             token_ids_for_price,
         )
-        .gas_price(GasPrice::Scaled(3.0))
+        .gas_price(
+            gnosis_safe_gas_station::get_gas_price()
+                .map(|gas_price| GasPrice::Value(gas_price.fast))
+                .unwrap_or(GasPrice::Scaled(3.0)),
```
fleupold: For fast I think we don't need to scale it with a factor 3. Let's start maybe with 1.5?

driver/src/contracts/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,82 @@
+use ethcontract::U256;
```
fleupold: This should probably not be in the contracts module as it is not really doing any smart contract interaction. I think it even deserves its own module.

driver/src/contracts/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,82 @@
...
+/// Retrieves gas prices from the Gnosis Safe Relay api.
+///
+/// Uses a timeout of 10 seconds.
+pub fn get_gas_price() -> Result<GasPrice, reqwest::Error> {
+    const URL: &str = "https://safe-relay.gnosis.io/api/v1/gas-station/";
+    const TIMEOUT: Duration = Duration::from_secs(10);
+    // It would be more efficient to reuse the client between calls. However, we
+    // only call this function once per batch when submitting a solution so this
+    // is not important at the moment.
+    let client = reqwest::blocking::ClientBuilder::new()
```
fleupold: As we consolidate our libraries to use Isahc can we share the same HTTP client across requests?

driver/src/contracts/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,82 @@
...
+use uint::FromDecStrErr;
+
+/// Result of https://safe-relay.gnosis.io/ api call `/v1/gas-station`.
+///
+/// Prices are in wei.
+#[derive(Deserialize, Debug, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GasPrice {
+    pub last_update: String,
+    #[serde(deserialize_with = "deserialize_u256_from_string")]
```
fleupold: Can we use [U256::from_dec_str])https://docs.rs/ethereum-types/0.2.3/ethereum_types/struct.U256.html#method.from_dec_str) here?

driver/src/contracts/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,82 @@
...
+    client.get(URL).send()?.json()
+}
+
+fn deserialize_u256_from_string<'de, D>(deserializer: D) -> Result<U256, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    U256::from_dec_str(&s)
+        .map_err(uint_error_to_string)
```
fleupold: nit: why not use debug format from `uint::FromDecStrErr` directly? Should we maybe also include the bad string in the error to make debugging easier?

fleupold (commented):
Looks good so far. Let's unify the HTTP request library though.

driver/src/contracts/stablex_contract.rs
e00E: It doesn't scale `gas_price.fast` by 3. If it cannot retrieve the gas price from the api then it uses ethcontract's built in `GasPrice::Scaled` which is the same behavior as before.
Or are you saying the old behavior should have been 1.5 all along?

driver/src/contracts/gnosis_safe_gas_station.rs
e00E: It is used by that function. It cannot be used at this level because this function doesn't have the correct parameter and return types.

fleupold: Oops, yeah I saw that and forgot to remove this comment. Never mind.

driver/src/contracts/stablex_contract.rs
fleupold: Yes, my bad. that's good

driver/src/contracts/gnosis_safe_gas_station.rs
e00E: Included the source string in the message in new commit.
In practice `Debug` would work but conceptually it isn't meant for the error message so it felt better to do it like this.

driver/src/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,86 @@
...
+/// Retrieves gas prices from the Gnosis Safe Relay api.
+///
+/// Uses a timeout of 10 seconds.
+pub fn get_gas_price() -> Result<GasPrice, isahc::Error> {
+    const URL: &str = "https://safe-relay.gnosis.io/api/v1/gas-station/";
+    const TIMEOUT: Duration = Duration::from_secs(10);
+    // It would be more efficient to reuse the client between calls. However, we
+    // only call this function once per batch when submitting a solution so this
+    // is not important at the moment.
+    let client = HttpClient::builder().timeout(TIMEOUT).build()?;
```
fleupold: Can we make the gas station a struct and pass this from above to the contract? Even though performance is not a main concern I think this is only a little bit of extra work to make the factoring here clean (the gas estimation is a dependency of the contract and should thus be injected and initializable with e.g. a mocked estimator).

driver/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Alexander Herrmann <alex@gnosis.io>", "Ben Smith <ben@gnosis.io>", "
 edition = "2018"
 
 [dependencies]
+anyhow = "1"
```
e00E: I have a question about this. I notice that we usually use exact versions, even for dependencies that are 1.0 . Why? Isnt just `1` enough?

driver/src/contracts/stablex_contract.rs
```diff
@@ -68,6 +69,7 @@ pub trait StableXContract {
         orders: Vec<Order>,
         solution: Solution,
         claimed_objective_value: U256,
+        get_gas_price: &(dyn GetGasPrice + 'static),
```
e00E: I had to add `'static` here when compiling in test mode. This seems to be a limitation of the current mockall that might get fixed in the future

driver/src/gnosis_safe_gas_station.rs
e00E: I made this mockable by introducing a struct and trait and passing it down.

driver/src/gnosis_safe_gas_station.rs
```diff
@@ -23,22 +28,30 @@ pub struct GasPrice {
...
+}
+
+pub struct GasStation {
+    client: HttpClient,
+    uri: Uri,
+}
+
+impl GasStation {
+    pub fn new(request_timeout: Duration, api_uri: &str) -> Result<GasStation> {
+        //let uri: Uri = uri.parse()?;
```
bh2smith: Should this be removed?

bh2smith (approved):
Looks nice and clean! Just one commented line that should go.

driver/Cargo.toml
fleupold: Looking at the json.lock, aren't we using anyhow 1.0.26?

driver/src/contracts/stablex_contract.rs
fleupold: Can't we make getGasPrice a dependency of the concrete implementation of the contract trait (pass it in new and store it on the struct) rather than passing it on the signature?

To me the way gas price estimation is done seems like an implementation detail of the contract and not something that needs to be part of the interface.

driver/src/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub standard: U256,
+    #[serde(deserialize_with = "deserialize_u256_from_string")]
+    pub fast: U256,
+    #[serde(deserialize_with = "deserialize_u256_from_string")]
+    pub fastest: U256,
+}
+
+#[cfg_attr(test, mockall::automock)]
+pub trait GetGasPrice {
+    /// Retrieves gas prices from the Gnosis Safe Relay api.
```
fleupold: This is the implementation detail of GasStation not specific to the GetGasPrice trait, which cold estimate the price based on any oracle it pleases (nit: I'd maybe name GasStation GnosisGasStation).

I personally also like -ing naming for interfaces (e.g. GasPriceEstimating) but that doesn't seem to be rust standard.

fleupold (approved):
Looks good, but I wouldn't make the GetGasPrice part of the interface but instead pass it at construction time.

driver/Cargo.toml
e00E: I think that means that its pulled in by a dependency but we are not using it since its not in Cargo.toml. But if there is no problem then I'll just keep the "anyhow = 1" .

driver/src/contracts/stablex_contract.rs
e00E: We `impl` on `BatchExchange` that comes from the generated `batch_exchange.rs` so there is  no struct to store things in. But we can make our own struct that contains `BatchExchange` and anything else we need to store.

driver/src/gnosis_safe_gas_station.rs
e00E: In my mind this is the gnosis safe gas station, not a general gas station as seen by the return struct that is specific to Gnosis.  The reason I made it a trait is so that it is mockable. I don't think we need to generalize it for potential other gas stations yet.

driver/Cargo.toml
nlordell: I'm a firm believer of `major.minor`. Note that they have different meanings:
- `1` <=> `^1.0.0` <=> `>=1.0.0 <2.0.0`
- `1.2` <=> `~1.2.0` <=> `>=1.2.0 <1.3.0`
- `1.2.3` <=> `~1.2.3` <=> `>=1.2.3 <1.3.0`

driver/src/solution_submission/mod.rs
```diff
@@ -47,7 +47,7 @@ pub struct StableXSolutionSubmitter<'a> {
     contract: &'a dyn StableXContract,
 }
 
-impl<'a> StableXSolutionSubmitter<'a> {
+impl<'a, 'b> StableXSolutionSubmitter<'a> {
```
nlordell: Is `'b` used anywhere?

driver/src/gnosis_safe_gas_station.rs
nlordell: > I personally also like -ing naming for interfaces (e.g. GasPriceEstimating) but that doesn't seem to be rust standard.

Rust standard is to use infinitive verb forms. So `Read` and `Write` and `Display`, here I would say `EstimateGasPrice` by Rust standard. That being said, I like the -ing standard we use for traits in our codebase and I would recommend that here.

driver/src/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,106 @@
...
+        .map_err(|err| format!("{}: {}", uint_error_to_string(err), s))
+        .map_err(serde::de::Error::custom)
+}
+
+fn uint_error_to_string(err: FromDecStrErr) -> &'static str {
+    match err {
+        FromDecStrErr::InvalidCharacter => "FromDecStrErr: invalid character",
+        FromDecStrErr::InvalidLength => "FromDecStrErr: invalid length",
+    }
+}
```
nlordell: Can we submit a PR to `uint` crate to implement the `Error` trait to avoid these shenanigans?

driver/src/main.rs
```diff
@@ -1,6 +1,7 @@
 mod contracts;
 mod driver;
 mod error;
+mod gnosis_safe_gas_station;
```
nlordell: I would call this module `gas_station` and the struct `GnosisGasStation`.

driver/src/gnosis_safe_gas_station.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fastest: U256,
+}
+
+#[cfg_attr(test, mockall::automock)]
+pub trait GasPriceEstimating {
+    /// Retrieves gas prices from the Gnosis Safe Relay api.
+    fn estimate_gas_price(&self) -> Result<GasPrice>;
+}
+
+pub struct GasPriceEstimatingImpl {
```
nlordell: Maybe this can be called `GnosisGasStation` or `GnosisSafeGasStation`?

nlordell (approved):
Looks good! Just some minor naming nits and annoyed that `uint::FromDecStrErr` does not implement `std::error::Error`.

driver/src/main.rs
```diff
@@ -55,8 +57,13 @@ fn main() {
     let optimization_model = OptimizationModel::from(optimization_model_string.as_str());
 
     let web3 = web3_provider(&ethereum_node_url).unwrap();
-    let contract = BatchExchange::new(&web3, network_id).unwrap();
-    info!("Using contract at {}", contract.address());
+    let gas_station = GasPriceEstimatingImpl::new(
+        Duration::from_secs(10),
```
nlordell: We can read this from the environment or even use the same timeout as transport?

e00E: That's true. I want to do this after we merge the environment and command line unification PR.

driver/src/gnosis_safe_gas_station.rs
e00E: I am going to do that.

driver/src/main.rs
```diff
@@ -95,9 +106,16 @@ fn main() {
...
+    let gas_station =
+        GnosisSafeGasStation::new(options.gas_station_timeout, gas_station::DEFAULT_URI).unwrap();
+    let contract = StableXContractImpl::new(
+        &web3,
+        options.private_key.clone(),
+        options.network_id,
+        &gas_station,
+    )
+    .unwrap();
+    info!("Using contract at {:?}", contract.instance().address());
```
fleupold: ubernit: I'm generally not a fan of getter chains (as they leak information and arguably violate [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter#In_object-oriented_programming)).

driver/src/contracts/stablex_contract.rs
```diff
@@ -23,23 +24,41 @@ lazy_static! {
 
 include!(concat!(env!("OUT_DIR"), "/batch_exchange.rs"));
 
-impl BatchExchange {
-    pub fn new(web3: &contracts::Web3, key: PrivateKey, network_id: u64) -> Result<Self> {
+pub struct StableXContractImpl<'a> {
```
fleupold: 👍

fleupold (approved):
Awesome! Looking forward to all those gas savings...

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FLCdPNT81vlv3y%2Fgiphy.gif&token=money" data-gitmeme-token="money"></a>

driver/Cargo.toml
e00E: Why specify the minor when semver guarantees that as long as the major is the same its compatible?

driver/src/main.rs
e00E: I changed it but do want to point out that we already had access to the whole `BatchExchange` before, just not through a getter, so that hasn't *changed*. At the same time it can still be better to hide it.

fleupold: yes good point. If there is a good reason to re-expose it I'm not against that.

bh2smith (commented):
⛽️ 🔥 🚒  Great!

PR #537, opened 2020-02-13
[Prometheus] Initializing metrics with 0 on startup

fleupold:
This PR makes it so that all the metrics we can report are properly initialised with 0. Metrics are reported for an arbitrary array of string labels (each metrics must have a value for each label) and the value themselves are also strings.

Thus there is no good built-in type safe way to make sure all metrics are initialised. This is an attempt of ensuring some type-safety (a lot of it could probably absorbed into a macro but that is beyond my skill set). Looking forward to your suggestions.

### Test Plan

Run the solver locally. See all metrics are initialized with 0:

```
# HELP dfusion_service_failures number of auctions failed
# TYPE dfusion_service_failures counter
dfusion_service_failures{stage="orders_fetched"} 0
dfusion_service_failures{stage="skipped"} 0
dfusion_service_failures{stage="solved"} 0
dfusion_service_failures{stage="started"} 4
dfusion_service_failures{stage="submitted"} 0
dfusion_service_failures{stage="verified"} 0
# HELP dfusion_service_orders number of orders in a batch
# TYPE dfusion_service_orders gauge
dfusion_service_orders{type="orderbook"} 0
dfusion_service_orders{type="solution"} 0
# HELP dfusion_service_success number of auctions successfully processed
# TYPE dfusion_service_success counter
dfusion_service_success{stage="orders_fetched"} 0
dfusion_service_success{stage="skipped"} 0
dfusion_service_success{stage="solved"} 0
dfusion_service_success{stage="started"} 0
dfusion_service_success{stage="submitted"} 0
dfusion_service_success{stage="verified"} 0
# HELP dfusion_service_tokens number of distinct tokens in a batch
# TYPE dfusion_service_tokens gauge
dfusion_service_tokens{type="orderbook"} 0
dfusion_service_tokens{type="solution"} 0
# HELP dfusion_service_users number of distinct users in a batch
# TYPE dfusion_service_users gauge
dfusion_service_users{type="orderbook"} 0
dfusion_service_users{type="solution"} 0
```

### DIFF
--- a/driver/src/metrics/stablex_metrics.rs
+++ b/driver/src/metrics/stablex_metrics.rs
@@ -18,76 +18,6 @@ pub struct StableXMetrics {
     users: IntGaugeVec,
 }
 
-trait InitializeableMetric: 'static + Sized + AsRef<str> {
-    const LABEL: &'static str;
-    const ALL_STAGES: &'static [Self];
-
-    fn initialize_counters(c: &IntCounterVec) {
-        for stage in Self::ALL_STAGES {
-            c.with_label_values(&[stage.as_ref()]).inc_by(0);
-        }
-    }
-
-    fn initialize_gauges(g: &IntGaugeVec) {
-        for stage in Self::ALL_STAGES {
-            g.with_label_values(&[stage.as_ref()]).set(0);
-        }
-    }
-}
-
-enum ProcessingStage {
-    Started,
-    OrdersFetched,
-    Solved,
-    Verified,
-    Submitted,
-    Skipped,
-}
-
-impl AsRef<str> for ProcessingStage {
-    fn as_ref(&self) -> &'static str {
-        match self {
-            Self::Started => "started",
-            Self::OrdersFetched => "orders_fetched",
-            Self::Solved => "solved",
-            Self::Verified => "verified",
-            Self::Submitted => "submitted",
-            Self::Skipped => "skipped",
-        }
-    }
-}
-
-impl InitializeableMetric for ProcessingStage {
-    const LABEL: &'static str = "stage";
-    const ALL_STAGES: &'static [Self] = &[
-        Self::Started,
-        Self::OrdersFetched,
-        Self::Solved,
-        Self::Verified,
-        Self::Submitted,
-        Self::Skipped,
-    ];
-}
-
-enum BookType {
-    Orderbook,
-    Solution,
-}
-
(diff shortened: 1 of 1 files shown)


bh2smith (approved):

driver/src/metrics/stablex_metrics.rs
```diff
@@ -24,38 +24,44 @@ impl StableXMetrics {
             "dfusion_service_processing_times",
             "timings between different processing stages",
         );
-        let processing_times = IntGaugeVec::new(processing_time_opts, &["stage"]).unwrap();
+        let processing_times =
+            IntGaugeVec::new(processing_time_opts, &[ProcessingStage::LABEL]).unwrap();
```
e00E: Should we initialize this too?

e00E (approved):

driver/src/metrics/stablex_metrics.rs
fleupold: Not sure if it is correct to have a processing time of 0 e.g. for solution submission if it has never actually happened. It could confuse people for metrics where 0s is a reasonable value.

e00E: I noticed in grafana for solution submission it keeps using the previous value when we don't actually submit a solution. Maybe it would be clearer to reset it to 0 at the start and for every batch where don't really reach the step. Or maybe not, just throwing it out there.

fleupold: Yes, I noticed that too. My concern about clearing is when would we do that? If we clear to early we might not get a sample at the actual value and thus miss the measurement. After all maybe a gauge is not the right metric type here?

e00E: The sampling problem we already have because technically the next batch could be solved so quickly that the old value cannot be measured even though this is unlikely.

This goes back to the discussion when the metrics were introduced that having these processing times as gauges does not fit the prometheus model. A histogram is more appropriate but then you cannot get the exact processing times and you cannot show just the most recent value.

What we really want is to log the processing times of each stage whenever the stage completes. This could go into a file or sql database table. Then that table contains exactly the information we want. I don't think this accomplishable with Prometheus. Maybe we can log into kibana or some postgres or sql server and have Grafana plot that.

fleupold: Yes, this goes however beyond the scope of this PR. I will create a separate issue for it.

PR #541, opened 2020-02-14
[Easy] store each solver result in a different folder

fleupold:
This PR makes it so that we don't override the result file in every auction. Instead we store them in separate folders similar to how we do it for input files.

This will allow us to sync the results with S3 and visualize them on the public S3 bucket.

### Test Plan

Run local instance with mainnet solver for two batches. See both result files on disk.

### DIFF
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -10,8 +10,6 @@ use std::io::{BufReader, BufWriter, Read, Write};
 use std::process::Command;
 use web3::types::H160;
 
-const RESULT_FOLDER: &str = "./results/tmp/";
-
 type PriceMap = HashMap<u16, u128>;
 
 mod solver_output {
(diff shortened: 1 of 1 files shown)


bh2smith (approved):

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -265,12 +265,13 @@ fn write_input(input_file: &str, input: &str) -> std::io::Result<()> {
...
 fn run_solver(
     input_file: &str,
+    result_folder: &str,
     optimization_model: OptimizationModel,
 ) -> Result<(), PriceFindingError> {
     let optimization_model_str = optimization_model.to_args();
     let output = Command::new("python")
         .args(&["-m", "batchauctions.scripts.e2e._run"])
-        .arg(RESULT_FOLDER)
+        .arg(result_folder)
```
e00E: Do we need to create the result folder being calling the solver?

fleupold: Good question! It's created by the solver.

PR #543, opened 2020-02-14
Custom HTTP Transport

### LINKED ISSUE #542: Add timeouts for all network communication
When the ethereum node is slow we can block indefinitely on calls to it. Instead we should configure a timeout so that we can handle this as an error.

nlordell:
Implements a custom HTTP transport with logging and a 10s timeout. Timeout can be made configurable and `eth_call` prometheus metrics can be added as well (in separate PRs :cactus:).

The custom HTTP transport was implemented because:
- better logging (we will be able to log the full response instead of just the result JSON / parsed error)
- allows us to work around the Ganache issue where immediately mined transactions that reverted were returning a result (with the Tx hash, as expected by the protocol) and an error
- allows us to use an HTTP client implementation's timeout functionality instead of rolling our own using timers (since there is no way to get access to the underlying HTTP client in the current transport implementation).

Closes #542 

### Test Plan

CI

### DIFF
--- a/driver/src/contracts/mod.rs
+++ b/driver/src/contracts/mod.rs
@@ -2,21 +2,20 @@ pub mod stablex_auction_element;
 pub mod stablex_contract;
 
 use crate::error::DriverError;
-use crate::transport::LoggingTransport;
+use crate::transport::HttpTransport;
 use ethcontract::contract::MethodDefaults;
-use ethcontract::web3::transports::{EventLoopHandle, Http};
 use ethcontract::{Account, PrivateKey};
 use log::Level;
 use std::env;
+use std::time::Duration;
 
-pub type Web3 = ethcontract::web3::api::Web3<LoggingTransport<Http>>;
+pub type Web3 = ethcontract::web3::api::Web3<HttpTransport>;
 
-pub fn web3_provider(url: &str) -> Result<(Web3, EventLoopHandle), DriverError> {
-    let (event_loop, http) = Http::new(&url)?;
-    let logging = LoggingTransport::new(http, Level::Debug);
-    let web3 = Web3::new(logging);
+pub fn web3_provider(url: &str) -> Result<Web3, DriverError> {
+    let http = HttpTransport::new(url, Level::Debug, Duration::from_secs(10))?;
+    let web3 = Web3::new(http);
 
-    Ok((web3, event_loop))
+    Ok(web3)
 }
 
 fn method_defaults(network_id: u64) -> Result<MethodDefaults, DriverError> {
--- a/driver/src/contracts/mod.rs
+++ b/driver/src/contracts/mod.rs
@@ -5,14 +5,13 @@ use crate::error::DriverError;
 use crate::transport::HttpTransport;
 use ethcontract::contract::MethodDefaults;
 use ethcontract::{Account, PrivateKey};
-use log::Level;
 use std::env;
 use std::time::Duration;
 
 pub type Web3 = ethcontract::web3::api::Web3<HttpTransport>;
 
 pub fn web3_provider(url: &str) -> Result<Web3, DriverError> {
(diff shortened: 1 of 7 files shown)


driver/src/transport.rs
```diff
@@ -1,79 +1,123 @@
...
-    /// Create a new `LoggingTranport` from an underlying transport and log
-    /// level.
-    pub fn new(inner: T, level: Level) -> Self {
-        LoggingTransport { inner, level }
+impl HttpTransport {
+    /// Creates a new HTTP transport with settings.
+    pub fn new(url: impl Into<String>, timeout: Duration) -> Result<HttpTransport, DriverError> {
+        let client = HttpClient::builder()
+            .timeout(timeout)
+            // NOTE: This is needed as curl with try to upgrade to HTTP/2 which
```
e00E: with -> will

driver/src/transport.rs
```diff
@@ -1,79 +1,123 @@
...
+use isahc::prelude::{Body, Request, Response};
+use isahc::{Error as HttpError, HttpClient, ResponseExt};
+use log::{debug, warn};
+use serde::Deserialize;
 use serde_json::Value;
+use std::fmt::{self, Debug, Formatter};
+use std::str::FromStr;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
```
bh2smith: Seems we are "using" a lot of stuff. Is there a more compact way of doing this? Are we using all of these imports or are they somehow being used indirectly here? Seems like there aren't even enough lines in this file to use all this stuff!

bh2smith (approved):
Nice work here! very clean.

driver/src/transport.rs
e00E: It should be a warning and thus an error because we don't allow warnings when there are unused imports.

e00E (approved):

driver/src/transport.rs
```diff
@@ -1,79 +1,123 @@
...
-            "sending request ID {}: {}",
-            id,
-            serde_json::to_string(&request).expect("request is invalid JSON")
-        );
-        LoggingFuture {
-            inner: self.inner.send(id, request),
-            id,
-            level: self.level,
+    /// Execute an HTTP JSON RPC request.
+    async fn execute(self: Arc<Self>, id: RequestId, request: Call) -> Result<Value, Web3Error> {
```
fleupold: nit: as this transport might be reused by other parts of the codebase where we make requests, maybe it's a good idea to name this with rpc in mind (e.g. `execute_rpc`)?

driver/src/contracts/mod.rs
```diff
@@ -2,21 +2,19 @@ pub mod stablex_auction_element;
...
 
-pub type Web3 = ethcontract::web3::api::Web3<LoggingTransport<Http>>;
+pub type Web3 = ethcontract::web3::api::Web3<HttpTransport>;
 
-pub fn web3_provider(url: &str) -> Result<(Web3, EventLoopHandle), DriverError> {
-    let (event_loop, http) = Http::new(&url)?;
-    let logging = LoggingTransport::new(http, Level::Debug);
-    let web3 = Web3::new(logging);
+pub fn web3_provider(url: &str) -> Result<Web3, DriverError> {
+    let http = HttpTransport::new(url, Duration::from_secs(10))?;
```
fleupold: Maybe a constant or even read from an env variable?

fleupold (approved):
Really nice stuff. As a follow-up, can we log roundtrip times to prometheus?

driver/src/transport.rs
nlordell: Its all used. Like @e00E said unused imports fail the build. Its just this transport is "gluing" a lot of separate pieces together, so it end up importing a lot.

Issue #544, opened 2020-02-15
request is not supported because your node is running with state pruning


e00E:
We get this message from time to time

> [driver] StableXDriver error: Driver Error: [method 'getEncodedUsersPaginated(address,uint16,uint16)' failure: web3 error: RPC error: Error { code: ServerError(-32000), message: "This request is not supported because your node is running with state pruning. Run with --pruning=archive.", data: None }]

This is probably related to us setting a fixed block number for the paginated order reading.  Maybe this only pops up now because the node is overloaded (getting gateway timeouts).

fleupold:
@denisgranha how hard would it be to provide an archive node for dƒusion so that we can query past blocks (realistically it would only be a few blocks in the past at most)

The challenge is that we are doing paginated queries and would like to get a consistent view of the state balances. If those queries happen in different blocks, there is a chance for inconsistency (balance might have changed).

However, considering that we are likely querying the balances in the same 5 minute batch anyways (for which the results are consistent) we could potentially not specify a past block (in particular as the services should ideally in the long run work with standard infura nodes as well)

denisgranha:
let me discuss it with the team and get back to you

nlordell:
Re-opening this issue as it affects our orderbook reading.

fleupold:
As mentioned on the PR, this should not be an issue. The reason it was an issue earlier was that we were behind a loadbalancer and would occasionally query blocks that the slower node had not yet synced with thus causing a "block does not exist" error.

PR #547, opened 2020-02-17
Do not set block number when reading order book

e00E:
We are currently having an issue where the node does not support queries
with block numbers in the past because it is pruning the state.
Until we decide whether this is fixable on the node's side we decided to
remove the fixed block number. This can in theory lead to reading the
orderbook in an inconsisent state when two pages come from different
batches. If this happens then we already cannot solve the auction
because there is no time left seeing as the batch changed.
A consequence of this is that we previously had a panic in the page
reading code when we notice that despite fixing the block number there
is still an inconsistency. The panic has been turned into a warning.
We could instead keep the check and return an error instad of panicking
but this makes the code more convoluted and there can still be
inconsistencies we cannot detect like including orders that no longer
exist.

related to #544

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 use std::env;
 
 use ethcontract::transaction::GasPrice;
-use ethcontract::{Address as H160, BlockNumber, U256};
+use ethcontract::{Address as H160, U256};
 use lazy_static::lazy_static;
 #[cfg(test)]
 use mockall::automock;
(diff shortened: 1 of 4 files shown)


fleupold (approved):
Given that we are also seeing errors of the form:

>  DEBG [driver::transport] request ID 1934 failed: Rpc(Error { code: InvalidParams, message: "Unknown block number", data: None })

It could be that the block has simply not yet been populated to all Full Nodes and given that we are querying behind a load balancer, we might be hitting different endpoints with different requests. So we might actually not be querying block in the future but instead in the past.

Anyways for now it's probably better to not query on a specific block.

bh2smith (approved):

PR #551, opened 2020-02-17
[Easy] Change e2e test to use same private key variable as the driver

nlordell:
The driver uses `PRIVATE_KEY` but the e2e tests use `PK`, switch to use the same environment variable as the driver.

### Test Plan

CI.


nlordell:
It seams like PK is needed by truffle as well, so I'm closing this.

PR #556, opened 2020-02-19
Reads additional token information from env

### LINKED ISSUE #555: Emergency backup for price data
As price feed might not be 100% reliable or simply do not offer all needed data, there should be the option to pass additional token information + their prices to the driver and solver via a new env variable.

The format for token information should be aligned with https://github.com/gnosis/dex-services/issues/507

josojo:
closes #555

---
testplan:
add the following two envs into common.env
```
SOLVER_TYPE=standard-solver
PRICE_FEED_INFORMATION={ "0": { "alias": "OWL", "decimals": 18, "externalPrice": 1000000000000000000}, "1": { "alias": "WETH", "decimals": 18, "externalPrice": 250000000000000000000 },  "2": { "alias": "USDT", "decimals": 6, "externalPrice": 1000000 }, "3":{ "alias": "TUSD", "decimals": 18, "externalPrice": 1000000000000000000 }, "8": { "alias": "sETH", "decimals": 18, "externalPrice": 250000000000000000000 }, "9":{ "alias": "sUSD", "decimals": 18, "externalPrice": 1000000000000000000 }, "15":{ "alias": "SNX", "decimals": 18, "externalPrice": 1000000000000000000 } }
```

witness the new format for the solver:
```
Solver input: {"tokens":{"T0000":{"decimals":18,"externalPrice":1000000000000000000},"T0001":{"decimals":18,"externalPrice":250000000000000000000}},"refToken":"T0000","accounts":{"90f8bf6a479f320ead074411a4b0e7944ea8c9c1":{"T0000":"3000000"},"ffcf8fdee72ac11b5c542428b35eef5769c409f0":{"T0001":"3000000"}},"orders":[{"accountID":"90f8bf6a479f320ead074411a4b0e7944ea8c9c1","sellToken":"T0000","buyToken":"T0001","sellAmount":"2000000","buyAmount":"999000"},{"accountID":"ffcf8fdee72ac11b5c542428b35eef5769c409f0","sellToken":"T0001","buyToken":"T0000","sellAmount":"999000","buyAmount":"1996000"}],"fee":{"token":"T0000","ratio":0.001}}
```


marcovc:
Question: Do we need to have the external prices as integers? Since we have the decimals, we could instead use the decimal point, and instead of:

```
{ "decimals": 18, "externalPrice": 1000000000000000000}, 
{ "decimals": 6, "externalPrice": 1000000 }
```

which is the cause of me having the screen full of fingerprints :), we could have:

```
{ "decimals": 18, "externalPrice": "1.0"}, 
{ "decimals": 6, "externalPrice": "1.0" }
```

cc/ @twalth3r

josojo:
>Question: Do we need to have the external prices as integers? Since we have the decimals, we could instead use the decimal point, and instead of:

No there is no need from the driver site. I am happy both ways.

marcovc:
Disadvantages of integer prices:
* I have to adapt solver code :)
* From your environment variable PRICE_FEED_INFORMATION, I'm guessing your idea is to have `externalPrice` field of a token T to be the amount of tokens T we can buy with 10^18 units of fee, right? It seems this would mean 10^18 is somehow hardcoded in this representation.

Disadvantages of float prices:
* seems somehow less consistent since everything else is in integers
* requires either passing floats as strings "1.00001" in the json, or if we want to pass then as floats in the json we have to hack into our json parser to use Decimals to parse floats (and you would likely have to do the same for your json writer).

twalth3r:
@marcovc - How about we take the prices as int mainly because of this argument:
> seems somehow less consistent since everything else is in integers

?

The int-representation is also what we ultimately want to use in our code.
Maybe there will even be another int-price in the future storing the most recent dfusion price.
We can of course add a human-readable representation ourselves (or have it given in addition by the driver). I strongly agree that this is super helpful!

I'd be in favour of having the number of decimals in any case!

josojo:
>From your environment variable PRICE_FEED_INFORMATION, I'm guessing your idea is to have externalPrice field of a token T to be the amount of tokens T we can buy with 10^18 units of fee, right? It seems this would mean 10^18 is somehow hardcoded in this representation.

yes, we would use here the same normalization of prices, as we use in final prices of the solution.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -339,6 +370,20 @@ pub mod tests {
 
     #[test]
     fn test_serialize_tokens() {
+        let mut token_data = HashMap::new();
+        let token_info_1 = Some(TokenInfo {
+            alias: String::from("T1"),
+            decimals: 18,
+            external_price: 1_000_000_000_000_000_000,
```
marcovc: Now I see why everybody seems to be fine with this large integers! :P

nlordell: I think python supports number separators as well:
```
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 1_000_000_000_000
1000000000000
```

marcovc: Oh wow I wasn't aware! Thanks!

nlordell: The only reason I know is that Rust copied its notation from Python :)

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -259,7 +290,7 @@ fn write_input(input_file: &str, input: &str) -> std::io::Result<()> {
     let file = File::create(&input_file)?;
     let mut writer = BufWriter::new(file);
     writer.write_all(input.as_bytes())?;
-    debug!("Solver input: {}", input);
+    println!("Solver input: {}", input);
```
fleupold: Please revert.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -104,17 +118,23 @@ pub struct OptimisationPriceFinder {
...
     optimization_model: OptimizationModel,
+    token_data: TokenData,
 }
 
 impl OptimisationPriceFinder {
-    pub fn new(fee: Option<Fee>, optimization_model: OptimizationModel) -> Self {
+    pub fn new(
+        fee: Option<Fee>,
+        optimization_model: OptimizationModel,
+        token_data: String,
```
fleupold: I think it would be cleaner and more typesafe to already pass the parsed data-structure here instead of an opaque string. We should try to avoid unwrapping in the lower level modules as much as possible.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -5,13 +5,24 @@ use crate::price_finding::price_finder_interface::{Fee, OptimizationModel, Price
...
 use std::fs::{create_dir_all, File};
 use std::io::{BufReader, BufWriter, Read, Write};
 use std::process::Command;
 
 type PriceMap = HashMap<u16, u128>;
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct TokenInfo {
+    pub alias: String,
```
fleupold: nit: do these have to be public?

nlordell:
@fleupold I started from the bottom up, so I first implemented a Kraken client and then the price aggregator (that should eventually support multiple sources) and the only bit left is getting the information to the solver.

This PR would help #530

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -5,13 +5,24 @@ use crate::price_finding::price_finder_interface::{Fee, OptimizationModel, Price
...
+    pub decimals: u32,
+    pub external_price: u128,
```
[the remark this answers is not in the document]
josojo: In the smart contract, we are using for the price representation u128. 
I wanted the provided token information as close as possible to the smart contract, as the smart contract kinda defines the precisions we should care about.

driver/src/price_finding/optimization_price_finder.rs
marcovc: It would be great if json format would accept that for integers :)

driver/src/price_finding/optimization_price_finder.rs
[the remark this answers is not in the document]
nlordell: Sorry, I didn't explain myself well. I agree that the internal representation of the price should be a `u128`. The issue, however, is that the JSON standard does not clearly state whether arbitrarily sized integers are allowed. Many JSON parser implementations will convert the large integer into a double precision floating point number.

As an example, in your browser console if you run:
```
> JSON.parse('123456789123456789123456789').toLocaleString()
"123,456,789,123,456,790,000,000,000"
```

Or even a contrived example Rust:
```
println!(
    "{:?}",
    serde_json::from_str::<serde_json::Value>(
        &serde_json::to_string(&123456789123456789123456789u128).unwrap()
    )
    .unwrap()
); // Outputs: Number(123456789123456790000000000.0)
```

That being said, I just checked with Python and it does support arbitrary precision integers and represents them :
```
>>> import json
>>> json.loads('123456789123456789123456789')
123456789123456789123456789
>>> type(json.loads('123456789123456789123456789'))
<class 'int'>
```

I also prefer using `u128` everywhere, but I just wanted to mention that the JSON spec doesn't officially support it, and it is dependent on JSON parser implementations.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -239,7 +275,7 @@ impl PriceFinding for OptimisationPriceFinder {
         state: &models::AccountState,
     ) -> Result<models::Solution, PriceFindingError> {
         let input = solver_input::Input {
-            tokens: serialize_tokens(&orders),
+            tokens: serialize_tokens(&orders, self.token_data.clone()),
```
nlordell: Since we only read the `token_data` for prices here, we can probably pass this in by reference instead of cloning.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -55,6 +57,28 @@ impl Serialize for TokenId {
 #[serde(transparent)]
 pub struct Num(#[serde(with = "display_fromstr")] pub u128);
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct TokenInfo {
+    alias: String,
+    decimals: u32,
```
nlordell: ubernit: this should probably be a u8 to match ERC20 contracts.

Issue #546, opened 2020-02-17
Pass OrderID in json to the solver


fleupold:
For debugging and particularly to blacklist "bad" orders, it would help a lot to include the orderId in the instance json that is passed on to the solver.

Moreover, I'd suggest to parse the resulting solution data based on the order IDs (at the moment we rely on the index in the result array). This would allow us on the solver side to only include orders in the result set that have a non-zero execution amount which would make analysing auction results much easier.

e00E:
Here is what I think we should do:

1. When communicating with the solver identify orders by their `accountId` and `orderId` instead of relying on the index in the list of orders. To do this we add the `order_id` field (`account_id` already exits) to the rust code.
2. We update the solver to use the new format which allows it to return the orders in the solution in any order and omit orders that were not touched.
3. We update the rust code to handle the new solver output format and bump the version of the solver at the same time so that we use the version that produces the new output.

For now we can only do step 1, then we need to synchronize with the solver being updated.

e00E:
Maybe someone from the solver could comment whether this makes sense. Not sure who I should ping. @twalth3r

twalth3r:
Generally, I agree that it would help us to be able to identify orders by ID. Currently, first thing the solver does is to give every order an ID (basically its index in the list with leading zeroes, e.g., '007'), and then stores them as dict for the rest of the execution. For this, I'd prefer an ID that is unique in itself over having a combination of (orderID, accountID).

An easy suggestion would be that you just add a field `orderID` as you proposed, which we don't touch but carry throughout our code; while we stick to our current internal ID.

Would that work for everyone?

e00E:
As long as the order id stays there in the solution it doesn't matter what happens internally and it should still be possible to omit orders that have zero execution amount in the solution.

fleupold:
My original idea was to restart populating the globally unique ID on the services side (as was the case before) and make it so that we can derive the per user ID from that value. However, the proposed solution here seems cleaner.
If @twalth3r is on board with passing this value through to the solution file, this seems like a good way to go.

twalth3r:
>If @twalth3r is on board with passing this value through to the solution file, this seems like a good way to go.

I'd be fine with that! It would just need to be called `orderID` (or whatever) and not just `ID` (which we use internally). Do we need to have this passing-through logic ready when this PR is merged, or are you not yet relying on it?

e00E:
[...]

PR #557, opened 2020-02-19
Use order_id for solver input

e00E:
first step of #546

The solver will propagate the order_id to the solution. This makes it easier to to see which order for example should be black listed. Later it will allow us to omit untouched orders in the solution.

### DIFF
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -46,7 +46,7 @@ mod solver_input {
         pub buy_token: String,
         pub sell_amount: String,
         pub buy_amount: String,
-        pub order_id: String,
+        pub order_id: u16,
     }
 
     #[derive(Serialize)]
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Love it!

nlordell (approved):

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -46,6 +46,7 @@ mod solver_input {
         pub buy_token: String,
         pub sell_amount: String,
         pub buy_amount: String,
+        pub order_id: String,
```
nlordell: Should this just be a u16?

e00E: True.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FRl9Yqavfj2Ula%2Fgiphy.gif&token=easy" data-gitmeme-token="easy"></a>

PR #558, opened 2020-02-19
First draft of fallback and standard solver

josojo:
The current choices between the optimization models MIP and NLP are not good, as the benchmarks have shown that NLP kind of always underperforms MIP.

This PR implements a real fallback solver: which uses as well the MIP and price feed queried by the solver.
Later on, we will use the query price feeds from the driver, this is really just an intermediate step for the trading challenge this evening.

The PR https://gitlab.gnosisdev.com/devops/gnosis-staging/merge_requests/181
does the necessary changes in the deployment configs 

---
testplan:
```
T1: docker-compose build --build-arg SOLVER_BASE=163030813197.dkr.ecr.us-east-1.amazonaws.com/dex-solver:v0.5.11 stablex
```
put SOLVER_TYPE=fallback-solver into common.env
```
docker-compose up stablex

T2: cargo test -p e2e ganache -- --nocapture
```
Check that a solution is provided.

PS: pricefeed work currently only for mainnet in dex-solver.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -14,7 +14,7 @@ use crate::contracts::{stablex_contract::BatchExchange, web3_provider};
 use crate::driver::stablex_driver::StableXDriver;
 use crate::metrics::{MetricsServer, StableXMetrics};
 use crate::orderbook::{FilteredOrderbookReader, PaginatedStableXOrderBookReader};
-use crate::price_finding::price_finder_interface::OptimizationModel;
+use crate::price_finding::price_finder_interface::SolverType;
 use crate::price_finding::Fee;
 use crate::solution_submission::StableXSolutionSubmitter;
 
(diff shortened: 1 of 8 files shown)


driver/src/price_finding/price_finder_interface.rs
```diff
@@ -23,29 +23,35 @@ impl Default for Fee {
...
+            SolverType::StandardSolver => &"--optModel=mip",
+            // TODO: Allow to hand over several args for the optimizer
+
+            // The fallback solver is also running --optModel=mip, as this is the default value
+            // although it is not handed over in the next line
+            SolverType::FallbackSolver => {
+                &"--tokenInfo=/app/batchauctions/scripts/e2e/token_info_mainnet.json"
+            }
+            SolverType::NaiveSolver => {
                 panic!("OptimizationSolver should not be called with naive solver")
```
bh2smith: Are we not using `NLP` anymore?

josojo: Yes, the benchmarks were so disencouraging that we abandon it for now

driver/src/main.rs
```diff
@@ -90,9 +91,8 @@ struct Options {
 fn main() {
     let options = Options::from_args();
 
-    let optimization_model_string: String =
-        env::var("SOLVER_TYPE").unwrap_or_else(|_| String::from("NaiveSolver"));
-    let optimization_model = SolverType::from(optimization_model_string.as_str());
+    let (_, _guard) = logging::init(&options.log_filter);
+    info!("using options: {:#?}", options);
```
bh2smith: I wonder if options it’s too vague here for a long statement. It looks like these correspond to runtime arguments, could we say something like that.

nlordell: I’ll open a PR for that, it’s already in master and would be an unrelated change.

PR #559, opened 2020-02-19
Some type-safety with solver input and output

nlordell:
This PR implements serialization types for solver input and output data. The main motivation behind this is that the new price estimation stuff would require its own conversion code in order to serialize information to the solver, and these types just make code reuse a bit easier.

I'm looking for feedback to see if you think this is useful, if not we can scrap it.

### Test Plan

CI


e00E (approved):
I like it and I think its an improvement even right now.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -119,116 +206,34 @@ impl OptimisationPriceFinder {
...
-fn serialize_tokens(orders: &[models::Order]) -> Vec<String> {
+fn serialize_tokens(orders: &[models::Order]) -> BTreeSet<TokenId> {
     // Get collection of all token ids appearing in orders
-    let mut token_ids = orders.iter().map(|o| o.buy_token).collect::<HashSet<u16>>();
-    token_ids.extend(orders.iter().map(|o| o.sell_token));
-
-    token_ids.into_iter().map(token_id).collect::<Vec<String>>()
+    orders
+        .iter()
+        .flat_map(|o| vec![TokenId(o.buy_token), TokenId(o.sell_token)])
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FLPAohuvTif0ek%2Fgiphy.gif&token=sugar" data-gitmeme-token="sugar"></a>

fleupold (approved):
Looks very nice. Is this going to conflict with #555 ?

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -119,116 +206,34 @@ impl OptimisationPriceFinder {
...
-fn token_id(token: u16) -> String {
-    format!("T{:04}", token)
-}
-
-fn account_id(account: H160) -> String {
-    format!("{:x}", account)
-}
-
-fn serialize_tokens(orders: &[models::Order]) -> Vec<String> {
+fn serialize_tokens(orders: &[models::Order]) -> BTreeSet<TokenId> {
```
josojo: Why do we use the BtreeSet here? Wouldn't a hashmap perform better?

josojo (approved):
Nice, I like the PR a lot.

Either we merge #555 or this PR first, both ways are fine for me.

driver/src/price_finding/optimization_price_finder.rs
nlordell: The solver expects a token collection that is in order and without duplicate token identifiers. Using a `BTreeSet` does both these things for in a way that is guaranteed by the data structure. Additionally, the JSON serialized array that this produces will also always be in token ID order and does not require any extra serialization code, or ensuring that the `Vec` is sorted and only contains unique elements before serializing.

The main motivation was purely to have a data structure that reflects exactly how the data is meant to look like. In terms of performance, I don't think there is a significant difference. Note that sorting a `Vec` is `O(n*log(n))` and inserting `n` items into a `BTreeSet` is also `O(n*log(n))` so, asymptotically, they should perform the same.

nlordell:
> Either we merge #555 or this PR first, both ways is fine for me.

I'm OK either way.

driver/src/price_finding/optimization_price_finder.rs
nlordell: Also, places like here: https://github.com/gnosis/dex-services/blob/3321a911e655c4a47be657bc1c5611823ff17d20/driver/src/price_finding/optimization_price_finder.rs#L169

I changed from a `HashMap` to a `BTreeMap` as the data is expected to be serialized in account ID order. While a `HashMap` would perform better for insertion and random access, when serializing we were collecting the `HashMap` into a `BTreeMap` to ensure that it would get serialized in order. Since the `solver_input::Input` type is only used for serialization, I thought it would be better to just create it with the `BTreeMap` directly and skip creating a `HashMap` only to create a `BTreeMap` afterwards.

If we were using this data internally in the driver as well, then using a `HashMap` would probably have performed better as you suggested.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -3,98 +3,181 @@ use crate::price_finding::error::{ErrorKind, PriceFindingError};
...
+                sell_token: TokenId(order.sell_token),
+                buy_token: TokenId(order.buy_token),
+                sell_amount: Num(order.sell_amount),
+                buy_amount: Num(order.buy_amount),
+            }
+        }
     }
 
-    pub type Accounts = HashMap<String, HashMap<String, String>>;
+    pub type Accounts = BTreeMap<H160, BTreeMap<TokenId, Num>>;
```
bh2smith: Whats the benefit of `BTree` instead of `HashMap`?

bh2smith (commented):
A lot to read on the last file, but overall looks a lot better!

driver/src/price_finding/optimization_price_finder.rs
nlordell: `BTreeMap`s keep the key value pairs in a sorted (on key) binary tree, the cool property about this is that a depth first traversal of the tree will return the key value pairs sorted on key.

As we wanted to serialize the accounts such that they are alphabetically sorted on the account address, using a `BTreeMap` here just gives us that for free.

PR #560, opened 2020-02-19
[Easy] Remove `Error::description`

nlordell:
`Error::description` is soft-deprecated as per the standard docs:https://doc.rust-lang.org/std/error/trait.Error.html#method.description. This PR removes all references to it.

### Test Plan

CI


e00E (approved):

bh2smith (commented):
Noice!

PR #562, opened 2020-02-20
Initial Kraken API implementation

nlordell:
This is the initial implementation of a Kraken HTTP API to be used by the price oracle. The full integration with the solver will come piecemeal

Note that the API is currently not used anywhere, but its a way to build up to a larger feature in manageable reviewable PRs.

### Test Plan

CI with new unit tests. Also, a test that does not run by default is included and can be run by:
```
cargo test online_kraken_api -- --ignored --nocapture
```

### DIFF
--- a/driver/src/macros.rs
+++ b/driver/src/macros.rs
@@ -1,4 +1,4 @@
-//! Module containing utility macros for sharing for in the crate.
+//! Module containing utility macros for sharing in the crate.
 
 /// Macro for instanciating a `HashMap`. Note that `ToOwned::to_owned` is called
 /// for keys, so things like `str` keys atomatically get turned into `String`s.
--- a/driver/src/macros.rs
+++ b/driver/src/macros.rs
@@ -1,0 +1,15 @@
+//! Module containing utility macros for sharing for in the crate.
+
+/// Macro for instanciating a `HashMap`. Note that `ToOwned::to_owned` is called
+/// for keys, so things like `str` keys atomatically get turned into `String`s.
+#[cfg(test)]
+macro_rules! hash_map {
+    ($( $key:expr => $value:expr ),* $(,)?) => {{
+        #[allow(unused_mut)]
+        let mut map = std::collections::HashMap::new();
+        $(
+            map.insert(($key).to_owned(), $value);
+        )*
+        map
+    }}
+}
--- a/driver/src/price_estimation/kraken/api.rs
+++ b/driver/src/price_estimation/kraken/api.rs
@@ -1,0 +1,265 @@
+use anyhow::{anyhow, Context, Result};
+use isahc::prelude::*;
+use serde::Deserialize;
+use serde_with::rust::display_fromstr;
+use std::collections::HashMap;
+
+/// A trait representing a Kraken API client.
+///
+/// Note that this is not the full API, only the subset required for the
+/// retrieving price estimates for the solver.
+#[cfg_attr(test, mockall::automock)]
+pub trait KrakenApi {
+    /// Retrieves the list of supported assets.
+    fn assets(&self) -> Result<HashMap<String, Asset>>;
+    /// Retrieves the list of supported asset pairs.
+    fn asset_pairs(&self) -> Result<HashMap<String, AssetPair>>;
+    /// Retrieves ticker information (with recent prices) for the given asset
+    /// pair identifiers.
+    fn ticker<'a>(&'a self, pairs: &'a [&'a str]) -> Result<HashMap<String, TickerInfo>>;
+}
+
+/// An HTTP Kraken API Client.
+#[derive(Debug)]
+pub struct KrakenHttpApi {
+    /// The base URL for the API calls.
+    base_url: String,
+    /// An HTTP client for all of the HTTP requests.
+    client: HttpClient,
+}
+
+/// The default Kraken API base URL.
+const DEFAULT_API_BASE_URL: &str = "https://api.kraken.com/0/public";
+
+impl KrakenHttpApi {
+    pub fn new() -> Result<Self> {
+        KrakenHttpApi::with_url(DEFAULT_API_BASE_URL)
+    }
+
+    pub fn with_url(base_url: &str) -> Result<Self> {
+        let client = HttpClient::new().context("failed to initialize HTTP client")?;
+        Ok(KrakenHttpApi {
+            base_url: base_url.into(),
+            client,
+        })
+    }
+}
+
+impl KrakenApi for KrakenHttpApi {
+    fn assets(&self) -> Result<HashMap<String, Asset>> {
+        self.client
+            .get(format!("{}/Assets", self.base_url))
+            .context("failed to retrieve list of assets from Kraken")?
+            .json::<KrakenResult<_>>()
+            .context("failed to parse assets JSON")?
+            .into_result()
+    }
+
+    fn asset_pairs(&self) -> Result<HashMap<String, AssetPair>> {
+        self.client
+            .get(format!("{}/AssetPairs", self.base_url))
+            .context("failed to retrieve list of asset pairs from Kraken")?
+            .json::<KrakenResult<_>>()
+            .context("failed to parse asset pairs JSON")?
+            .into_result()
+    }
+
+    fn ticker(&self, pairs: &[&str]) -> Result<HashMap<String, TickerInfo>> {
+        self.client
+            .get(format!("{}/Ticker?pair={}", self.base_url, pairs.join(",")))
+            .context("failed to retrieve ticker infos from Kraken")?
+            .json::<KrakenResult<_>>()
+            .context("failed to parse ticker JSON")?
+            .into_result()
+    }
+}
+
+/// The result type that is returned by Kraken on API requests. This type is
+/// only used internally.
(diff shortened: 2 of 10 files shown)


driver/src/macros.rs
```diff
@@ -0,0 +1,15 @@
+//! Module containing utility macros for sharing for in the crate.
```
josojo: one for too much

josojo (approved):
Thanks for splitting it up :)

Looks all good to me 👍

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,265 @@
...
+/// The default Kraken API base URL.
+const DEFAULT_API_BASE_URL: &str = "https://api.kraken.com/0/public";
+
+impl KrakenHttpApi {
+    pub fn new() -> Result<Self> {
+        KrakenHttpApi::with_url(DEFAULT_API_BASE_URL)
+    }
+
+    pub fn with_url(base_url: &str) -> Result<Self> {
+        let client = HttpClient::new().context("failed to initialize HTTP client")?;
```
e00E: The client should probably have a timeout.

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,265 @@
...
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct Asset {
+    pub altname: String,
+}
+
+impl Asset {
+    /// Create a new asset from an alternate name.
+    #[cfg(test)]
+    pub(crate) fn new(altname: &str) -> Asset {
```
e00E: What is the purpose of `pub(crate)` here? No need to comment that in the code. I would just like to know.

e00E (approved):

driver/src/price_estimation/kraken/api.rs
nlordell: Oh! that was probably a copy pasta booboo, I'll fix it.

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+}
+
+/// A price pair used in the ticker info, where the first field is today's price
+/// and the second field is from the last 24 hours.
+#[derive(Copy, Clone, Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct PricePair(
+    #[serde(with = "display_fromstr")] pub f64,
```
fleupold: nit: does this have to public?

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,272 @@
...
+/// The default timeout used for HTTP requests to Kraken.
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);
+
+impl KrakenHttpApi {
+    pub fn new() -> Result<Self> {
+        KrakenHttpApi::with_url(DEFAULT_API_BASE_URL, DEFAULT_TIMEOUT)
+    }
+
+    pub fn with_url(base_url: &str, timeout: Duration) -> Result<Self> {
+        let client = HttpClient::builder()
```
fleupold: Should we be using the same client as the GasEstimation/RPC transport and hide all isahc specific implementation details?

It would also be useful to measure the round trip times/sizes for all requests we make and report them into prometheus.

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,272 @@
...
+                "XETHZUSD" => AssetPair::new("XETH", "ZUSD"),
+            }
+        );
+    }
+
+    #[test]
+    fn parse_ticker_infos_json() {
+        // Sample retrieved from https://api.kraken.com/0/public/Ticker?pair=DAIUSD,ETHUSD
+        let value: HashMap<String, TickerInfo> = deserialize(
+            r#"{"error":[],"result":{"DAIUSD":{"a":["1.00105000","3091","3091.000"],"b":["0.99989000","628","628.000"],"c":["0.99986000","115.05510869"],"v":["267638.81097379","587455.25359912"],"p":["0.99938164","0.99862453"],"t":[1219,2709],"l":["0.99696000","0.99608000"],"h":["1.00300000","1.00300000"],"o":"0.99828000"},"XETHZUSD":{"a":["257.73000","124","124.000"],"b":["257.72000","1","1.000"],"c":["257.81000","6.60000000"],"v":["99246.07873952","130832.28052933"],"p":["251.77023","248.47638"],"t":[13177,18557],"l":["238.06000","230.57000"],"h":["259.48000","259.48000"],"o":"238.06000"}}}"#,
```
bh2smith: Seems like the only relevant values here are `"p":["0.99938164","0.99862453"]`. What is all this other stuff?

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,272 @@
...
+/// information.
+///
+/// Note that this is only a small subset of the data provided by the Kraken API
+/// and only the parts required for retrieving price estimates for the solver
+/// are included.
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct TickerInfo {
+    pub p: PricePair,
+}
```
bh2smith: I am confused by the name of this struct, since nothing about it seems to have anything to do with ticking of any sort.

driver/src/price_estimation/kraken/api.rs
```diff
@@ -0,0 +1,272 @@
...
+/// and the second field is from the last 24 hours.
+#[derive(Copy, Clone, Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct PricePair(
+    #[serde(with = "display_fromstr")] pub f64,
+    #[serde(with = "display_fromstr")] pub f64,
+);
+
+impl PricePair {
+    /// Retrieves the price since midnight.
```
bh2smith: Is this the price "at midnight" because prices fluctuate "since" a point in time and I would expect to see more than just a single value here... same question for last 24 hours.

bh2smith (commented):
Seems all fine, I am just confused by the naming and was surprised to see the equality assertions in the unit tests

driver/src/price_estimation/kraken/api.rs
nlordell: As mentioned in the comment: `Sample retrieved from https://api.kraken.com/0/public/Ticker?pair=DAIUSD,ETHUSD`. Its a sample JSON document returned by the Kraken API. This test just makes sure that it can be correctly parsed.

Since our implementation only cares about the `"p":["0.99938164","0.99862453"]`, that's all we parse from the JSON document.

driver/src/price_estimation/kraken/api.rs
nlordell: I think the word "ticker" is related to stock tickers, and represents the price widgets you see in the Kraken interface. The name was chosen based on the Kraken API documentation:
https://www.kraken.com/features/api#get-ticker-info

driver/src/price_estimation/kraken/api.rs
nlordell: From what I understood from the documentation price pairs always contain two values, one which is "today", or since 00:00, and the other which is for the "last 24h". Depending on what the price pair is for, it can mean different things. So the ticker info JSON that is retrieved from Kraken, it contains things like the low, high and weighted average prices and for each of these, the price pair respectively represents the low price since 00:00 and the low price in the last 24h, the high price since 00:00 and the high price in the last 24h, and the weighted average price since 00:00 and in the last 24h.

nlordell:
@bh2smith Sorry for the confusing names, I hope my explanations helped clear things up. Maybe a few more comments can be added in a follow up PR or something to make things more self-explanatory.

When I wrote this I tried to keep everything a 1 to 1 with the Kraken API documented [here](https://www.kraken.com/features/api) to make it easier to look through there documentation for help is something happens. That being said, more comments never hurt!

driver/src/price_estimation/kraken/api.rs
bh2smith: But what is the "p" does it stand for "price"? "pair"? why does it contain exactly two values?

nlordell: The `p` comes from the Kraken API: 
> ```
> <pair_name> = pair name
>     a = ask array(<price>, <whole lot volume>, <lot volume>),
>     b = bid array(<price>, <whole lot volume>, <lot volume>),
>     c = last trade closed array(<price>, <lot volume>),
>     v = volume array(<today>, <last 24 hours>),
>     p = volume weighted average price array(<today>, <last 24 hours>),
>     t = number of trades array(<today>, <last 24 hours>),
>     l = low array(<today>, <last 24 hours>),
>     h = high array(<today>, <last 24 hours>),
>     o = today's opening price
> ```

bh2smith: This is very nice thanks for going the extra mile!

bh2smith: Whoever wrote that API must be a python dev without any guidance (cf. single character identifiers).

PR #564, opened 2020-02-20
Change all occurences of `H160` to `Address`

nlordell:
:point_up: `Address` reads better in the code IMO.


fleupold (approved):
I'd also be fine changing all occurrences of H160 to Address as this type reads more intuitive.

nlordell:
> I'd also be fine changing all occurrences of H160 to Address as this type reads more intuitive.

I was hoping you'd say that :)

bh2smith (commented):
Yes amazing, this is exactly what I always wanted! Does this mean you recently introduced an `Address` type in ethcontracts? Is this just implemented as an alias to `H160`? Great thanks a million!

PR #567, opened 2020-02-20
Implement ERC20 token reading

nlordell:
This PR implements token reading from the block chain. This token information will be used by price estimations. Specifically the number of decimals and token symbols (used in order to search exchanges for prices).

### Test Plan

None

### DIFF
--- a/driver/src/contracts/erc20.rs
+++ b/driver/src/contracts/erc20.rs
@@ -1,0 +1,63 @@
+//! This module implements the ERC20 smart contract interface as well a trait
+//! and implementation to read it from the block chain.
+
+use super::Web3;
+use crate::models::TokenInfo;
+use crate::util::FutureWaitExt;
+use anyhow::Result;
+use ethcontract::H160;
+use lazy_static::lazy_static;
+use std::collections::HashMap;
+
+/// Trait for reading ERC20 token information from the block chain.
+#[cfg_attr(test, mockall::automock)]
+pub trait TokenReading {
+    /// Retrieve token information by address.
+    fn read_token_info(&self, token_address: H160) -> Result<TokenInfo>;
+}
+
+include!(concat!(env!("OUT_DIR"), "/erc20_detailed.rs"));
+
+/// Web3 `TokenReading` implementation.
+pub struct TokenReader {
+    web3: Web3,
+}
+
+impl TokenReader {
+    /// Create a new token reader from a web3 instnace.
+    #[allow(dead_code)]
+    pub fn new(web3: Web3) -> Self {
+        TokenReader { web3 }
+    }
+}
+
+lazy_static! {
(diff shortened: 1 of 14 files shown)


fleupold (commented):
What's the strategy for reading this information? Would we be doing this for all tokens on the exchange or only for a set of whitelisted one?

I'm worried about (a) the performance of querying n addresses and then running two lookups for each token and (b) about the possibility of adding a Fake token with a common symbol like DAI and thus influencing the Price finder in a potentially weird way.

What do you think?

driver/src/contracts/erc20.rs
```diff
@@ -0,0 +1,63 @@
...
+
+lazy_static! {
+    static ref SYMBOL_OVERRIDES: HashMap<String, String> = hash_map! {
+        "WETH" => "ETH".to_owned(),
+    };
+}
+
+impl TokenReading for TokenReader {
+    /// Retrieve token information by address.
+    fn read_token_info(&self, token_address: H160) -> Result<TokenInfo> {
```
josojo: Regarding Felix comment:
1) Performance issue:
If ee could do a promise all for many token addresses, and if do these requests only once a day (potentially even only at the instance start), then I would guess that performance is not an issue.

2) I would like to get all possible symbol information into the json-instance description. Even if some is fake, generally, I think that it can be helpful during debugging. 
But for sure, we should have a hardcoded list between SYMBOL and "real token", and only for tokens appearing in the list, I would propose to get a price feed. This would also allow us an easy identification of fake token symbols during the instance investigation.

nlordell:
> What's the strategy for reading this information? Would we be doing this for all tokens on the exchange or only for a set of whitelisted one?

These tokens should be filtered IMO. I'm OK with a whitelist here since we will probably whitelist all tokens that are commonly found on exchanges for which we can find prices.

> (a) the performance of querying n addresses and then running two lookups for each token

I had it implemented in my draft PR that this information would be read once and cached (since the exchange can't unlist tokens, this is super easy). Then just the number of tokens needs to be checked each run loop and potentially retry some of the failed requests.

nlordell:
That being said, if we are going to whitelist tokens, then reading this information might be kind of pointless. It might make more sense to just include it there.

nlordell:
Marking this on hold until #571 gets merged as this would be an extension to that.

bh2smith:
I see that this was "closed" by a bot, but it does not appear it was meant to be closed... rather merged.

nlordell:
@bh2smith I think the issue was stale, and from what I understood we don't really have a reason right now to get ERC20 token information from the block chain in the driver. That is because we only want to whitelist tokens for price estimation, and in doing that the required information that would need to get read from the block chain gets provided through the `PRICE_FEED_INFORMATION` environment variable anyway.

We can revisit this if it is needed in the future.

PR #568, opened 2020-02-20
Use `anyhow` errors

nlordell:
Tried converting crate error types to use `anyhow`. Its pretty convenient and allows adding arbitrary `.context()` to an error, which can be used to sort of create back traces and help distinguish error with the same type and message happening in different locations.

We already started using `anyhow` in other places, this just converts all errors to this new kind. It also avoid having to write a bunch of `From<Error> to Error` code.

### Test Plan

CI


fleupold (approved):
I like that it removes a lot of custom result types and boilerplate code. My concern would be that it's no harder to "handle" errors as they are not as typesafe/distinguishable as before (with custom types). That being said we don't really handle them in any sophisticated away (other than logging), so I guess having all information encoded as Strings is fine.

nlordell:
> My concern would be that it's no harder to "handle" errors as they are not as typesafe/distinguishable

Anyhow lets you do things like:
```
// If the error was caused by redaction, then return a tombstone instead
// of the content.
match root_cause.downcast_ref::<DataStoreError>() {
    Some(DataStoreError::Censored(_)) => Ok(Poll::Ready(REDACTED_CONTENT)),
    None => Err(error),
}
```
This keeps the full type information for the error, instead of converting the error message to a `String` like we currently do. It actually has a bunch of awesome features which make it ideal for being the "top-level" error type (which is what we want in our case as `driver` is an application and doesn't have methods that are meant to be consumed by other crates).

driver/src/driver/stablex_driver.rs
```diff
@@ -139,17 +137,22 @@ mod tests {
         let state = AccountState::with_balance_for(&orders);
 
         let batch = U256::from(42);
-        reader.expect_get_auction_index().return_const(Ok(batch));
+        reader
+            .expect_get_auction_index()
+            .return_once(move || Ok(batch));
```
bh2smith: Is this somehow related to anyhow? is `move` something native to anyhow?

driver/src/price_finding/naive_solver.rs
```diff
@@ -112,11 +112,7 @@ impl NaiveSolver {
 }
 
 impl PriceFinding for NaiveSolver {
-    fn find_prices(
-        &self,
-        orders: &[Order],
-        state: &AccountState,
-    ) -> Result<Solution, PriceFindingError> {
+    fn find_prices(&self, orders: &[Order], state: &AccountState) -> Result<Solution> {
```
bh2smith: So we no longer include Error as part of the `Result`?

driver/src/transport.rs
```diff
@@ -28,7 +28,7 @@ struct HttpTransportInner {
 
 impl HttpTransport {
     /// Creates a new HTTP transport with settings.
-    pub fn new(url: impl Into<String>, timeout: Duration) -> Result<HttpTransport, DriverError> {
+    pub fn new(url: impl Into<String>, timeout: Duration) -> Result<HttpTransport, Error> {
```
bh2smith: But here.. we do return Error as an option with Result...

driver/src/solution_submission/mod.rs
```diff
@@ -1,16 +1,13 @@
...
 use crate::models::{Order, Solution};
 
+use anyhow::Result;
+use ethcontract::U256;
 #[cfg(test)]
 use mockall::automock;
 
-use ethcontract::U256;
-
-type Result<T> = std::result::Result<T, DriverError>;
```
bh2smith: Does this have something to do with the awkward `Result<SometimesErrorSometimesNot>`?

bh2smith (commented):
Seems cool, just a few questions regarding my confusion surrounding `Result<Error>`

driver/src/price_finding/naive_solver.rs
nlordell: The `anyhow` has a result type alias like this: `type Result<T> = std::result::Result<T, anyhow::Error>`, it basically means all our results use the same error type.

driver/src/transport.rs
nlordell: This is because there is no `use anyhow::Result;` in this file, as we need to distinguish between different error types here.

driver/src/driver/stablex_driver.rs
e00E: `move` can be put in front of a closure (lambda function) to indicate that the variables it captures from the environment should be moved into the closure instead of being references. So in this case it means `bach` is moved into the closure so that the closure can return it . 
https://doc.rust-lang.org/stable/book/ch13-01-closures.html#capturing-the-environment-with-closures

PR #570, opened 2020-02-21
SOLVER_TIME_LIMIT env introduction

josojo:
This PR introduces a new env SOLVER_TIME_LIMIT that allows us to limit the time the solver is taking for running the optimization process.

Checkout the associated new yml files;
https://gitlab.gnosisdev.com/devops/gnosis-staging/merge_requests/185

---
testplan:

```
docker pull 163030813197.dkr.ecr.us-east-1.amazonaws.com/dex-solver:v0.5.13
docker-compose build --build-arg SOLVER_BASE=163030813197.dkr.ecr.us-east-1.amazonaws.com/dex-solver:v0.5.13 stabl
ex
```
insert into common.env:
```
SOLVER_TYPE=standard-solver
SOLVER_TIME_LIMIT=120
```
and run an e2e test:
```
T1:docker-compose down && docker-compose up stablex truffle
T2: cargo test -p e2e ganache -- --nocapture
```
observe that a solution can be found.


driver/src/main.rs
```diff
@@ -63,9 +63,15 @@ struct Options {
...
     solver_type: SolverType,
 
+    /// JSON encoded object of back-up price information in case the oracle functionality is
+    /// not working.
     #[structopt(long, env = "PRICE_FEED_INFORMATION", default_value = "{}")]
     backup_token_data: TokenData,
 
+    /// Number of seconds the solver should maximally use for the optimization process
+    #[structopt(long, env = "SOLVER_TIME_LIMIT", default_value = "150")]
+    solver_time_limit: u32,
```
fleupold: Since this and TokenData only apply to the Standard and Fallback Solver can we make it a field on the enum subtypes instead of passing more and more data into create_price_finder?

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -247,13 +259,19 @@ fn write_input(input_file: &str, input: &str) -> std::io::Result<()> {
     Ok(())
 }
 
-fn run_solver(input_file: &str, result_folder: &str, solver_type: SolverType) -> Result<()> {
+fn run_solver(
+    input_file: &str,
+    result_folder: &str,
+    solver_type: SolverType,
+    solver_time_limit: u32,
```
fleupold: This would also allow to encapsulate the logic inside the `solver_type.to_args()` method

driver/src/main.rs
```diff
@@ -60,7 +60,7 @@ struct Options {
     /// solver; 'MIP' for mixed integer programming solver; 'NLP' for non-linear
     /// programming solver.
     #[structopt(long, env = "SOLVER_TYPE", default_value = "NaiveSolver")]
-    solver_type: SolverType,
+    solver_type: String,
```
fleupold: nit: the other options are spelled fallback-solver & standard-solver, so maybe naive-solver?

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -23,37 +20,36 @@ impl Default for Fee {
...
+impl SolverConfig {
+    pub fn new(solver_type_str: &str, solver_time_limit: u32) -> Self {
         match solver_type_str.to_lowercase().as_str() {
-            "standard-solver" => Ok(SolverType::StandardSolver),
-            "fallback-solver" => Ok(SolverType::FallbackSolver),
+            "standard-solver" => SolverConfig::StandardSolver(solver_time_limit),
+            "fallback-solver" => SolverConfig::FallbackSolver(solver_time_limit),
             // the naive solver is the standard solver.
-            _ => Ok(SolverType::NaiveSolver),
+            _ => SolverConfig::NaiveSolver,
```
fleupold: I think it's more safe to panic in the default case (since this happens on startup) and require people to actually use valid configs. Otherwise a small type in standard-sovler will give you the naive solver without a warning.

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -23,37 +20,35 @@ impl Default for Fee {
...
+    pub fn new(solver_type_str: &str, solver_time_limit: u32) -> Self {
         match solver_type_str.to_lowercase().as_str() {
-            "standard-solver" => Ok(SolverType::StandardSolver),
-            "fallback-solver" => Ok(SolverType::FallbackSolver),
-            // the naive solver is the standard solver.
-            _ => Ok(SolverType::NaiveSolver),
+            "standard-solver" => SolverConfig::StandardSolver(solver_time_limit),
+            "fallback-solver" => SolverConfig::FallbackSolver(solver_time_limit),
+            "naive-solver" => SolverConfig::NaiveSolver,
+            _ => panic!("solver type does not exist"),
```
e00E: Nit: It is cleaner to continue to have this function return `Result`. Then we can panic at the point where that result is needed by unwrapping. To do this we would return `anyhow::Result<Self>` in this function and then `Err(anyhow!("solver type does not exit"))` instead of panic.

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -23,37 +20,35 @@ impl Default for Fee {
 }
 
 #[derive(Clone, Debug, Copy, PartialEq)]
-pub enum SolverType {
+pub enum SolverConfig {
     NaiveSolver,
-    StandardSolver,
-    FallbackSolver,
+    StandardSolver(u32), //params for --solverTimeLimit
+    FallbackSolver(u32),
```
e00E: Nit: I would make named like `StandardSolver{ time_limit: u32 }`.

driver/src/price_finding/mod.rs
```diff
@@ -5,20 +5,20 @@ pub mod price_finder_interface;
...
     price_oracle: impl PriceEstimating + 'static,
 ) -> Box<dyn PriceFinding> {
-    if solver_type == SolverType::NaiveSolver {
+    if solver_type == SolverConfig::NaiveSolver {
         info!("Using naive price finder");
         Box::new(NaiveSolver::new(fee))
     } else {
         info!(
-            "Using optimisation price finder with the args {:}",
+            "Using optimization price finder with the args {:?}",
```
bh2smith: Thanks for the "zed"!

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -523,7 +522,9 @@ pub mod tests {
             run_solver: |_, _, _| Ok(()),
             read_output: |_| Err(std::io::Error::last_os_error()),
             fee: Some(fee),
-            solver_type: SolverType::StandardSolver,
+            solver_type: SolverConfig::StandardSolver {
+                solver_time_limit: 180,
```
bh2smith: is this a default or did we just hardcode the three minutes in right here?

josojo: here, it is just a test parameter, which is not relevant for the actual test

PR #571, opened 2020-02-21
Kraken price oracle

### LINKED ISSUE #507: Implement external price feed
The logic implemented in https://github.com/gnosis/dex-solver/pull/121 breaks the interface between the services and the solver. The solver should not rely on external information that it has to fetch e.g. via network. It should work just based on the input json (cf. [comment](https://github.com/gnosis/dex-solver/pull/121#issuecomment-582513764))

This task captures to implement the centralized price oracle fetching logic in the services.

nlordell:
This PR implements the price oracle that estimates prices from Kraken and falls back to the prices provided from the environment token data.

Closes #507 

### Test Plan

New unit tests + CI

### DIFF
--- a/driver/src/price_estimation/mod.rs
+++ b/driver/src/price_estimation/mod.rs
@@ -188,9 +188,9 @@ mod tests {
 
     #[test]
     fn price_oracle_fetches_token_prices() {
-        let tokens = TokenData::test(hash_map! {
-            TokenId(1) => TokenInfo::test("WETH", 18, 0),
-            TokenId(2) => TokenInfo::test("USDT", 6, 0),
+        let tokens = TokenData::from(hash_map! {
+            TokenId(1) => TokenInfo::new("WETH", 18, 0),
+            TokenId(2) => TokenInfo::new("USDT", 6, 0),
         });
 
         let mut source = MockPriceSource::new();
@@ -199,7 +199,7 @@ mod tests {
             .withf(|tokens| {
                 let mut tokens = tokens.to_vec();
                 tokens.sort_unstable_by_key(|token| token.id);
-                tokens == [Token::test(1, "WETH", 18), Token::test(2, "USDT", 6)]
+                tokens == [Token::new(1, "WETH", 18), Token::new(2, "USDT", 6)]
             })
             .returning(|_| {
                 Ok(hash_map! {
@@ -220,8 +220,8 @@ mod tests {
             prices,
             btree_map! {
                 TokenId(0) => None,
-                TokenId(1) => Some(TokenInfo::test("WETH", 18, 0)),
-                TokenId(2) => Some(TokenInfo::test("USDT", 6, 1_000_000_000_000_000_000)),
+                TokenId(1) => Some(TokenInfo::new("WETH", 18, 0)),
+                TokenId(2) => Some(TokenInfo::new("USDT", 6, 1_000_000_000_000_000_000)),
                 TokenId(3) => None,
             }
         );
@@ -229,8 +229,8 @@ mod tests {
 
     #[test]
     fn price_oracle_ignores_source_error() {
-        let tokens = TokenData::test(hash_map! {
-            TokenId(1) => TokenInfo::test("WETH", 18, 0),
+        let tokens = TokenData::from(hash_map! {
+            TokenId(1) => TokenInfo::new("WETH", 18, 0),
         });
 
         let mut source = MockPriceSource::new();
@@ -245,7 +245,7 @@ mod tests {
             prices,
             btree_map! {
                 TokenId(0) => None,
-                TokenId(1) => Some(TokenInfo::test("WETH", 18, 0)),
+                TokenId(1) => Some(TokenInfo::new("WETH", 18, 0)),
             }
         );
     }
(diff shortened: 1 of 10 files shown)


driver/src/price_estimation/mod.rs
```diff
@@ -1,30 +1,151 @@
...
+
+/// A trait representing a price oracle that retrieves price estimates for the
+/// tokens included in the current orderbook.
+#[cfg_attr(test, mockall::automock)]
+pub trait PriceEstimating {
+    fn get_token_prices(&self, orders: &[Order]) -> Tokens;
+}
+
+pub struct PriceOracle {
+    /// The token data supplied by the environment. This ensures that only
```
josojo: I would prefer not assuming that a token is whitelisted for the external price feed if the token info is provided. Simply for the reason that we might want to provide data for tokens, which are legitimate, but still there is not price feed available. And easy example would be SNX, as there is barely any centralized exchange trading this token.

But we could add another field to the tokenInfo: `LegibleForAPIPriceProvision` TRUE/FALSE, and then split the list `(mut tokens_to_price, mut unpriced_token_ids)` with this boolean information

josojo (commented):
It is really fun reading your code :)

driver/src/price_estimation/mod.rs
nlordell: I added a `shouldEstimatePrice` flag that, when set to `true` will enable price estimations for that token. Note that it still parses the old token data format if the option is omitted, defaulting to false.

nlordell: I implemented this as part of a separate PR #578 to keep them easier to review.

josojo (approved):
jap, makes perfect sense that you implemented `useAPIPriceProvision` in another PR.

Looks really good!

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -223,7 +212,7 @@ impl PriceFinding for OptimisationPriceFinder {
         state: &models::AccountState,
     ) -> Result<models::Solution> {
         let input = solver_input::Input {
-            tokens: serialize_tokens(&orders, &self.token_data),
+            tokens: self.price_oracle.get_token_prices(&orders),
```
fleupold: Do we know how long this roughly takes? Given that it's on the time critical path for now it would be good to have an idea and maybe capture somewhere a follow up task to do the price estimation just before a batch can be solved.

driver/src/price_estimation/mod.rs
```diff
@@ -1,30 +1,151 @@
...
+            .iter()
+            .flat_map(|order| vec![order.buy_token, order.sell_token])
+            .map(TokenId)
+            // NOTE: Always include the reference token. This is done since the
+            //   solver input specifies the reference token, so for correctness
+            //   it should always be considered.
+            .chain(iter::once(TokenId::reference()))
+            .collect();
+
+        unique_token_ids.into_iter().fold(
```
fleupold: Why not https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FLVqKaedYeLrLa%2Fgiphy.gif&token=price" data-gitmeme-token="price"></a>

driver/src/price_estimation/mod.rs
fleupold: Actually never mind. Saw you response here: https://github.com/gnosis/dex-services/pull/578#discussion_r383028252

driver/src/price_finding/optimization_price_finder.rs
nlordell: I added a timer to the `KrakenClient` test, which tries to retrieve prices for 10 tokens and this is what I got:

```
Total elapsed time: 614.848532ms
```

For reference, the test is here: https://github.com/gnosis/dex-services/blob/09746e7382cf77bbf4df9ca6e6167715ac0ba0cb/driver/src/price_estimation/kraken.rs#L171-L206

nlordell: We might be able to cut down the time a bit if we get the asset and asset pairs together, and maybe check if there is any HTTP caching we can do.

nlordell: Created  #583

fleupold: Ok, well since it's less than 1s it's negligible

driver/src/price_estimation/mod.rs
```diff
@@ -1,30 +1,151 @@
...
+    /// symbol override map, and if it is, then that value is used instead. This
+    /// allows ERC20 tokens like WETH to be treated as ETH, since exchanges
+    /// generally only track prices for the latter.
     fn symbol(&self) -> &str {
-        &self.info.alias
+        lazy_static! {
+            static ref SYMBOL_OVERRIDES: HashMap<String, String> = hash_map! {
+                "WETH" => "ETH".to_owned(),
+            };
+        }
```
bh2smith: Interesting way of fetching `WETH` price!

bh2smith (commented):
Very nice contribution!

PR #574, opened 2020-02-21
[Easy] Token type cleanup

nlordell:
Just cleanup some of the token types that were recently introduced.

### DIFF
--- a/driver/src/price_estimation/mod.rs
+++ b/driver/src/price_estimation/mod.rs
@@ -40,7 +40,7 @@ impl Token {
     pub fn new(id: impl Into<TokenId>, symbol: impl Into<String>, decimals: u8) -> Self {
         Token {
             id: id.into(),
-            address: H160::repeat_byte(index as _),
+            address: H160::zero(),
             info: TokenInfo {
                 alias: symbol.into(),
                 decimals,
--- a/driver/src/price_estimation/mod.rs
+++ b/driver/src/price_estimation/mod.rs
@@ -35,11 +35,11 @@ impl Token {
         (usd_price * 10f64.powi(pow)) as _
     }
 
-    /// Creates a new token with a fictional address for testing.
+    /// Creates a new token from its parameters.
     #[cfg(test)]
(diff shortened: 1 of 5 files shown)


e00E (approved):

josojo (approved):
Clean of the day!

driver/src/price_estimation/mod.rs
```diff
@@ -68,10 +68,10 @@ mod tests {
...
     fn token_get_price() {
         for (token, usd_price, expected) in &[
-            (Token::test(4, "USDC", 6), 0.99, 0.99 * 10f64.powi(30)),
-            (Token::test(7, "DAI", 18), 1.01, 1.01 * 10f64.powi(18)),
-            (Token::test(42, "FAKE", 32), 1.0, 10f64.powi(4)),
-            (Token::test(99, "SCAM", 42), 10f64.powi(10), 10f64.powi(4)),
+            (Token::new(4, "USDC", 6), 0.99, 0.99 * 10f64.powi(30)),
+            (Token::new(7, "DAI", 18), 1.01, 1.01 * 10f64.powi(18)),
+            (Token::new(42, "FAKE", 32), 1.0, 10f64.powi(4)),
+            (Token::new(99, "SCAM", 42), 10f64.powi(10), 10f64.powi(4)),
```
bh2smith: I see that we label these tokens as fake and scam, but does the inflated price have anything to do with it or are we just demonstrating that we can have very large prices?

bh2smith (commented):
Squeaky clean.

driver/src/price_estimation/mod.rs
nlordell: It is to test tokens with more than 18 and more than 36 decimals respectively. More than 36 decimals is a special case as you would expect the OWL price the be 0 even when the USD price is `1.0`, this just tests that if the token has more than 36 decimals it still works for very large USD prices.

bh2smith: Oh wow. I didn't even think about ever having more than 18 decimals!

PR #575, opened 2020-02-21
[Easy] Adding tini as entrypoint

fleupold:
Instead of turning tini to the command, we can make it an entry point of the top level docker, which makes the change in the staging repo unnecessary.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       - '9586:9586'
     depends_on:
       - ganache-cli
-    command: ["/tini", "-v", "--", "cargo", "run"]
+    command: ["cargo", "run"]
     volumes:
       # Sync src folder to allow development without rebuilding on each change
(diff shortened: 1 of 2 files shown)


e00E (approved):

nlordell (approved):
Love it!

bh2smith (commented):
This seems great, but can we get some sort of testplan/instruction on usage for those of us who have no idea what `tini` is an have not encountered it within this project in the past?

nlordell:
@bh2smith https://github.com/krallin/tini

It just starts up the `driver` process and makes sure to properly handle signals. Specifically, it handles SIGTERM (used when the docker container wants to gracefully shutdown for example) which we were not properly handling previously and caused the K8 pods to get stuck in an "exiting" state.

From what I understand it shouldn't affect the development process for the driver, it is more of a DevOps fix.

PR #576, opened 2020-02-21
[Trivial] Move create_price_finder to price_finding module

nlordell:
This just moves the `create_price_finder` function to `price_finding` module so it is closer to the structs that it actually instantiates.

### Test Plan

Not needed, just moving code around.


josojo (approved):
Another clean up of the day. Thanks!

PR #578, opened 2020-02-21
Don't retrieve prices for all tokens

nlordell:
This PR implements a filter on the backup token data to indicate that it should provide the price from the environment and not attempt to retrieve it from Kraken.

### Test Plan

New unit tests + CI


josojo (approved):
Niiiice, exactly as discussed :)

driver/src/price_estimation/mod.rs
```diff
@@ -64,13 +69,13 @@ impl PriceOracle {
         unique_token_ids.into_iter().fold(
             (Vec::new(), Vec::new()),
             |(mut tokens_to_price, mut unpriced_token_ids), id| {
-                if let Some(info) = self.tokens.info(id) {
-                    tokens_to_price.push(Token {
+                match self.tokens.info(id).cloned() {
```
fleupold: Would [partition](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition) be a good use here?

driver/src/main.rs
```diff
@@ -75,10 +74,11 @@ struct Options {
     ///     "alias": "USDC",
     ///     "decimals": 6,
     ///     "external_price": 1000000000000000000000000000000,
+    ///     "should_estimate_price": true
     ///   }
     /// }'
     #[structopt(long, env = "PRICE_FEED_INFORMATION", default_value = "{}")]
-    backup_token_data: TokenData,
+    backup_token_data: TokenFallbackData,
```
fleupold: Is this really "fallback" data? Isn't the information in there (except for the price) used as primary and not fallback data? Also given that the option name is called `PRICE_FEED_INFORMATION` I'd expect a name more similar (or equal) to that.

fleupold (approved):

driver/src/price_estimation/mod.rs
nlordell: TIL! The only issue I see is that partition requires the data in both collections to be the same, here they are slightly different, one is a collection of `(TokenId, TokenInfo) <=> Token` and the other `(TokenId, Option<TokenInfo>)`.

driver/src/main.rs
nlordell: Good point, I'll rename it.

bh2smith (commented):
Cute. What would happen if we say `true` for a token that is not listed in the oracle exchange? For example, I don't believe `GUSD` is listed on Kraken.

nlordell:
@bh2smith It would probably not find a price. Price estimates are returned as a sparse map as not finding a price on an exchange is not considered an error. The following test tries to retrieve prices for 10 tokens and only finds about 4 (if I'm not mistaken):
https://github.com/gnosis/dex-services/blob/008dde51b0889350df62a3d4cf24ed1b659f5c65/driver/src/price_estimation/kraken.rs#L169-L206

PR #579, opened 2020-02-22
[Easy] use hardcoded gas amount

### LINKED ISSUE #561: Use hardcoded gas limit estimation for submitSolution
We ran out of gas in the following tx: https://etherscan.io/tx/0x6b07a0a518d8134138d498811aa1acb47d2eeed9a20aa4562a5c005037583db2 [2]

We were trying to submit a solution, but just before (in the same block) another solution had been successfully submitted by another solver: https://etherscan.io/tx/0x8ded631c94b37433fa1bd956f461815a4e655924120ba34b1fe7e5a9c2ffde4b [1]

While the second solution was sufficiently better to replace the first one, the gas estimation for tx [2] happened before the tx [1] was submitted. Since reverting a previous solution costs more gas than submitting the first valid solution to a batch this gas estimate was off and we are running out of gas.

It's likely better to pass a large (hard-coded) gas limit to the tx and rely on it not being fully used.

fleupold:
Closes #561 (see task for description). The number comes from our [worst case test in dex-contracts](https://github.com/gnosis/dex-contracts/blob/master/test/stablex/stablex_large_example.js#L12

### Test Plan

Build Bot, then don't expect failures like shown in the task.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -128,6 +128,8 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
                 prices,
                 token_ids_for_price,
             )
+            // Gas estimate might be off, as we race with other solution submissions and thus might have to revert trades which costs more gas than expected.
+            .gas(5_000_000.into())
             .call()
             .wait()?;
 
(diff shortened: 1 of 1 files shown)


josojo (approved):
Thanks for taking care!

(I have not executed the test plan, since I am very confident that this was the issue)

bh2smith (commented):
Fantastic! Great thinking and nice easy fix!

PR #586, opened 2020-02-24
Handle sparse solver output

e00E:
related to #546
The solver is going to output only orders that have been touched instead
of outputting all orders with untouched orders having 0 sell and buy
amount.

Before merging this PR the solver needs to be udpated, tagged and the tag used here changed appropriately.

Test Plan:
Manually build the solver docker from the correct tag like we do in .travis.yml and then run the e2e ganache test with this as the solver. There print out from the solution shows that we get `ExecutedOrder`s. More orders can be added to the test to see that there is no problem if not all are filled.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -11,7 +11,7 @@ use mockall::automock;
 
 use crate::contracts;
 use crate::gas_station::GasPriceEstimating;
-use crate::models::{Order, Solution};
+use crate::models::{ExecutedOrder, Solution};
 use crate::util::FutureWaitExt;
 
 lazy_static! {
@@ -75,14 +75,12 @@ pub trait StableXContract {
     fn get_solution_objective_value(
         &self,
         batch_index: U256,
-        orders: Vec<Order>,
         solution: Solution,
         block_number: Option<BlockNumber>,
     ) -> Result<U256>;
     fn submit_solution(
         &self,
         batch_index: U256,
-        orders: Vec<Order>,
         solution: Solution,
         claimed_objective_value: U256,
     ) -> Result<()>;
@@ -112,13 +110,11 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
     fn get_solution_objective_value(
         &self,
         batch_index: U256,
-        orders: Vec<Order>,
         solution: Solution,
         block_number: Option<BlockNumber>,
     ) -> Result<U256> {
         let (prices, token_ids_for_price) = encode_prices_for_contract(&solution.prices);
-        let (owners, order_ids, volumes) =
-            encode_execution_for_contract(orders, solution.executed_buy_amounts);
+        let (owners, order_ids, volumes) = encode_execution_for_contract(&solution.executed_orders);
         let mut builder = self
             .instance
             .submit_solution(
@@ -139,13 +135,11 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
     fn submit_solution(
         &self,
         batch_index: U256,
-        orders: Vec<Order>,
         solution: Solution,
         claimed_objective_value: U256,
     ) -> Result<()> {
         let (prices, token_ids_for_price) = encode_prices_for_contract(&solution.prices);
-        let (owners, order_ids, volumes) =
-            encode_execution_for_contract(orders, solution.executed_buy_amounts);
+        let (owners, order_ids, volumes) = encode_execution_for_contract(&solution.executed_orders);
         let gas_price = self.gas_price_estimating.estimate_gas_price();
         if let Err(ref err) = gas_price {
             log::warn!(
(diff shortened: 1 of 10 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -192,24 +178,18 @@ fn encode_prices_for_contract(price_map: &HashMap<u16, u128>) -> (Vec<u128>, Vec
...
-    );
     let mut owners = vec![];
     let mut order_ids = vec![];
     let mut volumes = vec![];
-    for (order_index, buy_amount) in executed_buy_amounts.into_iter().enumerate() {
-        if buy_amount > 0 {
+    for order in executed_orders {
+        if order.buy_amount > 0 {
             // order was touched!
             // Note that above condition is only holds for sell orders.
```
e00E: Does this comment still make sense? Can it be possible (even before this change) that an order has only one of sell amount and buy amount equal to zero but not the other?

driver/src/metrics/stablex_metrics.rs
```diff
@@ -106,29 +106,27 @@ impl StableXMetrics {
...
+                    solution
+                        .executed_orders
+                        .len()
+                        .try_into()
+                        .unwrap_or(std::i64::MAX),
+                );
                 self.tokens
                     .with_label_values(book_label)
-                    .set(tokens_from_orders(&touched_orders));
+                    .set(tokens_from_solution(solution));
```
e00E: @fleupold This does the same thing as before, right? (see the implementation of this function below)

e00E:
This can be reviewed now. Sorry for the PR being big. I tried but wasn't able to cut it down into multiple commits in a sensible way. A lot of changes are in test code because we are no longer passing `&[model::Orders]` around in several places.

driver/src/metrics/stablex_metrics.rs
fleupold: Is OWL included in the price vector? I know that it is not passed on the smart contract (as it's always assumed to be 10**18). But it should be considered a token for logging sake.

fleupold: Ok, it's filtered out here, so we should be fine: https://github.com/gnosis/dex-services/blob/4e7548e373d6c8909063229ea771d1d791c42fe2/driver/src/contracts/stablex_contract.rs#L182

driver/src/models/solution.rs
```diff
@@ -1,25 +1,34 @@
+use ethcontract::Address;
 use std::collections::HashMap;
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct ExecutedOrder {
+    pub account_id: Address,
```
fleupold: Would it make sure to contain the entire underlying Order in the struct? Not sure if there is a use case but would feel sensible and less repetitive from a modelling perspective. But then, it would complicate the solution parsing logic, as we have to find the orders in the list, so feel free to ignore.

driver/src/models/order.rs
```diff
@@ -41,4 +42,17 @@ pub mod test_util {
             sell_amount: 4,
         }
     }
+
+    pub fn order_to_executed_order(
```
fleupold: Should this be using the From or Into trait instead?

driver/src/price_finding/naive_solver.rs
```diff
@@ -115,48 +115,57 @@ impl PriceFinding for NaiveSolver {
...
         let mut prices = HashMap::new();
-        let mut exec_buy_amount: Vec<u128> = vec![0; orders.len()];
-        let mut exec_sell_amount: Vec<u128> = vec![0; orders.len()];
-
+        let mut executed_orders: Option<[ExecutedOrder; 2]> = None;
         let mut matching_orders_indices: Option<[usize; 2]> = None;
         'outer: for (i, x) in orders.iter().enumerate() {
-            for j in i + 1..orders.len() {
+            for (j, y) in orders.iter().enumerate().skip(i + 1) {
+                let mut set_matched_orders = |sell_amounts: [u128; 2], buy_amounts: [u128; 2]| {
```
fleupold: Can we outline this lambda please into a method. This is a bit hard to read.

driver/src/price_finding/naive_solver.rs
```diff
@@ -165,43 +174,45 @@ impl PriceFinding for NaiveSolver {
...
-            for i in matching_orders_indices
-                .expect("fee price was nonzero so we should have a match")
-                .iter()
-            {
-                let i = *i;
-                let order = &orders[i];
+            for i in 0..2 {
+                let expect_message = "fee price was nonzero so we should have a match";
+                let order = &orders[matching_orders_indices.expect(expect_message)[i]];
+                let executed_order = &mut executed_orders.as_mut().expect(expect_message)[i];
```
fleupold: Nit: should we expect outside for the loop?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FPhKhSXofSAm3e%2Fgiphy.gif&token=nice" data-gitmeme-token="nice"></a>

Nice! Please add to the description how to manually test this.

josojo (approved):
Good job!

Feel free to merge the PR in dex-solver and then push it through...

driver/src/models/order.rs
e00E: Doesn't work because there are additional parameters.

e00E: Even if it did work I am not sure if its correct to go into the whole From machinery when its not used this way.

driver/src/price_finding/naive_solver.rs
e00E: Not easily possible because it captures a lot from the environment which would have to be made parameters. I made it like this in order to introduce the least amount of changes to naive_solver. I'll do a cleanup PR of naive_solver after this one.

driver/src/models/solution.rs
e00E: Overall I prefer the current approach. It only contains then necessary information even if some fields are the same as in another struct. It is easier when parsing the solution and a lot of test code would have to include information that is not needed.

Issue #580, opened 2020-02-22
Don't error on expected failures


fleupold:
We are currently seeing a lot of errors for exepected failures. Two examples are:

1. `StableXDriver error: method 'submitSolution(uint32,uint256,address[],uint16[],uint128[],uint128[],uint16[])' failure: contract call reverted with message: Some("SafeMath: subtraction overflow")`
2. `StableXDriver error: method 'submitSolution(uint32,uint256,address[],uint16[],uint128[],uint128[],uint16[])' failure: contract call reverted with message: Some("New objective doesn\'t sufficiently improve current solution")`

We should filter for those types of errors both when verifying and submitting solution and make sure they don't get counted as real errors since they are technically not.

PR #588, opened 2020-02-25
Don't error on benign verification failures

fleupold:
First part of #580 

This PR makes it so that for somewhat "expected" failures during solution verification we don't emit an error but instead treat the batch as skipped. Those failures are for now

1. `Claimed objective doesn't sufficiently improve current solution` (when a better or evenly good solution is already applied - happens when the objective Value is provided in the submitSolution tx)
2. `New objective doesn\'t sufficiently improve current solution` (same as above but this happens when objectiveValue is estimated) 
3. `SafeMath: subtraction overflow` (when the solver produces a solution with negative overall objective value)

We still log those error into kibana and thus can get an idea of how often they happen. They should however not affect our success ratio as the batch was successfully processed.

As a next step we have to apply similar logic to the submitSolution transaction (this PR is for the call), which will be slightly more involved as the revert message only contains the failed tx hash and not a revert reason.

### Test Plan
Added a unit test.

Change the smart contract to always fail with an expected error inside submitSolution. Then run the ganache e2e tests and see the following output: 

```
Feb 25 11:13:31.787 INFO [driver] Using contract at 0xf204a4ef082f5c04bb89f7d5e6568b796096735a
Feb 25 11:13:31.788 INFO [driver] Using account 0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1
Feb 25 11:13:31.790 INFO [driver::price_finding] Using naive price finder
Feb 25 11:13:31.790 INFO [driver] Orderbook filter: OrderbookFilter { tokens: {}, users: {} }
Feb 25 11:13:31.904 INFO [driver::driver::stablex_driver] Computed solution: Solution { prices: {1: 1999997997997997998, 0: 1000000000000000000}, executed_buy_amounts: [999000, 1996001], executed_sell_amounts: [1999997, 999000] }
Feb 25 11:13:31.923 INFO [driver::driver::stablex_driver] Benign failure while verifying solution: Claimed objective doesn't sufficiently improve current solution
```

Lastly, we would expect none of the benign error to show up in our alerts channel as soon as this PR is merged.

### DIFF
--- a/driver/src/contracts/mod.rs
+++ b/driver/src/contracts/mod.rs
@@ -2,8 +2,10 @@ pub mod stablex_auction_element;
 pub mod stablex_contract;
 
 use crate::transport::HttpTransport;
-use anyhow::Result;
+use anyhow::{Error, Result};
 use ethcontract::contract::MethodDefaults;
+use ethcontract::errors::{ExecutionError, MethodError};
+
 use ethcontract::{Account, PrivateKey};
 use std::time::Duration;
 
@@ -25,4 +27,37 @@ fn method_defaults(key: PrivateKey, network_id: u64) -> Result<MethodDefaults> {
     };
 
     Ok(defaults)
+}
+
+const EXPECTED_ERRORS: &[&str; 3] = &[
+    "New objective doesn\'t sufficiently improve current solution",
+    "Claimed objective doesn't sufficiently improve current solution",
+    "SafeMath: subtraction overflow",
+];
+
+pub fn extract_benign_submission_failure(err: &Error) -> Option<String> {
+    err.downcast_ref::<MethodError>()
+        .and_then(|method_error| match &method_error.inner {
+            ExecutionError::Revert(Some(reason)) => {
+                if EXPECTED_ERRORS.contains(&&reason[..]) {
+                    Some(reason.clone())
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        })
+}
+
+#[cfg(test)]
+pub mod tests {
+    use super::{ExecutionError, MethodError};
+    use anyhow::anyhow;
+    pub fn benign_error() -> anyhow::Error {
+        anyhow!(MethodError::from_parts(
+            "submitSolution(uint32,uint256,address[],uint16[],uint128[],uint128[],uint16[])"
+                .to_owned(),
+            ExecutionError::Revert(Some("SafeMath: subtraction overflow".to_owned())),
+        ))
+    }
 }
--- a/driver/src/contracts/mod.rs
+++ b/driver/src/contracts/mod.rs
@@ -2,11 +2,13 @@ pub mod stablex_auction_element;
 pub mod stablex_contract;
 
 use crate::transport::HttpTransport;
-use anyhow::{Error, Result};
+use anyhow::{anyhow, Error, Result};
 use ethcontract::contract::MethodDefaults;
 use ethcontract::errors::{ExecutionError, MethodError};
 
 use ethcontract::{Account, PrivateKey};
+use std::convert::TryFrom;
+use std::fmt;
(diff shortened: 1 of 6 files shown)


driver/src/contracts/mod.rs
```diff
@@ -26,3 +28,36 @@ fn method_defaults(key: PrivateKey, network_id: u64) -> Result<MethodDefaults> {
...
     Ok(defaults)
 }
+
+const EXPECTED_ERRORS: &[&str; 3] = &[
+    "New objective doesn\'t sufficiently improve current solution",
+    "Claimed objective doesn't sufficiently improve current solution",
+    "SafeMath: subtraction overflow",
+];
+
+pub fn extract_benign_submission_failure(err: &Error) -> Option<String> {
```
e00E: Nit: This can probably return `Option<'static str>` or an `enum ExpectedError` (which implements a conversion to the exact error string). But I see that this would make the code a bit longer and this isn't a performance concern so feel free to disregard this comment.

e00E (approved):

bh2smith:
> Feb 25 11:13:31.923 INFO [driver::driver::stablex_driver] Benign failure while verifying solution: Claimed objective doesn't sufficiently improve current solution

Just wondering how this is "logged as an error in kibana" Looks like and info log in the test plan.

driver/src/contracts/mod.rs
```diff
@@ -26,3 +29,36 @@ fn method_defaults(key: PrivateKey, network_id: u64) -> Result<MethodDefaults> {
...
+        }
+    }
+}
+
+impl TryFrom<&str> for BenignSolutionFailure {
+    type Error = anyhow::Error;
+
+    fn try_from(reason: &str) -> Result<Self, Self::Error> {
+        match reason {
+            "New objective doesn\'t sufficiently improve current solution" => {
```
bh2smith: I would be happy if we could take this opportunity to fix the \ in the logs... perhaps a double slash would correctly escape this? or do we leave it this way so it is "searchable" in plain text?

bh2smith (approved):
Very nice!

driver/src/driver/stablex_driver.rs
```diff
@@ -85,11 +85,34 @@ impl<'a> StableXDriver<'a> {
...
+                },
+            }
+        } else {
             info!(
-                "Verified solution with objective value: {}",
-                objective_value
+                "Not submitting trivial solution for batch {}",
+                batch_to_solve
             );
+            None
```
nlordell: Ubernit: this branch and the one in lines 128-131 are equivalent. I would probably just add the metric and return false here right away.

nlordell (approved):
Love it! I think the way the types and logic was split-up is 👌

driver/src/contracts/mod.rs
fleupold: The backslash comes from the way the string is escaped in RPC I believe. I actually think it gets printed correctly on the console:

> Feb 25 16:39:29.975 INFO [driver::driver::stablex_driver] Benign failure while verifying solution: Claimed objective doesn't sufficiently improve current solution

driver/src/driver/stablex_driver.rs
fleupold: But we also return None in line 99 when it's a Benign error. We would have to log and return there as well, which I thought complicates the control flow (I'm overall not super happy how this control flow turned out, it's a bit messy)

nlordell: Oh I missed that, you are absolutely right!

driver/src/contracts/mod.rs
fleupold: Thanks for the suggestion. It actually pushed me into making this a more typesafe top-level return type of the solution submission.

PR #591, opened 2020-02-26
Query revert reason when tx fails

fleupold:
This PR is the first step to ignoring benign failures even when they happen upon actual solution submission (not just verification). This can happen in a race condition where two solvers verified their solution is better than the current best one and each submit their solution. They might end up mined in the same block and the later one could get reverted as it is not sufficiently better than the first one (e.g. https://etherscan.io/tx/0x9d6e7954efd1fa7c35ccd9022308b2a9ecb890cdc472e374b9d3220c6aa4a6b9)

ethcontracts only gives us the failed TX hash and not the revert reason. Therefore, we have to identify the block number in which this tx was mined and reissue `submit_solution` as a call (not a tx). This will give us the underlying revert reason.

In the next PR I will make it so that we treat benign errors similar as to how we do for verification failures.

### Test Plan

Added a unit test. 

Also changing the smart contract to revert with a benign reason if the objective value is small:

```solidity
require(claimedObjectiveValue > 1146456329082338568550207772363246612408613709560797663757005782256565639990, "Claimed objective doesn't sufficiently improve current solution");
```

We also need to change the code [here](https://github.com/gnosis/dex-services/blob/814920b265c8065cd914298f07e0ef00a0741b80/driver/src/contracts/stablex_contract.rs#L126) to that we don't supply the MAX_OBJECTIVE_VALUE if a block_number is specified (otherwise we won't get the same revert reason when we replay the verification)

Then we can run a ganache with "ERROR ON TRANSACTION FAILURE" disabled and the following e2e test:

`cargo test -p e2e ganache -- --nocapture`

In the logs we can then see:

> Feb 26 18:30:45.035 ERRO [driver] StableXDriver error: Benign Error: Claimed objective doesn't sufficiently improve current solution

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 
 use anyhow::Result;
 use ethcontract::transaction::GasPrice;
-use ethcontract::{Address, PrivateKey, U256};
+use ethcontract::{Address, BlockNumber, PrivateKey, U256};
 use lazy_static::lazy_static;
 #[cfg(test)]
 use mockall::automock;
@@ -77,6 +77,7 @@ pub trait StableXContract {
         batch_index: U256,
         orders: Vec<Order>,
         solution: Solution,
+        block_number: Option<BlockNumber>,
     ) -> Result<U256>;
     fn submit_solution(
         &self,
@@ -113,11 +114,12 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
         batch_index: U256,
         orders: Vec<Order>,
         solution: Solution,
+        block_number: Option<BlockNumber>,
     ) -> Result<U256> {
         let (prices, token_ids_for_price) = encode_prices_for_contract(&solution.prices);
         let (owners, order_ids, volumes) =
             encode_execution_for_contract(orders, solution.executed_buy_amounts);
-        let objective_value = self
+        let builder = self
             .instance
             .submit_solution(
                 batch_index.low_u32(),
@@ -128,9 +130,14 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
                 prices,
                 token_ids_for_price,
             )
-            .call()
-            .wait()?;
+            .view();
 
+        let builder = if let Some(block_number) = block_number {
+            builder.block(block_number)
+        } else {
+            builder
+        };
+        let objective_value = builder.call().wait()?;
         Ok(objective_value)
     }
 
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -119,7 +119,7 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
         let (prices, token_ids_for_price) = encode_prices_for_contract(&solution.prices);
         let (owners, order_ids, volumes) =
             encode_execution_for_contract(orders, solution.executed_buy_amounts);
-        let builder = self
+        let mut builder = self
             .instance
             .submit_solution(
                 batch_index.low_u32(),
@@ -131,12 +131,7 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
                 token_ids_for_price,
             )
             .view();
-
-        let builder = if let Some(block_number) = block_number {
-            builder.block(block_number)
-        } else {
-            builder
-        };
+        builder.block = block_number;
         let objective_value = builder.call().wait()?;
         Ok(objective_value)
     }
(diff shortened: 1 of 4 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -128,9 +130,14 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
                 prices,
                 token_ids_for_price,
             )
-            .call()
-            .wait()?;
+            .view();
 
+        let builder = if let Some(block_number) = block_number {
+            builder.block(block_number)
```
e00E: Isn't going to run into the same error we had when using the block number for paginated order book reading where the node did not support this properly?

fleupold: I think there the underlying problem was that we were load-balancing between multiple nodes and some nodes might have not yet seen the block. Thus we were actually querying for a future block (as far as that node was concerned). We checked the config and should have been able to query up to 64 blocks in the past.

I guess the view function should be fairly fast to execute so that a race-condition is unlikely (meaning we could just omit the block number). I'd be keen to give it another try and see how it behaves in staging.

What do you think?

driver/src/contracts/stablex_contract.rs
```diff
@@ -128,9 +130,14 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
                 prices,
                 token_ids_for_price,
             )
-            .call()
-            .wait()?;
+            .view();
 
+        let builder = if let Some(block_number) = block_number {
```
nlordell:
```suggestion
        builder.block = block_numer
```

The builder properties are public, the builder methods are just convenient in most cases.

driver/src/eth_rpc.rs
```diff
@@ -0,0 +1,32 @@
...
+use ethcontract::web3::futures::Future;
+use ethcontract::web3::types::TransactionReceipt;
+use ethcontract::H256;
+#[cfg(test)]
+use mockall::automock;
+
+/// Interface defining possible interactions via Ethereum RPCs
+/// Main purpose it to allow mocking the concrete Web3 implementation
+#[cfg_attr(test, automock)]
+pub trait EthRpc {
```
nlordell: Would it make sense for `ethcontract` to return the tx receipt? It already queried it to read the failure.

driver/src/solution_submission/mod.rs
```diff
@@ -106,7 +108,133 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+            .submit_solution(
+                batch_index.clone(),
+                orders.clone(),
+                solution.clone(),
+                claimed_objective_value,
+            )
+            .map_err(|err| {
+                extract_transaction_hash(&err)
+                    .and_then(|hash| {
+                        let receipt = self.rpc.get_transaction_receipts(hash).ok()??;
```
nlordell: `??!?!?!`

driver/src/contracts/stablex_contract.rs
e00E: I didn't realize it should have been possible. If 64 blocks should work then it's definitely fine.

nlordell (approved):
Great addition!

e00E (approved):

nlordell:
> Any idea how to test this other then "we do it live" would be appreciated. The fake contract address can be injected via failing_contract_networks.txt to try it locally.

Could we use Ganache and disable the `error` payload when transactions fail? That should cause the transaction to get sent, mined, and the confirmed. Then you can see if the subsequent call works?

driver/src/eth_rpc.rs
fleupold: Yes that would make sense. This way we'd save one round trip.

driver/src/solution_submission/mod.rs
fleupold: is there a better way? I'm getting a Result<Option<T>> and only want to continue if the inner option is one. Otherwise return None, ideally without adding another layer of nestedness.

fleupold:
> Could we use Ganache and disable the error payload when transactions fail?

Great idea. I managed to test it successfully and changed the test plan accordingly.

nlordell:
@fleupold, Now that I think about it, we might actually be able to run with ERROR ON TRANSACTION FAILURE on now because of https://github.com/gnosis/dex-services/blob/7db96e868b7e955910fdf2a590fc717007788c5a/driver/src/transport.rs#L92-L98

nlordell:
Looks like the CI failure was because the job started at the very end of a batch and we missed the solution submission window:
https://travis-ci.com/gnosis/dex-services/builds/150710056#L1911

Should we just restart the build?

PR #592, opened 2020-02-27
Don't error on benign solution submission failures

fleupold:
Based on top of #591 

Now that we can get the revert reason from a failed tx, we can decide to ignore it if the failure is benign (similar to #588)

### Test Plan

Added a unit test. Manual test is the same as is #591 but this time the log statement we say is of type INFO and no longer type ERROR. Thus it should get ignored in the alert channel:

> Feb 26 20:06:29.655 INFO [driver::driver::stablex_driver] Benign failure while submitting solution: Claimed objective doesn't sufficiently improve current solution

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 
 use anyhow::Result;
 use ethcontract::transaction::GasPrice;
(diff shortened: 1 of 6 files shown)


nlordell (approved):
Awesome!

bh2smith (approved):
This look ready to go!

PR #593, opened 2020-02-28
Skip first batch

nlordell:
As an interim solution, we skip the first batch to avoid missing the solution submission deadline and possibly cause the second batch to fail as well.

Skipped batches do not log errors but will add a increment the "failed" metric counter as we don't want to consider them successes.

A better solution is currently being worked on.

### Test Plan

CI

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -35,6 +35,19 @@ impl<'a> StableXDriver<'a> {
         }
     }
 
+    #[cfg(test)]
+    fn with_past_auction(
+        price_finder: &'a mut dyn PriceFinding,
+        orderbook_reader: &'a dyn StableXOrderBookReading,
+        solution_submitter: &'a dyn StableXSolutionSubmitting,
+        metrics: StableXMetrics,
+    ) -> Self {
+        let mut driver =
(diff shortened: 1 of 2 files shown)


e00E:
Is this PR still useful if #595 is close to done?

nlordell:
@e00E I think the value is that it unblocks other PRs.

driver/src/driver/stablex_driver.rs
```diff
@@ -47,6 +47,23 @@ impl<'a> StableXDriver<'a> {
             return Ok(false);
         }
 
+        #[cfg(not(test))]
```
fleupold: nit: can we move this below `auction_processing_started`? I feel like we could still measure when we "started" the processing.

driver/src/metrics/stablex_metrics.rs
```diff
@@ -73,6 +73,12 @@ impl StableXMetrics {
         }
     }
 
+    #[cfg(not(test))]
```
fleupold: is this macro needed to suppress a warning?

fleupold (approved):
Let's use this as a fix for now to give us time and do the more sophisticated logic properly.

e00E (approved):

fleupold:
Looks good. Seems like a perfect of use of captain "merge-when-green"!

PR #595, opened 2020-02-28
Calculate time limit for solver based on remaining time in batch.

nlordell:
The PR introduces time limit calculation base on the remaining in batch. This introduces a new command line/environment parameter for configuring the when the solver is expected to finish computing.

As a follow up PR we can add a metrics for how much time the solver was given to calculate a solution.

### Test Plan

CI

### DIFF
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -6,6 +6,7 @@ use std::thread;
 use std::time::{Duration, SystemTime, SystemTimeError};
 
 const BATCH_DURATION: Duration = Duration::from_secs(300);
+const SOLVING_WINDOW: Duration = Duration::from_secs(240);
 
 /// Wraps a batch id as in the smart contract to add functionality related to
 /// the current time.
@@ -44,9 +45,14 @@ pub struct AuctionTimingConfiguration {
     /// The offset from the start of a batch at which point we should start
     /// solving.
     pub target_start_solve_time: Duration,
+
     /// The offset from the start of the batch at which point there is not
     /// enough time left to attempt to solve.
     pub latest_solve_attempt_time: Duration,
+
+    /// The offset from the start of the batch to cap the solver's execution
+    /// time.
+    pub solver_time_limit: Duration,
 }
 
 pub struct Scheduler<'a> {
@@ -57,7 +63,7 @@ pub struct Scheduler<'a> {
 
 #[derive(Debug, Eq, PartialEq)]
 enum Action {
-    Solve(BatchId),
+    Solve(BatchId, Duration),
     Sleep(Duration),
 }
 
@@ -66,11 +72,21 @@ impl<'a> Scheduler<'a> {
         driver: &'a mut dyn StableXDriver,
         auction_timing_configuration: AuctionTimingConfiguration,
     ) -> Self {
-        assert!(auction_timing_configuration.latest_solve_attempt_time <= BATCH_DURATION);
+        assert!(
+            auction_timing_configuration.solver_time_limit < SOLVING_WINDOW,
+            "The solver time limit must be within the solving window",
+        );
+        assert!(
+            auction_timing_configuration.latest_solve_attempt_time
+                < auction_timing_configuration.solver_time_limit,
+            "The latest solve attempt time must be earlier than the solver time limit",
+        );
         assert!(
             auction_timing_configuration.target_start_solve_time
-                < auction_timing_configuration.latest_solve_attempt_time
+                < auction_timing_configuration.latest_solve_attempt_time,
+            "the target solve start time must be earlier than the latest solve time",
         );
+
         Self {
             driver,
             auction_timing_configuration,
@@ -92,9 +108,9 @@ impl<'a> Scheduler<'a> {
                 info!("Sleeping {}s.", duration.as_secs());
                 duration
             }
-            Ok(Action::Solve(batch_id)) => {
+            Ok(Action::Solve(batch_id, time_limit)) => {
                 info!("Starting to solve batch {}.", batch_id.0);
-                match self.driver.run(batch_id.0.into()) {
+                match self.driver.run(batch_id.0.into(), time_limit) {
                     DriverResult::Ok => {
                         info!("Batch {} solved successfully.", batch_id.0);
                         self.last_solved_batch.replace(batch_id);
@@ -143,8 +159,10 @@ impl<'a> Scheduler<'a> {
             let duration = intended_solve_start_time.duration_since(now).unwrap();
             Action::Sleep(duration)
         } else {
-            Action::Solve(solving_batch)
+            let time_limit = self.auction_timing_configuration.solver_time_limit - elapsed_time;
+            Action::Solve(solving_batch, time_limit)
         };
+
         Ok(action)
     }
 }
@@ -153,7 +171,6 @@ impl<'a> Scheduler<'a> {
 mod tests {
     use anyhow::anyhow;
     use ethcontract::U256;
-    use mockall::predicate::eq;
 
     use super::*;
     use crate::driver::stablex_driver::MockStableXDriver;
@@ -187,6 +204,7 @@ mod tests {
         let auction_timing_configuration = AuctionTimingConfiguration {
             target_start_solve_time: Duration::from_secs(10),
             latest_solve_attempt_time: Duration::from_secs(20),
+            solver_time_limit: Duration::from_secs(30),
         };
         let scheduler = Scheduler::new(&mut driver, auction_timing_configuration);
 
@@ -208,14 +226,14 @@ mod tests {
             scheduler
                 .determine_action(base_time + Duration::from_secs(10))
                 .unwrap(),
-            Action::Solve(BatchId(0))
+            Action::Solve(BatchId(0), Duration::from_secs(20))
         );
 
         assert_eq!(
             scheduler
                 .determine_action(base_time + Duration::from_secs(19))
                 .unwrap(),
-            Action::Solve(BatchId(0))
+            Action::Solve(BatchId(0), Duration::from_secs(11))
         );
 
         // Sleep because we are behind latest_solve_attempt_time:
@@ -241,6 +259,7 @@ mod tests {
         let auction_timing_configuration = AuctionTimingConfiguration {
             target_start_solve_time: Duration::from_secs(10),
             latest_solve_attempt_time: Duration::from_secs(20),
+            solver_time_limit: Duration::from_secs(30),
         };
         let mut scheduler = Scheduler::new(&mut driver, auction_timing_configuration);
         scheduler.last_solved_batch = Some(BatchId(0));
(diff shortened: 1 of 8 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -61,7 +61,13 @@ impl<'a> StableXContractImpl<'a> {
 
 #[cfg_attr(test, automock)]
 pub trait StableXContract {
-    fn get_current_auction_index(&self) -> Result<U256>;
+    /// Retrieve the index for the auction that is currently accepting orders.
+    fn get_current_auction_index(&self) -> Result<u32>;
+
+    /// Retrieve the amount of time remaining in the current batch in
+    /// milliseconds.
+    fn get_current_auction_remaining_seconds(&self) -> Result<U256>;
```
e00E: Comment says milliseconds, function says seconds. Should return Duration?

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -210,7 +211,22 @@ impl PriceFinding for OptimisationPriceFinder {
         &self,
         orders: &[models::Order],
         state: &models::AccountState,
+        time_limit: Duration,
     ) -> Result<models::Solution> {
+        const MINIMUM_SOLVER_TIME_LIMIT: Duration = Duration::from_secs(20);
+
+        // NOTE: A bear mininum amount of time the solver needs for calculating
```
e00E: typo: bare minimum

e00E (approved):

driver/src/contracts/stablex_contract.rs
e00E: I see that we return Duration from the orderbook. So its fine to just fix the comment.

e00E:
There are two ways to calculate the remaining time.

Use the smart contract. This is what this PR does. It means we call the ethereum node. The real remaining time might be lower depending on request delays and maybe how the node works internally.

Use the system clock to perform the same calculation as the smart contract. This is what we do for metrics. Has the advantage of not needing the node. Has the disadvantage that if the system clock is misconfigured we will get strange solver time limits.

I like the first approach more.

driver/src/driver/stablex_driver.rs
```diff
@@ -62,14 +62,30 @@ impl<'a> StableXDriver<'a> {
...
-            let solution = price_finder_result?;
-            info!("Computed solution: {:?}", &solution);
-
-            solution
+                .auction_solver_time_limit(&solver_time_limit_result);
+            let solver_time_limit = solver_time_limit_result?;
+
+            if solver_time_limit.as_secs() == 0 {
+                info!("Not enough time remaning in batch to compute solution");
+                Solution::trivial(0)
```
fleupold: I think this should not log a skipped event, because we actually don't process the batch. And so as far as our "success" metric is concerned we did not succeed (and it was kind of our fault for being late)

driver/src/main.rs
```diff
@@ -82,10 +82,6 @@ struct Options {
     #[structopt(long, env = "PRICE_FEED_INFORMATION", default_value = "{}")]
     backup_token_data: TokenData,
 
-    /// Number of seconds the solver should maximally use for the optimization process
-    #[structopt(long, env = "SOLVER_TIME_LIMIT", default_value = "150")]
```
fleupold: I think we can keep this as the target time when the solver returns (3 minutes) and then have the code compute how many seconds are left for that.

driver/src/metrics/stablex_metrics.rs
```diff
@@ -63,13 +65,24 @@ impl StableXMetrics {
         BookType::initialize_gauges(&users);
         registry.register(Box::new(users.clone())).unwrap();
 
+        let solver_time_limit_opts = HistogramOpts::new(
+            "dfusion_service_solver_time_limit",
```
fleupold: Nit: Couldn't this theoretically be computed by the time difference between solution and orders fetched?

driver/src/orderbook/mod.rs
```diff
@@ -46,9 +50,26 @@ impl<'a> PaginatedStableXOrderBookReader<'a> {
...
+        // NOTE: Solutions are due 1 minutes before the batch ends and we give
+        //   ourselves another 30 seconds for submitting the solution.
+        const SOLUTION_DEADLINE: Duration = Duration::from_secs(90);
+
+        let time_remaining = {
+            let seconds = self.contract.get_current_auction_remaining_seconds()?;
+            Duration::from_secs(seconds.as_u64())
+        };
+        let solution_time_remaining = time_remaining
+            .checked_sub(SOLUTION_DEADLINE)
```
fleupold: nit: saturating sub https://docs.rs/ethereum-types/0.3.1/ethereum_types/struct.U256.html#method.saturating_sub

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -210,7 +211,22 @@ impl PriceFinding for OptimisationPriceFinder {
...
+        //   a solution. If the time limit is smaller than this minimum, don't
+        //   start the solver at all. Note that this is not considered an error,
+        //   and the trivial solution is returned.
+        if time_limit < MINIMUM_SOLVER_TIME_LIMIT {
+            info!(
+                "Skipping solver run as time limit {} is too short",
+                time_limit.as_secs_f64()
+            );
+            return Ok(models::Solution::trivial(0));
```
fleupold: Again, I think this should be considered an error as we don't process the auction and it should count against our goal of 99% success rate. Not sure if it has to be a hard error, but maybe we can even make our restart scripts smart enough and only restart at the beginning of a batch.

A "real" trivial solution is a success case.

driver/src/metrics/stablex_metrics.rs
e00E: The difference is that it could be possible for the solver to not abide by the time limit. In this case the metrics would be different.

fleupold:
Regarding the two way to measure the time. I have no strong feeling and for now we should probably do what's simpler. At the end calling the blockchain is querying a single miners timestamp and I don't think we need to be synced on the millisecond.

In the long run I foresee however that we are not using the server to determine batchID and time_remaining (if we did they should become an atomic view function) and instead start solving as soon the 5 minute mark was passed on our local machine (+ some configurable security lag to consider potential reorgs). 
At the moment we are losing on average 20-30s (looking at grafana) until we start the order_fetching process by waiting for the first block to be mined in the new batch.

driver/src/contracts/stablex_contract.rs
nlordell: Nice catch!

driver/src/price_finding/optimization_price_finder.rs
nlordell: :bear:

driver/src/main.rs
nlordell: I'm not sure I understand what you mean. Shouldn't we just always compute how much time the solver has and give it more when we have more?

driver/src/metrics/stablex_metrics.rs
nlordell: Also I thought it would be nice to have a metric to see how much time we usually give the solver for producing a solution.

driver/src/orderbook/mod.rs
nlordell: `time_remaining` is a `Duration` which does not have a `saturating_sub` :cry:

driver/src/price_finding/optimization_price_finder.rs
nlordell: Sounds fair.

nlordell:
After talking with @fleupold we want to:
- Use the system clock as the time oracle
- Pass the solver time limit to the solver as is done here
- Add metrics (possibly split this up into a separate PR)

driver/src/driver/scheduler.rs
```diff
@@ -293,20 +316,21 @@ mod tests {
         let mut driver = MockStableXDriver::new();
         driver
             .expect_run()
-            .with(eq(U256::from(0)))
-            .returning(|_| DriverResult::Ok);
+            .withf(|b, t| *b == U256::from(0) && *t == Duration::from_secs(20))
```
fleupold: why not use eq here?

fleupold (approved):
Looks very nice! Blends in really nicely with the work that @e00E did.

⏲

driver/src/driver/scheduler.rs
nlordell: I couldn't find a way to use `eq` with more than one parameter. Maybe I'm missing something obvious.

fleupold: It should be `with(eq(U256...), eq(Duration...))`

nlordell: That was easy... Thanks!

nlordell:
> Looks very nice! Blends in really nicely with the work that @e00E did.

Yes! I think the scheduler was setup nicely to be extended for this kind of thing.

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -20,34 +22,34 @@ impl Default for Fee {
...
     pub fn to_args(self) -> Vec<String> {
         match self {
-            SolverConfig::StandardSolver { solver_time_limit } => {
-                vec![format!("--solverTimeLimit={:}", solver_time_limit)]
-            }
-            SolverConfig::FallbackSolver { solver_time_limit } => vec![
-                format!("--solverTimeLimit={:}", solver_time_limit),
+            SolverType::StandardSolver => vec![],
+            SolverType::FallbackSolver => vec![
+                String::from("--tokenInfo=/app/batchauctions/scripts/token_info_mainnet.json"),
```
e00E: I am curious about this line. It wasn't deleted anywhere, so it is new. Is this intended?

e00E (approved):

driver/src/price_finding/price_finder_interface.rs
nlordell: Good catch! That was a rebase error that snuck in.

driver/src/driver/scheduler.rs
```diff
@@ -66,11 +72,21 @@ impl<'a> Scheduler<'a> {
...
+        assert!(
+            auction_timing_configuration.latest_solve_attempt_time
+                < auction_timing_configuration.solver_time_limit,
+            "The latest solve attempt time must be earlier than the solver time limit",
+        );
         assert!(
             auction_timing_configuration.target_start_solve_time
-                < auction_timing_configuration.latest_solve_attempt_time
+                < auction_timing_configuration.latest_solve_attempt_time,
+            "the target solve start time must be earlier than the latest solve time",
```
bh2smith: ubernit: last message here is the only one whose first character is not capitalized. 

If this does wind up eventually being changed, I would even go as far as removing all of the "The " from each of the messages since they only contribute width.

bh2smith (commented):
Although I am tremendously late to the party here, I would be in favour of using system time directly from the machine this is running on since its essentially equivalent (given that the EVM time is the miner's system time). This would eliminate a completely unnecessary call to the EVM.

driver/src/driver/scheduler.rs
nlordell: > I would be in favour of using system time directly from the machine this is running on since its essentially equivalent (given that the EVM time is the miner's system time). This would eliminate a completely unnecessary call to the EVM.

I believe this PR was for calculating solver time limit using the system time and not using EVM time. As far as I can tell, there is no new EVM calls added by this PR.

> last message here is the only one whose first character is not capitalized.

Sorry, that was my bad! I tend to not capitalize log messages and errors and didn't realize that we were in the rest of the file. 

I think as a follow up, we could go over our logs and clean up the style everywhere so they are consistent.

Issue #581, opened 2020-02-23
Measure latency and size of requests


fleupold:
We should measure and report prometheus metrics on our request sizes and latencies for the HTTP requests we do, in particular for

1. RPC requests
2. GasStation estimates
3. Kraken Price information

This will help us identify if there are issues and measure performance improvements going forward.

PR #596, opened 2020-02-28
Use a shared HTTP client implementation

nlordell:
Use a shared HTTP client implementation. This is the first step for #581 so we can share metrics across HTTP services.

### Test Plan

CI

### DIFF
--- a/driver/src/http/mod.rs
+++ b/driver/src/http/mod.rs
@@ -1,0 +1,82 @@
+//! Module contains the implementation for a shared HTTP client for various
+//! driver components.
+
+use anyhow::{anyhow, Result};
+use isahc::http::{HttpTryFrom, Uri};
+use isahc::prelude::Request;
+use isahc::{HttpClientBuilder, ResponseExt};
+use serde::de::DeserializeOwned;
+use std::time::Duration;
+
+#[cfg(test)]
+lazy_static::lazy_static! {
+    /// An HTTP factory used for testing.
+    pub static ref TEST_FACTORY: HttpFactory = HttpFactory::new(Duration::from_secs(10));
+}
+
+/// A factory type for creating HTTP clients.
+#[derive(Debug)]
+pub struct HttpFactory {
+    default_timeout: Duration,
+}
+
+impl HttpFactory {
+    /// Creates a new HTTP client factory.
+    pub fn new(default_timeout: Duration) -> Self {
+        HttpFactory { default_timeout }
+    }
+
+    /// Creates a new HTTP client with the default configuration.
+    pub fn create(&self) -> Result<HttpClient> {
+        self.with_config(|builder| builder)
+    }
+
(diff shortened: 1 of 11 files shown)


e00E (approved):

driver/src/http/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+use isahc::http::{HttpTryFrom, Uri};
+use isahc::prelude::Request;
+use isahc::{HttpClientBuilder, ResponseExt};
+use serde::de::DeserializeOwned;
+use std::time::Duration;
+
+#[cfg(test)]
+lazy_static::lazy_static! {
+    /// An HTTP factory used for testing.
+    pub static ref TEST_FACTORY: HttpFactory = HttpFactory::new(Duration::from_secs(10));
```
fleupold: I'd expect a test factory to return mocked versions of HTTPClient (to set expectations without doing network calls). I guess this is a real factory used in the #[ignore] tests. Not sure if the name could capture this somehow. Would it maybe make sense to implement Default on HttpFactory?

fleupold (approved):
Any particular reason we want to use multiple clients instead of a single one?

driver/src/http/mod.rs
nlordell: Maybe we can rename this to `GLOBAL_FACTORY`?. This is really intended for creating real actual HTTP clients for the ignored tests.

nlordell:
> Any particular reason we want to use multiple clients instead of a single one?

Its so they can have separate configurations.

driver/src/http/mod.rs
nlordell: Just implemented `Default`, it makes the most sense.

nlordell:
@e00E do you mind taking a quick look at the changes I made to the DexAG code and let me know if you think they are OK? This changes it to use the new HTTP factory.

driver/src/price_estimation/dexag/api.rs
```diff
@@ -80,16 +75,14 @@ impl DexagApi for DexagHttpApi {
...
-            })
-            .boxed()
+        async move {
+            Ok(self
+                .client
+                .get_json_async(url.as_str())
+                .await
+                .context("failed to parse price json from dexag")?)
+        }
+        .boxed()
```
e00E: There is a problem here.

This function here returns f64 so `get_json_async` tries to deserialize into f64 but it should deserialize into `Price`.

```
async move {
            Ok(self
                .client
                .get_json_async::<_, Price>(url.as_str())
                .await
                .context("failed to parse price json from dexag")?
                .price)
        }
        .boxed()
```

e00E: That is the only problem I see.

nlordell: Fixed :)

Issue #585, opened 2020-02-24
New price feed (https://dex.ag/)


josojo:
In order to improve the liability and the number of prices, which we can get for our tokens, we should implement one more price feed from https://dex.ag/

e00E:
+ maybe price estimation outside of critical path

nlordell:
Here is the API docs: https://docs.dex.ag/#api

e00E:
For kraken we get the USD price for each token and use this as the OWL price. But dexag tracks neither USD nor OWL. So how should we calculate the price of a token in Owl?
It does track several stable coins (USDC, SUSD, USDT, TUSD) so one approach is to average them together to get an approximate USD price. Imo this would end up being too many requests per token.
Should we pick one specific stable coin?
Are there other solutions?

josojo:
Currently, in our ecosystem, the DAI has the highest liquidity against most of the coins traded on our exchange. Hence, I would query all prices against DAI and then report these prices.

The small fluctuations DAI vs USD on various exchanges should be negligible for our solver.

fleupold:
How does @marcovc deal with that at the moment in the dex-solver?

marcovc:
Honestly what I've done is too basic for a definitive solution to this problem. The main problem I was trying to solve was to get at least one price estimate for all tokens, and not really went into these considerations. I ended up organizing tokens in groups of tokens: I assumed that any two tokens in one of these groups can act as a proxy for each other and thus have an equivalent price.  OWL endes up in the same group as USD and other stable coins pegged to USD, if I recall correctly. Likely this could be improved by taking the mean of the group as @e00E sugest, but again, what was on my mind of mostly what to do when these aggregators don't work for a particular token. Currently this is no longer a big issue since @twalth3r adapted the price estimation to work in a hybrid way where some of the tokens have external price and others don't.

PR #597, opened 2020-02-28
Implement dex.ag api

e00E:
This is the first step for #585 .
After this I will implement the `PriceSource` trait for dexag.

Test Plan:
The code is not yet used in the project but it has unit tests.


driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,183 @@
...
+pub struct Price {
+    #[serde(with = "display_fromstr")]
+    pub price: f64,
+}
+
+/// Wrapper type of Address that implements deserialize from a hex string.
+#[derive(Clone, Debug)]
+pub struct Address_(pub Address);
+
+impl<'de> Deserialize<'de> for Address_ {
```
fleupold: The existing Deserialise or FromStr are not working?

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,183 @@
...
+        dex: Dex,
+        discluded: &str,
+    ) -> Result<Price>;
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Token {
+    pub name: String,
+    pub symbol: String,
+    pub address: Option<Address_>,
```
fleupold: I'm not a big fan of the wrapper type given that it would have to be publicly exposed. Can we not implement custom deserialisation for this field?

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// https://docs.dex.ag/api/tokens
+    fn token_list_full(&self) -> Result<Vec<Token>>;
+    /// https://docs.dex.ag/api/price
+    fn price(
+        &self,
+        from: &str,
+        to: &str,
+        amount: EitherAmount,
+        dex: Dex,
+        discluded: &str,
```
fleupold: what does discluded mean? Looks like it's [non-standard english for excluded](https://www.yourdictionary.com/disclude), maybe we can add a comment explaining its function.

fleupold: Checked on the documentation. Looks like this itself should be of type Vec<Dex> ?

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,183 @@
...
+            q.append_pair("from", from);
+            q.append_pair("to", to);
+            match amount {
+                EitherAmount::From(amount) => q.append_pair("fromAmount", &amount.to_string()),
+                EitherAmount::To(amount) => q.append_pair("toAmount", &amount.to_string()),
+            };
+            q.append_pair(
+                "dex",
+                match dex {
+                    Dex::Ag => "ag",
```
fleupold: nit: maybe implement toStr on the enum directly?

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,183 @@
...
+
+#[cfg_attr(test, mockall::automock)]
+/// Parts of the dex.ag api.
+pub trait DexagApi {
+    /// https://docs.dex.ag/api/tokens
+    fn token_list_full(&self) -> Result<Vec<Token>>;
+    /// https://docs.dex.ag/api/price
+    fn price(
+        &self,
+        from: &str,
```
fleupold: Shouldn't from/to be of type Token?

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,183 @@
...
+            Address::from_str("89d24a6b4ccb1b6faa2625fe562bdd9a23260359").unwrap()
+        );
+    }
+
+    #[test]
+    #[ignore]
+    fn online_dexag_api() {
+        let api = DexagApiImpl::new().unwrap();
+        let tokens = api.token_list_full().unwrap();
+        println!("{:?}", tokens);
```
fleupold: assert on top of print here? This helps readers to know what is the expected outcome without running it.

fleupold (commented):
Really good start! Some comments inline.

driver/src/price_estimation/dexag/api.rs
e00E: `Token` in this module is the result of the `token_list` call. For this function dex.ag uses token symbols which is just one part of the struct Token so I don't want to require filling a full Token struct. In my mind this module is trying to be a thin wrapper around the http api so I am happy this interface.

driver/src/price_estimation/dexag/api.rs
e00E: As in the other comment I want this to be a thin wrapper around their api which is why I kept the same name as their api docs use and did not write documentation comments. Technically this is vector but we are not going to use it so I don't think its worth it to write more code for this case until it's needed. I could remove both the dex and discluded parameter then we don't need the enum either.

driver/src/price_estimation/dexag/api.rs
e00E: Not a fan either. If this was a normal `Address` and not an `Option` then we could use the serde `deserialize_with` attribute but I don't think it works in the option case so as far as I can tell we would have to manually implement `Deserialize` for the whole token struct which I dislike even more than exposing the wrapper.

driver/src/price_estimation/dexag/api.rs
e00E: The existing deserialize expects the address hex string to not have the `0x` prefix. I haven't checked FromStr but I am going to.

driver/src/price_estimation/dexag/api.rs
e00E: Was thinking about that too but its only used in this function and it could be argued that its an implementation detail which is why I slightly prefer this way.

driver/src/price_estimation/dexag/api.rs
e00E: What should be asserted? We don't know what the call will return in the future and the current response is quite long so I don't want to have it in the source.
Only meaningful thing I can think of is asserting that there is at least one result.

driver/src/price_estimation/dexag/api.rs
e00E: The other way is to have two structs. One for the public trait api which doesn't implement deserialize and has the correct `Address` type. The other private struct has `address: String` and does derive Deserialize and then we convert between the structs in the function.
I like this idea. Will implement and see how it feels when its written down.

driver/src/price_estimation/dexag/api.rs
fleupold: In that case, let's please remove it. I'd prefer a concise but cleanly typed API and we can add things later as they come up.

driver/src/price_estimation/dexag/api.rs
fleupold: Can we not use deserialize_with on Option<Address> ?

driver/src/price_estimation/dexag/api.rs
fleupold: But we are only going to request the price for tokens that have been previously returned by `token_list_full`, aren't we? 
Thus I find it more logical to have the types be the same (gives the reader a clue that this output is used as an input here). We could also change `token_list_full` to return &str but I think having strong types here is better.

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,183 @@
...
+use serde::{de::Error as _, Deserialize, Deserializer};
+use serde_with::rust::display_fromstr;
+use std::time::Duration;
+use url::Url;
+
+#[cfg_attr(test, mockall::automock)]
+/// Parts of the dex.ag api.
+pub trait DexagApi {
+    /// https://docs.dex.ag/api/tokens
+    fn token_list_full(&self) -> Result<Vec<Token>>;
```
fleupold: ubernit: maybe call this `get_token_list` and `get_price`? (I don't see a specific meaning of *full list* as there seems to be no other type of token list to query)

driver/src/price_estimation/dexag/api.rs
e00E: I expected this to be more complex but it turns out simple.

josojo (approved):
Nice.. looks really good. No comments.

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,167 @@
+#![allow(dead_code)]
+
+use anyhow::{anyhow, Context, Result};
+use ethcontract::Address;
+use isahc::prelude::*;
```
fleupold: I'm not a huge fan of * imports. I think it's better to be explicit about what aspects we need.

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,167 @@
...
+pub const DEFAULT_BASE_URL: &str = "https://api-v2.dex.ag";
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);
+
+impl DexagApiImpl {
+    pub fn new() -> Result<Self> {
+        Self::custom(DEFAULT_BASE_URL, DEFAULT_TIMEOUT)
+    }
+
+    pub fn custom(base_url: &str, timeout: Duration) -> Result<Self> {
+        let client = HttpClient::builder()
```
fleupold: Would be nice if this could use #596, but I guess it will go in first. But then #596 should update this callsite as well.

driver/src/price_estimation/dexag/api.rs
fleupold: I see your point. Generally I think it's nice if I can see the expected format from looking at the code rather than having to actually run it to see what it prints, but you are right that here it's likely not really predictable.

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+fn deserialize_address<'de, D>(deserializer: D) -> Result<Option<Address>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    fn hex_string_to_address(string: &str) -> Result<Address> {
+        let prefix = "0x";
+        if !string.starts_with(prefix) {
+            return Err(anyhow!("does not start with {}", prefix));
```
fleupold: nit: Maybe it helps printing the actual string here so it's easier to debug the problem (but I see it's actually printed in the context below).

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FxT0BKJGfxmloVAqjss%2Fgiphy-downsized-medium.gif&token=go" data-gitmeme-token="go"></a>

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn get_price(&self, from: &Token, to: &Token, amount: EitherAmount) -> Result<Price>;
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Token {
+    pub name: String,
+    pub symbol: String,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_address")]
+    pub address: Option<Address>,
```
nlordell: Nit: `Address` should implement `Deserialize` as far as I know so I think this shouldn’t be needed.

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,167 @@
...
+    })
+}
+
+pub enum EitherAmount {
+    From(f64),
+    To(f64),
+}
+
+#[derive(Debug)]
+pub struct DexagApiImpl {
```
nlordell: Ubernit: maybe `DexagHttpApi`?

driver/src/price_estimation/dexag/api.rs
nlordell: I usually agree but prelude modules are usually designed to be *-imported this way. I’m ok with ether in this case.

driver/src/price_estimation/dexag/api.rs
```diff
@@ -0,0 +1,167 @@
...
+        let token: Token = serde_json::from_str(json).unwrap();
+        assert_eq!(token.name, "SAI old DAI (SAI)");
+        assert_eq!(token.symbol, "SAI");
+        assert!(token.address.is_none());
+    }
+
+    #[test]
+    #[ignore]
+    fn online_dexag_api() {
+        let api = DexagApiImpl::new().unwrap();
```
nlordell: Can you add a comment here with the command line needed to run this example? Just makes it easier for readers to run it and see the output.

nlordell (approved):
Looks good!

driver/src/price_estimation/dexag/api.rs
e00E: Problem is that the default implementation does not allow the Strings to start with `0x`.

driver/src/price_estimation/dexag/api.rs
e00E: I don't like that we have this same comment for every one of these tests then but I'll add it for now.

driver/src/price_estimation/dexag/api.rs
nlordell: AFAIK that is only true of the `FromStr` implementation. The `Deserialize` implementations require the address string to start with `”0x”`.

driver/src/price_estimation/dexag/api.rs
nlordell: Any particular reason you’re not a fan of these comments?

driver/src/price_estimation/dexag/api.rs
e00E: You're right.

PR #598, opened 2020-02-28
Implemented HTTP metrics

nlordell:
This PR adds metrics to the unified client per subsystem.

### Test Plan

CI, check that new metrics are produced on `http://localhost:9586`


driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// Initialize the metrics for a labeled subsystem.
+    fn initialize_labeled_subsystem<L: LabeledSubsystem + Any>(
+        &mut self,
+        registry: &Arc<Registry>,
+    ) -> Result<()> {
+        let name = L::name();
+
+        let latency = {
+            let opts = Opts::new(
+                format!("dfusion_service_http_{}_latency", name),
```
e00E: I  prefer the subsystem name being another label in addition to the "request type label".

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,249 @@
...
+use std::collections::HashMap;
+use std::fmt::{self, Debug, Formatter};
+use std::sync::Arc;
+use std::time::Duration;
+
+/// A registry for all HTTP related metrics.
+#[derive(Debug)]
+pub struct HttpMetrics {
+    labeled: HashMap<TypeId, SubsystemMetrics<IntGaugeVec>>,
+    unlabeled: HashMap<TypeId, SubsystemMetrics<IntGauge>>,
```
e00E: Has to be histogram instead of gauge.

e00E: I am wondering if the use of `TypeId` is worth it. The simplest solution is to use `&'static str` as the key which we get from the name of the metric. The name of the metric already has to be unique otherwise two metrics would override each other.

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        subsystem
+            .size
+            .with_label_values(&[label.label()])
+            .set(size as _);
+    }
+
+    /// Add a request latency and size measurement to the current HTTP metrics
+    /// registry.
+    pub fn request_unlabeled<L: UnlabeledSubsystem + Any>(&self, latency: Duration, size: usize) {
```
e00E: Is it worth it to have both labeled and unlabeled? Unlabeled can be implemented in terms of a labeled with only one request type. Has the disadvantage that code using this has to pass `MySubsystem::default_or_something_like_that` but makes the code here simpler and then we don't have duplicate `_unlabeled` methods.

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+impl Default for HttpMetrics {
+    fn default() -> Self {
+        HttpMetrics::new(&Default::default()).unwrap()
+    }
+}
+
+/// A trait abstracting a labeled subsystem, that is a specific use of an HTTP
+/// client where the HTTP requests should be differenciated.
```
e00E: differentiated

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,249 @@
...
+        for label in Self::all_labels() {
+            gauges.with_label_values(&[label]).set(0);
+        }
+
+        Ok(gauges)
+    }
+}
+
+/// A trait abstracting a labeled subsystem, that is a specific use of an HTTP
+/// client where the HTTP requests should not be differenciated.
```
e00E: differentiated

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn initialize_labeled_subsystem<L: LabeledSubsystem + Any>(
+        &mut self,
+        registry: &Arc<Registry>,
+    ) -> Result<()> {
+        let name = L::name();
+
+        let latency = {
+            let opts = Opts::new(
+                format!("dfusion_service_http_{}_latency", name),
+                format!("Latency in milliseconds for {} HTTP requests", name),
```
e00E: I prefer measuring latency in seconds as ``f64` and using `Duration::as_secs_f64`.

driver/src/metrics/http_metrics.rs
nlordell: Sorry, my prometheus-foo is very poor. I'm not sure I understand what you mean.

Do you want the all the HTTP latency and size metrics to be the same metric (`"dfusion_service_http_latency"`) and then have the subsystem be a label instead?

driver/src/metrics/http_metrics.rs
nlordell: From what I understand, histograms require us to know how to split it up into buckets, I'm not sure what sizes and latencies to use :(

nlordell: `TypeId` should be pretty free here. Under the hood its just a `u64` and it should be determined at compile-time and not runtime (since we don't use any `Box<Any>`).

driver/src/metrics/http_metrics.rs
nlordell: I thought of that, but the `labeled` variants require a value which doesn't make sense for unlabeled subsystems.

driver/src/metrics/http_metrics.rs
nlordell: Sounds good to me!

driver/src/metrics/http_metrics.rs
e00E: Yes

driver/src/metrics/http_metrics.rs
e00E: Gauge would not work in this case so it has to be Histogramm. With a Gauge you only have a single value so every time we did one http request we would overwrite the old value.

We do have to pick bucket sizes but I wouldn't overthink it. We can use this function https://docs.rs/prometheus/0.8.0/prometheus/fn.exponential_buckets.html to make buckets. `exponential_buckets(0.5, 2, 5)` would  return `[0.5, 1, 2, 4, 8]` second buckets which seems fine.

PR #601, opened 2020-03-02
Make e2e test solvable with standard-solver

e00E:
The standard solver expects 10**18 units of the fee token to be worth
roughly 1 USD so very low amounts as have been used in the test get
filtered out.

Test Plan: Run the ganache e2e the test with the real solver.

### DIFF
--- a/e2e/tests/stablex_test.rs
+++ b/e2e/tests/stablex_test.rs
@@ -97,10 +97,13 @@ fn test_with_ganache() {
     let balance_after = tokens[1]
         .balance_of(accounts[0])
         .wait_and_expect("Cannot get balance after");
-    assert_eq!(
-        balance_after - balance_before,
-        (999 * usd_price_in_fee).into()
-    )
+    let balance_change = (balance_after - balance_before).as_u128();
+    let expected_balance_change = 999 * usd_price_in_fee;
+    // With a non naive solver it is possible that the trade does not exactly
+    // match what is expected.
+    let allowed_difference = usd_price_in_fee;
+    let difference = ((balance_change as i128) - (expected_balance_change as i128)).abs();
+    assert!(difference < allowed_difference as i128);
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


e00E:
I would also like to have CI run the ganache test with the real solver. This can catch accidental incompatibilities between the driver and the solver. I thought this was already the case because we explicitly build the solver docker before running the e2e tests but it turns out that this is not true because it still uses `common.env` which specifies the fallback solver.
What is the correct way to accomplish this? We probably want to keep the naive solver by default for developers so we need to somehow set the `solver_type` argument / env variable. We could run the ganache test with both the naive and real solver.

e2e/tests/stablex_test.rs
```diff
@@ -92,7 +97,10 @@ fn test_with_ganache() {
     let balance_after = tokens[1]
         .balance_of(accounts[0])
         .wait_and_expect("Cannot get balance after");
-    assert_eq!(balance_after - balance_before, 999_000.into())
+    assert_eq!(
+        balance_after - balance_before,
+        (999 * usd_price_in_fee).into()
+    )
```
e00E: This assert fails with the standard solver because the balance change is 998.99... instead of exactly 999. Should we allow small fluctuations?

josojo: Would it be cleaner to read out the used solver from the env variables and then make two different assertions, depending on the solver?

josojo (commented):
I would be in favor of running both tests. They both add value: one makes sure that the primitive solver is still working and the other one makes sure that normal solver is working and finds a reasonable price

e2e/tests/stablex_test.rs
josojo: Actually, a range check seems reasonable to me: 
Since we sometimes change minimal parameters in the solver, the expected number can change ( e.g. we might change the buffer for token shifting etc). Adjusting each time the unit tests seems like an unnecessary burden.

bh2smith (approved):

e2e/tests/stablex_test.rs
fleupold: A range check is fine with me although I don't think this test has to necessarily pass if you are using it wit a non-designated solver.

fleupold (approved):
looks good...

PR #606, opened 2020-03-03
Only count prices > 0 in solution for metrics

e00E:
The solver seems to include 0 prices for tokens when they haven't been
touched. We probably only want to count tokens that have a  nonzero
price in the metric.

### DIFF
--- a/driver/src/metrics/stablex_metrics.rs
+++ b/driver/src/metrics/stablex_metrics.rs
@@ -205,13 +205,17 @@ fn tokens_from_orders(orders: &[Order]) -> i64 {
 }
 
 fn tokens_from_solution(solution: &Solution) -> i64 {
-    solution
-        .prices
-        .iter()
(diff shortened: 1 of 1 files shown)


e00E:
This is visible in Grafana. We see that after merging the sparse price map PR the number of tokens has stayed at 18.
With this PR it should stay at 1 because the fee token still has a price even in a trivial solution.
Before both PRs it stayed at 0. I am not sure whether 0 or 1 is more correct.

josojo (approved):
Good catch!

Both are fine, counting 1 or 0 prices for trivial solutions. 
However, I would lean to a 0 price-count for trivial solutions, as actually no meaningful prices are calculated or sent to the smart contract.

fleupold (approved):
Looks good. I'd also say the trivial solution should probably log 0 rather than 1 token, but a very simple trade should be logging 2 tokens. Thus logging 1 token should never happen.

PR #607, opened 2020-03-04
Implement PriceSource for Dexag

e00E:
Now that we have the http api we can turn it into a price source.

After this PR I still need to enable this as a price source and move price estimation out of the main loop.
Depending how the latter is implemented we might not need to parallelize the requests to Dexag anymore which we originally wanted to do because here we need one request per token.

I removed some parts from the dexag api because it turns out we do not need them.

### DIFF
--- a/driver/src/price_estimation/dexag/api.rs
+++ b/driver/src/price_estimation/dexag/api.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)]
-
 use anyhow::{anyhow, Context, Result};
 use ethcontract::Address;
 use isahc::prelude::*;
--- a/driver/src/price_estimation/dexag/api.rs
+++ b/driver/src/price_estimation/dexag/api.rs
@@ -12,7 +12,8 @@ pub trait DexagApi {
     /// https://docs.dex.ag/api/tokens
     fn get_token_list(&self) -> Result<Vec<Token>>;
     /// https://docs.dex.ag/api/price
-    /// Returns the price of one unit of `to` expressed in `from`.
+    /// Returns the price of one unit of `from` expressed in `to`.
+    /// For example `get_price("ETH", "DAI")` is ~220.
     fn get_price(&self, from: &Token, to: &Token) -> Result<f64>;
 }
 
--- a/driver/src/price_estimation/dexag/api.rs
+++ b/driver/src/price_estimation/dexag/api.rs
@@ -14,10 +14,11 @@ pub trait DexagApi {
     /// https://docs.dex.ag/api/tokens
     fn get_token_list(&self) -> Result<Vec<Token>>;
     /// https://docs.dex.ag/api/price
-    fn get_price(&self, from: &Token, to: &Token, amount: EitherAmount) -> Result<Price>;
+    /// Returns the price of one unit of `to` expressed in `from`.
+    fn get_price(&self, from: &Token, to: &Token) -> Result<f64>;
 }
 
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize, PartialEq)]
 pub struct Token {
     pub name: String,
     pub symbol: String,
@@ -27,9 +28,9 @@ pub struct Token {
 }
 
 #[derive(Clone, Debug, Deserialize)]
-pub struct Price {
+struct Price {
     #[serde(with = "display_fromstr")]
-    pub price: f64,
+    price: f64,
 }
 
 fn deserialize_address<'de, D>(deserializer: D) -> Result<Option<Address>, D::Error>
@@ -53,11 +54,6 @@ where
     } else {
         None
     })
-}
-
-pub enum EitherAmount {
-    From(f64),
-    To(f64),
 }
 
 #[derive(Debug)]
@@ -97,7 +93,7 @@ impl DexagApi for DexagHttpApi {
             .context("failed to parse token list json from dexag response")
     }
 
-    fn get_price(&self, from: &Token, to: &Token, amount: EitherAmount) -> Result<Price> {
+    fn get_price(&self, from: &Token, to: &Token) -> Result<f64> {
         let mut url = self.base_url.clone();
         url.set_path("price");
         {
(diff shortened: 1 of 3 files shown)


driver/src/price_estimation/dexag/api.rs
```diff
@@ -72,25 +66,23 @@ impl DexagApi for DexagHttpApi {
...
         {
             // This is in its own block because we are supposed to drop `q`.
             let mut q = url.query_pairs_mut();
             q.append_pair("from", &from.symbol);
             q.append_pair("to", &to.symbol);
-            match amount {
-                EitherAmount::From(amount) => q.append_pair("fromAmount", &amount.to_string()),
-                EitherAmount::To(amount) => q.append_pair("toAmount", &amount.to_string()),
-            };
+            q.append_pair("fromAmount", "1");
```
josojo: This means that we will get the prices for a sellAmount of 1 DAI, right? I think that is an excellent solution to the slippage challenge with this price feed.

josojo (approved):
Very clean code!

> After this PR I still need to enable this as a price source and move price estimation out of the main loop.

Putting the price-sourcing into its own thread would be my preferred solution. But I am also fine, if we just parallelize the requests for now. I don't think that we will hit the api limit of 120 requests / minute soon with our small token sets.

driver/src/price_estimation/dexag/api.rs
e00E: If this is called with `from == DAI` and `to == ETH` then this will give the price of 1 DAI in ETH which is something like 0.004. Writing this I realize that the comment I put on this function at the top in the trait is wrong.

josojo: > If this is called with from == DAI and to == ETH then this will give the price of 1 DAI in ETH which is something like 0.004. 

This means that there is some minimal slippage right? If I would sell 0.001 DAI, I would maybe get a better price.
While it is not important, I see that the `fromAmount` is optional. Maybe we should just leave it out and it will give us a price "without slippage"

driver/src/price_estimation/dexag/api.rs
```diff
@@ -14,25 +12,22 @@ pub trait DexagApi {
     /// https://docs.dex.ag/api/tokens
     fn get_token_list(&self) -> Result<Vec<Token>>;
     /// https://docs.dex.ag/api/price
-    fn get_price(&self, from: &Token, to: &Token, amount: EitherAmount) -> Result<Price>;
+    /// Returns the price of one unit of `from` expressed in `to`.
+    /// For example `get_price("ETH", "DAI")` is ~220.
```
fleupold: I assume, this comment will be funny to look at it a few years (either way).

driver/src/price_estimation/dexag/mod.rs
```diff
@@ -1 +1,178 @@
...
+{
+    pub fn with_api(api: Api) -> Result<Self> {
+        // We need to return prices in OWL but Dexag does not track it. OWL tracks
+        // USD so we use another stable coin as an approximate USD price.
+        const STABLE_COIN: &str = "DAI";
+
+        let tokens = api.get_token_list()?;
+        let mut tokens: HashMap<String, api::Token> = tokens
+            .into_iter()
+            .map(|token| (token.symbol.clone(), token))
```
fleupold: nit: feels like if we implemented hash and equality on token to just look at the symbol we wouldn't need a map here, but could use a set instead.

driver/src/price_estimation/dexag/mod.rs
```diff
@@ -1 +1,178 @@
...
+
+impl<Api> PriceSource for DexagClient<Api>
+where
+    Api: DexagApi,
+{
+    fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
+        Ok(tokens
+            .iter()
+            .filter_map(|token| {
+                let stable_coin_price = if token.symbol() == self.stable_coin.symbol {
```
fleupold: I'm a bit confused by the name. Is this name supposed to imply that this price is always in reference to the stable coin? Should it be called `priceInOwl` or something like that

driver/src/price_estimation/dexag/mod.rs
```diff
@@ -1 +1,178 @@
...
+            .with(
+                eq(api_tokens[0].clone()),
+                // Clippy claims we can remove the clone here but compilation
+                // actually fails if we follow the suggestion.
+                #[allow(clippy::redundant_clone)]
+                eq(api_tokens[0].clone()),
+            )
+            .returning(|_, _| Ok(1.0));
+
+        let client = DexagClient::with_api(api).unwrap();
```
fleupold: Should we test what happens with an API that doesn't return DAI?

driver/src/price_estimation/dexag/mod.rs
```diff
@@ -1 +1,178 @@
...
+        let prices = client.get_prices(&tokens).unwrap();
+        assert_eq!(
+            prices,
+            hash_map! {
+                TokenId(1) => tokens[0].get_owl_price(0.7) as u128,
+                TokenId(4) => tokens[1].get_owl_price(1.2) as u128,
```
fleupold: nit: why is tokenId 5 missing? 

Also, would it make sense to write the exact number we are expecting rather than invoking a function? While I agree that this function is tested elsewhere, we are kind of leaking the fact that we are relying on it while we could just assert on the expected output.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F3oz8xQQP4ahKiyuxHy%2Fgiphy.gif&token=ok" data-gitmeme-token="ok"></a>

driver/src/price_estimation/dexag/mod.rs
e00E: We need to store `tokens` in the struct anyway because it is used to look up `Api::Token` by `Token::symbol`.

driver/src/price_estimation/dexag/mod.rs
e00E: > Is this name supposed to imply that this price is always in reference to the stable coin?

That is the case. The reason for why there is a stable coin is explained in a comment earlier in the file.

driver/src/price_estimation/dexag/mod.rs
e00E: Added a test for it.

driver/src/price_estimation/dexag/mod.rs
e00E: TokenId 5 is missing because the test is setup to return an error for this token. I have added a comment to explain that.

`get_owl_price` is the correct result so it makes sense to me. If for some reason this struct internally wanted to use a difference function it should still return the same result.

fleupold: For the tokenID 5 it sounds like this could be a separate test case where the test name implies that errored responses are ignored. We should only be testing one thing per test.

e00E: Made a second test case.

PR #608, opened 2020-03-04
Refactor `NaiveSolver::find_prices` into smaller functions

e00E:
Test Plan: CI, tests havent' changed

The diff for the whole PR as shown in Github is hard to read. I suggest either reading `fn find_prices(&self, orders: &[Order], state: &AccountState) -> Result<Solution>` and following all new function calls or looking at each commit individually.

### DIFF
--- a/driver/src/price_finding/naive_solver.rs
+++ b/driver/src/price_finding/naive_solver.rs
@@ -153,7 +153,7 @@ fn find_first_match(orders: &[Order], state: &AccountState, fee: &Option<Fee>) -
 }
 
 fn create_executed_orders(first_match: &Match, fee: &Option<Fee>) -> (ExecutedOrderPair, PriceMap) {
-    fn create_order(order: &Order, sell_amount: u128, buy_amount: u128) -> ExecutedOrder {
+    fn create_executed_order(order: &Order, sell_amount: u128, buy_amount: u128) -> ExecutedOrder {
         ExecutedOrder {
             account_id: order.account_id,
             order_id: order.id,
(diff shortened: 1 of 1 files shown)


driver/src/price_finding/naive_solver.rs
```diff
@@ -111,113 +111,132 @@ impl NaiveSolver {
...
+    (executed_orders, prices)
+}
+
+/// Returns whether the solution should be trivial.
+fn update_with_fee(
+    orders: &OrderPair,
+    fee: &Fee,
+    executed_orders: &mut ExecutedOrderPair,
+    prices: &mut PriceMap,
+) -> bool {
```
fleupold: nit: I find this return type a bit un-intuitive. Should it maybe return an Option<()> so that `Some(())` means the update was successful? This would allow you to use a bunch of ? instead of match statements in this method. I wouldn't even mind if instead of mutating the orders and prices in place we returned an Option<(ExecutedOrderPair , PriceMap)> to make the funciton more pure.

driver/src/price_finding/naive_solver.rs
```diff
@@ -111,113 +111,132 @@ impl NaiveSolver {
     }
 }
 
+pub struct Match {
```
fleupold: Does this need to be public?

fleupold (approved):
Nice

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F5t1YusAYxmRn474Sid%2Fgiphy-downsized-medium.gif&token=cleanup" data-gitmeme-token="cleanup"></a>

driver/src/price_finding/naive_solver.rs
e00E: I made this take the arguments by value and return Solution. Then it is pure and can return `Solution::trivial()` itself.

PR #609, opened 2020-03-05
Use Dexag in addition to Kraken as price source

e00E:
If we get two prices they are averaged together.

Also parallelize dexag requests with Rayon (see comment in source code) because Dexag requires one request per token which can take tens of seconds otherwise.

### DIFF
--- a/driver/src/price_estimation/average_price_source.rs
+++ b/driver/src/price_estimation/average_price_source.rs
@@ -1,26 +1,34 @@
 use super::{PriceSource, Token};
 use crate::models::TokenId;
 use anyhow::{anyhow, Result};
+use crossbeam_utils::thread;
 use std::collections::HashMap;
+use std::sync::Mutex;
 
 /// Combines two prices into one average.
 pub struct AveragePriceSource<T0, T1> {
-    source_0: T0,
-    source_1: T1,
+    source_0: Mutex<T0>,
+    source_1: Mutex<T1>,
 }
 
 impl<T0, T1> AveragePriceSource<T0, T1> {
     pub fn new(source_0: T0, source_1: T1) -> Self {
-        Self { source_0, source_1 }
+        Self {
+            source_0: Mutex::new(source_0),
+            source_1: Mutex::new(source_1),
+        }
     }
 }
 
-impl<T0: PriceSource, T1: PriceSource> PriceSource for AveragePriceSource<T0, T1> {
+impl<T0: PriceSource + Send, T1: PriceSource + Send> PriceSource for AveragePriceSource<T0, T1> {
     fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
-        match (
-            self.source_0.get_prices(tokens),
-            self.source_1.get_prices(tokens),
-        ) {
+        let prices = thread::scope(|s| {
+            let handle_0 = s.spawn(|_| self.source_0.lock().unwrap().get_prices(tokens));
+            let handle_1 = s.spawn(|_| self.source_1.lock().unwrap().get_prices(tokens));
+            (handle_0.join().unwrap(), handle_1.join().unwrap())
+        })
+        .unwrap();
+        match prices {
             (Ok(p0), Ok(p1)) => Ok(average_prices(p0, p1)),
             (Ok(p), Err(e)) | (Err(e), Ok(p)) => {
                 log::warn!("one price source failed: {}", e);
--- a/driver/src/price_estimation/average_price_source.rs
+++ b/driver/src/price_estimation/average_price_source.rs
@@ -1,0 +1,78 @@
+use super::{PriceSource, Token};
+use crate::models::TokenId;
+use anyhow::{anyhow, Result};
+use std::collections::HashMap;
+
+/// Combines two prices into one average.
+pub struct AveragePriceSource<T0, T1> {
+    source_0: T0,
+    source_1: T1,
+}
+
+impl<T0, T1> AveragePriceSource<T0, T1> {
+    pub fn new(source_0: T0, source_1: T1) -> Self {
+        Self { source_0, source_1 }
+    }
+}
+
+impl<T0: PriceSource, T1: PriceSource> PriceSource for AveragePriceSource<T0, T1> {
+    fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
+        match (
+            self.source_0.get_prices(tokens),
+            self.source_1.get_prices(tokens),
+        ) {
+            (Ok(p0), Ok(p1)) => Ok(average_prices(p0, p1)),
+            (Ok(p), Err(e)) | (Err(e), Ok(p)) => {
+                log::warn!("one price source failed: {}", e);
+                Ok(p)
+            }
+            (Err(e0), Err(e1)) => Err(anyhow!("both price sources failed: {}, {}", e0, e1)),
+        }
+    }
+}
+
+fn average_prices(
+    prices_0: HashMap<TokenId, u128>,
+    mut prices_1: HashMap<TokenId, u128>,
+) -> HashMap<TokenId, u128> {
+    for (token_id, price_1) in prices_0 {
+        prices_1
+            .entry(token_id)
+            .and_modify(|price| *price = (*price + price_1) / 2)
+            .or_insert(price_1);
+    }
+    prices_1
+}
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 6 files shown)


driver/src/price_estimation/dexag/mod.rs
```diff
@@ -54,21 +53,34 @@ where
...
+        // `PriceSource` trait.
+        let pool = rayon::ThreadPoolBuilder::new()
+            .num_threads(8)
+            .build()
+            .unwrap();
+        let result = pool.install(|| {
+            tokens
+                .par_iter()
+                .filter_map(|token| {
+                    let stable_coin_price = if token.symbol() == self.stable_coin.symbol {
```
josojo: This variable stable_coin_price is only the stable coin price in one case, right? Maybe we can just call it 'price' or 'price_in_float_representation'

josojo (approved):
Another great PR.

I really like how strictly you build these PRs on top of each other, step by step.

driver/src/price_estimation/average_price_source.rs
```diff
@@ -0,0 +1,78 @@
...
+                Ok(p)
+            }
+            (Err(e0), Err(e1)) => Err(anyhow!("both price sources failed: {}, {}", e0, e1)),
+        }
+    }
+}
+
+fn average_prices(
+    prices_0: HashMap<TokenId, u128>,
+    mut prices_1: HashMap<TokenId, u128>,
```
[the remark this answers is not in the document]
e00E: From the outside is a pure function, it takes owernship of both parameters. It just so happens that we can reuse one of them but this isn't observable.

driver/src/price_estimation/dexag/mod.rs
e00E: We get some tokens as input for which we want the price in terms of owl. Every token is compared to `self.stable_coin` which acts as an approximation of owl. The reason there is an `if` is that it is possible that we are asked for the price token itself that we have picked as the stable coin. In this case it doesn't make sense to ask Dexag because it is the same coin. The price has to be 1.

josojo: Maybe it is just me, but if I am hearing stable_coin_price, I think of the price of a stable coin. but what we actually have is a price of an arbitrary token vs a stable coin. So maybe price_vs_stable_coin makes more sense?

e00E: Changed the variable name.

nlordell (commented):
Looks good! My only concern is using futures instead of a threadpool for parallelizing API requests.

Note that we can just return a `BoxFuture` for now to get around the no-`async` limitation in traits.

driver/src/price_estimation/average_price_source.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn new(source_0: T0, source_1: T1) -> Self {
+        Self { source_0, source_1 }
+    }
+}
+
+impl<T0: PriceSource, T1: PriceSource> PriceSource for AveragePriceSource<T0, T1> {
+    fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
+        match (
+            self.source_0.get_prices(tokens),
+            self.source_1.get_prices(tokens),
```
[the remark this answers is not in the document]
e00E: Kraken was so fast in my tests that I didn't think it was needed. But its no problem to do it.

driver/src/price_estimation/dexag/mod.rs
```diff
@@ -54,21 +53,34 @@ where
...
-                };
-                Some((token.id, token.get_owl_price(stable_coin_price)))
-            })
-            .collect())
+        use rayon::prelude::*;
+        // Use rayon as a quick and dirty way to make requests in parallel. It
+        // would be more efficient if we used async or Futures instead but for
+        // now this is fine.
+        // async is not currently supported in traits so we cannot use it in the
+        // `PriceSource` trait.
```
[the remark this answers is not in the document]
e00E: I totally didn't realize that I don't need to make the trait fn async. Just doing it here is enough. Will do.

driver/src/price_estimation/average_price_source.rs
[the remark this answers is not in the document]
nlordell: Let’s leave it as is for now then. When we expand to `n` sources then we can reevaluate

driver/src/price_estimation/dexag/mod.rs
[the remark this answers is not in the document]
nlordell: The HTTP Api trait might still need to support a `Future` return type, but you can use a `BoxFuture` and the `boxed()` extension method.

driver/src/price_estimation/average_price_source.rs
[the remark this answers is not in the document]
e00E: Too late. It's implemented :) .

driver/src/price_estimation/average_price_source.rs
```diff
@@ -0,0 +1,86 @@
+use super::{PriceSource, Token};
+use crate::models::TokenId;
+use anyhow::{anyhow, Result};
+use crossbeam_utils::thread;
+use std::collections::HashMap;
+use std::sync::Mutex;
+
+/// Combines two prices into one average.
+pub struct AveragePriceSource<T0, T1> {
```
fleupold: If T0 and T1 are both PriceSource + Send, can we just make it a statically sized array so we can iterate over it? Otherwise, as we grow I'm afraid the match logic below will explode in combinations. 
But we can also cross that bridge when we get there.

fleupold: ubernit: maybe AveragingPriceSource is a slightly better name?

driver/src/price_estimation/average_price_source.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl<T0: PriceSource + Send, T1: PriceSource + Send> PriceSource for AveragePriceSource<T0, T1> {
+    fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
+        let prices = thread::scope(|s| {
+            let handle_0 = s.spawn(|_| self.source_0.lock().unwrap().get_prices(tokens));
+            let handle_1 = s.spawn(|_| self.source_1.lock().unwrap().get_prices(tokens));
+            (handle_0.join().unwrap(), handle_1.join().unwrap())
+        })
+        .unwrap();
```
fleupold: There is a lot of unwrap going on here. Can we translate those into errors instead of panicking? 

Also if I understand [crossbeam::thread::scope](https://docs.rs/crossbeam-utils/0.6.6/crossbeam_utils/thread/fn.scope.html) correctly, it joins all spawned thread automatically and will return an error if one of them fails (thus we shouldn't have to join and at least not unwrap the join() call manually)

fleupold (approved):
Looks good, really nicely integrated behind the existing traits. Let's just not be so aggressive about unwrapping the thread results.

fleupold:
Can you add a small comment how you tested this manually (outside of the added unit test)?

e00E:
I tested it with another `#ignored` test that creates the price oracle that uses the AveragingPriceSource and runs it. I also added prints to the threads to see that they really execute at the same time. I didn't think it was useful to include this test so I deleted it afterwards.

driver/src/price_estimation/average_price_source.rs
e00E: These unwraps are all related to the threads panicking. Usually you cannot do anything about panics in Rust except when they happen in another thread. So if we did the same thing we are doing in the threads which querying the prices in the main thread we would crash in a panic. Thus it makes sense to unwrap the results since that just leads to a panic again.

e00E: The manual join is needed to get the queried prices back out of the `scope` function.

driver/src/price_estimation/average_price_source.rs
e00E: I agree that if we are going to add more price sources in the future and average all of them then we have to make this a list.
Until then this is simpler so let's not do this before this is necessary.

driver/src/price_estimation/average_price_source.rs
e00E: Hope this argument makes sense. If not we can revisit it later. I'll merge now it since we wanted to have this for the weekend and I want to see if there are any problems on staging before the evening.

fleupold: Yes, makes sense

PR #615, opened 2020-03-09
Fetch prices in background thread

e00E:
We want to not block the main loop every batch to get new prices. Instead we update the prices continuously in the background and use the most recent results.

Test Plan: CI with new unit tests. I also tested the new PriceOracle manually with printouts to see that it really updates in the expected interval.


driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Non blocking.
+    /// Infallible.
+    fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
+        let mut nbps = self.non_blocking_price_source.lock().unwrap();
+        nbps.track_tokens(tokens);
+        // At this point tokens that were not being tracked already have been
+        // added but they do not yet have prices. This is expected because this
+        // function is supposed to be non blocking.
+        let result = nbps.get_prices(tokens);
+        std::mem::drop(nbps);
```
fleupold: out of curiosity, any reason to not wrap in a block instead of dropping?

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,197 @@
...
+        // We do tell the thread to immediately update the prices instead of
+        // waiting a full update interval so that the prices are hopefully
+        // available the next `get_prices` is called.
+        self.channel_to_thread
+            .send(ThreadMessage::UpdateImmediately)
```
fleupold: Maybe the list of tokens we should fetch can be passed around as part of the message?

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        let nbps = non_blocking_price_source.clone();
+        // vk: Clippy notes that the following loop and match can be written as
+        // a `while let` loop but I find it clearer to make all cases explicit.
+        #[allow(clippy::while_let_loop)]
+        let join_handle = thread::spawn(move || loop {
+            match receiver.recv_timeout(update_interval) {
+                Ok(ThreadMessage::UpdateImmediately) | Err(RecvTimeoutError::Timeout) => {
+                    let now = Instant::now();
+                    let cutoff = now.checked_sub(update_interval).unwrap_or(now);
```
fleupold: Even if we update immediately we only update the tokens that were not updated within the update interval? I would have expected that in that case we update all tokens.

driver/src/main.rs
```diff
@@ -124,6 +124,15 @@ struct Options {
         parse(try_from_str = duration_millis),
     )]
     gas_station_timeout: Duration,
+
+    /// Time interval in seconds in which price sources should be updated.
+    #[structopt(
+        long,
+        env = "PRICE_SOURCE_UPDATE_INTERVAL",
+        default_value = "1000",
```
fleupold: Let's make the default 5 minutes given the volatility of prices and how cheap these requests are.

fleupold (commented):
I'm wondering if we can simplify the approach here. As far as I understand the complexity comes from the fact that you only know which tokens you need prices for in the moment where you need them (and thus we might not have fetched them before).

Given that the list of tokens we can fetch prices for is hardcoded at startup time, I wonder if we can just issue a load request every 5 minutes (towards the end of each batch) for this entire list. This way we should have up to date prices for all tokens we support.

This way we don't have to track tokens or update times and can just have a recurring timer to inject new prices into the non blocking PriceStore/Cache.

driver/src/price_estimation/threaded_price_source.rs
e00E: > I'm wondering if we can simplify the approach here.

That makes sense. I thought about this from the PriceSource perspective but we do know all the tokens. Will implement that.

e00E:
> I'm wondering if we can simplify the approach here.

That makes sense. I thought about this from the PriceSource perspective but we do know all the tokens. Will implement that.

driver/src/price_estimation/threaded_price_source.rs
e00E: I accidentally deleted your question what the difference between a new block and manually dropping is.

There is no functional difference. A block is more obvious but forces indentation and if you have intermediate variables created in the block that you want to keep outside you have to return them.

e00E:
> Given that the list of tokens we can fetch prices for is hardcoded at startup time, I wonder if we can just issue a load request every 5 minutes (towards the end of each batch) for this entire list. This way we should have up to date prices for all tokens we support.

The prices are updated in 5 minute intervals now but I don't think it's necessary to update them at a specific time. Surely an average of 2.5 minutes old is fine. This avoids adding an extra `update` method to PriceOracle forcing it to work only work with the ThreadedPriceSource.

driver/src/price_estimation/mod.rs
```diff
@@ -32,19 +37,29 @@ pub struct PriceOracle {
...
-        Ok(PriceOracle::with_source(
-            tokens,
-            AveragePriceSource::new(KrakenClient::new()?, DexagClient::new()?),
-        ))
+    pub fn new(tokens: TokenData, update_interval: Duration) -> Result<Self> {
+        Ok(if tokens.is_empty() {
+            // If there are no tokens we don't need to create a real price
+            // source. This avoids sending requests in the Dexag constructor.
+            struct EmptyPriceSource {};
+            impl PriceSource for EmptyPriceSource {
```
fleupold: While it's cool that we can inline implementations like this, I'd prefer to lift it outside and make it a top level implementation to keep this function more readable.

driver/src/price_estimation/mod.rs
```diff
@@ -32,19 +37,29 @@ pub struct PriceOracle {
...
     /// Creates a new price oracle from a token whitelist data.
-    pub fn new(tokens: TokenData) -> Result<Self> {
-        Ok(PriceOracle::with_source(
-            tokens,
-            AveragePriceSource::new(KrakenClient::new()?, DexagClient::new()?),
-        ))
+    pub fn new(tokens: TokenData, update_interval: Duration) -> Result<Self> {
+        Ok(if tokens.is_empty() {
+            // If there are no tokens we don't need to create a real price
+            // source. This avoids sending requests in the Dexag constructor.
```
fleupold: Looking at this comment now, I think it's weird to have the constructor perform a long running operation (they are usually non-blocking). Maybe not super critical now, but I'd prefer if we can move the token initialisation to be lazy.

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,134 @@
...
+        ps.expect_get_prices().returning(|_| Ok(HashMap::new()));
+        let (tps, handle) =
+            ThreadedPriceSource::new(TOKENS.to_vec(), ps, Duration::from_secs(std::u64::MAX));
+        // If the thread didn't exit we would wait forever.
+        join(tps, handle);
+    }
+
+    #[test]
+    fn update_triggered_by_interval() {
+        let mut ps = MockPriceSource::new();
```
fleupold: in test code it's fine, but generally let's please spell out our variable names. It's a few strokes more when you write it but makes comprehension much easier down the line.

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,134 @@
...
+        ];
+    }
+
+    /// Drops tps and joins the handle which ensures that the thread exits and
+    /// hasn't panicked. This is important because the mockall expectation
+    /// panics happen in the thread which would otherwise let the tests pass
+    /// when an expectation is violated.
+    fn join(tps: ThreadedPriceSource, handle: JoinHandle<()>) {
+        std::mem::drop(tps);
+        handle.join().unwrap();
```
fleupold: It's not great that we'd wait forever if the thread doesn't finish (specfically for CI where I don't believe we have timeouts). 
Would it be a lot of effort to signal back to the owning struct (either via a channel or a boolean) that we have finished running?

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,134 @@
...
+            Ok(if start_returning_.load(ORDERING) {
+                hash_map! {TOKENS[0].id => 1}
+            } else {
+                hash_map! {}
+            })
+        });
+
+        let (tps, handle) = ThreadedPriceSource::new(TOKENS.to_vec(), ps, Duration::from_millis(5));
+        assert_eq!(tps.get_prices(&TOKENS[..]).unwrap(), hash_map! {});
+        thread::sleep(Duration::from_millis(10));
```
fleupold: Using sleep statements in unit tests is a smell (it creates brittle tests which pass depending on environment/load).

Generally, I find unit testing threading behaviour hard to do and prefer separating the units so they can each be tested for logic individually (and then we trust that the threading library works as intended).

The way I would test this piece of logic is to move everything inside the thread closure into its own function and invoke this directly in the unit test (thus not testing if channel.recv_timeout works, but we can assume that).

fleupold (approved):
Nice and much simpler from a logical point of view. If you didn't want to roll your own threading logic you could use the [timer crate](https://docs.rs/timer/0.1.3/timer/struct.Timer.html#method.schedule_repeating) to schedule a recurring codeblock for execution. This might make testing the drop behaviour simpler.

In any case, I'd suggest to get rid of the sleep in the unit test and try to make it actually fail in case of failure rather than deadlocking.

driver/src/price_estimation/mod.rs
e00E: I implemented this in https://github.com/gnosis/dex-services/pull/626 .

driver/src/price_estimation/mod.rs
e00E: On the other hand if it is only needed in one function then it might be more readable to have it there. I don't mind either way so I'll move it out.

e00E: Actually this won't be needed with the lazy initialization of dexag PR.

driver/src/price_estimation/threaded_price_source.rs
e00E: The actual logic so simple (`old_prices.extend(new_prices)`) that it's not worth testing.

I agree that sleep statements are not optimal but afaik there is no other way to test multithreaded code. So the question is whether a brittle test is worse than no test. The answer is that I'm not sure. Maybe it would be acceptable to make the tests`#[ignore]`

driver/src/price_estimation/threaded_price_source.rs
e00E: I can hardcode a timeout by doing the joining in another thread and having the main thread panic if that thread does not exist within some time. That does have the brittle `sleep` again but it is better than hanging forever.

driver/src/price_estimation/threaded_price_source.rs
e00E: I have thought of a way. Only if we are in the testing configuration, I can add another channel on which the background signals that it has completed its loop. Then in the test we can listen to that channel instead of sleeping. Will implement to see how it looks.

fleupold: > I agree that sleep statements are not optimal but afaik there is no other way to test multithreaded code.

The way I have done it in the past is to separate the threading and the business logic by components or functions and then test the logic in isolation (or abstract the threading component so it can be mocked with a non-threaded executor).

Generally I think a potentially flaky test is worse than no test at all.

e00E:
Implemented without any sleeps now. The test is still slightly brittle because we have to pick some timeout for thread update loop which in theory in an extremely slow system could be hit. However in practice I don't believe this can be hit unless this is a VM which is put to sleep during the test or something like that.

With the timer crate I wasn't sure exactly how their threading model works and they mention that the callback should not take too long which is impractical in our case because we want to perform a network operation. So I prefer the current version.

driver/src/main.rs
```diff
@@ -124,6 +124,15 @@ struct Options {
         parse(try_from_str = duration_millis),
     )]
     gas_station_timeout: Duration,
+
+    /// Time interval in seconds in which price sources should be updated.
+    #[structopt(
```
fleupold: nit: add to readme

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,174 @@
...
+pub struct ThreadedPriceSource {
+    // Shared between this struct and the thread. The background thread writes
+    // the prices, the thread calling `get_prices` reads them.
+    price_map: Arc<Mutex<HashMap<TokenId, u128>>>,
+    // Allows the thread to notice when the owning struct is dropped.
+    _channel_to_thread: Sender<()>,
+
+    // To make testing easier we use another channel to which the thread sends
+    // a message whenever it completes one update loop.
+    #[cfg(test)]
```
fleupold: We could even add this in any case as it doesn't seem that perfomance critical.

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        let test_receiver = extract_test_receiver(tps);
+        let deadline = Instant::now() + THREAD_TIMEOUT;
+        loop {
+            match test_receiver.recv_timeout(THREAD_TIMEOUT) {
+                // The sender side of `test_receiver` has been dropped which
+                // only happens when the thread exits.
+                Err(RecvTimeoutError::Disconnected) => break,
+                // There is still a queued messages on the channel.
+                Ok(_) if Instant::now() <= deadline => continue,
```
fleupold: As you said we can still run into deadlock here if we happened to produce more messages than we can consume (e.g. if we produced fetched prices every 1ns or so).

Since multithreading is so hard, it might be nicer to inject a "periodic executor" into the threaded price source which can be instantiated with a closure and implements `execute_now()`. In the tests you could implement a synchronous version of it and for prod you use the thread + channel version.

In order to test the right drop behavior (which is a nice test) you could then keep the back-channel without the need of sending a message after each execution which should make it not stallable.

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        #[cfg(test)]
+        let (test_sender, test_receiver) = mpsc::channel::<()>();
+
+        let price_map_ = price_map.clone();
+        // vk: Clippy notes that the following loop and match can be written as
+        // a `while let` loop but I find it clearer to make all cases explicit.
+        #[allow(clippy::while_let_loop)]
+        let join_handle = thread::spawn(move || loop {
+            match receiver.recv_timeout(update_interval) {
```
nlordell: This is a super cleaver way to sleep for `update_interval` and allow waking to happen on demand! Awesome :metal:.

nlordell (approved):
:+1:

driver/src/price_estimation/threaded_price_source.rs
e00E: In case you find it interesting the "traditional" way without channels (like in c++) would be to use a https://doc.rust-lang.org/std/sync/struct.Condvar.html which is more lightweight than a channel but a channel is much nicer to use.

driver/src/price_estimation/threaded_price_source.rs
e00E: I find it clearer if it's only there for testing. It actually would be a problem too in non testing because the channel would buffer the update messages forever as no one is reading them.

driver/src/price_estimation/threaded_price_source.rs
e00E: There is no deadlock here because we are testing against deadline. If there are still messages after the deadline we panic.

Technically there is such a deadlock in the function below at "Clear previous messages". I changed this to also have a timeout.

With regards to the "executer": I specifically want to test the real threading behavior, not a synchronous test executor. Additionally consider that we can also have a deadlock in single threaded with an accidental infinite loop and we don't wrap all our other tests in timeouts either.

driver/src/price_estimation/threaded_price_source.rs
nlordell: That's how I would have done it. I also find the channels way nicer to read and to use.

One minor nit, now that I think of it. Maybe the channel names an better represent what they are for - like `update_prices_{rx,tx}`. I'm not sure how others feel about this.

driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -79,10 +86,15 @@ impl ThreadedPriceSource {
 }
 
 impl PriceSource for ThreadedPriceSource {
-    /// Non blocking.
-    /// Infallible.
     fn get_prices(&self, _tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
-        Ok(self.price_map.lock().unwrap().clone())
+        // NOTE: Return all the prices as they will get filtered by the caller
```
fleupold: Is there a technical reason for this? I find it a bit odd as it violates the contract and requires this component to "know" how it is being used (which can change and then maybe the implementation is not updated).

driver/src/price_estimation/threaded_price_source.rs
fleupold: I see your point about deadlocking on infinity loops. I don't like to use infinity loops (I remember a comment on a previous PR), unless e.g. in a top level thread that is not supposed to terminate or when it's absolutely obvious the logic will exit.
The advantage about using your own executor is that you can completely control the threading behaviour (things like when it gets scheduled compared to logic happening on another thread). In my opinion the actual thread::spawn is likely not test worthy. 
With the current approach we are having a potentially non-deterministic test, and understanding the test setup for a future maintainer is quite hard.

PR #619, opened 2020-03-09
Use system time instead of node to get current batch id

### LINKED ISSUE #616: Use system time to start batch processing
We are currently waiting for the first block to be mined with a timestamp in a batchID `t` before we start solving `t-1`. This leads to unpredictable start times and can be inefficient (sometimes a block isn't mined until one minute into the batch).

A better approach could be to use the system clock and wait for a certain amount of time (e.g. 30 seconds but could be configurable) before starting to process the batch. Even if there is no block mined within 30 seconds of the 5 minute cutoff, any future block is likely going to have timestamp on batch `t` and thus not affect the previous batch `t-1` in terms of orders (making it safe to start processing).

An even more optimal solution could use a hybrid, where we wait at most 30s but start if we see a block with the right timestamp before (yet this adds significant extra complexity and might not be worth it).

e00E:
Closes #616

Test Plan: CI

### DIFF
--- a/driver/src/driver/mod.rs
+++ b/driver/src/driver/mod.rs
@@ -1,1 +1,2 @@
 pub mod stablex_driver;
+pub mod scheduler;
--- a/driver/src/driver/mod.rs
+++ b/driver/src/driver/mod.rs
@@ -1,2 +1,2 @@
-pub mod stablex_driver;
 pub mod scheduler;
+pub mod stablex_driver;
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -144,11 +144,16 @@ impl<'a> Scheduler<'a> {
     }
 
     fn wait_for_next_batch_id(&self, currently_being_solved: BatchId) {
-        let duration = currently_being_solved
-            .next()
-            .solve_start_time()
+        let next = currently_being_solved.next();
+        let duration = (next.solve_start_time()
+            + self.auction_timing_configuration.target_start_solve_time)
             .duration_since(SystemTime::now())
             .unwrap_or(Duration::from_secs(1));
+        info!(
+            "Waiting {}s to handle batch {}.",
+            duration.as_secs(),
+            next.0
+        );
         thread::sleep(duration);
     }
 }
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -85,7 +85,7 @@ impl<'a> Scheduler<'a> {
     }
 
     /// Returns how long to sleep before starting the next iteration.
-    fn run_forever_single_iteration(&mut self, now: SystemTime) -> Duration {
+    fn run_once(&mut self, now: SystemTime) -> Duration {
         match self.determine_action(now) {
             Ok(Action::Sleep(duration)) => {
                 info!("Sleeping {}s.", duration.as_secs());
@@ -95,7 +95,7 @@ impl<'a> Scheduler<'a> {
                 info!("Starting to solve batch {}.", batch_id.0);
                 match self.driver.run(batch_id.0.into()) {
                     DriverResult::Ok => {
-                        info!("Batch {} solved without error.", batch_id.0);
+                        info!("Batch {} solved successfully.", batch_id.0);
                         self.last_solved_batch.replace(batch_id);
                         Duration::from_secs(0)
                     }
@@ -154,7 +154,7 @@ mod tests {
     use ethcontract::U256;
     use mockall::predicate::eq;
 
-    use super::super::stablex_driver::MockStableXDriver;
+    use crate::driver::stablex_driver::MockStableXDriver;
     use super::*;
 
     #[test]
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -1,10 +1,9 @@
 use super::stablex_driver::StableXDriver;
-use crate::orderbook::StableXOrderBookReading;
 use anyhow::{anyhow, Context, Result};
 use log::error;
 use log::info;
 use std::thread;
-use std::time::{Duration, Instant, SystemTime, SystemTimeError};
+use std::time::{Duration, SystemTime, SystemTimeError};
 
 const BATCH_DURATION: Duration = Duration::from_secs(300);
 
@@ -21,12 +20,12 @@ impl BatchId {
         Self::current(now).map(|batch_id| batch_id.prev())
     }
 
-    pub fn start_time(self) -> SystemTime {
+    pub fn order_collection_start_time(self) -> SystemTime {
         SystemTime::UNIX_EPOCH + Duration::from_secs(self.0 * BATCH_DURATION.as_secs())
     }
 
     pub fn solve_start_time(self) -> SystemTime {
-        self.start_time() + BATCH_DURATION
+        self.order_collection_start_time() + BATCH_DURATION
     }
 
     pub fn next(self) -> BatchId {
@@ -164,22 +163,22 @@ mod tests {
         let start_time = SystemTime::UNIX_EPOCH;
         let batch_id = BatchId::current(start_time).unwrap();
         assert_eq!(batch_id.0, 0);
-        assert_eq!(batch_id.start_time(), start_time);
+        assert_eq!(batch_id.order_collection_start_time(), start_time);
 
         let start_time = SystemTime::UNIX_EPOCH;
         let batch_id = BatchId::current(start_time + Duration::from_secs(299)).unwrap();
         assert_eq!(batch_id.0, 0);
-        assert_eq!(batch_id.start_time(), start_time);
+        assert_eq!(batch_id.order_collection_start_time(), start_time);
 
         let start_time = SystemTime::UNIX_EPOCH + Duration::from_secs(300);
         let batch_id = BatchId::current(start_time).unwrap();
         assert_eq!(batch_id.0, 1);
-        assert_eq!(batch_id.start_time(), start_time);
+        assert_eq!(batch_id.order_collection_start_time(), start_time);
 
         let start_time = SystemTime::UNIX_EPOCH + Duration::from_secs(300);
         let batch_id = BatchId::current(start_time + Duration::from_secs(299)).unwrap();
         assert_eq!(batch_id.0, 1);
-        assert_eq!(batch_id.start_time(), start_time);
+        assert_eq!(batch_id.order_collection_start_time(), start_time);
     }
 
     #[test]
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -100,6 +100,7 @@ impl<'a> Scheduler<'a> {
         info!("Running solver for batch {}.", batch_id.0);
         if let Err(err) = self.driver.run(batch_id.0.into()) {
             error!("StableXDriver error: {}", err);
+            thread::sleep(Duration::from_secs(10));
         } else {
             self.last_solved_batch.replace(batch_id);
         }
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -1,5 +1,5 @@
 use super::stablex_driver::{DriverResult, StableXDriver};
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use log::error;
 use log::info;
 use std::thread;
@@ -7,32 +7,34 @@ use std::time::{Duration, SystemTime, SystemTimeError};
 
 const BATCH_DURATION: Duration = Duration::from_secs(300);
 
+/// Wraps a batch id as in the smart contract to add functionality related to
+/// the current time.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 struct BatchId(u64);
 
 impl BatchId {
-    pub fn current(now: SystemTime) -> std::result::Result<Self, SystemTimeError> {
+    fn current(now: SystemTime) -> std::result::Result<Self, SystemTimeError> {
         let time_since_epoch = now.duration_since(SystemTime::UNIX_EPOCH)?;
         Ok(Self(time_since_epoch.as_secs() / BATCH_DURATION.as_secs()))
     }
 
-    pub fn currently_being_solved(now: SystemTime) -> std::result::Result<Self, SystemTimeError> {
+    fn currently_being_solved(now: SystemTime) -> std::result::Result<Self, SystemTimeError> {
         Self::current(now).map(|batch_id| batch_id.prev())
     }
 
-    pub fn order_collection_start_time(self) -> SystemTime {
+    fn order_collection_start_time(self) -> SystemTime {
         SystemTime::UNIX_EPOCH + Duration::from_secs(self.0 * BATCH_DURATION.as_secs())
     }
 
-    pub fn solve_start_time(self) -> SystemTime {
+    fn solve_start_time(self) -> SystemTime {
         self.order_collection_start_time() + BATCH_DURATION
     }
 
-    pub fn next(self) -> BatchId {
+    fn next(self) -> BatchId {
         self.0.checked_add(1).map(BatchId).unwrap()
     }
 
-    pub fn prev(self) -> BatchId {
+    fn prev(self) -> BatchId {
         self.0.checked_sub(1).map(BatchId).unwrap()
     }
 }
@@ -116,7 +118,7 @@ impl<'a> Scheduler<'a> {
     /// Return current batch id and what to do with it.
     fn determine_action(&self, now: SystemTime) -> Result<Action> {
         let solving_batch = BatchId::currently_being_solved(now)
-            .with_context(|| anyhow!("failed to get batch id currently being solved"))?;
+            .context("failed to get batch id currently being solved")?;
         let intended_solve_start_time = solving_batch.solve_start_time()
             + self.auction_timing_configuration.target_start_solve_time;
         // unwrap here because this cannot fail because the `solving_batch`'s
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -1,10 +1,8 @@
 use super::stablex_driver::StableXDriver;
-use crate::metrics::StableXMetrics;
 use crate::orderbook::StableXOrderBookReading;
 use anyhow::{anyhow, Context, Result};
 use log::error;
 use log::info;
-use std::collections::HashSet;
 use std::thread;
 use std::time::{Duration, Instant, SystemTime, SystemTimeError};
 
@@ -40,37 +38,44 @@ impl BatchId {
     }
 }
 
+#[derive(Debug)]
+pub struct AuctionTimingConfiguration {
+    /// The offset from the start of a batch at which point we should start
+    /// solving.
+    pub target_start_solve_time: Duration,
+    /// The offset from the start of the batch at which point there is not
+    /// enough time left to attempt to solve.
+    pub latest_solve_attempt_time: Duration,
+}
+
 pub struct Scheduler<'a> {
     driver: &'a mut dyn StableXDriver,
     orderbook_reader: &'a dyn StableXOrderBookReading,
-    metrics: &'a StableXMetrics,
-    batch_wait_time: Duration,
-    max_batch_elapsed_time: Duration,
-    past_auctions: HashSet<BatchId>,
+    auction_timing_configuration: AuctionTimingConfiguration,
 }
 
 #[derive(Debug, Eq, PartialEq)]
 enum Action {
-    AlreadyHandled,
     NotEnoughTimeLeft,
-    Solve,
+    SolveImmediately,
+    SolveAtTime(SystemTime),
 }
 
 impl<'a> Scheduler<'a> {
     pub fn new(
         driver: &'a mut dyn StableXDriver,
         orderbook_reader: &'a dyn StableXOrderBookReading,
-        metrics: &'a StableXMetrics,
-        batch_wait_time: Duration,
-        max_batch_elapsed_time: Duration,
+        auction_timing_configuration: AuctionTimingConfiguration,
     ) -> Self {
+        assert!(auction_timing_configuration.latest_solve_attempt_time <= BATCH_DURATION);
+        assert!(
+            auction_timing_configuration.target_start_solve_time
+                < auction_timing_configuration.latest_solve_attempt_time
+        );
         Self {
             driver,
             orderbook_reader,
-            metrics,
-            batch_wait_time,
-            max_batch_elapsed_time,
-            past_auctions: HashSet::new(),
+            auction_timing_configuration,
         }
     }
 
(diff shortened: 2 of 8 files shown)


e00E:
The problem with CI here is that in the ganache test we increment the time manually using `evm_increaseTime` which makes the batch id according to system time and batch id according to ganache out of sync. I'll have to think about the best solution.

driver/src/driver/stablex_driver.rs
```diff
@@ -25,52 +41,77 @@ impl<'a> StableXDriver<'a> {
...
+            }
+            thread::sleep(sleep_interval);
         }
+    }
 
-        self.metrics
-            .auction_processing_started(&batch_to_solve_result);
-        let batch_to_solve = batch_to_solve_result?;
+    fn run_internal(&mut self, now: SystemTime) -> Result<bool> {
+        let open_batch = BatchId::current(now)?;
```
fleupold: I like the BatchId abstraction. Maybe we can implement previous and into<U256> directly on it to avoid accessing the internals here.

fleupold (commented):
I'm a bit worried that we are mixing hard to test threading code (sleep/system time, etc) into our logic heavy driver component (which has a lot of tests associated with it).

Maybe it makes sense to introduce the concept of a scheduler here? This component can take care of sleeping and assess when is the right time to start the driver. The driver itself will then just run through it's state-machine and execute the different processing steps (fetch orderbook, run optimization, verify, submit solution).

e00E:
> Maybe it makes sense to introduce the concept of a scheduler here?

Very true. Will do.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,234 @@
...
+            max_batch_elapsed_time,
+            past_auctions: HashSet::new(),
+        }
+    }
+
+    pub fn run_forever(&mut self) {
+        loop {
+            let now = SystemTime::now();
+            match self.determine_action(now) {
+                Ok((batch_id, Action::AlreadyHandled)) => {
```
nlordell: If we're always waiting for batches to end, should this case really ever happen? If it shouldn't I would suggest it be an `Err` and not an `Ok`.

e00E: I thought I'd leave this in in case the system time is changed while the scheduler is running. If it is changed backwards while we are waiting on the next batch then it can happen.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /// Return current batch id and what to do with it.
+    fn determine_action(&self, now: SystemTime) -> Result<(BatchId, Action)> {
+        let open_batch =
+            BatchId::current(now).with_context(|| anyhow!("failed to get current batch id"))?;
+        let solving_batch = open_batch.prev();
+        // unwrap here because this cannot fail because the `solving_batch`'s
+        // start time is always before `now`.
+        let elapsed_time = now.duration_since(solving_batch.start_time()).unwrap() - BATCH_DURATION;
```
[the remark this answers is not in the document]
e00E: In this function we don't sample SystemTime. The only SystemTime used is the one passed in as `now` so I don't think this can happen.

e2e/src/common.rs
```diff
@@ -95,6 +95,8 @@ pub fn wait_for(web3: &Web3<Http>, seconds: u32) {
         .execute("evm_mine", vec![])
         .wait()
         .expect("Cannot mine to increase time");
+    // TODO vk:
+    std::thread::sleep(std::time::Duration::from_secs(seconds as u64 + 10));
```
e00E: Does anyone have a better idea for this? I don't like making the Ganache test run for 10 minutes but I don't see a way around it unless we add extra code to the scheduler just so the ganache test works.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+                        batch_id.0
+                    );
+                    self.wait_for_next_batch_id(batch_id);
+                }
+                Ok((batch_id, Action::NotEnoughTimeLeft)) => {
+                    info!(
+                        "Skipping batch {} because there is not enough time left.",
+                        batch_id.0
+                    );
+                    self.metrics.auction_skipped(batch_id.0.into());
```
fleupold: This method triggers a success log, which shouldn't be the case if we are running out of time (because we haven't actually processed the batch and thus it shouldn't count towards our success metric). 
Maybe we should rename this method to "processed_not_submitted" and have another "not_enough_time" metric for this case here.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+            thread::sleep(sleep_interval);
+        }
+    }
+
+    fn wait_for_next_batch_id(&self, currently_being_solved: BatchId) {
+        let mut duration = Duration::from_secs(1);
+        if let Ok(duration_) = currently_being_solved
+            .next()
+            .solve_start_time()
+            .duration_since(SystemTime::now())
```
fleupold: nit: why not unwrap_or()?

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+                    info!(
+                        "Skipping batch {} because there is not enough time left.",
+                        batch_id.0
+                    );
+                    self.metrics.auction_skipped(batch_id.0.into());
+                    self.wait_for_next_batch_id(batch_id);
+                }
+                Ok((batch_id, Action::Solve)) => {
+                    let deadline = batch_id.start_time() + self.batch_wait_time;
+                    if let Ok(duration) = deadline.duration_since(now) {
```
fleupold: Why do we have to wait here again? I'd expect Action::Solve to potentially provide the deadline as a field (so it can be unit tested and eventually passed down to the price_finder) and only invoke self.driver.run(). Since we rightfully don't unit test the run_forever method it should be free of meaningful logic (it should basically just wire things together).

We could have another action which is called Wait(time_to_wait)

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    pub fn prev(self) -> BatchId {
+        self.0.checked_sub(1).map(BatchId).unwrap()
+    }
+}
+
+pub struct Scheduler<'a> {
+    driver: &'a mut dyn StableXDriver,
+    orderbook_reader: &'a dyn StableXOrderBookReading,
```
fleupold: We are not really reading the orderbook here anymore. May be worthwhile splitting this in to a OrderBookReading and AuctionStateReading at some point (maybe a separate PR)?

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+    metrics: &'a StableXMetrics,
+    batch_wait_time: Duration,
+    max_batch_elapsed_time: Duration,
```
fleupold: Does it make sense to combine these into a AucitonTimingConfiguration struct (it would allow explicit documentation and not require to pass to identical types in this method which are easy to invert)?

Also a nit about naming here, can we call it `auction_processing_start_time` and `solution_submission_time`. I find batch_wait_time too generic in that it could mean anything.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+use log::error;
+use log::info;
+use std::collections::HashSet;
+use std::thread;
+use std::time::{Duration, Instant, SystemTime, SystemTimeError};
+
+const BATCH_DURATION: Duration = Duration::from_secs(300);
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+struct BatchId(pub u64);
```
fleupold: Not a huge fan of all the `batch_id.0`. Should we just implement into<u64> and make this field private?

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+            metrics,
+            batch_wait_time,
+            max_batch_elapsed_time,
+            past_auctions: HashSet::new(),
+        }
+    }
+
+    pub fn run_forever(&mut self) {
+        loop {
+            let now = SystemTime::now();
```
fleupold: Do we have to be careful with timezones here? Can we specify we want now in UTC somehow?

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+        ))
+    }
+
+    /// Wait until the batch id matches the batch id according to the order book
+    /// but at most until deadline.
+    fn wait_for_batch_id(&self, batch_id: BatchId, deadline: Instant) {
+        info!("Waiting for the next batch ({}) to begin.", batch_id.0);
+        let sleep_interval = Duration::from_secs(5);
+        loop {
+            if let Ok(index) = self.orderbook_reader.get_auction_index() {
```
fleupold: What is your though on having the dual logic "either new batch mined or deadline expired"? It adds a bunch of complexity and one could argue that only having the deadline makes the code more predictable and less prone to 1 block reorgs that could happen.

e2e/src/common.rs
fleupold: Why is this needed? Because we sleep for the entire time of the batch in one go? Maybe it's fine to break the scheduler into multiple smaller sleeps?

If we decide to remove the code that is checking for the most up to date batch Id and only rely on system time, we might want to implement our own SystemTime abstraction that on ganache we can replace with an object that uses the blockchain time.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn current(now: SystemTime) -> std::result::Result<Self, SystemTimeError> {
+        let time_since_epoch = now.duration_since(SystemTime::UNIX_EPOCH)?;
+        Ok(Self(time_since_epoch.as_secs() / BATCH_DURATION.as_secs()))
+    }
+
+    pub fn currently_being_solved(now: SystemTime) -> std::result::Result<Self, SystemTimeError> {
+        Self::current(now).map(|batch_id| batch_id.prev())
+    }
+
+    pub fn start_time(self) -> SystemTime {
```
fleupold: let's call this order_collection_start_time to be explicit.

fleupold (commented):
It's great to see how the Scheduler makes the driver much simpler. Some requests inline to fully separate the testable and non testable logic, but I think this is a really good way. 

Generally I feel this logic is still a bit too complicated. In particular, all these configuration parameters are starting to become hard to wrap my head around.

driver/src/driver/scheduler.rs
nlordell: From what I understand from the docs, `SystemTime` is completely opaque and cannot be directly inspected, but rather compared to another time. Since `UNIX_EPOCH` is timezone agnostic, I don't think its an issue here.

driver/src/driver/scheduler.rs
e00E: Removed metric from scheduler. Don't think we need a new metric for not enough time, especially when we have the threaded scheduer.

driver/src/driver/scheduler.rs
e00E: I removed the get_auction_index call in the scheduler for simplicity of always waiting a fixed time.

driver/src/driver/scheduler.rs
e00E: I came to the same conclusion. Time zones don't affect this.

driver/src/driver/scheduler.rs
e00E: I agree.

driver/src/driver/scheduler.rs
e00E: Made this a struct and changed names to target_start_solve_time and latest_solve_attempt_time.

driver/src/driver/scheduler.rs
e00E: Disagree. In this file we treat batch ids as u64 because it is more convenient than U256. Every u64 is a valid batch id and vice versa. The reason there is a wrapper is just to give this type a name and have one place to implement methods on it.

e2e/src/common.rs
e00E: I am still looking into this part. The problem is what you mentioned and that ganache does not advance time at all unless `evm_increaseTime` is called. So if you have the `docker-compose stablex` running for 15 minutes the expected current batch id in the scheduler is `x + 3` while in Ganache it is still `x`. 
I'll see if I can `evm_increaseTime` the proper amount at the start of the test.

fleupold: As far as I know, ganache increases time one each evm_mine so it should advance whenever a transaction is sent (even if it was manipulated to be in the future by using evm_increaseTime). In case there is no transaction I guess there is also no need to run the driver. 
There is still the problem of scheduling a long sleep based on an old time, but we could break the sleep statement into smaller batches and recheck time every 10s or so.

fleupold (approved):
Looks good to me.

driver/src/main.rs
```diff
@@ -124,6 +127,26 @@ struct Options {
         parse(try_from_str = duration_millis),
     )]
     gas_station_timeout: Duration,
+
+    /// The offset from the start of a batch in seconds at which point we
+    /// should start solving.
+    #[structopt(
+        long,
+        env = "BATCH_WAIT_TIME",
```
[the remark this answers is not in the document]
e00E: Absolutely. Forgot to rename the env.

driver/src/driver/scheduler.rs
e00E: True, doesn't have to be public and the pub did nothing anyway because the the wrapper itself wasn't public.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,246 @@
...
+                    // unwrap is ok because it would be a logic error in
+                    // determine_action to return SolveAtTime with a time that
+                    // is not actually in the future.
+                    let duration = target_time.duration_since(now).unwrap();
+                    info!(
+                        "Waiting {}s to handle batch {}.",
+                        duration.as_secs(),
+                        batch_id.0
+                    );
+                    thread::sleep(duration);
```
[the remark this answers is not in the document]
e00E: Good point. Implemented this. Instead of processed_instances set we just keep track of last_solved_batch.

e2e/src/common.rs
e00E: I have fixed this now. It should always work for the first time the ganache e2e test is executed but if the test is executed mutliple times without `docker-compose down` in between then it can fail.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+
+    fn run_solver(&mut self, batch_id: BatchId) {
+        info!("Running solver for batch {}.", batch_id.0);
+        if let Err(err) = self.driver.run(batch_id.0.into()) {
+            error!("StableXDriver error: {}", err);
+            thread::sleep(Duration::from_secs(10));
+        } else {
+            self.last_solved_batch.replace(batch_id);
```
fleupold: In the current logic we also skip the batch going forward if the error came from the price finding, verification or solution submission as in those case it's unlikely to succeed in a retry (and it might clog the next batch). So really we only retry if the order book fetching failed for some reason (that was handy when we were running into occasional timeouts).

In order to reproduce this behaviour we would have to differentiate between driver error on a more fine level (maybe a should_retry flag or the like) and I'm not sure that's worth it. So I'd suggest to always add the batchID as "last_solved_batch" even if there was an error.

What do you think?

e2e/tests/stablex_test.rs
```diff
@@ -69,15 +69,25 @@ fn test_with_ganache() {
         )
         .from(Account::Local(accounts[1], None))
         .wait_and_expect("Cannot place first order");
-    close_auction(&web3, &instance);
 
+    let batch_id = instance.get_current_batch_id().call().wait().unwrap();
+    close_auction(&web3, &instance);
+    let wait_time = Duration::from_secs(310);
```
fleupold: While this is fine for now, I still think it's a bummer that the ganache e2e test now takes > 5 minutes in the worst case. 

Is there a reason why you want to do the sleeps in "one go" instead of waking up every e.g. 5s and re-determining action? This would allow us to build this fake System clock based on the ganache

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2Fi5wNCqyMzY2Oc%2Fgiphy.gif&token=beautiful" data-gitmeme-token="beautiful"></a>

I really like the way the logic is split and tested in the scheduler. Let's address the retry behaviour and discuss how we could improve the e2e test (can be done in a later PR) and it's good to go. 

@nlordell would you also be able to take a look?

I noticed that the description and testplan are a bit weak (the issue explains the what but not really the how). For a change like this I think testing it manually against rinkeby should be part of the testplan.

driver/src/driver/mod.rs
```diff
@@ -1 +1,2 @@
 pub mod stablex_driver;
+pub mod scheduler;
```
nlordell: ubernit: missing EOF newline.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,274 @@
...
+use anyhow::{anyhow, Context, Result};
+use log::error;
+use log::info;
+use std::thread;
+use std::time::{Duration, SystemTime, SystemTimeError};
+
+const BATCH_DURATION: Duration = Duration::from_secs(300);
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+struct BatchId(u64);
```
nlordell: Could you add a doc comment for this type?

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,274 @@
...
+
+impl BatchId {
+    pub fn current(now: SystemTime) -> std::result::Result<Self, SystemTimeError> {
```
nlordell: None of these methods need to be `pub` as `BatchId` is only used in this module right?

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,274 @@
...
+                Ok(Action::Solve(batch_id)) => {
+                    info!("Solving batch {}.", batch_id.0);
+                    self.run_solver(batch_id);
+                }
+                Err(err) => {
+                    error!("Scheduler error: {}", err);
+                    thread::sleep(Duration::from_secs(10));
+                }
+            }
+        }
```
nlordell: The way this is implemented now is :ok_hand:

driver/src/driver/scheduler.rs
nlordell: I agree as well, we should log the error, and skip to the next batch in case of solver error. The `driver.run()` method can return something like a:
```
enum DriverResult {
    Ok,
    Retry(Error),
    Skip(Error),
}
```

That way we can differentiate between an error where we want to retry and a hard error where we want to skip the batch.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    fn run_solver(&mut self, batch_id: BatchId) {
+        info!("Running solver for batch {}.", batch_id.0);
+        if let Err(err) = self.driver.run(batch_id.0.into()) {
+            error!("StableXDriver error: {}", err);
+            thread::sleep(Duration::from_secs(10));
```
nlordell: I would rather all sleeps be in the `run_forever` loop. I'm not sure what the best way to do this would be though.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,274 @@
...
+        } else {
+            self.last_solved_batch.replace(batch_id);
+        }
+    }
+
+    /// Return current batch id and what to do with it.
+    fn determine_action(&self, now: SystemTime) -> Result<Action> {
+        let solving_batch = BatchId::currently_being_solved(now)
+            .with_context(|| anyhow!("failed to get batch id currently being solved"))?;
```
nlordell: nit: context can be a `&'static str`.

nlordell (approved):
Looks good, just some minor nits.

e2e/tests/stablex_test.rs
e00E: If I understand that approach correctly then we would have to introduce SystemClock trait, a new type that implements it based on the real SystemClock and one on Ganache (by sending it requests whenever the SystemTime is retrieved) and then command line options to switch.
To me that is too much extra logic that only exists for a test. If we had always been using the SystemClock and now wanted to create the Ganache test I don't think this would feel natural.
But if you think this is worth or I misunderstood the complexity then I can implement that. In that case let's move it into a separate PR after this one.

driver/src/driver/scheduler.rs
e00E: Implemented this.

driver/src/driver/scheduler.rs
e00E: This one is just so we don't busy loop too hard in case we would retry. Imo it's fine and its not hidden as its just a few lines below where the function is called.

nlordell: The motivation was more about keeping all sleeping logic in the main looping function. If we move sleeps to other functions they have side-effects that aren't immediately obvious to someone who is reading `run_forever` for example.

nlordell: One way to model this is that the `match` expression in `run_forever` evaluates to a `Option<Duration>` and then `thread::sleep` is called with that duration if it is `Some`. That way we only have one sleep and the various match branches sleep with different values.

If we do it this way we an even refactor out a single `run_forever` iteration into its own method and have unit test(s) to verify that the duration that is supposed to be slept is as intended.

e00E: Implemented this and added a test for it.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use anyhow::anyhow;
+    use ethcontract::U256;
+    use mockall::predicate::eq;
+
+    use super::super::stablex_driver::MockStableXDriver;
```
nlordell: nit:
```suggestion
    use crate::driver::stablex_driver::MockStableXDriver;
```

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    pub fn run_forever(&mut self) {
+        loop {
+            thread::sleep(self.run_forever_single_iteration(SystemTime::now()));
+        }
+    }
+
+    /// Returns how long to sleep before starting the next iteration.
+    fn run_forever_single_iteration(&mut self, now: SystemTime) -> Duration {
```
nlordell: nit:
```suggestion
    fn run_once(&mut self, now: SystemTime) -> Duration {
```

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,350 @@
...
+                info!("Sleeping {}s.", duration.as_secs());
+                duration
+            }
+            Ok(Action::Solve(batch_id)) => {
+                info!("Starting to solve batch {}.", batch_id.0);
+                match self.driver.run(batch_id.0.into()) {
+                    DriverResult::Ok => {
+                        info!("Batch {} solved without error.", batch_id.0);
+                        self.last_solved_batch.replace(batch_id);
+                        Duration::from_secs(0)
```
nlordell: ubernit: I would slightly prefer returning `None` in this case as its a little more clear that it should not sleep.

e2e/tests/stablex_test.rs
nlordell: I also agree this should be a separate PR.

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,350 @@
...
+                    }
+                    DriverResult::Retry(err) => {
+                        error!("StableXDriver retryable error: {}", err);
+                        Duration::from_secs(10)
```
nlordell: nit: Can this be a `const RETRY_TIMEOUT`?

nlordell (approved):
Love it! Just a couple minor nits.

driver/src/driver/scheduler.rs
nlordell: Just reviewed it! Looks good :)

nlordell:
@e00E The issue here does not appear to be related to Kraken but:
```
stablex_1_f2d7b83fde60 | Mar 12 09:55:53.326 INFO [driver::driver::scheduler] Batch 5280022 solved without error.
```

driver/src/driver/scheduler.rs
```diff
@@ -0,0 +1,350 @@
...
+        match self.determine_action(now) {
+            Ok(Action::Sleep(duration)) => {
+                info!("Sleeping {}s.", duration.as_secs());
+                duration
+            }
+            Ok(Action::Solve(batch_id)) => {
+                info!("Starting to solve batch {}.", batch_id.0);
+                match self.driver.run(batch_id.0.into()) {
+                    DriverResult::Ok => {
+                        info!("Batch {} solved without error.", batch_id.0);
```
nlordell:
```suggestion
                        info!("Batch {} solved successfully.", batch_id.0);
```

driver/src/driver/scheduler.rs
e00E: Not very important to me but I don't agree because with that argument a 0 Duration never makes sense so I'll leave it like this.

nlordell: He issue is that `thread::sleep(0)` sometimes behaves like a `yield` (I would imagine that it does in Rust as well but I’m not sure). That means sleeping for 0 and not sleeping at all may have slightly different semantics and from the code I don’t think yielding the thread is intended.

nlordell: Also, as it is right now, we would usually sleep for 0 and then sleep for the desired duration.

josojo:
I am very sorry for joining the discussion so late. But can I get your opinions here:
https://github.com/gnosis/dex-services/issues/616#issuecomment-598721774
?

e00E:
Conclusion was that we will merge this because we will soon switch to an orderbook based on events.

driver/src/driver/scheduler.rs
e00E: I am leaving this in because it is benign and it will change when we merge the each batch in own thread anyway.

bh2smith (commented):
This was a very nice read. Really great changes made here!

PR #620, opened 2020-03-09
Inlcude batchID in instance/result file name

fleupold:
This PR makes it so that our instance/result files include the batchID and are also split up by date. We now create instance filenames in the following format:

`instances/{day}/instance_{batchId}_{datetime}.json`

and results in the following format:

`results/{day}/instance_{batchId}_{datetime}/<json|visualization>`

### Test Plan

Tested locally that files are created and read accordingly. For synchronisation with S3 I have to slightly adjust the sync script to not be too strict about the matching.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,4 +35,3 @@ services:
     volumes:
       # Sync src folder to allow development without rebuilding on each change
       - ./driver:/app/driver
-      - ./dex-contracts:/app/dex-contracts
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,3 +35,4 @@ services:
     volumes:
       # Sync src folder to allow development without rebuilding on each change
       - ./driver:/app/driver
+      - ./dex-contracts:/app/dex-contracts
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -224,7 +224,8 @@ impl PriceFinding for OptimisationPriceFinder {
         };
 
         let now = Utc::now();
-        let batch_id = now.timestamp() / 300;
+        // We are solving the batch before the current one
+        let batch_id = (now.timestamp() / 300) - 1;
         let date = now.format("%Y-%m-%d");
         // We only need to create the folder for the input, the output
         // folder is created by the python script
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -176,7 +176,6 @@ impl OptimisationPriceFinder {
         solver_type: SolverConfig,
         price_oracle: impl PriceEstimating + 'static,
     ) -> Self {
-        create_dir_all("instances").expect("Could not create instance directory");
         OptimisationPriceFinder {
             write_input,
             run_solver,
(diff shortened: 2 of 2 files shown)


driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -223,9 +222,27 @@ impl PriceFinding for OptimisationPriceFinder {
             orders: orders.iter().map(From::from).collect(),
             fee: self.fee.as_ref().map(From::from),
         };
-        let current_time = Utc::now().to_rfc3339();
-        let input_file = format!("instances/instance_{}.json", &current_time);
-        let result_folder = format!("results/instance_{}/", &current_time);
+
+        let now = Utc::now();
+        let batch_id = now.timestamp() / 300;
```
e00E: Should it be `now.timestamp() / 300 - 1`?

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -223,9 +222,27 @@ impl PriceFinding for OptimisationPriceFinder {
...
+        let input_file = format!(
+            "{}/instance_{}_{}.json",
+            &input_folder,
+            &batch_id,
+            &now.to_rfc3339()
+        );
+        let result_folder = format!(
+            "results/{}/instance_{}_{}/",
+            &date,
+            &batch_id,
```
e00E: To have this ordered lexicographically correctly we should 0 pad the batch_id.

e00E (approved):

docker-compose.yml
```diff
@@ -35,3 +35,4 @@ services:
     volumes:
       # Sync src folder to allow development without rebuilding on each change
       - ./driver:/app/driver
+      - ./dex-contracts:/app/dex-contracts
```
nlordell: Is this change related to the PR? I'm not sure I understand why it is needed for the instance and result file name changes.

driver/src/price_finding/optimization_price_finder.rs
nlordell: Makes sense to me as we are solving for the previous batch.

driver/src/price_finding/optimization_price_finder.rs
nlordell: While that is in general true, I don't think that is needed here as we are currently at batch `5279419` and until batch `10000000` is about 44 years (if my math isn't wrong).

nlordell (approved):
LGTM

driver/src/price_finding/optimization_price_finder.rs
fleupold: Great catch!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F3o6ZsSRkXifoEJfpMA%2Fgiphy-downsized-medium.gif&token=great_catch" data-gitmeme-token="great_catch"></a>

driver/src/price_finding/optimization_price_finder.rs
fleupold: It's a valid point, but I wouldn't know how many 0s I should use for padding and it's quite annoying to look at.

docker-compose.yml
fleupold: It's not required, but more convenient as otherwise you are stuck with the contract definitions from the time you built the container (might be outdated). And rebuilding just to get the contracts in there is time consuming. Will make this a separate PR though.

PR #623, opened 2020-03-10
Reset processing times between batches

fleupold:
At the moment we "carry over" the processing times of previous batches. That is, if a certain stage (e.g. verification or submission) is not performed in the current batch its time from the last batch it was carried out is kept (cf. the red line for "solution submission" in this image):

Screen Shot 2020-03-10 at 15 07 30

This PR makes it so that processing times get reset to 0 whenever a batch is started.

### Test Plan

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FCDZwopbecAbIc%2Fgiphy-downsized-medium.gif&token=production" data-gitmeme-token="production"></a>

### DIFF
--- a/driver/src/metrics/stablex_metrics.rs
+++ b/driver/src/metrics/stablex_metrics.rs
@@ -74,6 +74,12 @@ impl StableXMetrics {
     }
 
     pub fn auction_processing_started(&self, res: &Result<U256>) {
+        // Reset values from previous batch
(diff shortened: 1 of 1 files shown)


e00E (approved):
Previously we worried that this could lead to use never observing the previous processing times. I guess this is still case the case but it is unlikely to happen so I don't have a problem with this.

PR #626, opened 2020-03-11
Lazily initialize dexag client

e00E:
We want to avoid doing heavy work on construction.

### DIFF
--- a/driver/src/price_estimation/dexag/mod.rs
+++ b/driver/src/price_estimation/dexag/mod.rs
@@ -2,24 +2,31 @@ mod api;
 
 use super::{PriceSource, Token};
 use crate::models::TokenId;
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use api::{DexagApi, DexagHttpApi};
 use futures::future::{self, BoxFuture};
+use std::cell::RefCell;
 use std::collections::HashMap;
 
-pub struct DexagClient<Api> {
-    api: Api,
+struct ApiTokens {
     // Maps Token::symbol to Token.
     // This is cached in the struct because we don't expect it to change often.
     tokens: HashMap<String, api::Token>,
     stable_coin: api::Token,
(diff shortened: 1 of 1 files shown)


driver/src/price_estimation/dexag/mod.rs
```diff
@@ -57,13 +81,26 @@ where
     Api: DexagApi + Sync,
 {
     fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
+        if tokens.is_empty() {
+            return Ok(HashMap::new());
+        }
+
+        self.initialize_if_needed()
```
fleupold: Can this method return the api token list as part of the result? This way we don't have to borrow and unwrap.

fleupold (commented):
Thanks for tackling this! What about using https://docs.rs/lazy-init/0.3.0/lazy_init/struct.Lazy.html? Looks like an abstract enough issue to be generalised.

driver/src/price_estimation/dexag/mod.rs
e00E: Agree that it's nicer. Implemented.

e00E:
I did not know about `lazy_init`. After trying it out, it is does not fit in this case because our initialization can fail whereas the crate expects it to never fail.

fleupold:
> whereas the crate expects it to never fail

I saw this too, but was thinking we can just make it return a Result<HashMap>

fleupold (approved):
No strong feeling about using the library, up to you...

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F3o7Z4tNIWOQWTlfGN2%2Fgiphy.gif&token=green_light" data-gitmeme-token="green_light"></a>

e00E:
But then if it returns an error the first time I won't update it the second time because it thinks that is already the correct result.

PR #627, opened 2020-03-12
Fix rinkeby e2e test

e00E:
`logs.to_lowercase().contains("erro")` was too general and got triggered
by this line

```
stablex_1_3fbc83eb8af8 | Mar 12 09:24:37.482 WARN [driver::price_estimation::average_price_source] one price source failed: Kraken API errors: ["EQuery:Unknown asset pair"]
```

https://travis-ci.com/github/gnosis/dex-services/builds/152950633

I am not sure if the best solution is, just throwing this out there. Even with this more specialized searching we can still get false positives. Maybe we should really parse line by line and make sure `ERRO` occurs at the correct position?
Why did we have `to_lowercase` in the first place? I removed it because I didn't understand it.

### DIFF
--- a/driver/src/logging.rs
+++ b/driver/src/logging.rs
@@ -1,3 +1,4 @@
+use slog::Level;
 use slog::{o, Drain, Logger, OwnedKVList, Record};
 use slog_async::Async;
 use slog_envlogger::LogBuilder;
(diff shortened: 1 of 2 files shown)


e00E:
I don't understand why my PR runs into this issue but not this one we just merged https://github.com/gnosis/dex-services/pull/596 .

nlordell:
It might have been something wrong with Kraken at the time. We shouldn't be getting "unknown asset pair" errors as the asset pairs are queried from Kraken before getting the prices and only ones that match the contract tokens are used. This means that the asset pair IDs (that the Kraken API is complaining doesn't exist) come from Kraken.

nlordell:
Maybe @fleupold should take a look as I am not sure why `to_lowercase` was used. I also am not a fan of this approach. I would rather we set `DFUSION_LOG=error` so only errors get logged and then check that `stdout` and `stderr` are both empty.

e00E:
The new idea is to divert error logs to a file, then check whether the file is empty.

fleupold:
> I am not sure why to_lowercase was used.

This was a clowny way to check for errors, from a time before we had proper logging with different severities. Your proposals are much better than that.

PR #628, opened 2020-03-13
Log only errors to stderr

e00E:
This fixes false positives in the e2e test where the test thinks an
error has occurred when it has not.

Test Plan: Add various logs in `main` then run the program while piping stdout and stderr to different destinations. Check that logs arrive at the correct place.

TODO: check that this change doesn't break the kubernetes environment

### DIFF
--- a/driver/src/logging.rs
+++ b/driver/src/logging.rs
@@ -1,3 +1,4 @@
+use slog::Level;
 use slog::{o, Drain, Logger, OwnedKVList, Record};
 use slog_async::Async;
 use slog_envlogger::LogBuilder;
@@ -9,7 +10,12 @@ const BUFFER_SIZE: usize = 1024;
 
 /// Initialize driver logging.
 pub fn init(filter: impl AsRef<str>) -> (Logger, GlobalLoggerGuard) {
-    let format = CustomFormatter::new(TermDecorator::new().stderr().build()).fuse();
+    // Log errors to stderr and lower severities to stdout.
+    let format = CustomFormatter::new(
+        TermDecorator::new().stderr().build(),
+        TermDecorator::new().stdout().build(),
+    )
+    .fuse();
     let drain = Async::new(LogBuilder::new(format).parse(filter.as_ref()).build())
         .chan_size(BUFFER_SIZE)
         .build();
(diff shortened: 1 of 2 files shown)


nlordell:
The issue with logging to two separate streams is that they can mangle each other and get mixed up. I'm also not sure how this will interact with how we collect logs for Kibana.

nlordell:
I would prefer keeping the current formatter, but being able to configure extra logs to go to file just for the ganache tests. This would allow us to not only not interfere with the current logging system but also have custom filters setup for the ganache tests where errors from certain modules can be ignored.

driver/src/logging.rs
```diff
@@ -21,47 +27,64 @@ pub fn init(filter: impl AsRef<str>) -> (Logger, GlobalLoggerGuard) {
     (logger, guard)
 }
 
-pub struct CustomFormatter<D: Decorator> {
-    decorator: D,
+/// Uses one decorator for `Error` and `Critical` log messages and the other for
+/// the rest.
+pub struct CustomFormatter<T0, T1> {
```
nlordell:
```suggestion
pub struct CustomFormatter<Out, Err> {
```

driver/src/logging.rs
```diff
@@ -21,47 +27,64 @@ pub fn init(filter: impl AsRef<str>) -> (Logger, GlobalLoggerGuard) {
...
-            decorator.flush()?;
+        decorator.start_whitespace()?;
+        write!(decorator, " ")?;
 
-            Ok(())
-        })
-    }
+        decorator.start_msg()?;
+        writeln!(decorator, "{}", record.msg())?;
+        decorator.flush()?;
```
nlordell: It flushes on every message :tada: So no worries about mangling the streams I think.

driver/src/logging.rs
```diff
@@ -21,47 +27,64 @@ pub fn init(filter: impl AsRef<str>) -> (Logger, GlobalLoggerGuard) {
...
     fn log(
         &self,
         record: &Record,
         values: &OwnedKVList,
     ) -> std::result::Result<Self::Ok, Self::Err> {
-        self.decorator.with_record(record, values, |mut decorator| {
-            decorator.start_timestamp()?;
-            slog_term::timestamp_utc(&mut decorator)?;
+        match record.level() {
+            Level::Error | Level::Critical => log_to_decorator(&self.err_decorator, record, values),
```
nlordell: can we include warnings here? It caught an issue in #615

nlordell (approved):
Awesome!

driver/src/logging.rs
e00E: I don't think we should. The way I imagined it was that anything on stderr is severe enough for us to get an alert on Slack. But warnings are not that severe.

nlordell: But then we would have merged #615 with the issue. I’m ok with CI being more restrictive than our alert system.

driver/src/logging.rs
fleupold: +1, nbd, but any reason this was not incorporated?

e00E: Imo this isn't an improvement because Err is also confusing because its usually used for Results. We can go all the way and call the types `StdoutDecorator` and `StderrDecorator` if you prefer that.

nlordell: I would just use `Stdout` and `Stderr` then. I'm not a fan of the numbered generic types here.

PR #631, opened 2020-03-13
Fix e2e test log error check

e00E:
`docker-compose logs` outputs both stdout and stderr logs to its own
stdout. This makes us unable to differentiate between the two and
introduced a bug where we would never detect errors.
`docker logs` on the other hand is aware of this functionality.

### DIFF
--- a/e2e/src/docker_logs.rs
+++ b/e2e/src/docker_logs.rs
@@ -1,52 +1,35 @@
 use anyhow::{anyhow, Context, Result};
 use std::process::Command;
 
-fn find_stablex_container_in_output(output: &str) -> Option<&str> {
-    output
-        .split('\n')
-        .skip(2)
-        .filter_map(|line| line.split(' ').next())
-        .find(|container_name| container_name.starts_with("dex-services_stablex_1"))
-}
-
-fn find_stablex_container() -> Result<String> {
-    let output = Command::new("docker-compose")
+fn find_container_id(container_name: &str) -> Result<String> {
+    let output = Command::new("docker")
         .arg("ps")
+        .arg("--quiet")
+        .arg("--filter")
+        .arg(format!("name={}", container_name))
         .output()
-        .context("failed to execute `docker-compose ps`")?;
+        .context("failed to execute `docker ps`")?;
     if !output.status.success() {
         return Err(anyhow!("status code is not success"));
     }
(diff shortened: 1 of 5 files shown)


e2e/src/docker_logs.rs
```diff
@@ -0,0 +1,52 @@
...
+fn find_stablex_container_in_output(output: &str) -> Option<&str> {
+    output
+        .split('\n')
+        .skip(2)
+        .filter_map(|line| line.split(' ').next())
+        .find(|container_name| container_name.starts_with("dex-services_stablex_1"))
+}
+
+fn find_stablex_container() -> Result<String> {
+    let output = Command::new("docker-compose")
```
nlordell: I would rather use something like `docker ps --quiet --filter '^dex-services_stablex_1'` to just get the ID directly, then we don't rely on the output being formatted in a specific way

e2e/tests/stablex_test.rs
```diff
@@ -202,12 +202,5 @@ fn test_rinkeby() {
...
 
-    // Make sure there was no error
-    let output = Command::new("docker-compose")
-        .arg("logs")
-        .output()
-        .expect("failed to execute process");
-    // Errors go to stderr while other messages go to stdout.
-    let logs = String::from_utf8(output.stderr).expect("failed to read logs");
-    assert!(logs.is_empty());
+    docker_logs::assert_no_errors_logged();
```
nlordell: maybe we can make this more generic by passing in the Docker compose service name (here `"stablex"`). The reason I say this is that the name `assert_no_errors_logged` doesn't capture that this is JUST for the `stablex` service.

nlordell (changes requested):

fleupold:
I'd also be fine with limiting output to only log error messages. The downside with that is that it makes debugging why the tests failed hard as we don't see any progress logs (such as INFO)

nlordell:
@fleupold We already introduced splitting logs into stdout and stderr in #628, I think this PR just uses the fact that only errors are on stderr to check for test success.

e2e/src/docker_logs.rs
e00E: Good suggestion. Didn't know about `--quiet`.

nlordell (approved):
:ship:

PR #633, opened 2020-03-16
Do not return prices for tokens that weren't asked for

e00E:
In the future if ThreadedPriceSource is used differently it could lead
to a bug if we return prices for tokens that were not part of the input.

### DIFF
--- a/driver/src/price_estimation/threaded_price_source.rs
+++ b/driver/src/price_estimation/threaded_price_source.rs
@@ -86,15 +86,15 @@ impl ThreadedPriceSource {
 }
 
 impl PriceSource for ThreadedPriceSource {
-    fn get_prices(&self, _tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
-        // NOTE: Return all the prices as they will get filtered by the caller
-        //   anyway.
-
-        Ok(self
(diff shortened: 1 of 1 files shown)


driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -86,15 +86,15 @@ impl ThreadedPriceSource {
...
+    fn get_prices(&self, tokens: &[Token]) -> Result<HashMap<TokenId, u128>> {
+        let price_map = self
             .price_map
             .lock()
-            .expect("mutex should never be poisoned")
-            .clone())
+            .expect("mutex should never be poisoned");
+        Ok(tokens
+            .iter()
+            .filter_map(|token| Some((token.id, *price_map.get(&token.id)?)))
```
nlordell: Love the use of `?`! Looks :ok_hand:

nlordell (commented):
Sounds good to me! That way the threaded price source respects the `PriceSource` interface better which is nice.

nlordell (approved):

fleupold (commented):
Awesome. Thanks for changing this!

🏆

fleupold (approved):

PR #634, opened 2020-03-16
Rename decorator types to be more readable

e00E:


driver/src/logging.rs
```diff
@@ -29,21 +29,23 @@ pub fn init(filter: impl AsRef<str>) -> (Logger, GlobalLoggerGuard) {
 
 /// Uses one decorator for `Error` and `Critical` log messages and the other for
 /// the rest.
-pub struct CustomFormatter<T0, T1> {
-    err_decorator: T0,
-    rest_decorator: T1,
+pub struct CustomFormatter<ErrDecorator, RestDecorator> {
```
nlordell: `LogDecorator`? I feel like `Rest` is a loaded word.

e00E: I was just reusing the variable name as CamelCase. We can go to ErrorLogsDecorator and OtherLogsDecorator.

nlordell: Hm, I don’t have a good name for this. Just leave it as is I think. It’s descriptive enough.

nlordell: Hm, I don’t have a good name for this. Just leave it as is I think. It’s descriptive enough.

PR #635, opened 2020-03-17
Add Sync bounds to allow usage of driver from multiple threads

e00E:
(and turn one mutable reference into immutable)

The Sync bounds are needed for a future PR that will solve each batch in
its own thread.

An alternative approach to adding Sync to all the trait bounds is to use generics instead of dyn traits. This would make each part more composable as it would still be individually usable even if it is not Sync which could help in tests. On the other hand it would mean more generics on every struct. I stuck with dyn traits because that is what already had but I don't mind either approach.


fleupold (approved):
Looks good!

PR #636, opened 2020-03-17
Use mockall with trait instead of functions

e00E:
This is inline with how we usually test things that we want to be
mockable.

The diff looks better if you use this option btw. Works like `git diff -w`.
image

### DIFF
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -162,7 +162,7 @@ mod solver_input {
 }
 
 #[cfg_attr(test, mockall::automock)]
-trait IoMethods {
+trait Io {
     fn write_input(&self, input_file: &str, input: &str) -> std::io::Result<()>;
     fn run_solver(
         &self,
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Fix of the week! Thanks!!!

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -160,11 +160,20 @@ mod solver_input {
     }
 }
 
+#[cfg_attr(test, mockall::automock)]
+trait IoMethods {
```
fleupold: super naming nit: not a huge fan of calling a trait XYMethods. Maybe this can be IoTrait and the concrete implementation FileIo/DefaultIo (theoretically we could add an IPC IO method later on).

fleupold (approved):
Looks great, thanks for the cleanup

✨

driver/src/price_finding/optimization_price_finder.rs
nlordell: Personally I slightly prefer `IoMethods` over `IoTrait`, since the latter seams a little redundant and `trait FooTrait` feels weird. I would rather just call it `Io` or `System` (and then `MockSystem` even sounds good :wink:).

PR #637, opened 2020-03-17
[Easy] Wait longer in e2e test worst case

e00E:
If the e2e runs within within the first 30 seconds of a batch then it could have failed because of `target_start_solve_time`.


fleupold (approved):
Looks inline with how the test is currently structured. Overall it would be nice if we can speed up this test or parallelise it with rinkeby (since batches are aligned they should take roughly the same time)

e00E:
Good point. Since we cannot speed up rinkeby anyway this would remove move of the disadvantage of having to sleep in ganache. I'll look into it.

PR #638, opened 2020-03-17
Http Metrics Revised

nlordell:
Revised HTTP metrics where histograms are used. This supersedes #598.

This PR introduces a new `HttpLabel` that must be specified per request and is used to determine which label to use when observing the HTTP request size and latency.

Previously, it was implemented with a different set of labels per HTTP susbystem (i.e. one for web3, one for gas station, etc.) but that added a lot of code complexity without much benefit, so a simpler approach was taken.

### Test Plan

CI, check that new metrics are produced on http://localhost:9586


driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    };
+}
+
+subsystem! {
+    /// A label for `web3` HTTP requests. This subsystem is used by the HTTP
+    /// transport implementation.
+    pub enum Web3Label as "web3" {
+        Call => "call",
+        EstimateGas => "estimate_gas",
```
fleupold: Should sendTx be also a specific type?

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn label(&self) -> &'static str;
+}
+
+/// A trait abstracting a labeled subsystem, that is a specific use of an HTTP
+/// client where the HTTP requests should not be differenciated.
+pub trait UnlabeledSubsystem: Default + LabelValues {
+    fn name() -> &'static str;
+}
+
+macro_rules! subsystem {
```
fleupold: Is this the fancy metaprogramming equivalent of https://github.com/gnosis/dex-services/blob/c24a3f31ce1ad3b255c24ef1817692cfd3fe8406/driver/src/metrics/stablex_metrics.rs#L282-L283 

What are the advantages?

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+impl HttpMetrics {
+    /// Create a new HTTP metrics registry.
+    pub fn new(registry: &Arc<Registry>) -> Result<Self> {
+        let latency = HttpMetrics::initialize_histogram(
+            registry,
+            "dfusion_service_http_latency",
+            "Latency in seconds for HTTP request",
+            vec![0.05, 0.1, 0.25, 0.5, 1.0, 2.0, 5.0, 10.0],
```
fleupold: Why not [DEFAULT_BUCKETS](https://docs.rs/prometheus/0.3.2/prometheus/constant.DEFAULT_BUCKETS.html)?

driver/src/price_estimation/kraken/api.rs
```diff
@@ -1,4 +1,5 @@
-use crate::http::{HttpClient, HttpFactory};
+use crate::http::{HttpClient, HttpFactory, UnlabeledHttpClient};
```
fleupold: Why are we using an UnlabeledClient here?

driver/src/http.rs
```diff
@@ -75,22 +94,165 @@ impl HttpClient {
     }
 
     /// Standard HTTP GET request that parses the result as JSON.
-    pub fn get_json<U, T>(&self, url: U) -> Result<T>
+    fn get_json<U, T>(&self, url: U) -> Result<(T, Duration, usize)>
```
fleupold: I'm not sure I'm a fan of this abstraction as it is leaking pretty obviously that the size and duration are for measurement purposes. What the advantage of wrapping this into two components over adding metric hooks directly into the inner client?

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,197 @@
...
+/// A common trait shared between labeled and unlabeled subsystems for getting
+/// the label values to use for metrics.
+pub trait LabelValues {
+    fn label_values(&self) -> [&'static str; 2];
+    fn initialize_histogram_labels(histograms: &HistogramVec);
+}
+
+/// A trait abstracting a labeled subsystem, that is a specific use of an HTTP
+/// client where the HTTP requests should be differenciated.
+pub trait LabeledSubsystem: LabelValues + Sized {
```
fleupold: It took me a while to understand the difference between a Labeled and Unlabelled subsystem. Why do we want to differentiate the two? Couldn't we just log the endpoint even if there is just one query? What's the advantage of doing the differentiation?

driver/src/http.rs
```diff
@@ -75,22 +94,165 @@ impl HttpClient {
...
+    metrics: Arc<HttpMetrics>,
+    _subsystem: PhantomData<Subsystem>,
+}
+
+/// Trait for extending HTTP client with labeled version of HTTP methods with
+/// metrics.
+///
+/// This trait allows `HttpClient` to require a label value when a subsystem is
+/// labeled, and to not require a label value when it is unlabeled.
+pub trait LabeledHttpClient<Subsystem> {
```
fleupold: Why do the Labeled and Unlabeled client have different trait signatures? I also don't see where they are being used (I assume it's somehow implicitly derived from whether a metric is labeled or unlabelled)?

fleupold (commented):
This PR is a strong demonstration of your rust skills (PhantomData, meta-programming such as macros and advanced generics).

See my questions inline, as I'm having a very hard time understanding what's going on and why we are doing things they way they are done (the PR description did not really help me).

driver/src/metrics/http_metrics.rs
nlordell: Don't think so, it just adds the tx to the mem-pool and is usually pretty fast and returns just a tx hash. If we were also tracking post body sizes, then yes we could definitely add it.

driver/src/metrics/http_metrics.rs
nlordell: Missed that! Thanks!

driver/src/metrics/http_metrics.rs
nlordell: The main advantage is that we generate code to enumerate the various labels: `all_labels()`. Adding a new label to the enum will automatically enumerate it.

driver/src/http.rs
nlordell: Good point! that was left-over from when everything had labeled and unlabeled equivalents. I'll fix it.

driver/src/metrics/http_metrics.rs
nlordell: Its minor, for unlabeled subsystems you can perform HTTP requests without a label (so, `http_client.get_json()` for example), while a labeled subsystem requires a label value (`http_client.get_json(Subsystem::SomeLabel)`).

driver/src/http.rs
nlordell: They are needed so that we can have methods with the same name but different signature:
`http_client.get_json()` and `http_client.get_json(Subsystem::Label)`.

driver/src/price_estimation/kraken/api.rs
nlordell: Because there isn't much value in differentiating the different Kraken HTTP requests.

nlordell:
> (the PR description did not really help me).

Sorry, just finishing up #595 so we have less flaky e2e tests and then I'll write a better description.

driver/src/http.rs
```diff
@@ -75,22 +92,36 @@ impl HttpClient {
...
         T: DeserializeOwned,
     {
-        Ok(self.inner.get(url)?.json()?)
+        let start = Instant::now();
+
+        let json = self.inner.get(url)?.text()?;
+        let size = json.len();
+        let result = serde_json::from_str(&json)?;
+
+        self.metrics.request(label, start.elapsed(), size);
```
fleupold: ubernit: here you are technically also timing the deserialization (negligible) but we could also just move it above the `let result ...` line.

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,120 @@
...
+            .observe(latency.as_secs_f64());
+        self.size
+            .with_label_values(&label.values())
+            .observe(size as _);
+    }
+}
+
+impl Default for HttpMetrics {
+    fn default() -> Self {
+        HttpMetrics::new(&Default::default()).unwrap()
```
fleupold: til Default::default is cool

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+macro_rules! labels {
```
fleupold: This is nice! Maybe we can have the StableXMetrics also use this macro in a follow up PR.

fleupold (approved):
Thanks for simplifying the approach a lot, so that an average Joe like me can also understand what's going on.

I hope I didn't take away all the fun that the other solution seemed to have.

Let's measure some HTTP metrics! ⌚️Can you also create the grafana dashboards once this is merged and deployed?

nlordell:
> Thanks for simplifying the approach a lot, so that an average Joe like me can also understand what's going on.

No worries, I think I went too far :stuck_out_tongue_closed_eyes: 

> Can you also create the grafana dashboards once this is merged and deployed?

Would love to!

driver/src/metrics/http_metrics.rs
nlordell: Less `use` statements FTW!

driver/src/http.rs
nlordell: Good point! Didn't think of that.

driver/src/metrics/http_metrics.rs
```diff
@@ -0,0 +1,120 @@
...
+            vec![
+                100.0,
+                1_000.0,
+                10_000.0,
+                100_000.0,
+                1_000_000.0,
+                10_000_000.0,
+                100_000_000.0,
+                1_000_000_000.0,
+            ],
```
e00E: We can use https://docs.rs/prometheus/0.8.0/prometheus/fn.exponential_buckets.html for this. But I don't mind writing it out manually either. Your choice.

nlordell: Thanks! I missed so many of these `prometheus` utility functions!

e00E (approved):

PR #641, opened 2020-03-18
Don't advance EVM time when waiting for solution

nlordell:
We are advancing the time (with `evm_increaseTime`) before waiting for a solution to be submitted. This is racy and causes situations where the orderbook and batch were queried before the `evm_increaseTime` and the solution was submitted for the next batch. This is causing intermittent CI failures with the Ganache test.

Note that this also requires increasing the wait time.

### Test Plan

CI

### DIFF
--- a/e2e/tests/stablex_test.rs
+++ b/e2e/tests/stablex_test.rs
@@ -73,7 +73,7 @@ fn test_with_ganache() {
     let batch_id = instance
         .get_current_batch_id()
         .wait_and_expect("Failed to query batch ID");
-    let wait_time = Duration::from_secs(340);
+    let wait_time = Duration::from_secs(640);
     println!(
         "Waiting {}s for the solver to submit a solution for batch {}",
         wait_time.as_secs(),
--- a/e2e/tests/stablex_test.rs
+++ b/e2e/tests/stablex_test.rs
@@ -5,7 +5,7 @@ use ethcontract::{Account, PrivateKey, U256};
 
 use futures::future::join_all;
 
-use e2e::common::{update_evm_time, wait_for_condition, FutureBuilderExt, FutureWaitExt};
+use e2e::common::{wait_for_condition, FutureBuilderExt, FutureWaitExt};
 use e2e::docker_logs;
 use e2e::stablex::{close_auction, setup_stablex};
 use e2e::{BatchExchange, IERC20};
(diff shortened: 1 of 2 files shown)


e00E:
I was investigating this too. I also tried this change with a lot of debug prints but it doesn't work. Something weird is going on with ganache. 
Without close_auction, the driver does not seem to ever see the new orders so it doesn't attempt to solve them. I think we might have to evm_mine at the point where the batch of the new orders becomes solvable.
If you want I can continue with this.

nlordell:
Oh!!! that makes sense, we need to have periodic `evm_mines` so the timestamp moves forward.

e00E (approved):
This should work.

e2e/tests/stablex_test.rs
```diff
@@ -81,6 +82,10 @@ fn test_with_ganache() {
     // wait for solver to submit solution
     wait_for_condition(
         || {
+            // NOTE: Make sure that we periodically update the EVM time, as no
+            //   new blocks are being mined to update the EVM timestamp.
+            update_evm_time(&web3);
```
fleupold: I think we should colocate this with the thread::sleep statement in `wait_for_condition`. Note that we also have `wait_for` in the common.rs which we might want to remove now since it doesn't seem to work with the new system time approach.

nlordell: The `wait_for` is still helpful for speeding up being able to withdraw the balance.

fleupold: Makes perfect sense!

fleupold (approved):
Looks great!

e00E:
Fails because should_mine is false because the real chain id is 1337 != 5777

nlordell:
> Fails because should_mine is false because the real chain id is 1337 != 5777

I forgot that the Ganache version we are using in CI doesn't fully support chain ID, updated to use the network ID instead (which we specifically set to be 5777).

e2e/tests/stablex_test.rs
```diff
@@ -70,16 +70,18 @@ fn test_with_ganache() {
         .from(Account::Local(accounts[1], None))
         .wait_and_expect("Cannot place first order");
 
-    let batch_id = instance.get_current_batch_id().call().wait().unwrap();
-    close_auction(&web3, &instance);
-    let wait_time = Duration::from_secs(310);
+    let batch_id = instance
+        .get_current_batch_id()
+        .wait_and_expect("Failed to query batch ID");
+    let wait_time = Duration::from_secs(340);
```
bh2smith: Is the change from `310` to `340` especially related to this PR?

nlordell: It is related although could have arguably been its own PR. Specifically, this PR was for fixing intermittent Ganache e2e test failures and the incorrect duration was causing errors. The main issue was advancing EVM time, but the incorrect duration was a minor secondary issue.

PR #642, opened 2020-03-18
[Trivial] Remove unused orderbook API methods

nlordell:
Since the orderbook is no longer responsible for determining the batch ID, remove the call from the the orderbook reading trait.

### Test Plan

It should build.


e00E (approved):
In that case we can also remove it from `StableXContract`.

nlordell:
> In that case we can also remove it from `StableXContract`.

I want to work on a EVM scheduler right now so that we can speed up the Ganache tests and potentially use it for our fallback solver on mainnet in case we see reorgs affecting our ability to submit orders.

fleupold (approved):
🍰

PR #643, opened 2020-03-18
Remove latest_solve_attempt config option

e00E:
If there is still time left in solver_time_limit then we should run the
solver regardless of how little time that is. Worst case is it doesn't
find a solution quickly enough which is the same result as if we hadn't
attempted to run it at all.


nlordell (approved):
LGTM!

fleupold (approved):
This makes a lot of sense and simplifies the configuration which is great.

Unrelated to this, but one thought I had reviewing is that we compute the "time to solve" before we fetch orders (which takes currently around 6s but could become worse/better in the future). Maybe we want to give it a Instant and determine how long that time is after we have fetched orders?

PR #644, opened 2020-03-18
Evm scheduler

nlordell:
This PR re-introduces the old EVM scheduler that relies on block chain data for executing the scheduler. This has the bonus of speeding up Ganache e2e tests and can also be used for the fallback solver in case there are concerns about reorgs or late blocks that affect the orderbook.

### Test Plan

CI: unit and e2e test. Note that the Rinkeby e2e test still runs with the system-time based scheduler, so both get exercised in the e2e tests.


nlordell:
# Note

This PR is currently set to merge into #645, this is to work around the fact that the GitHub diff UI doesn't seem to properly display when you `git mv` a file in a commit and then create a new file with the same name as the old one in another commit.

This was sub-optimal in this case as the diff UI would not correctly show the changes to the system time scheduler.

driver/src/driver/scheduler/evm.rs
```diff
@@ -0,0 +1,279 @@
...
+        if batch_id != verify_batch_id {
+            info!(
+                "batch ID changed during run loop ({} -> {}); retrying",
+                batch_id, verify_batch_id,
+            );
+            return Ok(());
+        }
+
+        let current_batch_time = BATCH_DURATION - time_remaining;
+        if current_batch_time > self.config.target_start_solve_time {
```
fleupold: Shouldn't this be latest_solve_attempt_time? We have seen in the past that we might not mine a block within the first 30s of a batch.

driver/src/driver/scheduler/evm.rs
```diff
@@ -0,0 +1,279 @@
...
+            .last_batch
+            .map(|last_batch| batch_id <= last_batch)
+            .unwrap_or_default()
+        {
+            debug!("skipping already processed batch {}", batch_id);
+            return Ok(());
+        }
+
+        let time_remaining = self.exchange.get_current_auction_remaining_time()?;
+        // NOTE: We need to take into account the asynchronous nature of web3
```
fleupold: Fine for now. We could deploy a wrapper contract that gives us this information? Or we could incorporate the batching API into ethcontracts which would allow to run multiple "calls" in the same context.

driver/src/driver/scheduler/evm.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    #[test]
+    fn scheduler_retries_when_batch_changes_during_run() {
+        let mut seq = Sequence::new();
+        let mut exchange = MockStableXContract::new();
+        exchange
+            .expect_get_current_auction_index()
+            .times(1)
+            .in_sequence(&mut seq)
```
fleupold: This is 🆒

fleupold (approved):
PR is quite bit, but very awesome!

🐎

driver/src/driver/scheduler/evm.rs
nlordell: If its OK with you, I'll do that as a separate PR.

driver/src/driver/scheduler/evm.rs
nlordell: :man_facepalming:

driver/src/driver/scheduler/evm.rs
nlordell: @fleupold Would this be a view method that I could add to `dex-contracts`?

driver/src/driver/scheduler/evm.rs
nlordell: Thought you might also like: https://github.com/gnosis/dex-services/blob/ad1fc0ed780b06a6b63bae81951b348521e7e1ea/driver/src/driver/scheduler.rs#L21

You mentioned that you did not like infinite loops too much, this `!` type can be used to indicate to the compiler that this method never returns. It even gives you cool things like compiler warnings in case you write statements after calling it.

e00E (approved):

driver/src/driver/scheduler/evm.rs
fleupold: Yes, I think it could be added to https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchangeViewer.sol

e2e/src/common.rs
```diff
@@ -98,42 +98,15 @@ pub fn wait_for(web3: &Web3<Http>, seconds: u32) {
...
             return Ok(());
         }
-
         std::thread::sleep(Duration::from_secs(1));
-        if should_mine {
-            web3.transport()
-                .execute("evm_mine", vec![])
-                .wait()
-                .expect("Cannot mine to update current time");
-        }
```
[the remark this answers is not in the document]
nlordell: Yes, so that code was a work-around for using the system scheduler where we would have to actually wait for time to go by. Once we brought back the EVM scheduler, we could just move time forward as we had done before.

PR #645, opened 2020-03-18
Evm scheduler - moving system scheduler to a submodule

nlordell:
This PR is a dummy PR to work around the fact that the GitHub diff UI seems to not always display diffs against `git mv`-ed files in the case that a file with the same name that gets created in the same PR.

This PR is not intented to be merged into master, but used for diffing purposes.


nlordell:
Closing as we are now merging #644

PR #646, opened 2020-03-18
[Trivial] Remove unused env keys

nlordell:
Just cleanup the `common*.env` files and remove some old env variables that are no longer used.


fleupold (approved):
🍭👶

PR #647, opened 2020-03-18
Discount orderbook retrieval from the solver time.

nlordell:
This PR implements discounting the orderbook retrieval time from the total time to solve for the solver. As the orderbook retrieval time grows, this can cause us to miss our deadline.

### Test Plan

New unit test.

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -3,10 +3,10 @@ use crate::models::{account_state::AccountState, order::Order, Solution};
 use crate::orderbook::StableXOrderBookReading;
 use crate::price_finding::PriceFinding;
 use crate::solution_submission::{SolutionSubmissionError, StableXSolutionSubmitting};
-use anyhow::{Error, Result};
+use anyhow::{anyhow, Error, Result};
 use ethcontract::U256;
 use log::info;
-use std::time::Duration;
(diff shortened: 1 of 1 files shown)


e00E (approved):
I am fine with this too but it would be cleaner to pass in a deadline from the beginning instead of converting between duration and instant several times.

nlordell:
> it would be cleaner to pass in a deadline from the beginning instead of converting between duration and instant several times.

Good point.

driver/src/driver/stablex_driver.rs
```diff
@@ -143,12 +143,23 @@ impl<'a> StableXDriverImpl<'a> {
...
+        let deadline = Instant::now() + time_limit;
+
         self.metrics.auction_processing_started(&Ok(batch_to_solve));
         let (account_state, orders) = match self.get_orderbook(batch_to_solve) {
             Ok(ok) => ok,
             Err(err) => return DriverResult::Retry(err),
         };
-        match self.solve(batch_to_solve, time_limit, account_state, orders) {
+
+        let actual_time_limit = match deadline.checked_duration_since(Instant::now()) {
```
fleupold: nit: maybe rename to price_finding_time_limit / remaining_time_limit?

fleupold (approved):
🍠

PR #652, opened 2020-03-19
Solve each batch in its own thread

e00E:
We want to solve each batch in its own thread so that one batch cannot make another start late. In the future this allows us to potentially start a solver optimistically even before its auction is up.

Test Plan: New unit tests, CI, manual test with the `ignored` test.

### DIFF
--- a/driver/src/driver/scheduler.rs
+++ b/driver/src/driver/scheduler.rs
@@ -21,7 +21,7 @@ pub trait Scheduler {
     fn start(&mut self) -> !;
 }
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug)]
 pub struct AuctionTimingConfiguration {
     /// The offset from the start of a batch at which point we should start
     /// solving.
--- a/driver/src/driver/scheduler/system.rs
+++ b/driver/src/driver/scheduler/system.rs
@@ -84,7 +84,12 @@ impl<'a> SystemScheduler<'a> {
                 batch_id,
                 &driver_result,
             ) {
-                SolveAgain::Yes { time_limit } => solver_time_limit = time_limit,
+                SolveAgain::Yes { time_limit } => {
+                    if time_limit > RETRY_SLEEP_DURATION {
+                        thread::sleep(RETRY_SLEEP_DURATION);
+                    }
+                    solver_time_limit = time_limit - RETRY_SLEEP_DURATION;
+                }
                 SolveAgain::No => break,
             }
         });
--- a/driver/src/driver/scheduler/system.rs
+++ b/driver/src/driver/scheduler/system.rs
@@ -90,42 +90,38 @@ impl<'a> SystemScheduler<'a> {
             }
         });
     }
-}
 
-/// Return current batch id and what to do with it.
-fn determine_action(
-    auction_timing_configuration: AuctionTimingConfiguration,
-    last_solved_batch: Option<BatchId>,
-    now: SystemTime,
-) -> Result<Action> {
-    let solving_batch = BatchId::currently_being_solved(now)
-        .context("failed to get batch id currently being solved")?;
-    let intended_solve_start_time =
-        solving_batch.solve_start_time() + auction_timing_configuration.target_start_solve_time;
-    // unwrap here because this cannot fail because the `solving_batch`'s
-    // start time is always before `now`.
-    let elapsed_time = now
-        .duration_since(solving_batch.solve_start_time())
-        .unwrap();
-
-    let action = if last_solved_batch == Some(solving_batch)
-        || elapsed_time >= auction_timing_configuration.solver_time_limit
-    {
-        let next = solving_batch.next();
-        let duration = (next.solve_start_time()
-            + auction_timing_configuration.target_start_solve_time)
-            .duration_since(now)
+    /// Return current batch id and what to do with it.
+    fn determine_action(&self, now: SystemTime) -> Result<Action> {
+        let solving_batch = BatchId::currently_being_solved(now)
+            .context("failed to get batch id currently being solved")?;
+        let intended_solve_start_time = solving_batch.solve_start_time()
+            + self.auction_timing_configuration.target_start_solve_time;
+        // unwrap here because this cannot fail because the `solving_batch`'s
+        // start time is always before `now`.
+        let elapsed_time = now
+            .duration_since(solving_batch.solve_start_time())
             .unwrap();
-        Action::Sleep(duration)
-    } else if now < intended_solve_start_time {
-        let duration = intended_solve_start_time.duration_since(now).unwrap();
-        Action::Sleep(duration)
-    } else {
-        let time_limit = auction_timing_configuration.solver_time_limit - elapsed_time;
-        Action::Solve(solving_batch, time_limit)
-    };
 
-    Ok(action)
+        let action = if self.last_solved_batch == Some(solving_batch)
+            || elapsed_time >= self.auction_timing_configuration.solver_time_limit
+        {
+            let next = solving_batch.next();
+            let duration = (next.solve_start_time()
+                + self.auction_timing_configuration.target_start_solve_time)
+                .duration_since(now)
+                .unwrap();
+            Action::Sleep(duration)
+        } else if now < intended_solve_start_time {
+            let duration = intended_solve_start_time.duration_since(now).unwrap();
+            Action::Sleep(duration)
+        } else {
+            let time_limit = self.auction_timing_configuration.solver_time_limit - elapsed_time;
+            Action::Solve(solving_batch, time_limit)
+        };
+
+        Ok(action)
+    }
 }
 
 fn should_attempt_to_solve_again(
@@ -160,11 +156,7 @@ impl<'a> Scheduler for SystemScheduler<'a> {
     fn start(&mut self) -> ! {
         crossbeam_utils::thread::scope(|scope| -> ! {
             loop {
-                match determine_action(
-                    self.auction_timing_configuration,
-                    self.last_solved_batch,
-                    SystemTime::now(),
-                ) {
+                match self.determine_action(SystemTime::now()) {
                     Ok(Action::Sleep(duration)) => {
                         info!("Sleeping {}s.", duration.as_secs());
                         thread::sleep(duration);
@@ -188,6 +180,7 @@ impl<'a> Scheduler for SystemScheduler<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::driver::stablex_driver::MockStableXDriver;
     use anyhow::anyhow;
 
     #[test]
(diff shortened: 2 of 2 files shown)


fleupold:
Is this PR ready for review? If so, could you please add a description and test-plan to it?

e00E:
Sorry, added.

driver/src/driver/scheduler/system.rs
```diff
@@ -62,92 +65,131 @@ impl<'a> SystemScheduler<'a> {
...
-        };
+fn should_attempt_to_solve_again(
+    auction_timing_configuration: AuctionTimingConfiguration,
+    now: SystemTime,
+    batch_id: BatchId,
+    driver_result: &DriverResult,
+) -> bool {
+    match driver_result {
+        DriverResult::Ok => false,
+        DriverResult::Retry(_) => match determine_action(auction_timing_configuration, None, now) {
```
fleupold: All we really want to know is if we still have time to solve batch_id again. While reusing `determine_action` happens to work, I found it hard to verify that (it has a lot more logic taking last_batch_id and the one that is currently being solved into account) and it might be simpler to just rewrite the logic in something like `can_still_solve_batch`. Maybe `determine_action` can call reuse that method.

fleupold (commented):
Looks pretty good already! Let's discuss what is the best way to organize the retry mechanism.

driver/src/driver/scheduler/system.rs
```diff
@@ -62,92 +65,131 @@ impl<'a> SystemScheduler<'a> {
...
+        &self,
+        batch_id: BatchId,
+        solver_time_limit: Duration,
+        scope: &Scope<'b>,
+    ) where
+        'a: 'b,
+    {
+        let driver = self.driver;
+        let auction_timing_configuration = self.auction_timing_configuration;
+        scope.spawn(move |_| loop {
```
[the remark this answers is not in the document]
e00E: It's not a do while loop, just kind of. For example you cannot `break` and `continue`.

driver/src/driver/scheduler/system.rs
```diff
@@ -1,11 +1,14 @@
 use super::{AuctionTimingConfiguration, Scheduler, BATCH_DURATION};
 use crate::driver::stablex_driver::{DriverResult, StableXDriver};
 use anyhow::{Context, Result};
+use crossbeam_utils::thread::Scope;
 use log::error;
 use log::info;
 use std::thread;
 use std::time::{Duration, SystemTime, SystemTimeError};
 
+const RETRY_SLEEP_DURATION: Duration = Duration::from_secs(10);
```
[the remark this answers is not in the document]
e00E: We can set it to 1 second. I do want some sleep in case the drivers returns keeps returning Error::Retry immediately so we don't busy loop.

driver/src/driver/scheduler/system.rs
```diff
@@ -62,92 +65,131 @@ impl<'a> SystemScheduler<'a> {
...
+        scope: &Scope<'b>,
+    ) where
+        'a: 'b,
+    {
+        let driver = self.driver;
+        let auction_timing_configuration = self.auction_timing_configuration;
+        scope.spawn(move |_| loop {
+            let driver_result = driver.run(batch_id.0.into(), solver_time_limit);
+            log_driver_result(batch_id, &driver_result);
+            if should_attempt_to_solve_again(
```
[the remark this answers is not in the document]
e00E: I find this way simpler. Each thread does whatever is needed for the batch id it was given, completely independently and stays alive until this is done.

driver/src/driver/scheduler/system.rs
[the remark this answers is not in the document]
e00E: If `should_attempt_to_solve_again` is correct then the loop is bounded. I don't find a counter helpful.

driver/src/driver/scheduler/system.rs
e00E: Agree. Will implement.

driver/src/driver/scheduler/system.rs
```diff
@@ -133,11 +128,69 @@ impl<'a> SystemScheduler<'a> {
     }
 }
 
-impl Scheduler for SystemScheduler<'_> {
-    fn start(&mut self) -> ! {
-        loop {
-            thread::sleep(self.run_once(SystemTime::now()));
+#[derive(Debug, Eq, PartialEq)]
+enum SolveAgain {
```
fleupold: ubernit: Semantically equivalent to an Option<Duration> but since it's private it doesn't matter much and if you think this type adds value feel free to leave like this.

driver/src/driver/scheduler/system.rs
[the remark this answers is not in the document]
fleupold: I just feel like we are adding a lot of potentially unbounded loops into the codebase, which makes it harder to reason about how and when things will terminate (we have quite some nested branches here and in `should_attempt_to_solve_again` that need to be understood).

On that note, would it make sense to add some metrics around how many threads we have and how long they run on average?

driver/src/driver/scheduler/system.rs
```diff
@@ -62,42 +65,34 @@ impl<'a> SystemScheduler<'a> {
...
-                        RETRY_TIMEOUT
-                    }
-                    DriverResult::Skip(err) => {
-                        error!("StableXDriver unretryable error: {}", err);
-                        self.last_solved_batch.replace(batch_id);
-                        Duration::from_secs(0)
+    fn start_solving_in_thread<'b>(
+        &self,
+        batch_id: BatchId,
+        mut solver_time_limit: Duration,
```
fleupold: nit: maybe it would make things slightly less coupled if we didn't pass in the solver time but made it the task of the thread to decide how long it has to solve (if it's None/0 it terminates). This way we would take a small chunk of the logic that is currently shared between threads (and mutated below) and isolate it in one place.

fleupold (approved):
Logic looks sounds. I'm still somewhat worried about the cognitive complexity added in its current form (another infinite loop, high number of branches).

Maybe the `start_solving_in_thread` method could look more like:

```rust
scope.spawn(move |_| for _ in 0..MAX_TRIES  {
  let retry_in = run_once(batch_id, auction_timing_configuration, now); // returns Option<Duration> and can be unittested
  match retry_in {
    Some(time) => thread::sleep(time),
    None => break
  }
}
```

It would have the advantage that the loop is for sure bounded and we can test all the logic inside the runloop.

driver/src/driver/scheduler/system.rs
e00E: This is true. The difference is that with an option I would write a comment saying what `None` and `Some` means but with the enum it is clear from the types. On the other hand I lose convenience methods implemented on Option. In this case I prefer the enum.

driver/src/driver/scheduler/system.rs
e00E: Thanks to your comment I realized this could be simplified a lot by passing a deadline `Instant` instead of a timeout `Duration`. Now there is no `mut` and there is a clear condition on which the loop exits.

driver/src/driver/scheduler/system.rs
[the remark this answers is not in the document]
e00E: We still expect there to only be one thread unless there is a bug with timeouts or the solver. A number of running threads metric would help up us notice if such a bug occurs because we would observe more than 1 thread.

Thread runtime is not useful because it is already measured by the processing times (and it would have the same problems as them if there were multiple threads running at the same time).

fleupold (approved):
I agree that's much simpler. Thanks for working through the changes!

driver/src/driver/scheduler/system.rs
```diff
@@ -62,42 +65,27 @@ impl<'a> SystemScheduler<'a> {
...
+    fn start_solving_in_thread<'b>(
+        &self,
+        batch_id: BatchId,
+        solver_time_limit: Instant,
+        scope: &Scope<'b>,
+    ) where
+        'a: 'b,
+    {
+        let driver = self.driver;
+        scope.spawn(move |_| loop {
```
nlordell:
```suggestion
        scope.spawn(move |_| while let Some(time_limit) = solver_time_limit.checked_duration_since(Instant::now()) {
```

That way we have a clear exit condition in the loop.

driver/src/driver/scheduler/system.rs
```diff
@@ -133,11 +121,41 @@ impl<'a> SystemScheduler<'a> {
...
+        DriverResult::Skip(err) => error!(
+            "Batch {} failed with unretryable error: {}",
+            batch_id.0, err
+        ),
+    }
+}
+
+impl<'a> Scheduler for SystemScheduler<'a> {
+    fn start(&mut self) -> ! {
+        crossbeam_utils::thread::scope(|scope| -> ! {
```
nlordell: Does `crossbeam` just give us a thread pool?

driver/src/driver/scheduler/system.rs
```diff
@@ -279,61 +295,38 @@ mod tests {
         );
     }
 
+    // Allows observing real behavior by looking at the log output.
```
nlordell: nit: Could you make this comment a bit more descriptive, maybe include some expected output. The reason is when I read it I understood that log messages were being intercepted somehow and used to determine if the test succeeded or not.

driver/src/driver/scheduler.rs
```diff
@@ -21,7 +21,7 @@ pub trait Scheduler {
     fn start(&mut self) -> !;
 }
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug)]
```
nlordell: nit: I don't think this should be `Copy` and should just be `Clone`.

nlordell (approved):
LGTM! just a couple of nits.

driver/src/driver/scheduler/system.rs
```diff
@@ -62,42 +65,27 @@ impl<'a> SystemScheduler<'a> {
...
-                    DriverResult::Skip(err) => {
-                        error!("StableXDriver unretryable error: {}", err);
-                        self.last_solved_batch.replace(batch_id);
-                        Duration::from_secs(0)
-                    }
-                }
+    fn start_solving_in_thread<'b>(
+        &self,
+        batch_id: BatchId,
+        solver_time_limit: Instant,
```
nlordell: naming nit: `solver_deadline`.

driver/src/driver/scheduler.rs
e00E: I made it copy because
> Types whose values can be duplicated simply by copying bits.

Is that not correct?

driver/src/driver/scheduler/system.rs
e00E: It's not a threadpool. It works like `std::thread::spawn` except that it doesn't require everything to be `'static` because crossbeam guarantees that the threads won't outlive the lifetime of the captured types.

driver/src/driver/scheduler.rs
nlordell: Hmm, good point.

driver/src/driver/scheduler/system.rs
nlordell: Cool! Oh, I understand why you use it now.

driver/src/driver/scheduler.rs
nlordell: I guess this config is small enough that it doesn't really matter.

PR #653, opened 2020-03-19
Use `--help` output to generate readme configuration options section

e00E:
So that we no longer need to manually repeat documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,21 +144,72 @@ Afterwards, when you run your environment e.g. with `docker-compose up stablex` 
 
 ## Configuration
 
-The following environment variables can be used to configure the behavior of the services:
+The binary can be configured via command line options and environment variables: `cargo run -- --help`
 
-- *AUCTION_DATA_PAGE_SIZE*: the page size with which to read orders from the smart contract
-- *DFUSION_LOG*: Log-level (e.g. `info,driver=debug`)
-- *ETHEREUM_NODE_URL*: Full-Node to connect to. Make sure the node allows view queries without a gas limit in order to fetch the entire orderbook at once.
-- *NETWORK_ID*: Network ID (e.g. 1 for mainnet, 4 for rinkeby, 5777 for ganache)
-- *OPTIMIZATION_MODEL*: Which style of solver to use (NAIVE for naive, MIP for mixed integer programming and NLP for the  non-linear programming solver)
-- *ORDERBOOK_FILTER*: json encoded object of which tokens/filters to ignore. See below for example filter.
-- *PRIVATE_KEY*: The key with which to sign transactions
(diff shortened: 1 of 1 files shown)


fleupold (approved):
Makes a lot of sense. If we want to be super fancy we could even write a script that checks if the README contains the up-to-date help string and allows updating it if not.

nlordell:
> Makes a lot of sense. If we want to be super fancy we could even write a script that checks if the README contains the up-to-date help string and allows updating it if not.

Or use emacs org-mode :nerd_face:

README.md
```diff
@@ -144,21 +144,72 @@ Afterwards, when you run your environment e.g. with `docker-compose up stablex`
...
+            [default: 300]
+    -k, --private-key <private-key>
+            The private key used by the driver to sign transactions [env: PRIVATE_KEY]
+
+        --rpc-timeout <rpc-timeout>
+            The timeout in milliseconds of web3 JSON RPC calls, defaults to 10000ms [env: WEB3_RPC_TIMEOUT=]  [default:
+            10000]
+        --scheduler <scheduler>
+            The kind of scheduler to use [env: SCHEDULER=]  [default: system]
+
```
nlordell: Are these seemingly random extra new-lines on purpose?

nlordell (approved):

fleupold:
> Or use emacs org-mode 🤓

And here I thought you were team vim 😞

README.md
e00E: That's the output of --help. I don't know why it outputs them like this but I'd rather have them in than remember to fix this manually.

README.md
```diff
@@ -144,21 +144,72 @@ Afterwards, when you run your environment e.g. with `docker-compose up stablex`
...
+        --solver-time-limit <solver-time-limit>
+            The offset from the start of the batch to cap the solver's execution time [env: SOLVER_TIME_LIMIT=]
+            [default: 210]
+        --solver-type <solver-type>
+            Which style of solver to use. Can be one of: 'NAIVE' for the naive solver; 'MIP' for mixed integer
+            programming solver; 'NLP' for non-linear programming solver [env: SOLVER_TYPE=]  [default: naive-solver]
+        --target-start-solve-time <target-start-solve-time>
```
bh2smith: Why does everything repeat itself `--thing <thing>`?

nlordell: I think that is the default behaviour of the automatically generated help when using `StructOpt` for generating command-line and environment variable parsing code.

We can use `name = NAME` to manually specify the name of the variable (so that name that shows up between angle brackets, `<NAME>`). The default behaviour is to use the field name.

nlordell: Would be a nice and easy cleanup PR :)

bh2smith: > Would be a nice and easy cleanup PR :)

#670

PR #655, opened 2020-03-19
Sleep longer to make sure solver gets started

nlordell:
Since moving to the system time scheduler for Rinkeby tests, it is likely that the solver never gets started as we only wait until the target start time (30s) of the next batch, and don't wait for the solver to actually get started.

Here is an example run where the solver was never even ran: https://travis-ci.com/github/gnosis/dex-services/builds/154083477#L1974

### Test Plan

CI and verify that the solver gets run.


e00E (approved):

fleupold (approved):
⏳

PR #656, opened 2020-03-19
Don't error when time limit exceeded on orderbook retrieval

nlordell:
The PR merged earlier today #647 causes false negative errors when CI starts right at the end of the solver time limit.

### Test Plan

CI, updated unit test.

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -155,7 +155,7 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
             Some(time_limit) => time_limit,
             None => {
                 self.metrics.auction_skipped(batch_to_solve);
-                warn!("account retrieval exceeded time limit");
+                warn!("orderbook retrieval exceeded time limit");
                 return DriverResult::Ok;
             }
         };
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -3,9 +3,9 @@ use crate::models::{account_state::AccountState, order::Order, Solution};
 use crate::orderbook::StableXOrderBookReading;
 use crate::price_finding::PriceFinding;
 use crate::solution_submission::{SolutionSubmissionError, StableXSolutionSubmitting};
-use anyhow::{anyhow, Error, Result};
+use anyhow::{Error, Result};
 use ethcontract::U256;
-use log::info;
+use log::{info, warn};
 use std::time::{Duration, Instant};
 
 #[derive(Debug)]
@@ -155,7 +155,8 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
             Some(time_limit) => time_limit,
             None => {
                 self.metrics.auction_skipped(batch_to_solve);
-                return DriverResult::Skip(anyhow!("account retrieval exceeded time limit"));
+                warn!("account retrieval exceeded time limit");
+                return DriverResult::Ok;
             }
         };
 
@@ -496,7 +497,7 @@ mod tests {
     }
 
     #[test]
-    fn skips_solving_when_orderbook_retrieval_exceedes_time_limit() {
+    fn does_not_invoke_price_finder_when_orderbook_retrieval_exceedes_time_limit() {
         let mut reader = MockStableXOrderBookReading::default();
         let submitter = MockStableXSolutionSubmitting::default();
         let pf = MockPriceFinding::default();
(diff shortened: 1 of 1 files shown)


e00E (approved):

driver/src/driver/stablex_driver.rs
```diff
@@ -155,7 +155,8 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
             Some(time_limit) => time_limit,
             None => {
                 self.metrics.auction_skipped(batch_to_solve);
-                return DriverResult::Skip(anyhow!("account retrieval exceeded time limit"));
+                warn!("account retrieval exceeded time limit");
```
fleupold: Nit: this should be orderbook retrieval not account retrieval

driver/src/driver/stablex_driver.rs
```diff
@@ -155,7 +155,8 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
             Some(time_limit) => time_limit,
             None => {
                 self.metrics.auction_skipped(batch_to_solve);
```
fleupold: I might have missed this in an earlier PR but this will cause us to log a successful processing (which is not true as we didn't attempt to solve this batch). Maybe we should rename the method to something like `processed_but_not_submitted` to make this clearer.

driver/src/driver/stablex_driver.rs
```diff
@@ -155,7 +155,8 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
...
-                return DriverResult::Skip(anyhow!("account retrieval exceeded time limit"));
```
fleupold: Maybe `Skip` here can also be misleading. Maybe the DriverResult should be ErrorRetryable, ErrorNonRetryable instead to make clear that Skip here is not benign

fleupold (approved):
LGTM, but noticed two things unrelated to this PR, which I think we should address.

driver/src/driver/stablex_driver.rs
nlordell: I'll do this in a separate PR.

driver/src/driver/stablex_driver.rs
nlordell: I noticed this as well but wanted to tackle this in a separate PR as I think it requires a bit of though. Right now we have some inconsistencies with how we handle timeouts:
- If the scheduler runs at the end of a batch past the `solver_time_limit`, it will silently (as in no metrics recorded) wait until the next batch
- If the scheduler runs at the end of a batch near the `solver_time_limit` then a couple things can happen:
  1. The orderbook retrieval is in time, but the solver is started with a very low time limit, the solver will probably not find a solution and it will increment the `auction_skipped` metric
  2. The orderbook retrieval puts us over the time limit, so we skip the solving and it will increment a `auction_skipped` metric.

To m,e both cases when the driver attempts to solve at the end of the batch, whether it starts the solver with a very low time limit (for arguments sake, lets say 0.1s) or skips it is about the same. 

A problem with this approach is that we don't really distinguish the case where because of some unforeseen issue, the orderbook retrieval takes, for argument's sake, 3 minutes and causes is to not run the solver. This should be an error but is currently equivalent to the second case mentioned above.

Also, the way we submit metrics is getting a little hard to follow for me, with many branches. I wanted to take a little time to clean this up and maybe address your concerns in the process.

nlordell: Let me know if you are OK with me addressing this in a separate PR.

fleupold: Yes, let's do that in a separate PR. Maybe we should re-introduce a minimum time in which we don't start the solver? We are also running into somewhat benign failures on rinkeby today. 

When we start fetching the orderbook with 3s remaining we expectedly run over time and still log an error.

driver/src/driver/stablex_driver.rs
```diff
@@ -155,7 +155,8 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
...
+                warn!("orderbook retrieval exceeded time limit");
```
bh2smith: It seems that the "account retrieval exceeded time limit" has become "orderbook retrieval exceeded time limit". Was this intended? Are these the same thing?

nlordell: It was intended. We are calling "`orderbook` retrieval" the process of retrieving the accounts (users + balances) and orders. The orderbook retrieval is what is taking too long and causing this warning message.

PR #658, opened 2020-03-20
Simplify AccountState

e00E:
Removed unused fields.

Turned hash map of hash map into a single hash map. This is simpler and
more efficient.

Exposed undlerying hash map directly.
AccountState wraps the hash map with convenience methods but there is no
reason to not allow direct access to other code.

Test Plan: existing unit tests

### DIFF
--- a/driver/src/models/account_state.rs
+++ b/driver/src/models/account_state.rs
@@ -1,32 +1,13 @@
-use ethcontract::{Address, H256, U256};
-use serde::{Deserialize, Serialize};
+use ethcontract::Address;
 use std::collections::HashMap;
 
-#[derive(Serialize, Default, Deserialize, Clone, Debug, PartialEq)]
-#[serde(rename_all = "camelCase")]
-pub struct AccountState {
-    pub state_hash: H256,
-    pub state_index: U256,
-    balances: HashMap<Address, HashMap<u16, u128>>, // UserId => (TokenId => balance)
-    pub num_tokens: u16,
-}
+/// Maps a user and a token id to the balance the user has of this token.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct AccountState(pub HashMap<(Address, u16), u128>);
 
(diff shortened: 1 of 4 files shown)


fleupold (approved):
Looks good. Great cleanup!

driver/src/models/account_state.rs
```diff
@@ -1,32 +1,13 @@
...
-#[serde(rename_all = "camelCase")]
-pub struct AccountState {
-    pub state_hash: H256,
-    pub state_index: U256,
-    balances: HashMap<Address, HashMap<u16, u128>>, // UserId => (TokenId => balance)
-    pub num_tokens: u16,
-}
+// (UserId TokenId) => balance
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct AccountState(pub HashMap<(Address, u16), u128>);
```
nlordell: nit: Doesn't need to be `pub`.

driver/src/models/account_state.rs
```diff
@@ -1,32 +1,13 @@
...
 
-#[derive(Serialize, Default, Deserialize, Clone, Debug, PartialEq)]
-#[serde(rename_all = "camelCase")]
-pub struct AccountState {
-    pub state_hash: H256,
-    pub state_index: U256,
-    balances: HashMap<Address, HashMap<u16, u128>>, // UserId => (TokenId => balance)
-    pub num_tokens: u16,
-}
+// (UserId TokenId) => balance
```
nlordell: nit: Maybe use a doc comment here that is a little more descriptive. I feel like just specifying the mapping doesn't add much value as that is already described by the type.

nlordell (approved):
Awesome cleanup!

driver/src/models/account_state.rs
e00E: Even if it is not needed, it should be but it is used in the paginated orderbook reader.

PR #660, opened 2020-03-23
Whitelist filter

### LINKED ISSUE #625: Enabling configuration for a conservative solver
There is a small risk that our solver might not perform well, in case a lot of tokens get suddenly added to the exchange and "unusual, unexpected" orders are being placed.

We can minimize the risk for an outage by running a conservative solver: This solver would only process orders between the tokens available in the front-end ( a conservative token set) and would try to provide reliable solutions for just them.

For this, the token filter already implemented needs to be amended such that we don't have to explicitly exclude tokens, but can just restrict the considered tokens to a list of tokens passed via an env variable.

josojo:
closes #625 

---
testplan: unit tests

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -25,10 +25,7 @@ use crate::eth_rpc::Web3EthRpc;
 use crate::gas_station::GnosisSafeGasStation;
 use crate::http::HttpFactory;
 use crate::metrics::{HttpMetrics, MetricsServer, StableXMetrics};
-use crate::orderbook::{
-    BlacklistOrderbookFilter, FilteredOrderbookReader, PaginatedStableXOrderBookReader,
-    WhitelistOrderbookFilter,
-};
+use crate::orderbook::{FilteredOrderbookReader, OrderbookFilter, PaginatedStableXOrderBookReader};
 use crate::price_estimation::{PriceOracle, TokenData};
 use crate::price_finding::{Fee, SolverType};
 use crate::solution_submission::StableXSolutionSubmitter;
@@ -94,22 +91,13 @@ struct Options {
     /// JSON encoded object of which tokens/orders to ignore.
     ///
     /// For example: '{
-    ///   "tokens": [1, 2],
+    ///   "tokens": {"Whitelist": [1, 2]},
     ///   "users": {
     ///     "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0A": { "OrderIds": [0, 1] },
     ///     "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0B": "All"
     ///   }
-    /// }'
-    #[structopt(long, env = "ORDERBOOK_BLACKLIST_FILTER", default_value = "{}")]
-    blacklist_orderbook_filter: BlacklistOrderbookFilter,
-
-    /// JSON encoded object of which tokens/orders to ignore.
-    ///
-    /// For example: '{
-    ///   "tokens": [1, 2],
-    /// }'
-    #[structopt(long, env = "ORDERBOOK_WHITELIST_FILTER", default_value = "{}")]
-    whitelist_orderbook_filter: WhitelistOrderbookFilter,
+    #[structopt(long, env = "ORDERBOOK_FILTER", default_value = "{}")]
+    orderbook_filter: OrderbookFilter,
 
     /// The private key used by the driver to sign transactions.
     #[structopt(short = "k", long, env = "PRIVATE_KEY", hide_env_values = true)]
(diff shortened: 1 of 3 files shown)


driver/src/main.rs
```diff
@@ -97,8 +100,16 @@ struct Options {
     ///     "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0B": "All"
     ///   }
     /// }'
-    #[structopt(long, env = "ORDERBOOK_FILTER", default_value = "{}")]
-    orderbook_filter: OrderbookFilter,
+    #[structopt(long, env = "ORDERBOOK_BLACKLIST_FILTER", default_value = "{}")]
+    blacklist_orderbook_filter: BlacklistOrderbookFilter,
+
+    /// JSON encoded object of which tokens/orders to ignore.
```
nlordell: nit: the comment is a little off, should be which tokens to allow.

nlordell (approved):
Nice!

I think having both a whitelist and a blacklist for tokens feels a little redundant. Maybe it would be better if instead of having separate filter options, the token filter options could be an enum like this:
```
pub struct OrderbookFilter {
    tokens: TokenFilter,
    users: HashMap<Address, UserOrderFilter>,
}

pub enum TokenFilter {
    Whitelist(HashSet<u16>),
    Blacklist(HashSet<u16>),
}
```

That way we can keep blacklisting users but have the ability to either whitelist or blacklist tokens. Just a suggestion, but I'm happy with how it is done how. What does everyone think?

driver/src/main.rs
```diff
@@ -97,8 +100,16 @@ struct Options {
...
+    ///
+    /// For example: '{
+    ///   "tokens": [1, 2],
+    /// }'
+    #[structopt(long, env = "ORDERBOOK_WHITELIST_FILTER", default_value = "{}")]
```
fleupold: I think it would make more sense to specify a blacklist/whitelist subtype on the token in the original filter list instead of creating a new input argument.

```
enum TokenFilter {
  Whitelist(Vec<u16>),
  Blacklist(Vec<u16>),
}
```

Right now you could theoretically specify a token whitelist and a token blacklist, which doesn't really make semantic sense.

The existing orderbook_filter could look like:

```
/// For example: '{
///   "tokens": {"Whitelist": [1, 2]},
///   "users": {
///     "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0A": { "OrderIds": [0, 1] },
///     "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0B": "All"	    
///   }
```

fleupold (changes requested):
Can we incorporate the distinction of black/whitelisting into the current type as having both at the same time is semantically not well defined.

driver/src/main.rs
```diff
@@ -91,12 +91,11 @@ struct Options {
     /// JSON encoded object of which tokens/orders to ignore.
     ///
     /// For example: '{
-    ///   "tokens": [1, 2],
+    ///   "tokens": {"Whitelist": [1, 2]},
```
bh2smith: I guess it is not possible to provide an example that uses both a whitelist and a blacklisted token... since these are mutually exclusive.  (Perhaps we could add a comment here or somewhere stating that blacklist could be used in the same way.

driver/src/main.rs
```diff
@@ -91,12 +91,11 @@ struct Options {
...
     ///   "users": {
     ///     "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0A": { "OrderIds": [0, 1] },
     ///     "0x7b60655Ca240AC6c76dD29c13C45BEd969Ee6F0B": "All"
     ///   }
-    /// }'
```
bh2smith: Not sure this was meant to be removed.

bh2smith (approved):
Cool stuff! I see someones Rust Lang skills coming along well!

driver/src/orderbook/filtered_orderbook.rs
```diff
@@ -50,23 +62,40 @@ impl<'a> FilteredOrderbookReader<'a> {
...
+        let user_filtered_orders = token_filtered_orders
             .into_iter()
             .filter(|o| {
-                let user_filter = if let Some(user_filter) = self.filter.users.get(&o.account_id) {
+                if let Some(user_filter) = self.filter.users.get(&o.account_id) {
                     match user_filter {
-                        UserOrderFilter::All => true,
-                        UserOrderFilter::OrderIds(ids) => ids.contains(&o.id),
+                        UserOrderFilter::All => false,
+                        UserOrderFilter::OrderIds(ids) => !ids.contains(&o.id),
```
fleupold: Nice simplification!

fleupold (approved):
Really clean, really nice! Great unittests!

driver/src/orderbook/filtered_orderbook.rs
```diff
@@ -7,12 +7,24 @@ use serde::Deserialize;
...
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
+enum TokenFilter {
+    Whitelist(HashSet<u16>),
+    Blacklist(HashSet<u16>),
+}
+
+impl Default for TokenFilter {
+    fn default() -> Self {
+        TokenFilter::Whitelist(HashSet::new())
```
nlordell: I think the default behaviour should be to consider all tokens (i.e. an empty black-list).

nlordell (approved):
Looks really good! Just one minor comment on the default behaviour.

driver/src/orderbook/filtered_orderbook.rs
```diff
@@ -50,23 +62,40 @@ impl<'a> FilteredOrderbookReader<'a> {
 impl<'a> StableXOrderBookReading for FilteredOrderbookReader<'a> {
     fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)> {
         let (state, orders) = self.orderbook.get_auction_data(index)?;
-        let filtered = orders
+        let token_filtered_orders = match &self.filter.tokens {
+            TokenFilter::Whitelist(token_list) => {
+                if !token_list.is_empty() {
```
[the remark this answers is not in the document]
josojo: You are right, I took it out.
While panicking might have made some sense, we would then also have to check that the whitelist is not just one token, etc.

bh2smith (approved):

PR #662, opened 2020-03-25
Event based orderbook prototype

e00E:
No need to review yet this but you can take a look if you want to.


driver/src/orderbook/mod.rs
```diff
@@ -8,6 +8,7 @@ use mockall::automock;
 use paginated_auction_data_reader::PaginatedAuctionDataReader;
 use std::convert::TryInto;
 
+mod event_based_orderbook;
```
nlordell: uber-naming-nit: `orderbook::event_based_orderbook` seams like a slightly redundant name to me (same for `filtered_orderbook`). How about `orderbook::streamed`?

driver/src/orderbook/event_based_orderbook/event.rs
```diff
@@ -0,0 +1,61 @@
+use super::*;
+use ethcontract::U256;
+
+#[derive(Clone, Copy, Debug)]
+pub enum Event {
```
nlordell: This will hopefully be generated by `ethcontract`. Although, the format would be slightly different:
```
enum EventData {
    Deposit(Deposit),
    Withdraw(Withdraw),
    // ...
}
```

This is because we reuse the data in the `all_events` and `event` streams.

driver/src/orderbook/event_based_orderbook/orderbook.rs
```diff
@@ -0,0 +1,110 @@
+// Ethereum events (logs) can be both created and removed. Removals happen if the chain reorganizes
+// and ends up not including block that was previously thought to be part of the chain.
+// However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
```
fleupold: Why can the orderbook state not remove events? What information is not reflected in the event data?

I'd assume there is a way to make sure events are always applied and reverted "in order".

e00E: Because not all operations are reversible. The simplest case is the deletion of an order. Once it has been deleted we can't undo the delete because we do not know what the original order was.

e00E:
> Nice, I wonder how this can be best broken up into reviewable pieces? Do you think it's possible to do 1 PR per account balances and a separate one for order?

We can do that.

nlordell:
@e00E Maybe because of the `eth_getFilterLogs` performance, it might make sense to fetch a snapshot and then incrementally update the orderbook instead of building it exclusively from events.

e00E:
The problems with that are
1. no way to get withdraw requests, deposits and balance, only "effective" balance
2. no way to get balance of users for tokens that they do not currently have an order for
4. no way to get all registered tokens
5. need to synchronize multiple getOrdersPaginated calls with events that happen at the same time

If we had access to the private fields of the contract it could work but afaik this is  do this in hacky ways. That is something we should consider solving if we ever redeploy the contracts.

e00E:
@fleupold What do you think about this?
Currently we have the problem that the only (?) way to get past events with a `getFilterLogs` requests. But this api is not paginated so the response can be very big and take a long time because it returns all past events in one request. The only way to improve this that I see is to get past events in multiple steps like all events for every 10000 blocks.
Any method that relies on past events is going to have a slower start up time over time as more events occur - but we cannot jump start without past events because of the problems outlined above.
Is the event based orderbook still the way to go if we need all past events? Do we want to use the more efficent orderbook retrieveal from the wrapper contract with the old paginated orderbook reading instead?

fleupold:
Do we know if paginated event querying is really slower than fetching the full orderbook (which also increases over time)?

I think a midterm improvement would be to serialize a snapshot (together with a blocknumber/hash) every so many blocks to a volume and read this snapshot at cold start (allowing to only subscribe to new events after that).

Generally I would be willing to tradeoff some startup time to then have quick and flexible orderbook calculation going forward. Maybe we could even start the system with the old slow approach, fill the incremental orderbook in the background and when ready swap them out?

As to the challenges with fetching a snapshot via the smart contract

> 1. no way to get withdraw requests, deposits and balance, only "effective" balance

EpochTokenLocker exposes `getPendingWithdraw` & `getPendingDeposit`. Would this be enough to get this information? We have to deal with similar logic when computing the balance for a specific "target batch" inside [BatchExchangeViewer](https://github.com/gnosis/dex-contracts/blob/f7d0ddbb23be717d845e3e9bc138b72bd666e14d/contracts/BatchExchangeViewer.sol#L247)

> 2. no way to get balance of users for tokens that they do not currently have an order for

Good point and even if we had a list of all listed tokens you could first deposit and later list the token. Can we "lazily" query the balance as soon as we see the first order?

> 3. no way to get all registered tokens

Tokens are registered with an incremental ID. We could query `tokenIdToAddressMap` with value from 0... until it fails and this way get all registered tokens.

> 4. need to synchronize multiple getOrdersPaginated calls with events that happen at the same time

Could we query at a given block (maybe slightly in the past) and then subscribe for events as of this block?

PR #670, opened 2020-03-30
[Trivial] Remove duplicate flag names

bh2smith:
As discussed in https://github.com/gnosis/dex-services/pull/653#discussion_r399933954 this cleans up the duplicate names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,49 +161,49 @@ FLAGS:
 
 
 OPTIONS:
-        --auction-data-page-size <auction-data-page-size>
+        --auction-data-page-size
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,49 +161,49 @@ FLAGS:
 
 
 OPTIONS:
-        --auction-data-page-size <auction-data-page-size>
```
nlordell: These should be updated in `main.rs` as well. The `#[structopt(...)]` attributes need to include a `name = 'some name'` since the goal here was to keep this auto-generated.

nlordell: You can then run `cargo run --package driver -- --help` to get the full help text.

bh2smith: Right thanks. This may wind up being less trivial than anticipated.

PR #674, opened 2020-03-31
Easy log update

bh2smith:
Simply removing the substring "error" from the success log for solution submission as it pollutes the Kibana logs

Screenshot 2020-03-31 at 9 26 55 PM

### DIFF
--- a/driver/src/driver/scheduler/system.rs
+++ b/driver/src/driver/scheduler/system.rs
@@ -121,7 +121,7 @@ impl<'a> SystemScheduler<'a> {
 
 fn log_driver_result(batch_id: BatchId, driver_result: &DriverResult) {
     match driver_result {
-        DriverResult::Ok => info!("Batch {} solved without error.", batch_id.0),
+        DriverResult::Ok => info!("Batch {} successfully solved.", batch_id.0),
         DriverResult::Retry(err) => {
             error!("Batch {} failed with retryable error: {}", batch_id.0, err)
         }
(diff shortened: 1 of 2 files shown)


driver/src/main.rs
```diff
@@ -47,14 +47,14 @@ use url::Url;
     rename_all = "kebab"
 )]
 struct Options {
-    /// The log fiter to use.
+    /// The log filter to use.
```
nlordell: Unrelated, but love it!

fleupold:
> Conceptually I disagree with the idea that we should never use the word error in any log messages because it it pollutes Kibana.

I think it's fine to use the word error, however it seems a bit weird to use it in the negated form (since there was also no warning, fatal, and a bunch of other things that didn't happen while solving this batch).

bh2smith:
For some reason travis hasn't started here yet...

driver/src/driver/scheduler/system.rs
```diff
@@ -121,7 +121,7 @@ impl<'a> SystemScheduler<'a> {
 
 fn log_driver_result(batch_id: BatchId, driver_result: &DriverResult) {
     match driver_result {
-        DriverResult::Ok => info!("Batch {} solved without error.", batch_id.0),
+        DriverResult::Ok => info!("Batch {} successfully solved.", batch_id.0),
```
bh2smith, line 124:
```suggestion
        DriverResult::Ok => info!("Batch {} solved successfully.", batch_id.0),
```

PR #678, opened 2020-04-03
[event based orderbook] First step towards event based orderbook

e00E:
Introduces a new streamed orderbook module and a way to build it based
on events only.
In the future we will add the code that actually retrieves the events.

There are two `unimplemented!` in the code for which I will make another PR because that part is more complicated.

Test Plan: new unit tests (but this code isn't live yet)

Before merging we need to release a new ethcontract version and move the ethcontract dependency back to crates.io.


driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,561 @@
...
+            .ok_or(Error::MathUnderflow)?;
+        Ok(())
+    }
+
+    fn token_listing(&mut self, event: &TokenListing) -> Result<(), Error> {
+        self.tokens.0.push((event.id, event.token));
+        Ok(())
+    }
+
+    fn order_placement(&mut self, event: &OrderPlacement) -> Result<(), Error> {
```
fleupold, line 139: naming nit: maybe these can all be imperatives? E.g. place_order/request_withdraw or process_order_placement if you prefer keeping the event name.

driver/src/orderbook/streamed/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+type OrderId = u16;
+type TokenId = u16;
+type BatchId = u32;
+
+#[cfg(test)]
+mod tests {
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Deserialize, Serialize)]
+    pub struct EventBackup {
```
fleupold, line 18: how is this used?

fleupold (approved):
Looking really good and clean. Some small suggestions inline. Can we add a test coverage tool to our repo? For the state module I think it would make sense to try and get 100% branch coverage.

driver/src/solution_submission/mod.rs
```diff
@@ -131,7 +131,9 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
 fn extract_transaction_hash(err: &Error) -> Option<H256> {
     err.downcast_ref::<MethodError>()
         .and_then(|method_error| match &method_error.inner {
-            ExecutionError::Failure(tx_hash) => Some(*tx_hash),
+            ExecutionError::Failure(transaction_receipt) => {
```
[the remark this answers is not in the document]
bh2smith: Would this be the task related to the fact that ethcontracts now returns transactions receipts for failed transactions? If so, I believe it would be my responsibility to create a task here that makes use of this returned receipt.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,561 @@
...
+        event: &SolutionSubmission,
+        block_batch_id: BatchId,
+    ) -> Result<(), Error> {
+        let fee_token = self.tokens.address(0).ok_or(Error::UnknownToken)?;
+        let balance = self
+            .balances
+            .entry((event.submitter, fee_token))
+            .or_default();
+        balance.update_deposit_balance(block_batch_id);
+        balance.balance += event.burnt_fees;
```
fleupold, line 186: Similar to https://github.com/gnosis/dex-services/pull/679#discussion_r403889834, here as well we should count this as a "deposit" since the fees are not yet spendable in the current batch.

fleupold (changes requested):
Requesting changes to fix the way the fee is credited.

driver/src/solution_submission/mod.rs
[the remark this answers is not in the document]
fleupold: Yes, that would be great. We can quickly tackle this one outside of this PR.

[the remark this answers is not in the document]
e00E: Once the next ethcontract version has been released we can move this out to a separate "Update ethonctract" PR. I wanted to include it here to make the build succeed.

driver/src/orderbook/streamed/mod.rs
e00E: I didn't mean to include this. I was using it to test locally.

driver/src/orderbook/streamed/state.rs
e00E: I don't mind but is it really an improvement? These functions are named exactly like the event type they handle but in snake_case.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,561 @@
...
+        match self.pending_deposit.as_mut() {
+            Some(deposit) => {
+                deposit.amount += new_deposit.amount;
+                deposit.batch_id = new_deposit.batch_id
+            }
+            None => self.pending_deposit = Some(new_deposit),
+        }
+    }
+    fn update_deposit_balance(&mut self, current_batch_id: BatchId) {
+        match self.pending_deposit {
```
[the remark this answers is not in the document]
e00E: No because it only gets set to None `if deposit.batch_id < current_batch_id`.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,561 @@
...
+    fn deposit(&mut self, event: &Deposit, block_batch_id: BatchId) -> Result<(), Error> {
+        self.balances
+            .entry((event.user, event.token))
+            .or_default()
+            .deposit(
+                Flux {
+                    amount: event.amount,
+                    batch_id: event.batch_id,
+                },
+                block_batch_id,
```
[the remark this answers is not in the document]
e00E: The contract currently only emits deposit events with batch_id set to getCurrentBatchId, so yes. On the other hand the event still does have this field which indicates it might be possible to set up a deposit further in the future.
Should I add a check and turn it into an error?

driver/src/orderbook/streamed/state.rs
e00E: I think I have a good solution for these pending balances from solution submission and trades. Since that is related to how trades are handled I will move this in the other PR and this PR will have solution submission unimplemented.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,561 @@
...
+
+// Most types, fields, functions in this module mirror the smart contract because we need to
+// emulate what it does based on the events it emits.
+
+/// The orderbook state as built from received events.
+///
+/// Note that there is no way to revert an event. The order in which events are received matters
+/// Applying events `A B` does not always result in the same state as `B A`. For example, a Trade
+/// can only be processed if the OrderPlacement for the traded order has previously been observed.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
```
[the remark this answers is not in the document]
e00E: It makes sense because the fee token emits a TokenListing event like all the other tokens which means it will be included in the history of all events.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,561 @@
...
+            token: address(0),
+            id: 0,
+        };
+        state.event(&Event::TokenListing(event), 0).unwrap();
+        state
+    }
+
+    fn account_state(state: &State, batch_id: BatchId) -> AccountState {
+        AccountState(state.account_state(batch_id).collect())
+    }
```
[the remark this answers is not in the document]
e00E: There is no practical difference. To me this just felt more natural and it is slightly less to type because it doesn't start a new `impl` block.
Even for non test `new` you could implement it as a free standing function but then it would be more awkward with imports but in the test there is no importing.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,561 @@
...
+            .entry((event.submitter, fee_token))
+            .or_default();
+        balance.update_deposit_balance(block_batch_id);
+        balance.balance += event.burnt_fees;
+        Ok(())
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum Error {
```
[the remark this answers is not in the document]
e00E: `thiserror` makes a lot of sense when you're forwarding other error types with `#[from]` but in this case all it would do is implement `Display` for us. I wouldn't add thiserror as a new dependency just for that but since we already have it I might as well use it for that too.

e00E:
> Can we add a test coverage tool to our repo? For the state module I think it would make sense to try and get 100% branch coverage.

I'll look into this in a separate PR.

driver/src/orderbook/streamed/state.rs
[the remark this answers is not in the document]
nlordell: It also implements `std::error::Error` for integration with `anyhow::Error` and other Error libraries in general. I think it has value here regardless.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,531 @@
...
+    OrderAlreadyExists,
+    #[error("math underflow")]
+    MathUnderflow,
+}
+
+/// Bidirectional map between token id and token address.
+///
+/// Std does not have this type so we use a vector. Alternatively we could find a crate.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+struct Tokens(Vec<(TokenId, TokenAddress)>);
```
nlordell, line 184: This could be done with `(HashMap<TokenId, TokenAddress>, HashMap<TokenAddress, TokenId>)`, but I'm not sure its worth it (not for the size of the token list we currently have at least).

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,531 @@
...
+
+impl Tokens {
+    fn get_address_by_id(&self, id: TokenId) -> Option<Address> {
```
nlordell, line 187: Can't this be re-written as `self.0[id as usize].1`?

driver/src/solution_submission/mod.rs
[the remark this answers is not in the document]
nlordell: Already done! #691

nlordell (approved):
Looks good

driver/src/orderbook/streamed/state.rs
nlordell: Might not make sense to change it though, as the current implementation is a little easier to understand and relies on less assumptions. What does everyone think?

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,531 @@
...
+    fn get_id_by_address(&self, address: TokenAddress) -> Option<TokenId> {
+        self.0
+            .iter()
+            .find(|(_, address_)| *address_ == address)
+            .map(|(token_id, _)| *token_id)
+    }
+}
+
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+struct Order {
```
nlordell, line 203: nit: This is a data type, and I think it might make sense for it to live with the other type aliases `TokenId` and `BatchId`. Maybe `Flux` and `Balance` as well? That way this module focuses more on the logic around keeping the data in sync with the EVM.

nlordell (approved):
Still looks good!

driver/src/orderbook/streamed/state.rs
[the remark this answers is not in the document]
nlordell: Here I would argue `state.account_state` should return an `AccountState`.

driver/src/orderbook/streamed/state.rs
nlordell: For consistency with `apply_event`, it could be `apply_order_placement`.

nlordell: I don't feel strongly either way.

driver/src/orderbook/streamed/state.rs
e00E: It is specific to this file (not pub) so I want to keep it here.

nlordell: Ah, good point. never mind.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,531 @@
...
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::iter::Iterator;
+use thiserror::Error;
+
+// TODO: Should we handle https://github.com/gnosis/dex-contracts/issues/620 ?
+// There is a workaround detailed in the issue it hasn't been implemented.
+
+// Most types, fields, functions in this module mirror the smart contract because we need to
+// emulate what it does based on the events it emits.
```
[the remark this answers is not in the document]
e00E: I was thinking that this isn't relevant to users of the struct (and module) because it is about implementation details. It is only developers editing this file. Not that I mind either way but a normal comment felt more appropriate.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,531 @@
...
+    balances: HashMap<(UserId, TokenAddress), Balance>,
+    tokens: Tokens,
+}
+
+impl State {
+    /// Can be used to create a `crate::models::AccountState`.
+    pub fn account_state(
+        &self,
+        batch_id: BatchId,
+    ) -> impl Iterator<Item = ((UserId, TokenId), u128)> + '_ {
```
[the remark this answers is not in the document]
e00E: I'll leave it as is for now and if I only use it as AccountState later then I'll change the function. I implemented it this way to be as flexible as possible without forcing consumers to pay for a hashmap if they don't need it.
I predict that we might filter 0 balances out for example.

driver/src/orderbook/streamed/state.rs
e00E: I prefer the current way in order to not make assumptions about token ids that we would then have to check in other places.

driver/src/orderbook/streamed/state.rs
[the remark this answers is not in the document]
nlordell: Maybe it shouldn't be public then?

driver/src/orderbook/streamed/state.rs
nlordell: Definitely not worth it.

driver/src/orderbook/streamed/state.rs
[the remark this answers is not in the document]
e00E: There has to be some way to get the orders and account state so it is going to be public even though it's not needed in this PR specifically. This is a tradeoff between flexibility (iterator) and convenience (AccountState). As I said I'll reevaluate this in the PR that actually makes use of the function.

driver/src/orderbook/streamed/state.rs
```diff
@@ -0,0 +1,531 @@
...
+        // Orders are allowed to be deleted multiple times.
+        self.orders.remove(&(event.owner, event.id));
+        Ok(())
+    }
+}
+
+#[derive(Clone, Copy, Debug, Error)]
+pub enum Error {
+    #[error("unknown token")]
+    UnknownToken,
```
fleupold, line 171: For debugging purposes it might be nice if these had the IDs as part of the enum type and particularly the error message.

E.g.     `#[error("Unknown Token `{0}`")]`

fleupold (approved):
Very cool!

driver/src/orderbook/streamed/state.rs
e00E: This is implemented in the next PR.

driver/src/orderbook/streamed/state.rs
e00E: I'll rename it to `apply_` in the next PR (will make merging easier).

bh2smith (approved):

bh2smith:
Is this meant to remain open? Seems like all these changes have already been merged... or were they all merged into here and this is the final change?

e00E:
> Is this meant to remain open? Seems like all these changes have already been merged... or were they all merged into here and this is the final change?

I started merging other PRs that depend on this one into this one and wanted to use this as the final pr. I realize now that it is better to do it the way Felix described and I am probably creating needless notifications for you when I merge into here.

PR #679, opened 2020-04-03
Handle trades in event based orderbook

e00E:
This is based on https://github.com/gnosis/dex-services/pull/678 .

### DIFF
--- a/driver/src/orderbook/streamed/balance.rs
+++ b/driver/src/orderbook/streamed/balance.rs
@@ -59,6 +59,7 @@ impl Balance {
                 .checked_add(self.deposit.amount)
                 .ok_or(Error::MathOverflow)?;
             self.balance = new_balance;
+            self.deposit.batch_id = deposit_batch_id;
             self.deposit.amount = amount;
         } else {
             let new_deposit_amount = self
@@ -156,9 +157,9 @@ impl Balance {
         batch_id: BatchId,
         operation: Operation,
     ) -> Result<(), Error> {
-        let new_balance = self.get_balance(batch_id)?;
         match self.pending_solution.batch_id.cmp(&batch_id) {
             Ordering::Less => {
+                let new_balance = self.get_balance_internal(batch_id, false)?;
                 self.balance = new_balance;
                 self.clear_pending_deposit_and_solution_if_possible(batch_id);
                 self.pending_solution.batch_id = batch_id;
--- a/driver/src/orderbook/streamed/balance.rs
+++ b/driver/src/orderbook/streamed/balance.rs
@@ -8,7 +8,7 @@ pub struct Balance {
     balance: U256,
     deposit: Flux,
     withdraw: Flux,
-    pending_solution: PendingSolutionBalance,
+    pending_proceeds: PendingProceeds,
 }
 
 #[derive(Copy, Clone, Debug, Default, Deserialize, Serialize)]
@@ -17,14 +17,16 @@ struct Flux {
     amount: U256,
 }
 
+/// Balance change from a potential solution that might still get replaced by a better solution in
+/// the same batch.
 #[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
-struct PendingSolutionBalance {
+struct PendingProceeds {
     batch_id: BatchId,
     increase: U256,
     decrease: U256,
 }
 
-impl PendingSolutionBalance {
+impl PendingProceeds {
     fn get_field(&mut self, operation: Operation) -> &mut U256 {
         match operation {
             Operation::Sell => &mut self.decrease,
@@ -127,11 +129,11 @@ impl Balance {
                 .ok_or(Error::MathOverflow)?;
         }
         // TODO: this could temporarily fail while not all trades for solution have been received.
-        if self.pending_solution.batch_id < batch_id {
+        if self.pending_proceeds.batch_id < batch_id {
             balance = balance
-                .checked_add(self.pending_solution.increase)
+                .checked_add(self.pending_proceeds.increase)
                 .ok_or(Error::MathOverflow)?
-                .checked_sub(self.pending_solution.decrease)
+                .checked_sub(self.pending_proceeds.decrease)
                 .ok_or(Error::MathOverflow)?;
         }
         if include_withdraw && self.withdraw.batch_id < batch_id {
@@ -145,9 +147,9 @@ impl Balance {
         if self.deposit.batch_id < batch_id {
             self.deposit.amount = 0.into();
         }
-        if self.pending_solution.batch_id < batch_id {
-            self.pending_solution.increase = 0.into();
-            self.pending_solution.decrease = 0.into();
+        if self.pending_proceeds.batch_id < batch_id {
+            self.pending_proceeds.increase = 0.into();
+            self.pending_proceeds.decrease = 0.into();
         }
     }
 
@@ -157,16 +159,16 @@ impl Balance {
         batch_id: BatchId,
         operation: Operation,
     ) -> Result<(), Error> {
-        match self.pending_solution.batch_id.cmp(&batch_id) {
+        match self.pending_proceeds.batch_id.cmp(&batch_id) {
             Ordering::Less => {
                 let new_balance = self.get_balance_internal(batch_id, false)?;
                 self.balance = new_balance;
                 self.clear_pending_deposit_and_solution_if_possible(batch_id);
-                self.pending_solution.batch_id = batch_id;
-                *self.pending_solution.get_field(operation) = amount.into();
+                self.pending_proceeds.batch_id = batch_id;
+                *self.pending_proceeds.get_field(operation) = amount.into();
             }
             Ordering::Equal => {
-                let field = self.pending_solution.get_field(operation);
+                let field = self.pending_proceeds.get_field(operation);
                 *field = field
                     .checked_add(amount.into())
                     .ok_or(Error::MathOverflow)?
@@ -182,12 +184,12 @@ impl Balance {
         batch_id: BatchId,
         operation: Operation,
     ) -> Result<(), Error> {
-        if self.pending_solution.batch_id != batch_id {
+        if self.pending_proceeds.batch_id != batch_id {
             return Err(Error::RevertingNonExistentTrade);
         }
         let field = match operation {
-            Operation::Sell => &mut self.pending_solution.decrease,
-            Operation::Buy => &mut self.pending_solution.increase,
+            Operation::Sell => &mut self.pending_proceeds.decrease,
+            Operation::Buy => &mut self.pending_proceeds.increase,
         };
         *field = field
             .checked_sub(amount.into())
--- a/driver/src/orderbook/streamed/balance.rs
+++ b/driver/src/orderbook/streamed/balance.rs
@@ -1,0 +1,198 @@
+use super::*;
+use error::Error;
+use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
+
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+pub struct Balance {
+    balance: U256,
+    deposit: Flux,
+    withdraw: Flux,
+    pending_solution: PendingSolutionBalance,
+}
+
+#[derive(Copy, Clone, Debug, Default, Deserialize, Serialize)]
+struct Flux {
+    batch_id: BatchId,
+    amount: U256,
+}
+
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+struct PendingSolutionBalance {
+    batch_id: BatchId,
+    increase: U256,
+    decrease: U256,
+}
+
+impl PendingSolutionBalance {
+    fn get_field(&mut self, operation: Operation) -> &mut U256 {
+        match operation {
+            Operation::Sell => &mut self.decrease,
+            Operation::Buy => &mut self.increase,
+        }
+    }
+}
+
+enum Operation {
+    Sell,
+    Buy,
+}
+
+impl Balance {
+    pub fn add_balance(&mut self, amount: U256) -> Result<(), Error> {
+        self.balance = self
+            .balance
+            .checked_add(amount)
+            .ok_or(Error::MathOverflow)?;
+        Ok(())
+    }
+
+    pub fn deposit(
+        &mut self,
+        amount: U256,
+        deposit_batch_id: BatchId,
+        current_batch_id: BatchId,
+    ) -> Result<(), Error> {
+        if self.deposit.batch_id < current_batch_id {
+            let new_balance = self
+                .balance
+                .checked_add(self.deposit.amount)
+                .ok_or(Error::MathOverflow)?;
+            self.balance = new_balance;
+            self.deposit.amount = amount;
+        } else {
+            let new_deposit_amount = self
+                .deposit
+                .amount
+                .checked_add(amount)
+                .ok_or(Error::MathOverflow)?;
+            self.deposit.batch_id = deposit_batch_id;
+            self.deposit.amount = new_deposit_amount;
+        }
+        Ok(())
+    }
+
+    pub fn withdraw_request(&mut self, amount: U256, batch_id: BatchId) {
+        self.withdraw.batch_id = batch_id;
+        self.withdraw.amount = amount;
+    }
+
+    pub fn withdraw(&mut self, amount: U256, batch_id: BatchId) -> Result<(), Error> {
+        if self.withdraw.batch_id >= batch_id {
+            return Err(Error::WithdrawEarlierThanRequested(self.withdraw.batch_id));
+        }
+        if self.withdraw.amount < amount {
+            return Err(Error::WithdrawMoreThanRequested(self.withdraw.amount));
+        }
+        let balance = self.get_balance_internal(batch_id, false)?;
+        let new_balance = balance
+            .checked_sub(amount)
+            .ok_or(Error::WithdrawMoreThanBalance(balance))?;
+        self.balance = new_balance;
+        self.clear_fluxes_and_pending_solution_if_possible(batch_id);
+        Ok(())
+    }
+
+    pub fn get_balance(&self, batch_id: BatchId) -> Result<U256, Error> {
+        self.get_balance_internal(batch_id, true)
+    }
+
+    pub fn sell(&mut self, amount: u128, batch_id: BatchId) -> Result<(), Error> {
+        self.sell_buy_internal(amount, batch_id, Operation::Sell)
+    }
+
+    pub fn buy(&mut self, amount: u128, batch_id: BatchId) -> Result<(), Error> {
+        self.sell_buy_internal(amount, batch_id, Operation::Buy)
+    }
+
+    pub fn revert_sell(&mut self, amount: u128, batch_id: BatchId) -> Result<(), Error> {
+        self.revert_sell_buy_internal(amount, batch_id, Operation::Sell)
+    }
+
+    pub fn revert_buy(&mut self, amount: u128, batch_id: BatchId) -> Result<(), Error> {
+        self.revert_sell_buy_internal(amount, batch_id, Operation::Buy)
+    }
+
+    fn get_balance_internal(
+        &self,
+        batch_id: BatchId,
+        include_withdraw: bool,
+    ) -> Result<U256, Error> {
+        let mut balance = self.balance;
+        if self.deposit.batch_id < batch_id {
+            balance = balance
+                .checked_add(self.deposit.amount)
+                .ok_or(Error::MathOverflow)?;
+        }
+        // TODO: this could temporarily fail while not all trades for solution have been received.
+        if self.pending_solution.batch_id < batch_id {
+            balance = balance
+                .checked_add(self.pending_solution.increase)
+                .ok_or(Error::MathOverflow)?
+                .checked_sub(self.pending_solution.decrease)
+                .ok_or(Error::MathOverflow)?;
+        }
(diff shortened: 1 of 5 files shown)


driver/src/orderbook/streamed/state.rs
```diff
@@ -186,6 +212,61 @@ impl State {
...
     }
+
+    fn apply_trade(
+        &mut self,
+        user: UserId,
+        order: OrderId,
+        executed_sell_amount: u128,
+        executed_buy_amount: u128,
+        batch_id: BatchId,
+        revert: bool,
```
fleupold, line 223: I'm not a fan of passing this boolean around, but I guess otherwise we would have to duplicate a ton of code? 

Alternatively we could have one method that gets a combined result of all dependencies (buyToken, sellToken, balances, order) and then do the +/- in the functions directly? 

But since it's an implementation detail it's also not a big deal

driver/src/orderbook/streamed/state.rs
```diff
@@ -186,6 +212,61 @@ impl State {
...
+    ) {
+        let balance = self.balances.entry((user, token)).or_default();
+        balance.update_deposit_balance(batch_id);
+        // It possible that we get an underflow here because the user did not have enough
+        // deposited for the sale but in the same batch filled another order that increased their
+        // balance but whose event we have not yet processed.
+        if subtract {
+            balance.balance = balance.balance.overflowing_sub(executed_amount.into()).0;
+        } else {
+            balance.balance = balance.balance.overflowing_add(executed_amount.into()).0;
```
fleupold, line 267: add should not be able to overflow though, right?

driver/src/orderbook/streamed/state.rs
```diff
@@ -186,6 +212,61 @@ impl State {
...
+    fn apply_balance_from_trade(
+        &mut self,
+        user: UserId,
+        token: TokenAddress,
+        executed_amount: u128,
+        batch_id: BatchId,
+        subtract: bool,
+    ) {
+        let balance = self.balances.entry((user, token)).or_default();
+        balance.update_deposit_balance(batch_id);
```
fleupold, line 260: Here again I think it would be nice if balance itself would decide when to update deposits.

fleupold (approved):

driver/src/orderbook/streamed/state.rs
```diff
@@ -186,6 +212,61 @@ impl State {
...
+            } else {
+                order.used_amount += executed_sell_amount;
+            }
+        }
+        self.apply_balance_from_trade(user, sell_token, executed_sell_amount, batch_id, !revert);
+        self.apply_balance_from_trade(user, buy_token, executed_buy_amount, batch_id, revert);
+        Ok(())
+    }
+
+    fn apply_balance_from_trade(
```
fleupold, line 251: After thinking more about https://github.com/gnosis/dex-services/issues/684 I think we need to treat the balance update as a deposit (valid as of next batch) as if you query the orderbook for the current batch you should not assume this balance is available (it will be reverted when you try to use it)

fleupold (changes requested):
I missed a subtle issue about crediting the balance of a trade that needs fixing.

driver/src/orderbook/streamed/state.rs
e00E: It can for the same reason. First the balance underflows to a very large value, then the add overflows it back to a small value.

driver/src/orderbook/streamed/state.rs
e00E: I have implemented a solution for this. This not only applies to balances but also to the order used amounts by the way.

driver/src/orderbook/streamed/state.rs
```diff
@@ -11,47 +11,63 @@ use std::collections::HashMap;
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct State {
     orders: HashMap<(UserId, OrderId), Order>,
     balances: HashMap<(UserId, TokenAddress), Balance>,
     tokens: Tokens,
+    pending_solution: PendingSolution,
 }
 
 impl State {
+    pub fn new() -> Self {
```
nlordell, line 37: nit: why not implement `Default` instead of having a `new` method?
```
impl Default for State {
    fn default() -> Self {
```

driver/src/orderbook/streamed/state.rs
```diff
@@ -11,47 +11,63 @@ use std::collections::HashMap;
...
+            balances: HashMap::new(),
+            tokens: Tokens(Vec::new()),
+            pending_solution: PendingSolution::AccumulatingTrades(Vec::new()),
+        }
+    }
+
     /// Can be used to create a `crate::models::AccountState`.
     pub fn account_state(
-        &self,
+        &mut self,
```
nlordell, line 48: nit: not a fan of having getter methods mutable. I think a separate call to `apply_pending_solution` should be the responsibility of the caller, and that the accessor should fail if there is a pending solution to apply.

driver/src/orderbook/streamed/state.rs
```diff
@@ -11,47 +11,63 @@ use std::collections::HashMap;
...
                 Some((
                     (*user_id, token_id),
                     balance.current_balance(batch_id).low_u128(),
                 ))
             })
     }
 
-    pub fn orders(&self, batch_id: BatchId) -> impl Iterator<Item = ModelOrder> + '_ {
+    pub fn orders(&mut self, batch_id: BatchId) -> impl Iterator<Item = ModelOrder> + '_ {
+        self.apply_pending_solution_if_needed(batch_id);
```
nlordell, line 67: nit: Same as above.

driver/src/orderbook/streamed/state.rs
```diff
@@ -158,11 +177,150 @@ impl State {
...
+        let balance = self.balances.entry((user, token)).or_default();
+        balance.update_deposit_balance(batch_id);
+        // It possible that we get an underflow here because the user did not have enough
+        // deposited for the sale but in the same batch filled another order that increased their
+        // balance but whose trade event we have not yet processed.
+        balance.balance = if subtract {
+            balance.balance.overflowing_sub(executed_amount.into()).0
+        } else {
+            balance.balance.overflowing_add(executed_amount.into()).0
+        };
```
nlordell, line 322: I feel like overflowing should be an error (if it doesn't overflow back the other way), but I don't see an easy way to implement this.

driver/src/orderbook/streamed/state.rs
```diff
@@ -277,6 +443,21 @@ impl Balance {
     }
 }
 
+/// The current solution for the batch but not yet finale as it can still be replaced by a better
+/// solution.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+enum PendingSolution {
+    /// We observed Trade events but no SolutionSubmission.
+    AccumulatingTrades(Vec<Trade>),
```
nlordell, line 451: This is cool!

driver/src/orderbook/streamed/state.rs
```diff
@@ -158,11 +177,150 @@ impl State {
...
+        event: &SolutionSubmission,
+        block_batch_id: BatchId,
+    ) -> Result<(), Error> {
+        if self.tokens.get_address_by_id(0).is_none() {
+            return Err(Error::SolutionWithoutFeeToken);
+        }
+        let trades = match &mut self.pending_solution {
+            PendingSolution::AccumulatingTrades(trades) => std::mem::replace(trades, Vec::new()),
+            // This would be weird because it means that the previous solution had no trades. Not
+            // sure if we should error.
```
nlordell, line 228: Yes, as that would mean it is a solution with 0 utility which is not allowed and something went wrong.

driver/src/orderbook/streamed/state.rs
```diff
@@ -158,11 +177,150 @@ impl State {
...
+        let trades = match &mut self.pending_solution {
+            PendingSolution::Submitted {
+                batch_id,
+                submitter,
+                burnt_fees,
+                trades,
+            } if *batch_id < block_batch_id => {
+                // Cannot fail because we ensure there is a fee token when a solution event is
+                // received.
+                let fee_token = self.tokens.get_address_by_id(0).unwrap();
```
nlordell, line 250: Maybe `fee_token` can be part of the `Submitted` variant, then the unwrap would not be necessary.

driver/src/orderbook/streamed/state.rs
```diff
@@ -158,11 +177,150 @@ impl State {
...
+        user: UserId,
+        order: OrderId,
+        executed_sell_amount: u128,
+        executed_buy_amount: u128,
+        batch_id: BatchId,
+    ) -> Result<(), Error> {
+        let order = self
+            .orders
+            .get_mut(&(user, order))
+            .ok_or(Error::UnknownOrder)?;
```
nlordell, line 287: nit: `UnknownOrder` and `UnknownToken` should include `order_id` and `token_id`.

nlordell (commented):
Some small nits, but looks good in general!

driver/src/orderbook/streamed/state.rs
```diff
@@ -11,47 +11,75 @@ use std::collections::HashMap;
...
+    pending_solution: PendingSolution,
+}
+
+impl Default for State {
+    fn default() -> Self {
+        Self {
+            orders: HashMap::new(),
+            balances: HashMap::new(),
+            tokens: Tokens(Vec::new()),
+            pending_solution: PendingSolution::AccumulatingTrades(Vec::new()),
```
fleupold, line 42: nit: would it make sense to implement default on PendingSolution instead and derive it here?

driver/src/orderbook/streamed/state.rs
```diff
@@ -83,17 +123,20 @@ impl State {
     /// `block_batch_id` is the current batch based on the timestamp of the block that contains the
     ///  event.
     pub fn apply_event(&mut self, event: &Event, block_batch_id: BatchId) -> Result<(), Error> {
+        self.apply_pending_solution_if_needed(block_batch_id);
```
fleupold, line 126: What happens if there is a batch without an event at all? Would we then not be able to get the account state as nothing would trigger the apply_pending_solution if needed?

Generally, it would be great if we could query the currently foreseen balance state for a future batch so that we can optimistically fetch state before the batch is finalized.

driver/src/orderbook/streamed/state.rs
```diff
@@ -66,7 +94,19 @@ impl State {
...
+    }
+
+    /// Returns whether the state needs to be updated with a pending solution that becomes the
+    /// accepted solution because the batch id has incremented. In this case
+    /// `apply_pending_solution_if_needed` has to be called before `account_state` and `orders` can
+    /// be used.
+    pub fn needs_to_apply_solution(&self, batch_id_: BatchId) -> bool {
+        match self.pending_solution {
+            PendingSolution::Submitted { batch_id, .. } if batch_id < batch_id_ => true,
+            _ => false,
```
fleupold, line 108: Can we distinguish the variable name a little bit more clearly please? One could be pending_solution_batch_id and the other one could be target_batch_id or just batch_id

driver/src/orderbook/streamed/state.rs
```diff
@@ -11,47 +11,75 @@ use std::collections::HashMap;
...
-                    (*user_id, token_id),
-                    balance.current_balance(batch_id).low_u128(),
-                ))
-            })
-    }
-
-    pub fn orders(&self, batch_id: BatchId) -> impl Iterator<Item = ModelOrder> + '_ {
-        self.orders
+    ) -> Result<impl Iterator<Item = ((UserId, TokenId), u128)> + '_, Error> {
+        if self.needs_to_apply_solution(batch_id) {
```
fleupold, line 53: Can we optimistically apply the solution here in memory (assuming nothing more would happen)? 

Later, we might want to expose an API later that allows subscribing to account_state updates.

fleupold (commented):
Generally I'm a bit worried about the added complexity that comes with the pending solution enum (Maybe we can just store the latest fee and BatchId for reversion purposes).

I wonder if trades could be handled similar to deposits that will automatically get applied when we query the state for a future batch but are not yet counted for the current batch. Maybe we should get together synchronously tomorrow morning to discuss the design?

fleupold:
Summary of synchronous discussion between @e00E @nlordell @bh2smith and myself.

### Problem: 
We have solutions coming in at times where we don’t know if they are going to be the final one. Thus, we don’t want to credit them yet in case they get reverted. Also any proceeds would not yet be valid if we tried to re-solve the same batch again. 

This is not only true for balances but also the usedAmounts on each order.

### Current implementation
We introduced a separate struct for the pending solution which is a state machine that can be either “complete” or in the process of “accumulating trades”. The pending solution struct gets applied to the state when it is known to be irevertible.

The advantage of using such a struct is that reversion can be implemented by simply throwing away the existing one and creating a new one.

The current approach has the drawback that we can sometimes be in a limbo state: We cannot query the order book if a pending solution needs to be applied for that batch (e.g. when pre-querying the state for the coming batch) and are relying on the liveliness of the system (receiving another event) to perform the required application.

Also, the state machine adds a bunch of `match` statements and different behaviours for its different variants which increases the complexity of the code.

### Proposed Changes

Generally we would like to treat trades more like we treat deposits (they are initiated by an event but can become “valid” without any further liveness of the contract).

Instead of blocking queries when the state is not yet finalised we return an “optimistic state” for the requested batchId with the pending proceeds/usedAmounts applied if they matter. Persisting the pending proceeds/usedAmounts can happen lazily on write.

By adding the pendingProceeds as a field on the user balance rather than in an orthogonal pendingSolution struct we hope to linearise the control flow and co-locate pending proceeds and deposit logic. We will see if this is actually better than keeping the extra struct as it requires implementing a similar concept for orders (to store their pending used Amounts)

We will still need a reduced struct for the pending solution because the submitter and feeAmount are not recoverable, yet need to be reverted when a new solution is submitted for the same batch.

bh2smith (approved):
Looks pretty good. I like the function rename.

e00E:
We can review this again. I apologize for making the PR so big. It got bigger because the new approach takes more code but also has a lot more sanity checks and because I refactored some types into their own file because `state.rs` got too big.

There is one todo left with regards to how we want to handle the situation where a solution has not yet been fully received. I will include that in a separate PR.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -0,0 +1,199 @@
...
+                self.pending_proceeds.batch_id = batch_id;
+                *self.pending_proceeds.get_field(operation) = amount.into();
+            }
+            Ordering::Equal => {
+                let field = self.pending_proceeds.get_field(operation);
+                *field = field
+                    .checked_add(amount.into())
+                    .ok_or(Error::MathOverflow)?
+            }
+            Ordering::Greater => return Err(Error::TradeForPastBatch),
```
fleupold, line 176: We never reset the batch_id. Is this not a problem when events are reverted due to reorgs?

driver/src/orderbook/streamed/balance.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn revert_sell_buy_internal(
+        &mut self,
+        amount: u128,
+        batch_id: BatchId,
+        operation: Operation,
+    ) -> Result<(), Error> {
+        if self.pending_proceeds.batch_id != batch_id {
+            return Err(Error::RevertingNonExistentTrade);
+        }
+        let field = match operation {
```
fleupold, line 190: nit: use `get_field` here?

driver/src/orderbook/streamed/balance.rs
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+
+    fn sell_buy_internal(
+        &mut self,
+        amount: u128,
+        batch_id: BatchId,
+        operation: Operation,
+    ) -> Result<(), Error> {
+        match self.pending_proceeds.batch_id.cmp(&batch_id) {
```
fleupold, line 162: Maybe we can only check if Ordering::Greater in which case we fail. Otherwise we unconditionally update_balance_internal (which would reset the pending proceeds) and then do the get_field + checked_add unconditionally? This would reduce the amount of logic that happens in the match arms to just a simple consistency check.

driver/src/orderbook/streamed/order.rs
```diff
@@ -0,0 +1,125 @@
...
+                .ok_or(Error::MathOverflow)?
+        } else {
+            used_amount
+        };
+        // Make sure the invariant still holds.
+        if new_pending_used_amount > self.price_denominator {
+            return Err(Error::TradeByMoreThanOrderLimit(self.price_denominator));
+        }
+        self.pending_used_amount = (batch_id, new_pending_used_amount);
+        self.used_amount = new_used_amount;
```
fleupold, line 63: Why is this written? I don't see it being changed in this method. After reviewing get_used_amount carefully I can see that it takes the pending_used_amount into account which in this method might get overridden and thus we have to write this value otherwise we lose this information.

This logic is extremely subtle and hidden in this code. Instead I'd suggest we turn `get_used_amount ` into `update_used_amount`, which takes care of applying and resetting the pending_used_amount if necessary. This way we can also get rid of the if/else statement when computing the new_pending_amount as if it was reset it would be 0 so we can always add `used_amount` to it.

driver/src/orderbook/streamed/state.rs
```diff
@@ -35,37 +33,40 @@ impl State {
...
-                    id: *order_id,
-                    account_id: *user_id,
-                    buy_token: order.buy_token,
-                    sell_token: order.sell_token,
-                    buy_amount,
-                    sell_amount,
-                }
+            .filter(move |(_, order)| order.is_valid_in_batch(batch_id))
+            .map(move |((user_id, order_id), order)| {
+                order.as_model_order(batch_id, *user_id, *order_id)
```
fleupold, line 69: I really like these helper methods!

driver/src/orderbook/streamed/state.rs
```diff
@@ -84,97 +85,204 @@ impl State {
...
+        event: &OrderDeletion,
+        block_batch_id: BatchId,
+    ) -> Result<(), Error> {
+        if let Some(order) = self.orders.get(&(event.owner, event.id)) {
+            if order.is_valid_in_batch(block_batch_id - 1) {
+                return Err(Error::DeletingValidOrder);
+            } else {
+                self.orders.remove(&(event.owner, event.id));
+            }
+        }
```
fleupold, line 164: Nit: the comment below is outdated now. We could still unconditionally remove the order if you prefer.

driver/src/orderbook/streamed/state.rs
```diff
@@ -84,97 +85,204 @@ impl State {
...
+            .ok_or(Error::UnknownToken(order.sell_token))?;
+        let buy_token = self
+            .tokens
+            .get_address_by_id(order.buy_token)
+            .ok_or(Error::UnknownToken(order.buy_token))?;
+
+        // This looks awkward because we make sure that no modifications take place if an error
+        // occurs.
+
+        let mut new_order = *order;
```
fleupold, line 186: Don't we have to copy the order as well?

driver/src/orderbook/streamed/state.rs
```diff
@@ -84,97 +85,204 @@ impl State {
...
+            .get_address_by_id(order.sell_token)
+            .ok_or(Error::UnknownToken(order.sell_token))?;
+        let buy_token = self
+            .tokens
+            .get_address_by_id(order.buy_token)
+            .ok_or(Error::UnknownToken(order.buy_token))?;
+
+        let mut new_order = *order;
+        let sell_balance_key = (event.owner, sell_token);
+        let mut sell_balance = *self
```
fleupold, line 231: Why no copy here?

driver/src/orderbook/streamed/state.rs
```diff
@@ -84,97 +85,204 @@ impl State {
...
+            .tokens
+            .get_address_by_id(0)
+            .ok_or(Error::SolutionWithoutFeeToken)?;
+        match self.last_solution.batch_id.cmp(&block_batch_id) {
+            Ordering::Less => {
+                let balance = self
+                    .balances
+                    .entry((self.last_solution.submitter, fee_token))
+                    .or_default();
+                balance.add_balance(self.last_solution.burnt_fees)?;
```
fleupold, line 267: Wouldn't this technically have to happen at potentially every event? Otherwise the solution submitter would not be able to claim their fee until another solution has come in (where in reality they could withdraw as soon as in the next batch even if that one isn't solved).

driver/src/orderbook/streamed/state.rs
```diff
@@ -84,97 +85,204 @@ impl State {
...
+            Ordering::Equal => (),
+            Ordering::Greater => return Err(Error::SolutionForPastBatch),
+        }
+
+        // Make sure that the submitter gets an entry an balances so their balance shows up when
+        // place an order for the fee token. Otherwise this would not happen because we only create
+        // balances on trades.
+        self.balances
+            .entry((event.submitter, fee_token))
+            .or_default();
```
fleupold, line 278: I don't understand this reasoning. What's the advantage of having a default entry in the balance list? Wouldn't we be able to lazily create it on read?

fleupold (changes requested):
This PR was extremely difficult for me to review. The size of the change paired with the amount of crucial logic is just a bit too much for me. I'd really appreciate if we can continue the discussion on smaller parts of this logic.

Maybe we can break this PR into three parts: the balance logic, the refactored order logic (I really like the logic you pulled into the Orders class) and how everything gets integrated into the code that we already have. Would that make sense?

e00E:
> Maybe we can break this PR into three parts: the balance logic, the refactored order logic (I really like the logic you pulled into the Orders class) and how everything gets integrated into the code that we already have. Would that make sense?

Okay. I'll make this PR contain only the balance logic then.

PR #680, opened 2020-04-04
Open solver integration part1

josojo:
This PR integrates the open-solver. 

While building the docker images, there will be choice between building on the closed-source standard-solver or the open-source standard solver. Depending on the choice different images will be created: either the `stablex-binary-public` or `stablex-binary-private`. These new images, we can then deploy with kubernetes, as a second PR will show.

Please note that this process of creating two driver dockers takes much more time and hence Travis is significantly slower. But the parallelization of these builds is out of the scope of this PR.

testplan: travis only

-- 

I am not particularly proud of all these new config files. Maybe someone has a suggestion for solving it smarter.


docker-compose.open-solver.yml
```diff
@@ -0,0 +1,4 @@
+version: "3.6"
+services:
+  stablex:
+    env_file: common-open-solver.env
```
fleupold, line 4: I think this file should also specify the image. This way you shouldn't need to differentiate between docker-compose.binary.public/private.yml but can instead keep a single docker-compose.binary.yml .

Also does exposing the `SOLVER_TYPE` you need inside `environment` not work? This way you wouldn't need the new docker-compose.binary.private.yml file

docker/rust/Dockerfile
```diff
@@ -1,5 +1,5 @@
-# With this image, only the naive solver will work
-ARG SOLVER_BASE=ubuntu:bionic
+# With this image, only the naive and open solver will work
+ARG SOLVER_BASE=gnosispm/dex-open-solver:v0.0.9-dev
```
fleupold, line 2: I would slightly prefer to keep the original base image. This would also make the commands of public/private solver more similar as both would require the SOLVER_BASE argument.

docker/rust/release/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM stablex-binary
+FROM stablex-binary-private
```
fleupold, line 1: Where does this change come from?

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -36,16 +37,50 @@ impl FromStr for SolverType {
...
-            SolverType::StandardSolver => vec![],
-            SolverType::FallbackSolver => vec![String::from("--useExternalPrices")],
+            SolverType::OpenSolver => open_solver_args,
+            SolverType::StandardSolver => standard_solver_args,
+            SolverType::FallbackSolver => {
+                standard_solver_args.extend(vec![String::from("--useExternalPrices")]);
+                standard_solver_args
+            }
+            SolverType::NaiveSolver => {
+                panic!("OptimizationSolver should not be called with naive solver")
```
fleupold, line 75: nit: this string is/was already off (here and below).

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -281,16 +281,13 @@ impl Io for DefaultIo {
...
         time_limit: Duration,
     ) -> Result<()> {
         let time_limit = (time_limit.as_secs_f64().round() as u64).to_string();
         let output = Command::new("python")
-            .args(&["-m", "batchauctions.scripts.e2e._run"])
-            .arg(input_file)
-            .args(&["--outputDir", result_folder])
-            .args(&["--solverTimeLimit", &time_limit])
-            .args(solver_type.to_args())
+            .current_dir(solver.dir())
```
fleupold, line 289: Is there any advantage of making the caller have to change to the directory before? 

It introduces some implicit coupling between the methods (current_dir needs to be invoked together with args). If there is, I'd prefer if the solver could return the entire Command argument but keep that logic internal to the solver type.

driver/src/price_finding/mod.rs
```diff
@@ -19,7 +19,7 @@ pub fn create_price_finder(
     } else {
         info!(
             "Using optimization price finder with the args {:?}",
-            solver_type.to_args()
+            solver_type.to_args("RESULT_FOLDER", "INPUT_FOLDER", String::from("TIME_LIMIT"))
```
fleupold, line 22: Not a big fan of these dummy placeholders. Can we expose the real values here or otherwise make the params optional?

fleupold (commented):
Can we try to get rid off `common-open-solver.env` and merge docker-compose.binary.private/public.yml back into docker-compose.binary.yml?

.travis.yml
```diff
@@ -34,13 +34,21 @@ matrix:
...
         # Build and publish compact image with compiled binary
-        - docker build --tag stablex-binary --build-arg SOLVER_BASE=163030813197.dkr.ecr.us-east-1.amazonaws.com/dex-solver:v0.6.0 --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
-        # StableX e2e Tests (Ganache)
-        - docker-compose -f docker-compose.yml -f docker-compose.binary.yml up -d stablex
+        - docker build --tag stablex-binary-public --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
+        # StableX e2e Tests (Ganache) - open solver
+        - docker-compose -f docker-compose.yml -f docker-compose.open-solver.yml -f docker-compose.binary.public.yml  up -d stablex
         - cargo test -p e2e ganache -- --nocapture
         - docker-compose logs
+        # StableX e2e Tests (Ganache) - naive solver
```
fleupold, line 42: Do we really want to keep the naive solver test? Not sure this is required.

fleupold (commented):
It would be nice to state the impact on build time in the summary (to assess the impact). As far as I can tell it's about 5 minutes extra time.

I noticed that we are running [this step](https://github.com/gnosis/dex-services/blob/c8ed7f607653218b7477de24449de441b1f4781e/docker/rust/Dockerfile#L30-L33) twice although we probably wouldn't have to (e.g. if we built the binary once and then just copied it into an image that combines the solver base and the binary, but maybe that's out of scope for now).

docker/rust/release/Dockerfile
josojo: In travis, I renamed the docker builds to stablex-binary-private and stablex-binary-public, depending on the solver

.travis.yml
josojo: I would say we either keep the test or cut out the naive solver completely.

driver/src/price_finding/optimization_price_finder.rs
josojo: We have to change the dictionary, as the command:
```
python -m open_solver/src/match
```
would give me the error message that the module open_solver does not exist. It only works for batchauctions dir, as we use the virtual env in the image

docker/rust/Dockerfile
bh2smith, line 2: This is an interesting change. We were using a standard ubuntu image before, but now this is some specialised base image? What are some of the main differences here?

driver/src/price_finding/mod.rs
bh2smith, line 22: Its werd to see `String::from` on only one of these input parameters. Was there a reason for this?

.travis.yml
fleupold: Good point. I think we should keep it as it's the easiest way to run the entire system without any downloading any additional image

driver/src/price_finding/optimization_price_finder.rs
fleupold: Why don't we have virtual env for the open solver?

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -36,16 +37,50 @@ impl FromStr for SolverType {
...
             "naive-solver" => Ok(SolverType::NaiveSolver),
+            "open-solver" => Ok(SolverType::OpenSolver),
             _ => Err(anyhow!("solver type does not exit")),
         }
     }
 }
 
 impl SolverType {
-    pub fn to_args(self) -> Vec<String> {
+    pub fn to_args(self, result_folder: &str, input_file: &str, time_limit: String) -> Vec<String> {
```
bh2smith, line 47: I see here that two of the parameters are `&str` and the other is `String`. Looks a bit strange. Could it be avoided?

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -36,16 +37,50 @@ impl FromStr for SolverType {
...
+                result_folder.to_owned(),
+                "06_solution_int_valid.json",
+            ),
+            String::from("best-token-pair"),
+        ];
         match self {
-            SolverType::StandardSolver => vec![],
-            SolverType::FallbackSolver => vec![String::from("--useExternalPrices")],
+            SolverType::OpenSolver => open_solver_args,
+            SolverType::StandardSolver => standard_solver_args,
```
bh2smith, line 69: I would propose that there is no reason to declare these args variables and that they could go directly within the match statement (since they don't appear to be used anywhere else).

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -36,16 +37,50 @@ impl FromStr for SolverType {
...
+        let open_solver_args = vec![
+            String::from("-m"),
+            String::from("src.match"),
+            format!("{}{}", "/app/", input_file.to_owned()),
+            format!(
+                "--solution={}{}{}",
+                "/app/",
+                result_folder.to_owned(),
+                "06_solution_int_valid.json",
+            ),
```
bh2smith, line 64: Slightly off topic, but where is the source code for the open solver located?

bh2smith (approved):
Nice work! Only a few stylistic suggestions.

driver/src/price_finding/price_finder_interface.rs
josojo: in the dir: /app/open_solver/ and then in the module src.match

driver/src/price_finding/price_finder_interface.rs
josojo: I want to use the variable `standard_solver_args`. I use it in the next line

bh2smith: Ok, if it is used twice, but what about the `open_solver_args`?

driver/src/price_finding/price_finder_interface.rs
bh2smith: I don't see any `/app` directory. Is it in a different project somewhere? Is the repo public?

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -36,16 +37,50 @@ impl FromStr for SolverType {
...
+            String::from("best-token-pair"),
+        ];
         match self {
-            SolverType::StandardSolver => vec![],
-            SolverType::FallbackSolver => vec![String::from("--useExternalPrices")],
+            SolverType::OpenSolver => open_solver_args,
+            SolverType::StandardSolver => standard_solver_args,
+            SolverType::FallbackSolver => {
+                standard_solver_args.extend(vec![String::from("--useExternalPrices")]);
```
nlordell, line 71: nit: You can use `vec.push(String::from(...))` here.

driver/src/price_finding/price_finder_interface.rs
josojo: https://github.com/gnosis/dex-open-solver

check out the docker file and you will see the app/

docker/rust/Dockerfile
josojo: The new image contains the open-solver

driver/src/price_finding/optimization_price_finder.rs
josojo: The switch in the Dir has the additional advantage that we don't have to import our python modules with complicated try statements as we currently do in the dex-solver: https://github.com/gnosis/dex-solver/blob/master/scripts/e2e/B_convert_input_int2float.py#L27

The virtual env would not solve this.

fleupold: Ok, what do you think about instead of exposing current_dir implement std::convert::Into<Command> on solver type so that this can stay an implementation detail of the solving strategy? Not a big deal if not but I think it would be cleaner to not make the caller deal with the ordering of the calls.

docker-compose.open-solver.yml
```diff
@@ -0,0 +1,13 @@
...
+    command: ./stablex
+    build:
+      context: .
+      dockerfile: docker/rust/Dockerfile
+      args:
+        - RUST_BASE=rust-binary
+    image: stablex-binary-public
+    environment:
+      - SOLVER_TYPE=open-solver
+      - RUST_BACKTRACE=1
```
fleupold, line 13: Can we not keep the parts that are common between the private and the open solver inside docker-compose.binary and add two configs that only specify the image and SOVLER_TYPE env variable? This way we would be duplicating less code.

driver/src/price_finding/mod.rs
josojo: I delete them here, as we will get better printing in this PR: #685. It makes more sense to print them later, especially, as the time-limit is variable

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -36,20 +40,51 @@ impl FromStr for SolverType {
...
+        let solver_command = match self {
+            SolverType::OpenSolver => command_open_solver,
+            SolverType::StandardSolver => command_standard_solver,
+            SolverType::FallbackSolver => {
+                command_standard_solver.arg(String::from("--useExternalPrices"));
+                command_standard_solver
+            }
             SolverType::NaiveSolver => {
-                panic!("OptimizationSolver should not be called with naive solver")
+                panic!("fn to_args should not be called by the naive solver")
```
fleupold, line 83: Already outdated again 😉

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -286,24 +285,11 @@ impl Io for DefaultIo {
...
-                .args(&["--solverTimeLimit", &time_limit])
-                .args(solver_type.to_args());
-
-            debug!("Using solver command `{:?}`", command);
-            command
-        };
-
-        let output = command.output()?;
+        let time_limit = (time_limit.as_secs_f64().round() as u64).to_string();
+        let output = solver.execute(result_folder, input_file, time_limit)?;
```
fleupold, line 292: Thanks. I like this much better.

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -36,20 +40,51 @@ impl FromStr for SolverType {
...
         }
     }
 }
 
 impl SolverType {
-    pub fn to_args(self) -> Vec<String> {
-        match self {
-            SolverType::StandardSolver => vec![],
-            SolverType::FallbackSolver => vec![String::from("--useExternalPrices")],
+    pub fn execute(
```
fleupold, line 50: Not important, as it's pretty internal but I find this method a but messy (>30 lines and some variables which are only necessary in some cases make this hard to follow).

Maybe we can have private helpers `execute_open_solver` and `execute_private_solver(use_external_prices)` and then only one match statement in this method that calls the helpers or panics.

docker/rust/release/Dockerfile
fleupold, line 1: Should we maybe use public as the default here to avoid accidentally exposure of the private one?

fleupold (approved):

docker/rust/release/Dockerfile
josojo: This PR intended to not change this logic, but just to keep it as before. The logic will be touched in https://github.com/gnosis/dex-services/pull/681

PR #681, opened 2020-04-04
Open solver integration part2 - deploy public image

josojo:
This PR contains the deployment for the public image.

testplan:
Please double check that I would not publish any private code to dockerhub.


docker/rust/release/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM stablex-binary-private
+ARG BINARY_BASE=stablex-binary-public
+
+FROM ${BINARY_BASE}
```
fleupold, line 3: Given this file is so small, maybe it makes sense to have two of them (public/private) and also export some env vars (e.g. SOLVER_TYPE should be set to open_solver in the public image).

fleupold (approved):
> Please double check that I would not publish any private code to dockerhub.

I think you are not but it's getting a bit hard to reason about that. As soon as this is merge please download the first pushed image and verify it doesn't have solver code.

bh2smith (approved):
seems legit.

docker/rust/release/Dockerfile
josojo: >SOLVER_TYPE should be set to open_solver in the public image

I could do that. But wouldn't it feel strange, that we are setting the solvertype for the one image,  but not for the other image? ( We can't know whether we have the standard-solver or fallback-solver?) Or should we always assume the standard-solver and overwrite this variable just for the fallback solver?

fleupold: I think that would be good. The less manual configuration one has to do the easier it will be to run this for external people.

nlordell (approved):

PR #685, opened 2020-04-06
Improve Solver Debugging

### LINKED ISSUE #676: Better Logging (e.g. file not found error)
We are conveniently using `?` to unwrap expectations in places that seem obvious, but could benefit from a bit more information in places such as the following:

https://github.com/gnosis/dex-services/blob/539efcfd0fecd55fd05430e0ce497d4e2b54bbee/driver/src/price_finding/optimization_price_finder.rs#L306-L313

Reason for this particular example stems from a recent alert bug that took a bit longer to investigate because of the missing attempted file name in the log.

```
mainnet-standard-solver  (occurred in both prod and dev)
    Batch 5285955 failed with unretryable error: 
        No such file or directory (os error 2)
```

If we had known what file was missing, it may have directed us to the problem a bit earlier!

nlordell:
This PR adds additional context to the optimization solver errors and adds a log message to record the exact commandline that was used when invoking the solver.

Closes #676 

### Test Plan

CI

### DIFF
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -2,7 +2,7 @@ use crate::models::{self, TokenId, TokenInfo};
 use crate::price_estimation::PriceEstimating;
 use crate::price_finding::price_finder_interface::{Fee, PriceFinding, SolverType};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use chrono::Utc;
 use log::{debug, error};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -284,14 +289,21 @@ impl Io for DefaultIo {
...
+                .arg(input_file)
+                .args(&["--outputDir", result_folder])
+                .args(&["--solverTimeLimit", &time_limit])
+                .args(solver_type.to_args());
+
+            debug!("Using solver command `{:?}`", command);
+            command
+        };
+
+        let output = command.output()?;
```
e00E, line 306: Should this get a context too?

e00E (approved):

bh2smith (approved):
Cool!

driver/src/price_finding/optimization_price_finder.rs
nlordell: The `IO::run_solver` has error context already, and if the exit status is non-zero there is an `error!` log message.

josojo:
Both, the solver type and the command are printed during the initialization of the "main" and the initialization of  `priceFinder`

But I agree that it might be more useful to have it printed every time, so that we don't have to search the beginning.

Maybe you wanna erase the printing at the beginning during the price finder initialization.

nlordell:
> Both, the solver type and the command are printed during the initialization of the "main" and the initialization of priceFinder

The `debug` statement prints the exact command line invocation including parameters like the input file name, output directory, serialized time limit. These change per run of the solver.

fleupold (approved):

PR #686, opened 2020-04-06
Vendor tini

### LINKED ISSUE #648: Cache or vendor `tini`
We should consider caching or vendoring `tini` as we got an error like this:
image

e00E:
fixes #648

It seems like my editor also automatically removed a few trailing whitespaces. Imo we can just keep that in the commit.


docker/rust/Dockerfile
```diff
@@ -38,7 +38,10 @@ ONBUILD COPY target/debug/driver stablex
 FROM ${RUST_BASE}
 
 # Add Tini
-ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+# We ran into github rate limiting using this url so we now keep a local copy of
+# tini instead.
+# ENV TINI_VERSION v0.18.0
+# ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+Add docker/rust/tini /tini
```
fleupold, line 45: ubernit: ADD instead of Add

docker/rust/Dockerfile
```diff
@@ -38,7 +38,10 @@ ONBUILD COPY target/debug/driver stablex
 FROM ${RUST_BASE}
 
 # Add Tini
-ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+# We ran into github rate limiting using this url so we now keep a local copy of
+# tini instead.
+# ENV TINI_VERSION v0.18.0
+# ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
```
fleupold, line 44: Not sure if it's worth keeping the old command around, but doesn't matter much.

fleupold (approved):

bh2smith (approved):

docker/rust/Dockerfile
bh2smith: I think its nice to know where it came from.

nlordell: I would rather add this information to `docker/rust/README.md`

nlordell (approved):

docker/rust/Dockerfile
e00E: I prefer it in this place because it right at the position where it is relevant so I'll leave it as is for now.

PR #687, opened 2020-04-06
Wait for solved batch to accept solutions

nlordell:
This PR implements polling the exchange's current batch ID before submitting the solution to make sure that the solved batch is already accepting solutions.

This is done as querying the objective value on the `pending` block is not alwasy enough, as the pending block might still have a timestamp earlier than the next block.

### Test Plan

New unit test.


driver/src/solution_submission/mod.rs
```diff
@@ -92,8 +103,15 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
         batch_index: U256,
         solution: Solution,
     ) -> Result<U256, SolutionSubmissionError> {
+        // NOTE: Compare with `>=` as the exchange's current batch index is the
+        //   one accepting orders and does not yet accept solutions.
+        while batch_index.as_u32() >= self.contract.get_current_auction_index()? {
```
fleupold, line 108: Why don't we have to `SolutionSubmissionError::from` here, but below?

driver/src/solution_submission/mod.rs
```diff
@@ -92,8 +103,15 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+            debug!("Solved batch is not yet accepting solutions, waiting for next batch.");
```
fleupold, line 109: ubernit: I'd suggest info-ing this

driver/src/solution_submission/mod.rs
```diff
@@ -92,8 +103,15 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+            thread::sleep(POLL_TIMEOUT);
```
fleupold, line 110: 😢 Yet another thread sleep in the middle of the code.

fleupold (approved):
Looks good, thanks for the unit-test!

driver/src/solution_submission/mod.rs
nlordell: `?` implicitly calls `From::from` for error conversion. The line below is for mapping the `Result<_, E>` to `Result<_, SolutionSumissionError>` - it could be re-written (and is equivalent to):
```
Ok(self.contract.get_solution_objective_value(batch_index, solution, None)?)
```

driver/src/solution_submission/mod.rs
nlordell: I know :sob: I couldn't think of a nicer way without making a larger change.

driver/src/solution_submission/mod.rs
fleupold: Thanks!

👨‍🏫

e00E (approved):

driver/src/solution_submission/mod.rs
nlordell: @fleupold I just had an idea. We can make `"Solutions are no longer accepted for this batch"` a benign error if we are in the previous block and have the scheduler re-run the driver.

nlordell: The downside here is that it will re-fetch the orderbook and look for a solution again, but I don't think this should be a big issue.

fleupold: That's a nice idea. We briefly discussed it over slack. It would require fetching the last block before verifying and then return a benign retryable error.

I'm somewhat concerned about running the solver again unnecessarily, but if we run into this it's already a sign that we were finishing too early (however this might become the norm if we start solving before the batch is finalized). We could also introduce a caching layer for the price finder that can quickly return the same result if the input files are the same.

bh2smith:
Looks like this could be merged anytime after the conflicts are resolved? Is there anything special we are waiting on here?

nlordell:
> Looks like this could be merged anytime after the conflicts are resolved? Is there anything special we are waiting on here?

The discussion here: https://github.com/gnosis/dex-services/pull/687#discussion_r404188080

driver/src/solution_submission/mod.rs
nlordell: > however this might become the norm if we start solving before the batch is finalized

I imagine at that point, the scheduler would be responsible for starting the solution submitter.

> We could also introduce a caching layer for the price finder that can quickly return the same result if the input files are the same.

That would be useful, especially when facing retryable errors.

fleupold: Overall I think the other approach might be a bit more work to implement, so I'd be in favor of merging it as is to reduce the amount of false warnings we will have next week and then look into improving this.

nlordell: Created #707 to capture this work.

PR #690, opened 2020-04-06
Added price graph crate with benchmarking

nlordell:
This PR adds a the price graph crate that for now just implements orderbook decoding and includes a benchmark test.

This is the first step to doing using graph pathfinding algorithms to find price estimates.

### Test Plan

Checkout locally and run
```
cargo bench -p pricegraph
```


fleupold (approved):
Cool, I can see how it needs to be very low level in order to compile down to WASM eventually.

e00E (approved):

PR #698, opened 2020-04-07
[Pricegraph] Build orderbook from encoded orders

nlordell:
This PR implements building a graph representation of an orderbook from encoded orders and computing whether or not the orderbook is overlapping by looking for negative cycles in the graph.

### Test Plan

Unit tests and new benchmark for `is_overlapping()`, which competes in:
```
Orderbook::is_overlapping
                        time:   [4.1024 us 4.1123 us 4.1230 us]
```

### DIFF
--- a/pricegraph/src/encoding.rs
+++ b/pricegraph/src/encoding.rs
@@ -1,21 +1,65 @@
 //! This module implements decoding for the standard `BatchExchange` contract
 //! encoded orders.
 
-use crate::{TokenId, UserId};
-use primitive_types::U256;
+use primitive_types::{H160, U256};
 
+/// The stride of an orderbook element in bytes.
 pub const ELEMENT_STRIDE: usize = 112;
 
+/// A type alias for a batch ID.
 pub type BatchId = u32;
 
+/// A type alias for a token ID.
+pub type TokenId = u16;
+
+/// A type alias for a user ID.
+pub type UserId = H160;
+
+/// A struct representing a buy/sell token pair.
+#[cfg_attr(test, derive(Debug, PartialEq))]
+pub struct TokenPair {
+    /// The buy token.
+    pub buy: TokenId,
+    /// The sell token.
+    pub sell: TokenId,
+}
+
+/// A struct representing the validity of an order.
+#[cfg_attr(test, derive(Debug, PartialEq))]
+pub struct Validity {
+    /// The batch starting from which the order is valid.
+    pub from: BatchId,
+    /// The last batch the order is valid for.
+    pub to: BatchId,
+}
+
+/// A price expressed as a fraction of buy and sell amounts.
+#[cfg_attr(test, derive(Debug, PartialEq))]
+pub struct Price {
+    /// The price numerator, or the buy amount.
+    pub numerator: u128,
+    /// The price denominator, or the sell amount.
+    pub denominator: u128,
+}
+
+/// An orderbook element that is retrieved from the smart contract.
 #[cfg_attr(test, derive(Debug, PartialEq))]
 pub struct Element {
+    /// The user that placed the order.
     pub user: UserId,
+    /// The user's sell token balance.
     pub balance: U256,
-    pub pair: (TokenId, TokenId),
-    pub valid: (BatchId, BatchId),
-    pub price: (u128, u128),
-    pub amount: u128,
+    /// The token pair for which this order was placed.
+    pub pair: TokenPair,
+    /// The validity of the order.
+    pub valid: Validity,
+    /// The price fraction for the order.
+    ///
+    /// Note that if one of the prices is `UNBOUNDED_AMOUNT` then that signals
+    /// that the order is an unbounded order and has no amount limit.
+    pub price: Price,
+    /// The remaining sell amount available to this order.
+    pub remaining_sell_amount: u128,
 }
 
 impl Element {
@@ -38,6 +82,9 @@ impl Element {
                 (U256) => {
                     U256::from_big_endian(&read!(32))
                 };
+                (H160) => {
+                    H160(read!(20))
+                };
                 ($n:expr) => {{
                     let mut buf = [0u8; $n];
                     buf.copy_from_slice(&chunk[..$n]);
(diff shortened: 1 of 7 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -0,0 +1,207 @@
...
+
+impl Order {
+    /// Compare two orders by descending price order.
+    ///
+    /// This method is used for sorting orders, instead of just sorting by key
+    /// on `price` field because `f64`s don't implement `Ord` and as such cannot
+    /// be used for sorting. This be because there is no real ordering for
+    /// `NaN`s and `NaN < 0 == false` and `NaN >= 0 == false` (cf. IEEE 754-2008
+    /// section 5.11), which can cause serious problems with sorting.
+    fn cmp_decending_prices(a: &Order, b: &Order) -> cmp::Ordering {
```
fleupold, line 147: typo: descending

pricegraph/src/orderbook.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        let mut projection = DiGraph::new();
+        let token_nodes = (0..=max_token)
+            .map(|token_id| projection.add_node(token_id))
+            .collect::<Vec<_>>();
+        projection.extend_with_edges(orders.iter().flat_map(|(sell, orders)| {
+            let sell_node = token_nodes[*sell as usize];
+            orders.iter().map({
+                let token_nodes = &token_nodes;
```
fleupold, line 59: What exactly is this called? I didn't know you can do assignments inside the block but before the function definition starts?

pricegraph/src/orderbook.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// including fees of the sell token expressed in the buy token.
+    ///
+    /// Specifically, this is the sell amount over the buy amount or price
+    /// denominator over price numerator, which is the inverse price as
+    /// expressed by the exchange.
+    pub price: f64,
+    /// The weight of the order in the graph. This is the base-2 logarithm of
+    /// the price including fees. This enables transitive prices to be computed
+    /// using addition.
+    pub weight: f64,
```
fleupold, line 136: Should we just compute this on the fly when needed instead of storing a derivative field?

fleupold (approved):
This is pretty awesome! Even accounts for the fee, which I was not expecting.

pricegraph/src/orderbook.rs
bh2smith: Good question. I suppose it depends on how many times these values will be accessed relative to the complexity of evaluation. I tried looking for the complexity here https://en.wikipedia.org/wiki/Binary_logarithm#Computational_complexity 

but it doesn't appear to be explicitly stated.

Intuitively, I would say, if we are going to save on computation using addition here (as opposed to multiplication) then we might find the combined complexity of computing the log followed by addition at each access might run us into the same complexity of simply multiplying in the first place. With this in mind, I might be inclined to think that we only gain any advantage by computing the log once and storing it here to be used in future operations. Same argument would also imply that if we only access this stored value once then we haven't really gained anything.

I could get more precise about this intuition if necessary, but first I think we should determine approximately how often these values will be accessed.

bh2smith (approved):
Very nice stuff here.

pricegraph/src/orderbook.rs
nlordell: I just create a scope to help the compiler figure out lifetimes - the `token_nodes` borrow only lasts for the scope I defined.

pricegraph/src/orderbook.rs
nlordell: > I could get more precise about this intuition if necessary, but first I think we should determine approximately how often these values will be accessed.

I have benchmarks, I think the most reliable way to make optimizations is to benchmark against actual orderbooks. I'll make computed on the fly for now, and optimize later.

pricegraph/src/encoding.rs
```diff
@@ -1,21 +1,62 @@
 //! This module implements decoding for the standard `BatchExchange` contract
 //! encoded orders.
 
-use crate::{TokenId, UserId};
-use primitive_types::U256;
+use primitive_types::{H160, U256};
 
+/// The stride of an orderbook element in bytes.
```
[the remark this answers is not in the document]
nlordell: Its a technical memory term. Its used for when you have a collection of some data laid out in memory. The length of the data is the size of all the structs sequentially laid out one after the other, the stride is the size of a single struct including padding.

pricegraph/src/encoding.rs
```diff
@@ -48,12 +92,21 @@ impl Element {
...
-                user: read!(20),
+                user: read!(H160),
                 balance: read!(U256),
-                pair: (read!(u16), read!(u16)),
-                valid: (read!(u32), read!(u32)),
-                price: (read!(u128), read!(u128)),
-                amount: read!(u128),
+                pair: TokenPair {
+                    buy: read!(u16),
+                    sell: read!(u16),
```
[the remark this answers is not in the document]
nlordell: I define it a few lines up :)

pricegraph/src/orderbook.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// The effective sell price for this order, that is price for this order
+    /// including fees of the sell token expressed in the buy token.
+    ///
+    /// Specifically, this is the sell amount over the buy amount or price
+    /// denominator over price numerator, which is the inverse price as
+    /// expressed by the exchange.
+    pub price: f64,
+    /// The weight of the order in the graph. This is the base-2 logarithm of
+    /// the price including fees. This enables transitive prices to be computed
+    /// using addition.
```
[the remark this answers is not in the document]
nlordell: @fleupold 's idea!

pricegraph/src/orderbook.rs
nlordell: I think the term is a block expression.

nlordell:
This PR implements building a graph representation of an orderbook from encoded orders and computing whether or not the orderbook is overlapping by looking for negative cycles in the graph.

### Test Plan

Unit tests and new benchmark for `is_overlapping()`, which competes in:
```
Orderbook::is_overlapping
                        time:   [4.1024 us 4.1123 us 4.1230 us]
```

PR #700, opened 2020-04-08
[Easy] New solver config: standard solver should also use external prices

josojo:
In the solver-team, we realized that the solutions found under consideration of the external prices are almost always better than the other solutions.

Hence, we decided to use external price feeds always. (cp: https://gnosisinc.slack.com/archives/CS6NV17HP/p1585220425012100)

The main difference between the standard and the fallback solver will be the whitelist of tokens, which I will configure in gnosis-staging.

### DIFF
--- a/driver/src/price_finding/price_finder_interface.rs
+++ b/driver/src/price_finding/price_finder_interface.rs
@@ -54,55 +54,55 @@ impl SolverType {
         time_limit: String,
     ) -> Result<Output> {
         match self {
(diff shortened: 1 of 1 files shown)


fleupold (approved):
nit: execute_private_solver/execute_open_solver don't need to be implemented on self.

Don't forget adding the token information for the standard solver as well, as we currently only fetch prices for specifically listed tokens.

e00E (approved):

PR #702, opened 2020-04-08
[Pricegraph] Find negative cycle

nlordell:
This PR introduces a modified Bellman-Ford implementation that returns the path of the negative cycle that if finds.

### Test Plan

New unit test, `cargo bench` and see performance not heavily affected.

### DIFF
--- a/pricegraph/src/graph.rs
+++ b/pricegraph/src/graph.rs
@@ -1,0 +1,5 @@
+//! This module contains graph related data types and algorithms that are either
+//! not included in the `petgraph` crate or are modified versions of the crate's
+//! implementation.
+
+pub mod bellman_ford;
--- a/pricegraph/src/graph/bellman_ford.rs
+++ b/pricegraph/src/graph/bellman_ford.rs
@@ -61,7 +61,7 @@ where
             let w = *edge.weight();
             if distance[ix(i)] + w < distance[ix(j)] {
                 // NOTE: The following 4 lines are the only modifications made
-                // to the original algorithm and where originally:
+                // to the original algorithm and were originally:
                 // ```diff
                 // -//println!("neg cycle, detected from {} to {}, weight={}", i, j, w);
                 // -return Err(NegativeCycle(()));
@@ -98,12 +98,11 @@ where
     let mut current = start;
     visited[ix(current)] = true;
     loop {
-        let next = predecessor[ix(current)]?;
-        if visited[ix(next)] {
+        current = predecessor[ix(current)]?;
+        if visited[ix(current)] {
             break;
         }
-        visited[ix(next)] = true;
-        current = next;
+        visited[ix(current)] = true;
     }
 
     // NOTE: `current` is now guaranteed to be on the cycle, so just walk
@@ -144,6 +143,6 @@ pub mod tests {
 
         let NegativeCycle(path) = search(&graph, 0.into()).unwrap_err();
 
-        assert_eq!(path, &[2.into(), 3.into(), 1.into()]);
+        assert_eq!(path, &[1.into(), 2.into(), 3.into()]);
     }
 }
--- a/pricegraph/src/graph/bellman_ford.rs
+++ b/pricegraph/src/graph/bellman_ford.rs
@@ -1,0 +1,150 @@
+//! The module contains a slightly modified version of the `petgraph`
+//! implementation of the Bellman-Ford graph search algorigthm that returns the
+//! detected negative cycle on error.
+
+use petgraph::algo::FloatMeasure;
+use petgraph::visit::{
+    Data, EdgeRef, GraphBase, IntoEdges, IntoNodeIdentifiers, NodeCount, NodeIndexable,
+};
+
+/// A type definition for the result of a Bellman-Ford shortest path search
+/// containing the weight of the shortest paths and the predecessor vector for
+/// the shortests paths.
+pub type Paths<G> = (
+    Vec<<G as Data>::EdgeWeight>,
+    Vec<Option<<G as GraphBase>::NodeId>>,
+);
+
+/// A negative cycle error with a path representing the cycle.
+pub struct NegativeCycle<G: Data>(pub Vec<G::NodeId>);
+
+/// This implementation is taken from the `petgraph` crate with a small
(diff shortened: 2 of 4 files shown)


pricegraph/src/graph/bellman_ford.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    // check for negative weight cycle
+    for i in g.node_identifiers() {
+        for edge in g.edges(i) {
+            let j = edge.target();
+            let w = *edge.weight();
+            if distance[ix(i)] + w < distance[ix(j)] {
+                // NOTE: The following 4 lines are the only modifications made
+                // to the original algorithm and where originally:
```
fleupold, line 64: nit typo: were

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -0,0 +1,149 @@
...
+/// A type definition for the result of a Bellman-Ford shortest path search
+/// containing the weight of the shortest paths and the predecessor vector for
+/// the shortests paths.
+pub type Paths<G> = (
+    Vec<<G as Data>::EdgeWeight>,
+    Vec<Option<<G as GraphBase>::NodeId>>,
+);
+
+/// A negative cycle error with a path representing the cycle.
+pub struct NegativeCycle<G: Data>(pub Vec<G::NodeId>);
```
fleupold, line 19: Would it make sense to also expose the weight as it relates to the surplus that can be extracted from the cycle?

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -0,0 +1,149 @@
...
+    let mut visited = vec![false; g.node_bound()];
+    let mut current = start;
+    visited[ix(current)] = true;
+    loop {
+        let next = predecessor[ix(current)]?;
+        if visited[ix(next)] {
+            break;
+        }
+        visited[ix(next)] = true;
+        current = next;
```
fleupold, line 106: nit: I think only one of current/next should be needed as in the loop below.

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -0,0 +1,149 @@
...
+    g: G,
+    start: G::NodeId,
+    predecessor: &[Option<G::NodeId>],
+) -> Option<Vec<G::NodeId>>
+where
+    G: NodeCount + NodeIndexable,
+{
+    let ix = |i| g.to_index(i);
+
+    // NOTE: First find a node that is actually on the cycle, this is done
```
fleupold, line 94: But are we guaranteed to find it by walking backwards? This is what we doing, right?

fleupold (approved):
🌀

pricegraph/src/graph.rs
```diff
@@ -0,0 +1,5 @@
+//! This module contains graph related data types and algorithms that are either
+//! not included in the `petgraph` crate or are modified versions of the crate's
+//! implementation.
+
+pub mod bellman_ford;
```
bh2smith, line 5: I understand the Bellmann-Ford is very similar Dijkstra for the same problem but can handle negative values in edges weights. I also see a lot of reference to negative cycles which can only be the result of having negative edges. Does this mean our order graph has negative edges? I can't think of any edge weights in an orderbook that would be negative.

If there are no negative edges, it is my understanding that Dijkstra's algorithm would be the superior choice.

pricegraph/src/graph/bellman_ford.rs
bh2smith: Nice find.

bh2smith (approved):
Dude. This is killer! I'd still like to see an explanation for these negative edges. Oh no... do the negative edges arise form the taking the log of the prices? If this is the reason, we might want to bite the bullet, not take logs and use Dijkstra...

pricegraph/src/graph/bellman_ford.rs
nlordell: :see_no_evil:

pricegraph/src/graph/bellman_ford.rs
nlordell: :eagle: :eye:

pricegraph/src/graph/bellman_ford.rs
nlordell: We can add it if we end up using it. We actually need to walk to cycle to fill it (since we need to find the capacity of the cycle - i.e. the minimum amount that can be sent around) and for that we use the prices and not the log of the prices to do the calculations.

nlordell: So for now we don't use it and it would add extra complexity to the `find_cycle` so I imagine we can leave it out for now.

pricegraph/src/graph/bellman_ford.rs
nlordell: I think so... It is what we are doing.

nlordell: So Bellman-Ford is guaranteed to complete in |V| - 1 iterations, and if it doesn't that means that there is a negative cycle somewhere. They are detected by finding a node that relaxes on the |V|th iteration where each iteration relaxes all adjacent nodes. This means, intuitvely, that a node relaxed on the |V|th iteration must be connected to the negative cycle or it would not be relaxable (since its calculated distance would already be guaranteed to be the lowest). So by walking backwards you are guaranteed to stumble upon the negative cycle.

The question becomes is it possible that the negative cycle does not appear in the `predecessors` vector? That may be possible. It *might* be that the |V|th iteration must be done to completion so that the negative cycle is in `predecessor`s although my intuition says it doesn't and that just finding one node that relaxes on the |V|th cycle is enough to ensure that it is complete in the `predecessors` vector.

pricegraph/src/graph.rs
nlordell: Negative edges happen because we use the logarithm of the price which can be negative. Dijkstra won't work here, as multiplication with factors < 1 are equivalent to addition with negative edges which does not work with Dijkstra.

It would be nice if we could as in general Dijkstra is faster for dense graphs.

PR #703, opened 2020-04-08
[Pricegraph] User balances

nlordell:
This PR additionally reads user balances from the encoded orderbook and creates a mapping from user ID to token balances

### Test Plan

Builds in CI.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -6,7 +6,7 @@
 //! (i.e. orders) connecting a token pair.
 
 use crate::encoding::{Element, InvalidLength, Price, TokenId, TokenPair, UserId};
-use crate::graph::bellman_ford::{self, NegativeCycle};
+use crate::graph::bellman_ford;
 use petgraph::graph::{DiGraph, NodeIndex};
 use primitive_types::U256;
 use std::cmp;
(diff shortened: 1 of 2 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -185,6 +219,43 @@ fn as_effective_sell_price(price: &Price) -> f64 {
...
+
+impl User {
+    /// Adds an encoded orderbook element to the user data, including the
+    /// reported balance and incrementing the number of orders.
+    fn include_order(&mut self, element: &Element) -> usize {
+        let order_id = self.num_orders;
+
+        self.balances
+            .entry(element.pair.sell)
+            .or_insert_with(|| u256_to_f64(element.balance));
```
fleupold, line 242: Is it worth asserting here that if the element already exists it's equal to `element.balance`?

pricegraph/src/orderbook.rs
```diff
@@ -185,6 +219,43 @@ fn as_effective_sell_price(price: &Price) -> f64 {
...
+
+/// Convert an unsigned 256-bit integer into a `f64`.
+fn u256_to_f64(u: U256) -> f64 {
+    let (u, factor) = match u {
+        U256([_, _, 0, 0]) => (u, 1.0),
+        U256([_, _, _, 0]) => (u >> 64, 2.0f64.powi(64)),
+        U256([_, _, _, _]) => (u >> 128, 2.0f64.powi(128)),
+    };
+    (u.low_u128() as f64) * factor
+}
```
fleupold, line 257: That's a sweet way to do it.

pricegraph/src/orderbook.rs
```diff
@@ -185,6 +219,43 @@ fn as_effective_sell_price(price: &Price) -> f64 {
...
+    /// User balances per token.
+    balances: HashMap<TokenId, f64>,
+    /// The number of orders this user has.
+    num_orders: usize,
+}
+
+impl User {
+    /// Adds an encoded orderbook element to the user data, including the
+    /// reported balance and incrementing the number of orders.
+    fn include_order(&mut self, element: &Element) -> usize {
```
fleupold, line 237: naming nit: I'd find update_balance[_from_order] slightly more descriptive.

fleupold (approved):
🉑

pricegraph/src/orderbook.rs
nlordell: :smile:

pricegraph/src/orderbook.rs
nlordell: It would validate the data a bit which is nice.

pricegraph/src/orderbook.rs
nlordell: My only issue with that name is that it doesn't capture that it also counts the orders to determine the order ID.

pricegraph/src/orderbook.rs
nlordell: So, I looked into bubbling up that error and its fairly disruptive. I'll add it as a separate PR as it would affect the code a bit.

In that case we should also validate other order parameters like making sure that the orders are actually valid and that the the price denominator is non-zero, etc.

nlordell: Created #710

pricegraph/src/orderbook.rs
nlordell: Maybe `add_order` or `visit_order`? I can't think of a good name :(.

fleupold: Your point about counting orderId makes sense, I didn't catch that.

pricegraph/src/orderbook.rs
fleupold: Fine to do later. We might even be able to panic instead of bubbling up some error, as this is really just a sign of inconsistent data.

PR #704, opened 2020-04-08
[Pricegraph] Introduce submodules to orderbook

nlordell:
This PR just moves code around into submodules without actually changing anything. This is because a lot of code will get added soon and the orderbook.rs file will get too big.

### Test Plan

CI


bh2smith (approved):
🌶

fleupold (approved):
🗃️➡️🗄️

PR #709, opened 2020-04-09
[Pricegraph] Implemented updating projection graph edges by removing filled orders

nlordell:
This implements an important sub-step in the algorithm that updates the projection graph's edges to reflect the updated weights once orders get used up.

### Test Plan

New unit test and benchmark.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -8,11 +8,11 @@
 mod order;
 mod user;
 
-use self::order::{Order, OrderMap};
+use self::order::{Order, OrderCollector, OrderMap};
 use self::user::UserMap;
 use crate::encoding::{Element, InvalidLength, TokenId, TokenPair};
 use crate::graph::bellman_ford;
-use petgraph::graph::{DiGraph, NodeIndex};
+use petgraph::graph::{DiGraph, EdgeIndex, NodeIndex};
 use std::cmp;
 use std::f64;
 
@@ -42,7 +42,7 @@ impl Orderbook {
     /// Creates an orderbook from an iterator over decoded auction elements.
     fn from_elements(elements: impl IntoIterator<Item = Element>) -> Orderbook {
         let mut max_token = 0;
-        let mut orders = OrderMap::default();
+        let mut orders = OrderCollector::default();
         let mut users = UserMap::default();
 
         for element in elements {
@@ -56,14 +56,7 @@ impl Orderbook {
             orders.insert_order(Order::new(element, order_id));
         }
 
-        // NOTE: Sort the orders per token pair in descending order, this makes
-        // it so the last order is the cheapest one and can be determined given
-        // a token pair in `O(1)`, and it can simply be `pop`-ed once its amount
-        // is used up and removed from the graph in `O(1)` as well.
-        for (_, pair_orders) in orders.all_pairs_mut() {
-            pair_orders.sort_unstable_by(Order::cmp_descending_prices);
-        }
-
+        let orders = orders.collect();
         let mut projection = DiGraph::new();
         for token_id in 0..=max_token {
             let token_node = projection.add_node(token_id);
@@ -79,8 +72,8 @@ impl Orderbook {
                     .last()
                     .expect("unexpected token pair in orders map without any orders");
                 (
-                    node_index(pair.sell),
                     node_index(pair.buy),
+                    node_index(pair.sell),
                     cheapest_order.weight(),
                 )
             }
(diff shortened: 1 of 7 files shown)


pricegraph/src/orderbook/order.rs
```diff
@@ -29,12 +56,51 @@ impl OrderMap {
...
+    fn all_pairs_mut(&mut self) -> impl Iterator<Item = (TokenPair, &'_ mut Vec<Order>)> + '_ {
         self.0.iter_mut().flat_map(|(&sell, o)| {
             o.iter_mut()
                 .map(move |(&buy, o)| (TokenPair { sell, buy }, o))
         })
     }
+
+    /// Returns the orders for an order pair. Returns `None` if that pair has
+    /// no orders.
+    fn pair_orders(&self, pair: TokenPair) -> Option<&[Order]> {
```
fleupold, line 68: naming nit: since pair is also a verb it confused me in the beginning. Maybe `orders_for_pair` is more clear?

pricegraph/src/orderbook/order.rs
```diff
@@ -29,12 +56,51 @@ impl OrderMap {
...
+    }
+
+    /// Returns a mutable reference to orders for an order pair. Returns `None`
+    /// if that pair has no orders.
+    fn pair_orders_mut(&mut self, pair: TokenPair) -> Option<&mut Vec<Order>> {
+        self.0.get_mut(&pair.sell)?.get_mut(&pair.buy)
+    }
+
+    /// Returns a reference to the cheapest order given an order pair.
+    pub fn pair_order(&self, pair: TokenPair) -> Option<&Order> {
```
fleupold, line 79: nit: Can the selection criteria be hinted at by the name? best_order_for_pair

pricegraph/src/orderbook/order.rs
```diff
@@ -1,15 +1,20 @@
...
 
-/// Type definition for a mapping of orders between buy and sell tokens.
+/// A type for collecting orders and building an order map that garantees that
+/// per-pair orders are sorted for optimal access.
 #[derive(Debug, Default)]
-pub struct OrderMap(HashMap<TokenId, HashMap<TokenId, Vec<Order>>>);
+pub struct OrderCollector(HashMap<TokenId, HashMap<TokenId, Vec<Order>>>);
 
-impl OrderMap {
+impl OrderCollector {
```
fleupold, line 17: it's like an OrderMapBuilder, isn't it? Any reason why you wanted this extra abstraction? So that the actual ordermap can be immutable?

pricegraph/src/orderbook/order.rs
```diff
@@ -113,5 +187,5 @@ fn is_unbounded(element: &Element) -> bool {
 /// Calculates an effective price as a `f64` from a price fraction.
 fn as_effective_sell_price(price: &Price) -> f64 {
     const FEE_FACTOR: f64 = 1.0 / 0.999;
-    FEE_FACTOR * (price.denominator as f64) / (price.numerator as f64)
+    FEE_FACTOR * (price.numerator as f64) / (price.denominator as f64)
```
fleupold, line 190: 🤦 Who is reviewing this code? 😉

pricegraph/src/orderbook.rs
```diff
@@ -110,6 +103,59 @@ impl Orderbook {
         let fee_token = node_index(0);
         bellman_ford::search(&self.projection, fee_token).is_err()
     }
+
+    /// Updates the projection graph for every token pair.
+    pub fn update_projection_graph(&mut self) {
```
fleupold, line 108: This will be called incrementally for each found path/cycle?

fleupold (approved):

pricegraph/src/orderbook/order.rs
nlordell: Its so nothing outside of the ordermap can modify the order of the per-pair orders and break the assumptions that they are always sorted.

pricegraph/src/orderbook/order.rs
nlordell: I only noticed it because I got to a point where I was removing a way too many negative cycles, then I realized that I had the prices the wrong way :facepalm:

pricegraph/src/orderbook.rs
nlordell: No, this is called once when removing negative cycles, then we do it per-edge per path.

pricegraph/src/orderbook/order.rs
nlordell: Its just a type for collecting the orders, and then once they are collected, then the order map can be built and the orders are guaranteed to stay sorted.

pricegraph/src/orderbook.rs
nlordell: It would be really bad if it had to be called at each iteration!

PR #711, opened 2020-04-09
[Pricegraph] Fill Path

nlordell:
This PR implements filling a path with to its maximum capacity, effectively removing it from the graph.

### Test Plan

A carefully crafted unit test.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -12,6 +12,7 @@ use self::order::{Order, OrderCollector, OrderMap};
 use self::user::UserMap;
 use crate::encoding::{Element, InvalidLength, TokenId, TokenPair};
 use crate::graph::bellman_ford;
+use crate::num;
 use petgraph::graph::{DiGraph, EdgeIndex, NodeIndex};
 use std::cmp;
 use std::f64;
@@ -104,7 +105,6 @@ impl Orderbook {
         bellman_ford::search(&self.projection, fee_token).is_err()
     }
 
-<<<<<<< HEAD
     /// Updates the projection graph for every token pair.
     pub fn update_projection_graph(&mut self) {
         let pairs = self
@@ -150,6 +150,53 @@ impl Orderbook {
         self.projection.find_edge(buy, sell)
     }
 
+    /// Fills a trading path through the orderbook to maximum capacity, reducing
+    /// the remaining order amounts and user balances along the way, returning
+    /// the amount of flow that left the first node in the path or `None` if the
+    /// path was invalid.
+    ///
+    /// Note that currently, user buy token balances are not incremented as a
+    /// result of filling orders along a path.
+    pub fn fill_path(&mut self, path: &[TokenId]) -> Option<f64> {
+        let Self {
+            ref mut orders,
+            users,
+            ..
+        } = self;
+
+        // First determine the capacity of the path expressed in the first and
+        // final token of the path.
+        let mut capacity = f64::INFINITY;
+        let mut transient_price = 1.0;
+        for pair in pairs_on_path(path) {
+            let order = orders.pair_order(pair)?;
+            capacity = num::min(
+                capacity,
+                order.get_effective_amount(users) / transient_price,
+            );
+            transient_price *= order.price;
+        }
+
+        // Now that the capacity of the path has been determined, push the flow
+        // around the orderbook deducting from the order amounts and user
+        // balances.
+        let mut transient_price = 1.0;
+        let mut smallest_fill_amount = f64::INFINITY;
+        for pair in pairs_on_path(path) {
+            let order = orders.pair_order_mut(pair).expect("missing order on path");
+            let fill_amount = capacity / transient_price;
+            order.amount -= fill_amount;
+            users
+                .get_mut(&order.user)
+                .expect("missing user with order")
+                .deduct_from_balance(order.pair.sell, fill_amount);
+            transient_price *= order.price;
+            smallest_fill_amount = num::min(smallest_fill_amount, fill_amount);
+        }
+
+        Some(smallest_fill_amount)
+    }
+
     /// Retrieve the weight of an edge in the projection graph. This is used for
     /// testing that the projection graph is in sync with the order map.
     #[cfg(test)]
@@ -162,6 +209,14 @@ impl Orderbook {
 /// Create a node index from a token ID.
 fn node_index(token: TokenId) -> NodeIndex {
     NodeIndex::new(token.into())
+}
+
+/// Returns an iterator with all pairs on a path.
+fn pairs_on_path(path: &[TokenId]) -> impl Iterator<Item = TokenPair> + '_ {
+    path.windows(2).map(|segment| TokenPair {
+        sell: segment[0],
+        buy: segment[1],
+    })
 }
 
 #[cfg(test)]
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -9,7 +9,7 @@ mod order;
 mod user;
 
 use self::order::{Order, OrderCollector, OrderMap};
-use self::user::UserMap;
+use self::user::{User, UserMap};
 use crate::encoding::{Element, InvalidLength, TokenId, TokenPair};
 use crate::graph::bellman_ford;
 use crate::num;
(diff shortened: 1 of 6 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -149,12 +162,88 @@ impl Orderbook {
...
+    /// starting token. Returns `None` if the path doesn't exist.
+    fn find_maximum_capacity(&self, path: &[TokenId]) -> Option<f64> {
+        let mut capacity = f64::INFINITY;
+        let mut transient_price = 1.0;
+        for pair in pairs_on_path(path) {
+            let order = self.orders.best_order_for_pair(pair)?;
+            transient_price *= order.price;
+
+            let sell_amount = order.get_effective_amount(&self.users);
+            capacity = num::min(capacity, sell_amount * transient_price);
```
fleupold, line 190: Why is the sell_amount (in some token along the path) multiplied with the transient_price, which as I understand includes information about all previous edges already. 

I would think, capacity is related to the amount at the current end of the path and also already incorporates all information about the previous prices.

Thus I'd expect

capacity = num::min(capacity, sell_amount) * order.price;

pricegraph/src/orderbook.rs
```diff
@@ -149,12 +162,88 @@ impl Orderbook {
...
+    fn best_order_with_user_for_pair_mut(
+        &mut self,
+        pair: TokenPair,
+    ) -> Option<(&'_ mut Order, &'_ mut User)> {
+        let Self { orders, users, .. } = self;
+
+        let order = orders.best_order_for_pair_mut(pair)?;
+        let user = users
+            .get_mut(&order.user)
+            .expect("missing user for existing order");
```
fleupold, line 244: nit: Would it make debugging easier if we included the user/order pair here?

pricegraph/src/orderbook.rs
```diff
@@ -222,6 +328,18 @@ mod tests {
         }};
     }
 
+    impl Orderbook {
```
fleupold, line 331: Should this only be available in tests?

fleupold (commented):
Something weird happened while I was reviewing and now the PR is way too big (766 lines). Will wait with further review until this is rebased to only include the relevant changes.

pricegraph/src/orderbook.rs
nlordell: We are trying to find the capacity of the path expressed in the starting token, this makes things much easier when doing successive shortest path as we want to find enough capacity over `n` paths to completely drain an amount.

pricegraph/src/orderbook.rs
nlordell: Good point! although I REALLY hope this assertion never fails :P

pricegraph/src/orderbook.rs
nlordell: Yes, its a testing utility method.

nlordell:
> Something weird happened while I was reviewing and now the PR is way too big (766 lines). Will wait with further review until this is rebased to only include the relevant changes.

Sorry @fleupold was merging and rebasing the previous PRs. Should be ready now!

pricegraph/src/orderbook.rs
fleupold: Right, this makes perfect sense! But don't we have to divide sellAmount with transient price in this case?

E.g. Take the order to "buy at least 1A for at most 2B" --> price 1/2 --> transient_price = 1* 1/2 = 1/2

Now let's say we have an order selling 1A. In order to convert that sell amount into starting token (B) we would have to divide by the transient price not multiply, wouldn't we?

pricegraph/src/orderbook.rs
fleupold: Oh sorry, I missed it was under the test module.

pricegraph/src/orderbook.rs
```diff
@@ -255,18 +384,124 @@ mod tests {
...
+    }
+
+    #[test]
+    fn fills_path_and_updates_projection() {
+        //             /---1.0---v
+        // 0 --1.0--> 1          2 --1.0--> 3 --1.0--> 4
+        //             \---2.0---^                    /
+        //                        \--------1.0-------/
+        //
+        // NOTE: Higher prices are worse.
```
fleupold, line 403: I might be wrong, but I would have thought it's the other way around. The numerator is the buyAmount and the denominator the sellAmount (https://github.com/gnosis/dex-contracts/blob/ca45b5d0fd2cbfc46cc14631e16d700dce2b2c57/contracts/BatchExchange.sol#L570). The more you sell the smaller the fraction (price) becomes and at the the same time the worse off you are (you are selling more while receiving the same).

pricegraph/src/orderbook/order.rs
```diff
@@ -63,6 +61,17 @@ impl OrderMap {
         })
     }
 
+    /// Returns an iterator over the orders matching a given sell token.
+    pub fn pairs_for_sell_token(
```
fleupold, line 65: nit: maybe integrate the fact that we return orders in the name? `pairs_and_orders_with_sell_token` or something like that?

pricegraph/src/orderbook/user.rs
```diff
@@ -30,8 +30,23 @@ impl User {
...
 
-    /// Retrieves the user's balance for a token
+    /// Return's the user's balance for the specified token.
     pub fn balance_of(&self, token: TokenId) -> f64 {
         self.balances.get(&token).copied().unwrap_or(0.0)
     }
+
+    /// Deducts an amount from the balance for the given token. Returns the new
+    /// balance or `None` if the user no longer has any balance.
+    pub fn deduct_from_balance(&mut self, token: TokenId, amount: f64) -> Option<f64> {
```
fleupold, line 40: Is this ok to do nothing if there is no or not enough balance?

fleupold (commented):
Looks all really good. Just a blocking question about the ordering of the price (and if you have to multiply or divide)

pricegraph/src/orderbook.rs
nlordell: I had price graph nightmares from thinking about this so much...

So the order "buy at least 1A for at most 2B" does indeed have a price of 1/2, but in our price graph, since we are pushing tokens through the graph, the graph direction is from the buy token to the sell token, since we can only offer token A to a funded order that is willing to buy it. In this situation, the sell amount is 2, so the maximum amount of token A we can send between the `A->B` order is 1 (`(sell_amount = 2) * (transient_price = 0.5)`).

pricegraph/src/orderbook/user.rs
nlordell: It should never happen since our capacities that we deduct always respect order limits and balances, but you are correct, there should be a debug assertion here.

pricegraph/src/orderbook.rs
nlordell: So higher prices are worse for pushing an amount through the graph, but are better for the owners of the order. That means that, when filling a path through the graph with the lowest cost, we want the lowest price (so highest sell amount) since they benefit us (the one pushing a flow through the graph) more by virtue of using existing orders that are worse off for the people who placed them.

nlordell: So the point of view matters, the higher prices are worse for the person trying to place an order that takes advantage of the existing orders.

pricegraph/src/orderbook.rs
nlordell: What I meant by:
> the graph direction is from the buy token to the sell token, since we can only offer token A to a funded order that is willing to buy it

Is that if we are trying to answer the question "what is the best value for `x` for my order: buy 1A for at most `x`B", then we do so by examining all orders that are "buy at most `b`B for at least `a`A" so that we can "push" our B token to A.

Given this, as you said, the best order for us is a small as possible `x` value (so the price goes up), and in order to achieve that, we need minimize the price of the other orders (hence the smaller price is better for getting a lower cost pushing some token through the graph) which is to say, minimize `a/b` (so higher `b` is better for this point of view).

pricegraph/src/orderbook.rs
nlordell: What I meant by:
> So the point of view matters, the higher prices are worse for the person trying to place an order that takes advantage of the existing orders.

Is that if we are trying to answer the question "what is the best value for `x` for my order: buy 1A for at most `x`B", then we do so by examining all orders that are "buy at most `b`B for at least `a`A" so that we can "push" our B token to A.

Given this, as you said, the best order for us is a small as possible `x` value (so the price goes up), and in order to achieve that, we need minimize the price of the reciprocal orders (hence the smaller price is better for getting a lower cost pushing some token through the graph) which is to say, minimize `b/a`, so higher `a`, the sell amount in this case, is better for our point of view.

pricegraph/src/orderbook.rs
nlordell: I will try to illustrate this with an example, if we want to find the best possible `x` for the order:
- "buy at least `x`A for at most 2B"

Given the orders:
- "buy at least 1B for 2A"
- "buy at least 1B for 3C"
- "buy at least 6C for 3A"

This means, in order to fill your order A->B, you can take the path B->A or B->C->A.

1. For the first path, the transient price is 1/2 (1B / 2A), and so the amount of B you can send through that path is `((sell_amount = 2A) * (1B / 2A)) = 1B`. This means you still have 1B of volume to push through the graph.
2. For the second path, there are two pairs:
  1. The first segment has a transient price of `1B / 3C`, and a capacity of 1B
  2. The second segment has a transient price of `1B * 6C / 3C * 3A = 2B / 3A` and a capacity of `((sell_amount = 3A) * (2B / 3A)) = 2B`
  3. This makes the final transient price be `2B / 3A` and the total capacity to be `min(1B, 2B) = 1B`.

I hope this helps explain it a bit.

pricegraph/src/orderbook/user.rs
nlordell: Added some `debug_assert!` statements to make sure its working as intended.

pricegraph/src/orderbook.rs
fleupold: I think I see it now. Maybe you can emphasize in the note that this is from the point of view of someone trying to push flow against the existing orders.

fleupold (approved):

PR #715, opened 2020-04-13
[Pricegraph] Estimate price by filling market order in the orderbook graph

nlordell:
This PR implements filling a market order in the orderbook graph using the successive shortest path algorithm. It allows for price estimation as it returns the price of the final shortest path used to fill the remaining volume.

### Test Plan

Unit tests (run with `--nocapture` to see price estimates for DAI/WETH on a real orderbook) and very promising benchmarks:
```
Orderbook::fill_market_order/100000000000000000                                                                            
                        time:   [488.01 us 492.00 us 496.36 us]
Orderbook::fill_market_order/1000000000000000000                                                                            
                        time:   [489.00 us 492.21 us 495.73 us]
Orderbook::fill_market_order/10000000000000000000                                                                            
                        time:   [491.00 us 495.20 us 499.58 us]
Orderbook::fill_market_order/100000000000000000000                                                                            
                        time:   [519.72 us 522.40 us 525.17 us]
Orderbook::fill_market_order/1000000000000000000000                                                                            
                        time:   [552.44 us 556.99 us 561.93 us]

Orderbook::fill_market_order(reduced)/100000000000000000                                                                            
                        time:   [22.840 us 23.260 us 23.614 us]
Orderbook::fill_market_order(reduced)/1000000000000000000                                                                            
                        time:   [23.277 us 23.766 us 24.182 us]
Orderbook::fill_market_order(reduced)/10000000000000000000                                                                            
                        time:   [23.047 us 23.449 us 23.784 us]
Orderbook::fill_market_order(reduced)/100000000000000000000                                                                            
                        time:   [37.304 us 38.050 us 38.645 us]
Orderbook::fill_market_order(reduced)/1000000000000000000000                                                                            
                        time:   [83.246 us 85.535 us 87.633 us]
```

WIth a reduced orderbook (so if the orderbook gets reused for multiple estimates) it takes 20-90 microsecond to perform a price estimate (depending on how much slippage there is) :rocket:. 
Even when using a non-reduced orderbook where the cycles need to be removed beforehand, the computation takes less than a millisecond.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -145,9 +145,9 @@ impl Orderbook {
     /// flow from the sell token to the buy token. The min-cost is calculated
     /// using the successive shortest path algorithm with Bellman-Ford.
     ///
-    /// Note that this method is not idempotent, successive calls with the same
-    /// token pair and volume yield different results as orders get filled with
-    /// each match changing the orderbook.
+    /// Note that in general this method is not idempotent, successive calls
+    /// with the same token pair and volume may yield different results as
+    /// orders get filled with each match changing the orderbook.
     pub fn fill_market_order(&mut self, pair: TokenPair, volume: f64) -> Option<f64> {
         self.update_projection_graph();
 
@@ -155,13 +155,27 @@ impl Orderbook {
         let predecessors = self.reduced_shortest_paths(sell);
         let mut path = path::find_path(&predecessors, sell, buy)?;
 
+        // NOTE: The transient price of a path is the price of the sell token
+        // (i.e. `sell_amount / buy_amount`). Since we are trying to find the
+        // best price for an order for the specified token pair (i.e. and order
+        // that would create a cycle of weight 0 going from the speicifed sell
+        // token to the buy token over `path` and then directly back to the sell
+        // token), we need to invert the price and account for the fees required
+        // for the final order in the cycle for the specified token pair.
+        fn invert_price(price: f64) -> f64 {
+            1.0 / (price * FEE_FACTOR)
+        }
+
         if volume <= 0.0 {
-            // NOTE: Intuitively an infinite price can be used for a 0 volume
-            // since it is always filled. This check is performed after finding
-            // the shortest path between a token pair to make sure that even
-            // when a volume of 0 is specified, `None` is returned if the tokens
-            // aren't connected.
-            return Some(f64::INFINITY);
+            // NOTE: For a 0 volume we simulate sending an tiny epsilon of value
+            // through the network without actually filling any orders.
+            let (_, price) = self.find_path_capacity_and_price(&path).unwrap_or_else(|| {
+                panic!(
+                    "failed to fill detected shortest path {}",
+                    format_path(&path),
+                )
+            });
+            return Some(invert_price(price));
         }
 
         let mut remaining_volume = volume;
@@ -183,11 +197,7 @@ impl Orderbook {
             path = path::find_path(&predecessors, sell, buy)?;
         }
 
-        // NOTE: The transient price is the price of the sell token in terms of
-        // the buy token. Since we are trying to find an order that would create
-        // a cycle of weight 0 that starts at the buy token and ends at the sell
-        // token, we want the inverse price with fees removed.
-        Some(1.0 / (last_transient_price * FEE_FACTOR))
+        Some(invert_price(last_transient_price))
     }
 
     /// Calculates the shortest paths from the start token to all other tokens
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -12,6 +12,8 @@ use self::order::{Order, OrderCollector, OrderMap};
 use self::user::{User, UserMap};
 use crate::encoding::{Element, InvalidLength, TokenId, TokenPair};
 use crate::graph::bellman_ford::{self, NegativeCycle};
+use crate::graph::path;
+use crate::graph::subgraph::{ControlFlow, Subgraphs};
 use crate::num;
 use petgraph::graph::{DiGraph, EdgeIndex, NodeIndex};
 use std::cmp;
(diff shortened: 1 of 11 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -129,15 +130,80 @@ impl Orderbook {
...
-                Ok((_, predecessor)) => break negative_cycle_predecessor.unwrap_or(predecessor),
-                Err(NegativeCycle(predecessor, start)) => {
+        let (sell, buy) = (node_index(pair.sell), node_index(pair.buy));
+        let predecessors = self.reduced_shortest_paths(sell);
+        let mut path = path::find_path(&predecessors, sell, buy)?;
+
+        if volume == 0.0 {
+            // NOTE: Intuitively an infinite price can be used for a 0 volume
+            // since it is always filled. This check is performed after finding
+            // the shortest path between to token pair to make sure that even
```
fleupold, line 161: typo: two token pairs

pricegraph/src/orderbook.rs
```diff
@@ -129,15 +130,80 @@ impl Orderbook {
...
-        Subgraphs::new(self.projection.node_indices()).for_each(|token| loop {
-            let mut negative_cycle_predecessor = None;
-            match bellman_ford::search(&self.projection, token) {
-                Ok((_, predecessor)) => break negative_cycle_predecessor.unwrap_or(predecessor),
-                Err(NegativeCycle(predecessor, start)) => {
+        let (sell, buy) = (node_index(pair.sell), node_index(pair.buy));
+        let predecessors = self.reduced_shortest_paths(sell);
+        let mut path = path::find_path(&predecessors, sell, buy)?;
+
+        if volume == 0.0 {
```
fleupold, line 158: Should this be the first thing the function does as none of the stuff above is necessary?

pricegraph/src/orderbook.rs
```diff
@@ -129,15 +130,80 @@ impl Orderbook {
...
+    /// each match changing the orderbook.
+    pub fn fill_market_order(&mut self, pair: TokenPair, volume: f64) -> Option<f64> {
         self.update_projection_graph();
 
-        Subgraphs::new(self.projection.node_indices()).for_each(|token| loop {
-            let mut negative_cycle_predecessor = None;
-            match bellman_ford::search(&self.projection, token) {
-                Ok((_, predecessor)) => break negative_cycle_predecessor.unwrap_or(predecessor),
-                Err(NegativeCycle(predecessor, start)) => {
+        let (sell, buy) = (node_index(pair.sell), node_index(pair.buy));
```
fleupold, line 154: Maybe my logic is wrong but doesn't this have to be inversed? If you are trying to fill a market order selling ETH for DAI, wouldn't you be looking in the graph for paths that are willing to buy ETH for DAI and check their price?

pricegraph/src/orderbook.rs
```diff
@@ -233,12 +298,13 @@ impl Orderbook {
...
+        Some((capacity, price))
     }
 
-    /// Gets the maximum capacity of a path expressed in an amount of the
-    /// starting token. Returns `None` if the path doesn't exist.
-    fn find_maximum_capacity(&self, path: &[NodeIndex]) -> Option<f64> {
+    /// Finds a transient trade along a path and returns the maximum capacity of
+    /// the path expressed in an amount of the starting token and the transient
+    /// price. Returns `None` if the path doesn't exist.
+    fn find_transient_trade(&self, path: &[NodeIndex]) -> Option<(f64, f64)> {
```
fleupold, line 307: naming nit: What about this trade is transient? Maybe calling it `find_maximum_capacity_and_price` or `find_best_trade`

pricegraph/src/orderbook.rs
```diff
@@ -478,6 +553,97 @@ mod tests {
...
+                    token 1 => 1_000_000,
+                    token 2 => 100_000_000,
+                }
+                @3 {
+                    token 1 => 1_000_000,
+                    token 2 => 100_000_000,
+                }
+            }
+            orders {
+                owner @1 buying 1 [1_000_000] selling 2 [99_000_000],
```
fleupold, line 581: I didn't notice this in the PRs before, but this DSL is really fancy.

fleupold (commented):
One question about the ordering of buy/sell tokens. I see that the unit tests pass but am looking for an explanation why we have to look for paths in the same direction as the orders that are already in the orderbook rather than searching for path against that direction.

pricegraph/src/orderbook.rs
nlordell: :see_no_evil:

pricegraph/src/orderbook.rs
nlordell: I mentioned this in the comment that I do the stuff above first so that we return `None` when there is no path, regardless of volume. I'm OK either way.

I had also considered having a volume of 0 be used so that you get the lowest transitive price between two orders without modifying the orderbook.

Let me know what you think is better.

pricegraph/src/orderbook.rs
nlordell: That is correct, it is why we are searching from the sell token since orders produce graph edges that go from the buy token to the sell token.

pricegraph/src/orderbook.rs
nlordell: It makes writing the orderbooks easier and more expressive.

nlordell:
> I see that the unit tests pass but am looking for an explanation why we have to look for paths in the same direction as the orders that are already in the orderbook rather than searching for path against that direction.

Its because the direction of the orders in the graph is a little unintuitive going from the buy token to the sell token, so if we search from a token A to find a path to token B, that gives us the cheapest (i.e. lowest price) transient order buying A and selling B. If we are trying to buy B for A, then this is the best transitive order we can match with.

pricegraph/src/orderbook.rs
fleupold: Oh I see now. I missed the ? which actually returns something before this code is hit. I think I'd slightly prefer to get the lowest possible transitive price since for technically for 0 volume any price is correct (the buy amount would always be 0 = (sellAmount = 0)* price). 

Maybe the notion of sending the smallest possible epsilon of flow through the network makes more sense?

fleupold (approved):
🚢

PR #717, opened 2020-04-13
[Pricegraph] Add WASM bindings to pricegraph crate for price estimation

nlordell:
This PR creates WASM bindings for the `pricegraph` crate so that it can be called from the browser and NodeJS :tada: 

### Test Plan

Build the WASM module with `(cd pricegraph/wasm; wasm-pack build --scope gnosis.pm)`. This will build a standalone npm package that can be used in NodeJS and in the browser to estimate prices for a given orderbook.

Additionally, a NodeJS integration test was added that estimates the price of 100 WETH's worth of DAI on an actual orderbook. This can be run with `(cd pricegraph/wasm; wasm-pack test --node)`.

### DIFF
--- a/pricegraph/wasm/src/lib.rs
+++ b/pricegraph/wasm/src/lib.rs
@@ -1,0 +1,37 @@
+//! This crate provides a thin WASM-compatible wrapper around the `pricegraph`
+//! crate and can be used for estimating prices for a given orderbook.
+
+use pricegraph::{Orderbook, TokenId, TokenPair};
+use wasm_bindgen::prelude::*;
+
+/// A graph representation of a complete orderbook.
+#[wasm_bindgen]
+pub struct PriceEstimator {
+    orderbook: Orderbook,
+}
+
+#[wasm_bindgen]
+impl PriceEstimator {
+    /// Creates a `PriceEstimator` instance by reading an orderbook from encoded
+    /// bytes. Returns an error if the encoded orders are invalid.
+    #[wasm_bindgen(constructor)]
+    pub fn new(bytes: &[u8]) -> Result<PriceEstimator, JsValue> {
+        console_error_panic_hook::set_once();
+
+        let mut orderbook = Orderbook::read(bytes).map_err(|err| err.to_string())?;
+        orderbook.reduce_overlapping_orders();
+
+        Ok(PriceEstimator { orderbook })
+    }
+
+    /// Estimates price for the specified trade. Returns `undefined` if the
+    /// volume cannot be fully filled.
+    #[wasm_bindgen(js_name = "estimatePrice")]
+    pub fn estimate_price(&self, buy: TokenId, sell: TokenId, volume: f64) -> Option<f64> {
+        // NOTE: Make sure to use a copy of the orderbook so that successive
+        // calls to `estimate_price` do not affect eachother.
+        self.orderbook
+            .clone()
+            .fill_market_order(TokenPair { buy, sell }, volume)
+    }
+}
--- a/pricegraph/wasm/tests/nodejs.rs
+++ b/pricegraph/wasm/tests/nodejs.rs
@@ -1,0 +1,32 @@
+extern crate wasm_bindgen_test;
+
+#[path = "../../data/mod.rs"]
+mod data;
+
+use pricegraph_wasm::PriceEstimator;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_test::*;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = Date)]
+    fn now() -> f64;
+}
+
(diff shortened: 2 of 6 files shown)


pricegraph/wasm/tests/nodejs.rs
```diff
@@ -0,0 +1,37 @@
...
+            .estimate_price(7, 1, 100.0 * 10.0f64.powi(18))
+            .unwrap()
+    });
+
+    console_log!(
+        "DAI-WETH price for selling 100 WETH: 1 WETH = {} DAI (load {}ms, estimate {}ms)",
+        price,
+        load_time,
+        estimate_time,
+    );
```
fleupold, line 36: How is this actually run in js and what is the unwrap translated to?

fleupold (approved):
This is true 🚀🔬

pricegraph/wasm/tests/nodejs.rs
nlordell: The `nodejs.rs` gets compiled to WASM and then a `node` executable is spun up that executes the WASM module's `#[wasm_bindgen_test]` methods.

In `PriceEstimator::new` there is a magic call to `console_error_panic_hook::set_once();` that handles panics by logging to `console.error`. This is pretty much the best we can do with panics since they produce an `unreachable` instruction that causes a trap the WASM runtime and makes the method return a `RuntimeError` right away (meaning you can't `catch_unwind` these panics and wrap it in a `Result`).

`Result`s in general produce JavaScript exceptions on `Err`, which is what we use for normal errors (in the `PriceEstimator::new` method for example). The hope is that the code should never panic (as that currently *should* only happen if some of the assumptions I made were wrong or there is a logic bug, and not on, say bad user input)

I must admit the amount of magic happening here is pretty crazy - gotta love `proc_macro`s and code generation!

nlordell: Here are some resource I found helpful:
- https://rustwasm.github.io/docs/book/
- https://rustwasm.github.io/docs/wasm-bindgen/

I also found running `wasm-pack build pricegraph/wasm --scope gnosis.pm` and then inspecting the `pricegraph/wasm/pkg` directory quite interesting to see how the JS/Rust interface works under the hood.

pricegraph/wasm/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+        let mut orderbook = Orderbook::read(bytes).map_err(|err| err.to_string())?;
+        orderbook.reduce_overlapping_orders();
+
+        Ok(PriceEstimator { orderbook })
+    }
+
+    /// Estimates price for the specified trade. Returns `undefined` if the
+    /// volume cannot be fully filled.
+    #[wasm_bindgen(js_name = "estimatePrice")]
+    pub fn estimate_price(&self, buy: TokenId, sell: TokenId, volume: f64) -> Option<f64> {
```
anxolin, line 30: This is too advanced for my rust knowledge, but even risking on making a silly question, why do you return an `Option` here?

Maybe is because it needs to warm up the estimation, but if that's the case, and given that you use `unwrap` in line `28` and that you said you cannot handle panics and they are just console.logged. 

Shouldn't we handle the None case so it doesn't panics, and we return that wrapped result you mentioned in the comment?

anxolin (commented):
Amazing, this is going to open so many possibilities :) 

I'm seeing it, it's a matter of time, this will evolve into a browser solver

pricegraph/wasm/src/lib.rs
nlordell: >  but even risking on making a silly question

No such thing!

> This is too advanced for my rust knowledge, but even risking on making a silly question, why do you return an Option here?

`Option` in rust gets converted to a nullable type in TS, so its like it returns `number?`. This is needed because estimates aren't always possible since there isn't always a "path" from the buy token to the sell token that allows for trading. The most trivial example of this is if you are trying to estimate a price for buying A and selling B, but there are no open orders that are buying B - so it doesn't matter how much you offer, your order won't get matched.

> Shouldn't we handle the None case so it doesn't panics, and we return that wrapped result you mentioned in the comment?

`unwrap` is bad because panics can't be caught in WASM since there is no stack unwinding (which is how they are implemented in Rust) and instead they function more like a C `abort` which can't be caught. That being said, Rust allows for "panic hooks" in WASM that run arbitrary code right before a panic happens and the code aborts. The `unwrap` you saw is in the integration tests module, where the test runner takes care to add a panic hook that signals the test has failed in case it panics which is what we want in that case.

anxolin: Ahhhh! okay, it's from the test module. My concern was just that I thought it was unwrapping in the module, so if there's price it returns the price, and if it doesn't have a price, it was panicking (so console.error)

Thanks for explaining!

PR #718, opened 2020-04-14
Retry with increased gas price

### LINKED ISSUE #677: Submit Solution: Retry with increased gas price
The likely cause of the following failed transaction is that the gas price was not high enough to get this transaction processed in time.

https://etherscan.io/tx/0xc87c495f1d05203d4d1a48ba785a883e506c4cadd64ea8d88e9e70b4175f53e1

A reasonable fix here is to observe that the transaction is taking too long and increase the gas price to avoid the dreadful error "Solutions are no longer accepted for this batch".

An incremental solution proposed by @fleupold is to double the gas prices up to a maximum bound.

bh2smith:
This PR introduces a few fundamental changes to `StableXContract.submit_solution` so to have a natural implementation of `retry_with_increased_gas_price` that is easily testable.

- `GasStation` is moved into the `SolutionSubmitter` instead of `StableXContract`
- Introduce a new function `retry_with_gas_price_increase` which doubles the gas price if solution is not mined within one block (to a maximum cap of 60 GWei).
- This functionality is tested with one test showing the gas price incremented (to the cap) and another which persists in not being mined. 

Note that the initial gas estimate cannot exceed the cap and that the the block timeout is reset to default on the last submitted transaction.

Closes #677

**TestPlan** Unit tests.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -86,7 +86,7 @@ pub trait StableXContract {
         solution: Solution,
         claimed_objective_value: U256,
         gas_price: U256,
-        timeout: Option<usize>,
+        block_timeout: Option<usize>,
     ) -> Result<(), MethodError>;
 }
 
@@ -150,7 +150,7 @@ impl StableXContract for StableXContractImpl {
         solution: Solution,
         claimed_objective_value: U256,
         gas_price: U256,
-        timeout: Option<usize>,
+        block_timeout: Option<usize>,
     ) -> Result<(), MethodError> {
         let (prices, token_ids_for_price) = encode_prices_for_contract(&solution.prices);
         let (owners, order_ids, volumes) = encode_execution_for_contract(&solution.executed_orders);
@@ -172,7 +172,7 @@ impl StableXContract for StableXContractImpl {
             .gas(5_500_000.into());
 
         method.tx.resolve = Some(ResolveCondition::Confirmed(ConfirmParams {
-            block_timeout: timeout,
+            block_timeout,
             ..Default::default()
         }));
         method.send().wait()?;
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -5,16 +5,17 @@ use std::collections::HashMap;
 use std::time::Duration;
 
 use anyhow::Result;
-use ethcontract::transaction::GasPrice;
+use ethcontract::transaction::{GasPrice, ResolveCondition};
 use ethcontract::{Address, BlockNumber, PrivateKey, U256};
 use lazy_static::lazy_static;
 #[cfg(test)]
 use mockall::automock;
 
 use crate::contracts;
-use crate::gas_station::GasPriceEstimating;
 use crate::models::{ExecutedOrder, Solution};
 use crate::util::FutureWaitExt;
+use ethcontract::errors::MethodError;
+use ethcontract::transaction::confirm::ConfirmParams;
 
 lazy_static! {
     // In the BatchExchange smart contract, the objective value will be multiplied by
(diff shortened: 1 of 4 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -93,10 +85,12 @@ pub trait StableXContract {
         batch_index: U256,
         solution: Solution,
         claimed_objective_value: U256,
-    ) -> Result<()>;
+        gas_price: U256,
+        timeout: Option<usize>,
```
e00E, line 89: Either explain in a comment or rename the variable so that the unit of the timeout is clear.

driver/src/solution_submission/mod.rs
```diff
@@ -120,48 +128,189 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+            solution.clone(),
+            claimed_objective_value,
+            self.gas_price_estimating,
+            60_000_000_000u64.into(),
+        )
+        .map_err(|err| {
+            extract_transaction_receipt(&err)
+                .and_then(|tx| {
+                    let block_number = tx.block_number?;
+                    match self.contract.get_solution_objective_value(
```
e00E, line 143: What is the purpose of this?

driver/src/solution_submission/mod.rs
```diff
@@ -120,48 +128,189 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+            } else {
+                Some(BLOCK_TIMEOUT)
+            },
+        );
+        // Breaking condition for our loop
+        gas_price < gas_cap
+            && matches!(
+                result,
+                Err(MethodError {
+                    inner: ExecutionError::ConfirmTimeout,
```
e00E, line 203: It might be useful to log this, at least in the beginning. I'm afraid we might unexpectedly always get a timeout when using `block_timeout == 1` or something unexpected like that.

e00E (approved):

driver/src/solution_submission/mod.rs
nlordell: This was there before, the diff is just struggling to show it correctly. It was used to get the revert message in case of a transaction failure.

nlordell:
@bh2smith Commandeering this since you will be traveling

driver/src/solution_submission/mod.rs
```diff
@@ -120,48 +128,193 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+fn retry_with_gas_price_increase(
+    contract: &dyn StableXContract,
+    batch_index: U256,
+    solution: Solution,
+    claimed_objective_value: U256,
+    gas_price_estimating: &dyn GasPriceEstimating,
+    gas_cap: U256,
+) -> Result<(), MethodError> {
+    const INCREASE_FACTOR: u32 = 2;
+    const BLOCK_TIMEOUT: usize = 1;
```
fleupold, line 167: Isn't 1 block timeout a bit too aggressive? I'm worried that our tx might not yet even have broadcasted until the next block is mined. Maybe we can start with 2?

In the future it would also be good if this could be a function of time remaining as with e.g. 3 minutes or so we could event try the normal safe estimate first in oder to save gas.

driver/src/solution_submission/mod.rs
```diff
@@ -120,48 +128,193 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+    const DEFAULT_GAS_PRICE: u64 = 15_000_000_000;
+
+    let mut gas_price = match gas_price_estimating.estimate_gas_price() {
```
fleupold, line 170: Do you have any thoughts if this be done as well inside each loop in case the gas price changed in the meantime?

fleupold (approved):
Looks great ⛽📈 

Generally, we could also abstract gas_estimation into a strategy object that can act based on the time remaining and maybe on the time tried before if we feel it becomes too much logic to keep it inlined in this module.

driver/src/solution_submission/mod.rs
nlordell: I had talked to @bh2smith and if mainnet transactions get mined pretty far apart. We figured that if the block timeout were 2 we wouldn't really retry more than once.

I can bump this up to 2 and if we realize the transaction isn't really getting re-sent than we can try for a different strategy (like one where timeout is in function of time remaining).

driver/src/solution_submission/mod.rs
nlordell: Do you think the gas price could double in 2 blocks?

nlordell: I implemented this, which I guess will help when the gas price is increasing rapidly

bh2smith: Unfortunately I didn't catch this at the time. I meant to only make the estimation once. Not exactly sure how this slipped by.

bh2smith: Oh, I see this was introduced in the last commit. I think it would be best if we only estimate once and increase from our current estimate.

PR #719, opened 2020-04-14
[DexAG] Make price estimation match case insensitive on symbol

fleupold:
[Tom pointed out](https://github.com/gnosis/dex-solver/issues/201#issuecomment-613308193) that our price estimation for sETH and sUSD wasn't working correctly.
Despite being enabled in the staging configs, we would only use the default price 250$ which is far off from the current market price.

It turns out the reason for this is that the symbols on DEX.ag are listed in all uppercase (SETH, SUSD), whereas our configuration labels the token as sETH/sUSD.

This PR makes the dex/ag price estimator case insensitive to the token symbol by always comparing uppercased symbols.

### Test Plan

Added a unit test. Also making sure that `cargo test -- --nocapture --ignored online_dexag_client` no longer return undertermined for sETH and sUSD.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,9 +18,9 @@ checksum = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.10"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8716408b8bc624ed7f65d223ddb9ac2d044c0547b6fa4b0d554f3a9540496ada"
+checksum = "743ad5a418686aad3b87fd14c43badd828cf26e214a00f92a384291cf22e1811"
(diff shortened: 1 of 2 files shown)


nlordell (approved):
Looks good, but I don't think the Cargo.lock changes are necessary here.

nlordell:
Build is failing on a clippy error:
```
error: redundant clone
   --> driver/src/price_estimation/dexag/mod.rs:296:40
    |
296 |             let api_tokens = api_tokens.clone();
    |                                        ^^^^^^^^ help: remove this
    |
    = note: `-D clippy::redundant-clone` implied by `-D warnings`
note: this value is dropped without further use
   --> driver/src/price_estimation/dexag/mod.rs:296:30
    |
296 |             let api_tokens = api_tokens.clone();
    |                              ^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
```

e00E (approved):

PR #721, opened 2020-04-15
Move error into own file

e00E:
As mentioned in the big that PR I closed I am going to create several smaller ones. This is the first one.

### DIFF
--- a/driver/src/orderbook/streamed/error.rs
+++ b/driver/src/orderbook/streamed/error.rs
@@ -1,0 +1,13 @@
+use thiserror::Error;
+
+#[derive(Clone, Copy, Debug, Error)]
+pub enum Error {
+    #[error("unknown token")]
+    UnknownToken,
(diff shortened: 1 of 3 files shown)


driver/src/orderbook/streamed/error.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[derive(Clone, Copy, Debug, Error)]
+pub enum Error {
+    #[error("unknown token")]
+    UnknownToken,
+    #[error("unknown order")]
+    UnknownOrder,
+    #[error("order already exists")]
+    OrderAlreadyExists,
+    #[error("math underflow")]
```
fleupold, line 11: You mentioned that the debug information around which order/token is causing troubles was in another PR? Up to you if you want to add it here.

fleupold (approved):
Nice and easy 👍

PR #722, opened 2020-04-15
Solver bump

josojo:


fleupold:
@josojo you are probably aware, but since we don't require accepts for merging (since we want people to be able to merge in cases of emergency) merge_when_green will merge even if nobody reviewed your PR.

This might have been intended because it was urgent (since also no reviewers were specified), just in the regular case I think it's nicer to still get a quick thumbs up even for version bumps just so that people can be aware of this change.

josojo:
I was aware :) but I thought it is not worth bothering you guys. 

I think that no approval is needed for such a trivial change. While we should not do it from the perspective of the "broken glass theory", I think that people anyways would just approve and not bother looking into it, just because it is not worth their time.

 But if you guys wanna have this notification, I can do it in the future.

Issue #684, opened 2020-04-06
Reliably Fetch Account State


bh2smith:
**TL;DR** Smart Contract revert with "Amount exceeds user's balance.". A competing solver was attempting to spend pending balances from an already accepted solution to the current batch. When the solution was submitted and existing trades reverted the balances were no longer sufficient.

Last week our driver software experienced an interesting and unexpected failure to submit solution to Batch 5286263. The smart contract failed with the message "Amount exceeds user's balance." Note that, because the driver queries the return value of `submitSolution` before actually submitting the transaction, this error occurred without an existing failed transaction to be linked here.

After careful investigation, it was determined that a competing solver (our fallback-solver) was working off of an incorrect state of accounts. In particular, there has already been a solution submitted before the competing solver had fetched or queried the `Orderbook` (i.e. account balances and active orders). The accounts that the solver was working from, were those that had already bee updated by an accepted solution. After finding another solution and submitting its proposal, the smart contract internally reverted the previously accepted solution (hence, reverting the balances to their previous state) and suddenly, the solution proposed by the competing solver was no longer valid (due to insufficient balance). For more details on the issue with links to relevant transactions and solutions search for "Batch 5286263" in our [oncall report](https://docs.google.com/document/d/1WjlqerD4P1Tn-oLmaCZiPWFZDtiyh25ElVTHmwzNgP8/edit#)

There are likely a few possible way to tackle this new problem, so I will propose here a few thoughts and see how the discussion develops.

One universal fact that must be considered in our solution is that competing solvers should remain independent. That is, since this fix will be made to the driver (i.e. from within `dex-services`) any independent application running an instance of this project will not know anything else about other instances.
  
**Some ideas**

1. Configure driver to fetch account state separately from orders so it is returned earlier. This may already be in place.
2. Configure driver to fetch orderbook from the first published block after batch has closed. 

_Note_, however, that it is technically possible that this first block published after batch has closed could already contain a successful solution submission. In fact, it is not even unlikely that this could be the case. For example, suppose there were two obviously matching orders placed at the beginning of the batch. It is now obvious that a non-trivial solution to this auction exists and could be submitted (and accepted) as early as the first published block after batch closure.

3. Configure driver to fetch account state from last block published before batch closure. 
4. Query for orderbook from block number minus one of an already accepted solution... seems like a game of cat and mouse.

bh2smith:
I think that, if possible. Option 3 in combination with 1 would be a reasonable approach.

[...]

PR #725, opened 2020-04-15
[WIP] Initial reliable orderbook retrieval

nlordell:
This PR starts implementing a reliable orderbook retrieval mechanism where it concurrently:
- scans for the last block in the solving batch and uses it to query the orderbook.
- listens for incoming events (in case of reorgs or the orderbook is being fetched optimistically assuming that no meaningful changes to the orderbook will happen) and signals along a channel to the orderbook querying thread to restart in case a meaningful change to the orderbook was detected.

This PR assumes the existence of an unimplemented `StableXContractAsync` trait that abstracts the contract and EVM interactions required for this implementation.

It has been modified from the original PR to no longer track solution submissions (in order to "unapply" trades) by ensuring that the orderbook gets queried at the latest on the last block of the current batch, making it impossible for solutions have been submitted.

This would solve #684 

Note that the current paginated orderbook collection abstraction fit in perfectly and didn't require any changes, which is super awesome!

### Test Plan

Unit tests to be implemented in order to validate the event and querying logic. If this works, put it on staging early in the week to see if it performs well or not. We can use Graphana metrics to measure if by starting the system scheduler a littler earlier, which is now safe since a change to the orderbook would be detected and cause the orderbook retrieval to restart, the solvers would have additional time to produce solutions.

### DIFF
--- a/driver/src/orderbook/retrying_retrieval.rs
+++ b/driver/src/orderbook/retrying_retrieval.rs
@@ -1,0 +1,260 @@
+//! This module implements a retrying orderbook reader that accounts for
+//! submitted solutions as well as batches rolling over while the orderbook is
+//! being read.
+
+#![allow(dead_code)]
+
+use crate::contracts::stablex_contract::batch_exchange;
+use crate::contracts::stablex_contract::batch_exchange::event_data::{SolutionSubmission, Trade};
+use crate::models::{AccountState, Order};
+use crate::orderbook::paginated_auction_data_reader::PaginatedAuctionDataReader;
+use anyhow::{anyhow, Result};
+use ethcontract::web3::types::BlockId;
+use ethcontract::{Address, Event, EventData, H256};
+use futures::channel::{mpsc, oneshot};
+use futures::future::{BoxFuture, FutureExt};
+use futures::stream::{BoxStream, StreamExt};
+use futures::{select, try_join};
+use std::collections::HashMap;
+use std::convert::TryInto;
+
+/// Type definition for complete solution data.
+pub type SolutionData = (SolutionSubmission, Vec<Trade>);
+
+/// A trait for querying contract data with futures and streams.
+#[cfg_attr(test, mockall::automock)]
+// NOTE: Clippy complains that the lifetime can be elided with `'_` for the
+//   various streams and futures, but doing so results in a compiler errors.
+#[allow(clippy::needless_lifetimes)]
+pub trait StableXContractAsync: Sync {
+    /// Retrieves the batch for a specific block number.
+    fn batch_id_at_block<'a>(&'a self, block: BlockId) -> BoxFuture<'a, Result<u32>>;
+    /// Retrieves the latest solution for the given batch index.
+    fn latest_solution<'a>(&'a self, batch_id: u32) -> BoxFuture<'a, Result<Option<SolutionData>>>;
+    /// Streams pages of encoded orders.
+    fn encoded_orders_paginated<'a>(
+        &'a self,
+        previous_page_user: Address,
+        previous_page_user_offset: u16,
+        page_size: u16,
+    ) -> BoxFuture<'a, Result<Vec<u8>>>;
+    /// Streams incomming events from the contract that may signal that the
+    /// orderbook retrieval must be retried as the data has become stale.
+    fn all_events<'a>(&'a self) -> BoxStream<'a, Result<Event<batch_exchange::Event>>>;
+}
+
+fn read_orderbook(
+    contract: &dyn StableXContractAsync,
+    batch_id: u32,
+    page_size: u16,
+) -> Result<(AccountState, Vec<Order>)> {
+    futures::executor::block_on(async {
+        // NOTE: Use a bounded channel with a 0 sized buffer so that the sender
+        //   (i.e. the future that is polling the incomming events) can only
+        //   send 1 retry at a time.
+        let (retry_tx, retry_rx) = mpsc::channel::<()>(0);
+        let (done_tx, done_rx) = oneshot::channel::<()>();
+
+        let (solution_event, latest_solution, (account_state, orders)) = try_join!(
+            poll_incoming_events(contract, batch_id, retry_tx, done_rx),
+            contract.latest_solution(batch_id),
+            read_orderbook_async(contract, batch_id, page_size, retry_rx, done_tx),
+        )?;
+
+        if let Some((solution, trades)) = solution_event.or(latest_solution) {
+            todo!("undo {:?} and from {:?} from orderbook", solution, trades);
+        }
+
+        Ok((account_state, orders))
+    })
+}
+
+async fn read_orderbook_async(
+    contract: &dyn StableXContractAsync,
+    batch_id: u32,
+    page_size: u16,
+    mut retry: mpsc::Receiver<()>,
+    _done: oneshot::Sender<()>,
+) -> Result<(AccountState, Vec<Order>)> {
+    let mut reader = PaginatedAuctionDataReader::new(batch_id.into());
+    loop {
+        let mut page_future = contract
+            .encoded_orders_paginated(
+                reader.pagination().previous_page_user,
+                reader
+                    .pagination()
+                    .previous_page_user_offset
+                    .try_into()
+                    .expect("user cannot have more than u16::MAX orders"),
+                page_size,
+            )
+            .fuse();
+
+        let page = select! {
+            page = page_future => page?,
+            _ = retry.next() => {
+                reader = PaginatedAuctionDataReader::new(batch_id.into());
+                continue;
+            },
+            complete => break Err(anyhow!("error processing incoming events")),
+        };
+
+        let number_of_orders: u16 = reader
+            .apply_page(&page)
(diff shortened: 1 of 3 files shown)


josojo:
>The current latest solution that needs to be removed from the order book for it to be correct.

Do we really need this complicated "un-applying" method for solutions? Why don't we read the order book right before the solution was provided? E.g. in the first block of the new batch at tx index 0. (With tenderly one can do it, not sure how hard it is to code in ethcontracts)

Otherwise, I think that we need exactly such a structure for retrieving and retrying to get the order book even with the event-based order book. Hence, I would say go ahead.
(But please take my comments with a grain of salt, I have not been involved too much in the earlier discussions)

nlordell:
>  E.g. in the first block of the new batch at tx index 0

I think we used to have the orderbook reader do something like this, but we were getting errors that the block didn't exist anymore or somthing because of state pruning. Maybe @e00E can comment more.

e00E:
Yes, we used to set a fixed block number and it turned out that our node didn't support it the way we wanted it. I don't remember the details but even for recent blocks it wouldn't be able to get the result.

driver/src/orderbook/retrying_retrieval.rs
```diff
@@ -0,0 +1,260 @@
...
+    batch_id: u32,
+    mut retry: mpsc::Sender<()>,
+    done: oneshot::Receiver<()>,
+) -> Result<Option<SolutionData>> {
+    use batch_exchange::Event::*;
+    use Batch::*;
+    use EventData::*;
+
+    let mut done = done.fuse();
+    let mut events = contract.all_events().fuse();
```
fleupold, line 128: Is this giving us all events from the beginning of time or just since we call this one? I assume the first, as we want to  get the latest solution. But this means we would retry the fetching initially a lot of times as we iterate over all past OrderCancellations.

Maybe we should query the current block ID, base all our fetch queries on that block and only restart (with a new blocknumber) if there is an order cancellation newer than our starting block but still in the past batch.

fleupold (commented):
I like the overall approach. Maybe we can limit from when we fetch events to something like 20 blocks in the past (~5 minutes) and make the invalidation logic a bit less verbose, but overall I think it's a nice idea.

driver/src/orderbook/retrying_retrieval.rs
nlordell: > Maybe we can limit from when we fetch events to something like 20 blocks in the past

I think this is a really good idea. We currently bail if we get an event in the next batch, but we can also listen on block number.

> and make the invalidation logic a bit less verbose

Do you mean the logic that determines if a retry is needed?

> Is this giving us all events from the beginning of time or just since we call this one?

It is from the call onwards, we have a separate future getting the last applied solution that we join on. We take the latest solution (from incoming events) or the latest solution from the aforementioned future. I did this to avoid the issue you just mentioned.

> Maybe we should query the current block ID, base all our fetch queries on that block

I was under the impression we ran into issues with this because of state pruning.

e00E:
Based on your description in the OP I think it should work (I don't understand all of the code) BUT I wonder if we should instead try to address the problem of why we can't query the node at a fixed block. If that worked it would be so much easier for our code, as we would just change one line the way Alex described.
I found the issue with state pruning: https://github.com/gnosis/dex-services/issues/544

nlordell:
Thanks for finding the issue! I'll investigate a little.

fleupold:
I believe the reason this was originally not working was that we were behind a load balancer and thus would sometimes query node B for a block that we read on node A but that was not yet propagated to node B.

@denisgranha looked at the config and tried a test command showing that we have up to 64 blocks of history (cf. [slack](https://gnosisinc.slack.com/archives/CU37HR35W/p1581946641018900))

Thus I believe the error was that we were querying future blocks (from the viewpoint of the slower node) rather than blocks that were too far in the past.

driver/src/orderbook/retrying_retrieval.rs
fleupold: > Do you mean the logic that determines if a retry is needed?

Makes sense if we are only getting events from the point of subscription onwards. I originally thought we are querying events from genesis.

nlordell:
I re-implemented this without undoing solutions by querying the orderbook on, at the latest, the last block of the solving batch (which would include all events that could potentially change the orderbook).

This uses the same event watcher as before, but now just for events that would invalidate the solution (see inline comments). If this works, then we could even have the drivers start optimistically retrieving the orderbook earlier as it would restart the retrieval if any events come in that would invalidate the orderbook. This would give our solver a little bit more time to solve.

driver/src/orderbook/retrying_retrieval.rs
```diff
@@ -0,0 +1,222 @@
...
+            },
+            _ = done => break Ok(()),
+        };
+
+        let block_number = match event.meta.as_ref() {
+            Some(meta) => meta.block_number,
+            _ => return Err(anyhow!("unexpected event on pending block in event stream")),
+        };
+
+        let requires_retry = match &event.data {
```
fleupold, line 171: Could we simplify the logic, by checking if the event's timestamp is from the next batch or the one we are currently trying to solve. In the former case we know that `requires_retry` will be false regardless of the event type, wouldn't we? We then would only have to look at the event details for order placements.

fleupold (commented):
Nice, before merging it would be great if you could add some tests to the logic, but it seems solid to me (I understand you wanted to get some feedback on the general approach first). 

One issue with starting early is that the balances might be off (not respecting already requested withdraws/deposits that will be valid in the batch we are starting to solve).

We can workaround that by changing to this method for fetching data: https://github.com/gnosis/dex-contracts/blob/6a72b1546cb09f4115300e18b5f7d7e320b6b42e/contracts/BatchExchangeViewer.sol#L42

driver/src/orderbook/retrying_retrieval.rs
nlordell: Sounds like a good plan to me.

driver/src/orderbook/retrying_retrieval.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn encoded_orders_paginated<'a>(
+        &'a self,
+        previous_page_user: Address,
+        previous_page_user_offset: u16,
+        page_size: u16,
+        block_number: BlockNumber,
+    ) -> BoxFuture<'a, Result<Vec<u8>>>;
+    /// Streams incomming events from the contract that may signal that the
+    /// orderbook retrieval must be retried as the data has become stale.
+    fn all_events<'a>(&'a self) -> BoxStream<'a, Result<Event<batch_exchange::Event>>>;
```
bh2smith, line 39: Could we run into trouble with this on restart? Wouldn't it replay the contract's entire event history? Or is it only meant to return event from restart onward...

driver/src/orderbook/retrying_retrieval.rs
```diff
@@ -0,0 +1,222 @@
...
+fn read_orderbook(
+    contract: &dyn StableXContractAsync,
+    batch_id: u32,
+    page_size: u16,
+) -> Result<(AccountState, Vec<Order>)> {
+    futures::executor::block_on(async {
+        // NOTE: Use a bounded channel with a 0 sized buffer so that the sender
+        //   (i.e. the future that is polling the incomming events) can only
+        //   send 1 retry at a time.
+        let (retry_tx, retry_rx) = mpsc::channel::<u64>(0);
```
bh2smith, line 51: What are `tx` and `rx`?

driver/src/orderbook/retrying_retrieval.rs
bh2smith, line 171: Perhaps I have missed the goal here, but it seems we are missing a key element to this match statement. Specifically, aren't we meant to retry when a Solution comes in, or even more precisely when "Trade" event is emitted. It is my understanding that the original issue arose from fetching the account state after a solution had been submitted and thus having incorrect balances.

bh2smith (commented):
Looks really great and nice scaffolding for the `StableXContractAsync`. Just wondering if there is a missing element in the retry match statement. That is, most fundamentally, do we not want to start over when "Trade" even has been emitted? Also, I am not able to see any reference to the account state retrieval, only orderbook. I suspect the balances are part of this, but I recall suggesting that they be separated.

driver/src/orderbook/retrying_retrieval.rs
nlordell: This sets up an event listener for future events. I'll rename this.

driver/src/orderbook/retrying_retrieval.rs
nlordell: Transmit and receive. They are commonly used abbreviations.

driver/src/orderbook/retrying_retrieval.rs
nlordell: Good point, solutions coming in could affect the balances!

bh2smith: Specifically "Trade" and "TradeReverted" events.

nlordell:
Closing as we are going with #736 for now.

PR #729, opened 2020-04-16
New benign error for not accepted solution due to fetching the orderbook after a solution has been applied

josojo:
This PR is controversial.

Since we don't have a solution for this issue: https://github.com/gnosis/dex-services/issues/684 until we get the event-based order book fetching and it keeps popping up in our alerts, this PR will silence the notifications for this error.

The obvious risk is that this PR silences other errors with the same error message as well.


nlordell (approved):
I started working on a solution #725 that should solve this problem. I think we can merge this for the  being though to clean up the alerts channel.

e00E (approved):
I am okay with this because I am sure anyone that sees this alert on slack already ignores it anyway.

PR #731, opened 2020-04-16
Add more helpers to balance and catch more errors

e00E:
Does not yet handle trades.


driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,63 +2,104 @@ use super::*;
...
-        }
+    pub fn deposit(
+        &mut self,
+        amount: U256,
+        deposit_batch_id: BatchId,
+        current_batch_id: BatchId,
+    ) -> Result<(), Error> {
+        self.apply_existing_deposit(current_batch_id)?;
+        // Works like in the smart contract: If there is an existing deposit we override the
+        // batch id and add to the amount. If there is no existing deposit then amount is already 0.
```
e00E, line 22: @fleupold 
> Do you need current_batch_id here? Couldn't we just compare `self.deposit.batch_id` to `deposit_batch_id`?
> 
> Otherwise I feel the `self.deposit.batch_id = deposit_batch_id;` might override some batchId that is in between `deposit_batch_id` and `current_batch_id`

I changed this a little bit compared to when you originally commented maybe it is more clear now.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,63 +2,104 @@ use super::*;
...
+            .ok_or(Error::MathOverflow)?;
+        self.deposit.batch_id = deposit_batch_id;
+        self.deposit.amount = deposit_amount_including_this_deposit;
+        Ok(())
+    }
+
+    pub fn withdraw_request(&mut self, amount: U256, batch_id: BatchId) {
+        self.withdraw.batch_id = batch_id;
+        self.withdraw.amount = amount;
     }
```
e00E, line 36: @fleupold 
> Can we assert here that the withdraw has been cleared before this method is invoked?

No. It is possible to for example request withdrawal twice in the same batch id. In this case the newer request overrides the older.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,63 +2,104 @@ use super::*;
...
+        }
+        let balance_excluding_withdraw = self.get_balance_internal(batch_id, false)?;
+        let new_balance = balance_excluding_withdraw
             .checked_sub(amount)
-            .ok_or(Error::MathUnderflow)?;
+            .ok_or(Error::WithdrawMoreThanBalance(balance_excluding_withdraw))?;
+        self.balance = new_balance;
+        // If there is a valid deposit then new_balance already includes it so we need to clear it
+        // now.
+        self.clear_pending_deposit_if_possible(batch_id);
```
e00E, line 52: @fleupold 
> The thing I dislike about this method is that it can only be safely called if before we have called get_balance_internal(batch_id, false) and re-write the result back to balance afterwards (otherwise we lose the pending deposit and and solution). This expectation is very implicit.

Because it is called "clear" and the implementation is simple I disagree that this is too implicit.

> Would it be possible to rename get_balance_internal into updated_and_get_balance and do the persistence as well as the clearing directly in there?

get_balance should not be mutable but I will add another function that does this.

e00E: I couldn't' do this in a way that I find to improve the code so I am keeping it like this. Compared to when you made your review there is a new comment in the code explaining why we are clearing the pending deposit.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,63 +2,104 @@ use super::*;
...
+            .checked_add(amount)
+            .ok_or(Error::MathOverflow)?;
+        self.deposit.batch_id = deposit_batch_id;
+        self.deposit.amount = deposit_amount_including_this_deposit;
+        Ok(())
+    }
+
+    pub fn withdraw_request(&mut self, amount: U256, batch_id: BatchId) {
+        self.withdraw.batch_id = batch_id;
+        self.withdraw.amount = amount;
```
nlordell, line 35: Once a balance withdrawal is requested, is the balance still available to the exchange and the orderbook for trading or is it "useless"?

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,63 +2,104 @@ use super::*;
...
     }
 
-    pub fn withdraw(&mut self, amount: U256, current_batch_id: BatchId) -> Result<(), Error> {
-        self.update_deposit_balance(current_batch_id);
-        self.pending_withdraw = None;
-        self.balance = self
-            .balance
+    pub fn withdraw(&mut self, amount: U256, batch_id: BatchId) -> Result<(), Error> {
+        if self.withdraw.batch_id >= batch_id {
+            return Err(Error::WithdrawEarlierThanRequested(self.withdraw.batch_id));
```
nlordell, line 40: Does this actually happen or is this just being careful that the exact order in which events come in from the event stream is not really documented?

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,63 +2,104 @@ use super::*;
...
+        }
+        if self.withdraw.amount < amount {
+            return Err(Error::WithdrawMoreThanRequested(self.withdraw.amount));
+        }
+        let balance_excluding_withdraw = self.get_balance_internal(batch_id, false)?;
+        let new_balance = balance_excluding_withdraw
             .checked_sub(amount)
-            .ok_or(Error::MathUnderflow)?;
+            .ok_or(Error::WithdrawMoreThanBalance(balance_excluding_withdraw))?;
```
nlordell, line 48: Maybe include the amount here as well so that we can see the exact overflow.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,63 +2,104 @@ use super::*;
...
-            }
+        Ok(balance)
+    }
+
+    fn apply_existing_deposit(&mut self, current_batch_id: BatchId) -> Result<(), Error> {
+        if self.deposit.batch_id < current_batch_id {
+            let new_balance = self
+                .balance
+                .checked_add(self.deposit.amount)
+                .ok_or(Error::MathOverflow)?;
```
nlordell, line 86: Maybe we can add backtraces to some of these errors? I feel like if we get a `MathOverflow` error in the logs it might be really hard to figure out where its coming from.

driver/src/orderbook/streamed/balance.rs
nlordell: Just curious, is it possible to `apply_deposit_if_possilbe` and then apply the withdrawal separately? The reason I suggest this is to reduce the number of places that mutating the deposit flux can happen.

driver/src/orderbook/streamed/error.rs
```diff
@@ -8,6 +9,12 @@ pub enum Error {
...
-    #[error("math underflow")]
-    MathUnderflow,
+    #[error("math under or overflow")]
+    MathOverflow,
+    #[error("withdraw batch does not match withdraw request {0}")]
+    WithdrawEarlierThanRequested(BatchId),
+    #[error("withdraw amount is larger than withdraw request {0}")]
+    WithdrawMoreThanRequested(U256),
+    #[error("withdraw amount is larger than balance {0}")]
+    WithdrawMoreThanBalance(U256),
```
nlordell, line 19: I think it might be worth wrapping this error type in something that identifies the event (by block number and log index for example) so that the exact event that caused the problem can be inspected on, say, etherscan.

nlordell: For a separate PR though!

nlordell (approved):
I think splitting out the balance logic into its own module was a great move! I think the implementation looks good to me but I do think that the logic around applying a withdrawal is a little complex. I think I'm missing something obvious about why it can't just be (in pseudo-code) `self.apply_deposit_if_possible(); self.balance = self.balance - amount;`

nlordell:
https://media.giphy.com/media/xT0xeJpnrWC4XWblEk/giphy.gif

nlordell:
mind blown

nlordell:
mind blown

nlordell:
I have a few questions related to code that was added to this PR and how it will fit in to the bigger picture. Specifically:
- How will this work with respect to removed events?
- Some code seams to be there to account for uncertainty regarding the order in which events come in. I think in general it would be better to order to events before applying so that reasoning about the code is easier.
- How will we handle failures? I think this is something that we should keep in mind since if we just completely fail orderbook retrieval if there is some unaccounted for weirdness then the whole system would go down. It might make sense to keep user balances/orders/etc. distinct from one another so that if updating one of these fail, then the rest of the orders and users keep on going so things mostly work.

driver/src/orderbook/streamed/balance.rs
e00E: Do you mean once the effective balance is 0 because of the valid withdraw request is >= balance? The smart contract uses `getBalance` so it will also see 0 and can't deduct any more.

driver/src/orderbook/streamed/balance.rs
e00E: I haven't seen it happen. It's just another thing we can check that must hold for the real smart contract so I am checking it.

driver/src/orderbook/streamed/balance.rs
e00E: I have thought about that. My conclusion is that the person adding the event already knows this (because it is in the event) so they would add the event to the context of the error or somethinhg like that if they get an error when applying it. It's the same for `UnknownOrder(OrderId)` where I no longer think the order id makes sense.

driver/src/orderbook/streamed/balance.rs
e00E: I could add more variants to all the places where an overflow can occur but it feels like overkill and premature. If we get an error on the real events we will know at which events it happens and can replay that to debug.
What exactly do you mean by backtrace, how do I get one? Maybe I could use the line number to help us debug.

e00E:
>     * How will this work with respect to removed events?

This `State` does not handle removed events. The "state manager" is going to get events from the node and if there is a revert it will recreate the state from a previous snapshot and replay the new correct chain of events.

>     * Some code seams to be there to account for uncertainty regarding the order in which events come in. I think in general it would be better to order to events before applying so that reasoning about the code is easier.

It is just checking as much as possible that we know has to hold for the real smart contract. Unrelated to reorderings. The events will get ordered before applying.

>     * How will we handle failures? I think this is something that we should keep in mind since if we just completely fail orderbook retrieval if there is some unaccounted for weirdness then the whole system would go down. It might make sense to keep user balances/orders/etc. distinct from one another so that if updating one of these fail, then the rest of the orders and users keep on going so things mostly work.

That is up to the "state manager". If it has 10 new events and the first one fails to apply it can still go ahead an apply the other ones hoping that the first one didn't influence state of the later oners.

driver/src/orderbook/streamed/error.rs
e00E: As I said above, this should be done by the "state manager" because it knows on which event the error happened and can add that information.

nlordell:
Another thought I've had: what if we organize the streamed orderbook so that it looks something like this:
```
/------
| Events |-----------------\
\--------/                 |
                           v 
                     /-----------------\         /-----------------\
                     | Event Collector |-------->| State Snapshots |
                     \-----------------/         \-----------------/
                           ^     |                        |
/--------\                 |     \----------[+]-----------/
| Blocks |-----------------/                 |
\--------/                                   |
                                             v
                                     ( Current State )

```

This way the state component won't ever have to deal with removed events, since it will only ever be applying added events (in the correct order) for a specific batch.

This would also reduce the complexity of the balance calculation since it wouldn't have to store state related to pending fluxes since it would know which events to apply and which ones to ignore (and it can just discard withdrawal (requests) and deposits that mature in future batches). For things like deposits where more than one can happen in the same batch, the events collection could be flattened to only include the latest one. This would make balances applied

This would also make it a bit more of a "pure" function that would apply changes to an orderbook snapshot (that we can even serialize to disk and use as a starting point for unit tests in the case where an unexpected error pops up).

Snapshots can be periodically collected based on block and batch. So, say, 5 blocks after a new batch, the previous block can make a new snapshot and apply all events for that block on top of it. This means that whenever events are being applied, all events are known so the code doesn't have to create state to account for future events to be applied (like storing a deposit as a pending flux).

Maybe you already thought about something like this but there are things that this doesn't model well, I'd be curious to hear your thoughts.

driver/src/orderbook/streamed/balance.rs
e00E: The problem is that I want to guarantee that we don't modify anything if an error happens. We can error if the withdraw amount is larger than `balance_excluding_withdraw` in the `checked_sub`. If I had applied the deposit then the deposit would stay applied even though there was an error.

nlordell:
> This State does not handle removed events. The "state manager" is going to get events from the node and if there is a revert it will recreate the state from a previous snapshot and replay the new correct chain of events.

I see! It looks like we came to the same conclusion about snapshots :smile: and not dealing with past events.

> It is just checking as much as possible that we know has to hold for the real smart contract. Unrelated to reorderings. The events will get ordered before applying.

:+1:

> That is up to the "state manager". If it has 10 new events and the first one fails to apply it can still go ahead an apply the other ones hoping that the first one didn't influence state of the later oners.

Sounds reasonable. Maybe it makes sense for the state manager to "blacklist" certain user `Balance`s and exclude them entirely from the account state when they fail, since it means that future events will probably be modifying a incorrect state. Same goes for `Order`s.

nlordell:
From reading what you posted it looks like we came largely to the same conclusions. I think the only major difference would be that I would group events into batches before applying so that extra state like pending fluxes and the logic around that doesn't need to be included.

driver/src/orderbook/streamed/balance.rs
nlordell: But if we error doesn't that likely mean that this current `Balance` state will be in an incorrect state anyway?

driver/src/orderbook/streamed/balance.rs
nlordell: Sorry, I mean if user A has a balance of T OWL and makes a withdrawal request of T in batch X (that matures in batch X+1), then is user A's balance in batch X+1 T or 0?

The question is should withdrawal requests be updating the balance or should withdrawals be? In the UI it looks like a withdrawal request makes it so your balance is no longer in the exchange.

nlordell: But at the same time I know that there was a withdrawal denial attack where if a balance gets touched while there is a withdrawal request, then it changes its maturing batch to the one after (or something like that - I don't remember the exact details).

driver/src/orderbook/streamed/balance.rs
nlordell: :+1:

driver/src/orderbook/streamed/balance.rs
nlordell: Good point!

driver/src/orderbook/streamed/balance.rs
nlordell: If you use [`thiserror`](https://docs.rs/thiserror/1.0.15/thiserror/) you can add a `std::backtrace::Backtrace` field to your variants and the generated code takes care of the rest. It will include line number and the complete call stack in the error.

nlordell: This makes it so you need less variants since the actual call hierarchy gets captured by the event. It's basically the backtrace you see when you panic with `RUST_BACKTRACE=1`.

driver/src/orderbook/streamed/error.rs
nlordell: Perfect! That was exactly what I meant.

e00E:
> From reading what you posted it looks like we came largely to the same conclusions.

Yes.

> I think the only major difference would be that I would group events into batches before applying so that extra state like pending fluxes and the logic around that doesn't need to be included. 

> This would also reduce the complexity of the balance calculation since it wouldn't have to store state related to pending fluxes since it would know which events to apply and which ones to ignore (and it can just discard withdrawal (requests) and deposits that mature in future batches). For things like deposits where more than one can happen in the same batch, the events collection could be flattened to only include the latest one. This would make balances applied

I don't think this works. For multiple deposits we would need to accumulate together because each one increases the balance. For withdraw requests, we can't ignore them because they deduct from the available balance. Then there are trades and trade reverts too. In the end we have ended up entangling the "state manager" with the details of how `State` works. I'd rather not do that.
Also, one of the ideas for the event based orderbook was that we can optimistically start solving the next batch early by assuming that no more solutions are provided for the current batch. This won't work if we only give state the events after the batch is over.

nlordell:
> For multiple deposits we would need to accumulate together because each one increases the balance.

But we are applying one batch at a time, so we would get all the deposit events for that batch and be able to add them up together.

> For withdraw requests, we can't ignore them because they deduct from the available balance.

I meant ignoring multiple withdrawal requests without a withdrawal to clear it.

> Then there are trades and trade reverts too. In the end we have ended up entangling the "state manager" with the details of how State works.

I think I didn't explain myself too well - the idea would be that the state manager wouldn't do a `balance.deposit(...); balance.withdrawal(...);` but rather a `balance.apply_events(&events)` where events are guaranteed to be in order for a given batch and to all be matured in the current batch (so the state doesn't need to know what the current batch is ever - I think this makes things cleaner, although I may be ignoring some complexity I haven't thought about). The state manager would just be in charge or collecting events (and removing them on reorgs) and grouping them into batches and then the state would decide how to process a batch of events, so the state - state manager separation still holds.

> Also, one of the ideas for the event based orderbook was that we can optimistically start solving the next batch early by assuming that no more solutions are provided for the current batch. This won't work if we only give state the events after the batch is over.

This would be for snapshots, for the optimistic orderbook we would just take the latest snapshot, and apply all pending events to it.

nlordell:
> > For withdraw requests, we can't ignore them because they deduct from the available balance.

> I meant ignoring multiple withdrawal requests without a withdrawal to clear it.

Oh, I think I see the problem, withdrawal requests can span multiple batches... so maybe the withdrawal flux will always be needed.

nlordell:
> For multiple deposits we would need to accumulate together because each one increases the balance.

Correct, sorry my deposits example was incorrect.

driver/src/orderbook/streamed/balance.rs
e00E: I would like getting an error to be inexpensive. Getting a backtrace might be expesnive and it is going to be heap allocated. On the other hand errors are still rare and I haven't measured performance. Maybe we can keep this in mind if it turns out to would have been useful in the future.

nlordell: Errors should be exceptional cases, I don't think there is value optimizing the error path in our case. Also expensive is a relative term. In this case the cost is tiny compared to the time and memory it takes to perform the web3 request.

I would rather include this right away instead to make this section of code more debuggable. Also, `anyhow` creates backtraces on all errors, so its not like we don't already have that cost in other places.

driver/src/orderbook/streamed/balance.rs
e00E: In this code it is supposed to work like in the smart contract, so that the solver sees the balance that the smart contract also sees (like in the paginated orderbook). This is `get_balance`. In your example in batch X+1 `get_balance` returns 0 because the withdrawal request is valid.
If I understand you correctly you're asking whether we can simplify the code by deducting the withdrawal request balance immediately and ignoring withdraw events.
We cannot. Example:

1. it is batch X
2. user U has balance B0
3. user U requests withdrawal RW > B0
4. user U gains balance B1 from a trade
5. it is batch X + 1
6. user withdraws RW

The correct final balance here is `(B0 + B1) - RW`.

But if we ignored withdraws and applied withdraw requests immediately then we would set the balance to 0 in step 3, increase it to B1 from 0 in step 4, ignore the withdraw in step 6 and en dup with balance B1.

nlordell: > The correct final balance here is (B0 + B1) - RW.

But this is regardless of step 6 right? so the balance is `(B0 + B1) - RW` even without "user withdraws RW"?

driver/src/orderbook/streamed/balance.rs
e00E: It means that the event is not consistent with the state but the state is still consistent with the pats events it got.
From the perspective of just `State` doing it this way makes more sense. If you do not yet know how future code will use it then it makes sense that you don't invalidate the state on error because that code might want to reuse the state and try a different event. If we invalidated state on error this wouldn't be possible.
Even with specific knowledge of how state is probably going to be used in our code with the snapshot we might still want this. Like let's say we get a new event and it fails to apply to state and we decide to "blacklist" this event and keep going. Without the guarantee we would have to reconstruct from the snapshot because state got invalidated. With the guarantee we can keep using state for the next event.

driver/src/orderbook/streamed/balance.rs
e00E: Yes. I included that step to show that the user withdrew RW and not just B0. The point is that we cannot remove the withdraw request until we get the corresponding withdraw to see how much was actually withdrawn.

nlordell: So, lets say the balance gets updated with all the events in a batch at once, such that the `Balances` state no longer needs to expect any other event that will come in for a given batch. Can't the new balance simply be computed by as `new_balance = balance + sum(deposit amounts + trade buy amounts) - sum(withdrawal amounts + trade sell amounts)` and the pending withdrawal be `pending_withdrawal = if withdrawn since last request { 0 } else { last(withdrawal request) }` making `available balace () => new_balance.saturating_sub(pending_withdrawal)`

But I see your point that withdrawal events are also needed since they clear the pending withdrawals.

I still think having events come in a collection is easier to manage cognitively and is in general more powerful since the balances state will have all relevant information upfront.

driver/src/orderbook/streamed/balance.rs
e00E: I mean `State` specifically because it will have a lot of events flowing through it. It still just doesn't feel right to include the backtrace in an error enum but it is not a strong opinion of mine. So if @fleupold also prefers it I'll add it.

driver/src/orderbook/streamed/balance.rs
nlordell: So my point is that if state gets updated with event collections instead of singe events, then they can express their updates in simpler terms since have more information (the complete event list for a specific batch).

Note that this should still work as a continuous orderbook since the event collection will at that point just be the known events.

Another thing worth mentioning is that we can use `FutureExt::immediate` to get "ready" events so that the event stream can synchronously, without blocking, produce all events that came in a single `eth_getFilterChanges` call - so all the events within a block so you don't have to worry about partially applied blocks.

driver/src/orderbook/streamed/balance.rs
nlordell: > It means that the event is not consistent with the state but the state is still consistent with the past events it got.

This is not true though, because the `State` will not be valid for that batch anymore since it is ignoring events. So it should arguably not be considered at all anymore for any batch after the last successful snapshot it was included in.

> might want to reuse the state and try a different event.

In that case not modifying the state does make sense, but I think it makes sense to do it atomically by doing a `Balances::clone()` and then applying the event, and then replacing the balances in case it succeeds, as it would make the code easier to follow and have the state have to think less about side-effects of functions and in what order those side-effects get applied.

> With the guarantee we can keep using state for the next event.

But that would mean that ultimately the state at the end of the batch just ignored an event and is invalid anyway, and using that invalid state and passing it to the solver may lead to invalid solutions which is worse than just not including it, which could only lead to suboptimal solutions.

nlordell: I feel like this is another situation where the state would benefit from getting all of the events for a batch at once so it can apply things in the order that is easiest for it.

driver/src/orderbook/streamed/balance.rs
nlordell: The backtrace is boxed so it only takes `usize` size in the error and only gets allocated on error. This is nice since errors are exceptional cases and so this will, most of the time, have 0 cost.

driver/src/orderbook/streamed/balance.rs
e00E: I think this would work (but it's hard to tell without trying it) but the simplification comes from our not checking some errors that we are checking in the current version. If we don't check that a withdraw matches a withdraw request or doesn't withdraw too much then code is simpler but also less accurate. It would accept some order of events that could not really have occurred. I don't know how important this is but I went for accuracy when possible which is why there are so many error variants.
The one thing the batching removes is keeping track of deposits because indeed we can apply them all together. Pending withdraws we still need to keep track of because they can be for a future batch.
One problem I see now with batching is that we won't know when a batch is over. We will know if we get a new event that is in the next batch but we won't know know until then.

nlordell: > I went for accuracy when possible 

Good point, but I still think it would be easier to be accurate with an event collection as there is more information available to the state.

> One problem I see now with batching is that we won't know when a batch is over. We will know if we get a new event that is in the next batch but we won't know know until then.

We do based on block number. We can setup a block stream and then whenever we get a new block decide whether or not we can commit a snapshot (something like 5 blocks into the next batch would be ultra safe).

nlordell: > Pending withdraws we still need to keep track of because they can be for a future batch.

Correct, but you can flatten the collection of pending withdrawals since you only care about the last one.

nlordell: I do like your approach of checking every event as it comes in - I think its the best way to guarantee that the events are being applied correctly since the state would become invalid at some point if that weren't the case.

That being said, I think if we take the simpler less correct approach we can get up and running a bit quicker, and then still validate each event within the context of a single `apply_events` with some future added logic.

nlordell: > but the simplification comes from our not checking some errors that we are checking in the current version

I agree, the end goal should be to check everything so we can be sure our state is coherent, but I would suggest having that come later.

driver/src/orderbook/streamed/balance.rs
e00E: > In that case not modifying the state does make sense, but I think it makes sense to do it atomically by doing a Balances::clone() 

True. By not holding up the guarantee in Balance but in State we can simplify the code in Balance. It still feels better to me to do this in Balance too but I agree it would be simpler outside of it. If Balance contained heap data I would be more opposed to it but maybe I am too picky here. On the other hand, I don't think this function is that complicated either. We could take this point to the extreme and say we don't need to guarantee this in State either because we can clone State before applying an event. In that case clearly it would be too expensive - but for Balance not.

nlordell: Its copying a stack copy of less than 100 bytes, so basically free :wink:.

>  On the other hand, I don't think this function is that complicated either

you're right it isn't but I still strongly believe there is no point holding a `Balance` struct around once it errors - If any event is not applied to it then it is wrong and can't be included in the orderbook so I think we should just modify and if there is an error black list it.

The only way this extra complexity will be worth it is if the events are being applied in the wrong order, and then the events need to tentatively be applied before. In that case you will need to keep a queue of events that need to be re-applied for that specific balance in order for it to "catch back up". I think this will add a lot of extra complexity that is not necessarily useful yet.

driver/src/orderbook/streamed/balance.rs
nlordell: When `anyhow` wraps an error it either reuses the backtrace of the error if there is one, or creates its own. So I think in either case a backtrace will be created, if it is part of the `MathOverflow` error for example, then it will just be more correct.

fleupold (commented):
Damn, I forgot to submit my comments which I made yesterday or even the day before 😢 
It would be awesome if github could nag you that you have pending but unsubmitted comments.

Removed the ones that were outdated and added some thoughts on the good ongoing discussions.

e00E:
We decided to move to anyhow errors because they include a backtrace, remove the requirement that the state is not modified in case of error.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -1,64 +1,108 @@
...
-            }
-            None => self.pending_deposit = Some(new_deposit),
-        }
+    pub fn deposit(
+        &mut self,
+        amount: U256,
+        batch_id: BatchId,
+        current_batch_id: BatchId,
+    ) -> Result<()> {
+        ensure!(
```
nlordell, line 20: Feels like an `assert!` woot!

driver/src/orderbook/streamed/balance.rs
```diff
@@ -1,64 +1,108 @@
...
+            batch_id == current_batch_id,
```
nlordell, line 21: Should this be `batch_id <= current_batch_id`? I guess it depends on how the event will be applied.

driver/src/orderbook/streamed/state.rs
```diff
@@ -79,57 +81,47 @@ impl State {
 
     /// Apply an event to the state, modifying it.
     ///
-    /// In case of error no modifications take place.
+    /// Consumes and returns back `self` because it cannot be reused in case of error.
     ///
     /// `block_batch_id` is the current batch based on the timestamp of the block that contains the
     ///  event.
-    pub fn apply_event(&mut self, event: &Event, block_batch_id: BatchId) -> Result<(), Error> {
+    pub fn apply_event(mut self, event: &Event, block_batch_id: BatchId) -> Result<Self> {
```
nlordell, line 88: This very cool! Passing self by value ensures that we don't use invalid states on errors. I love it!

nlordell (approved):
Nice! I'm beginning to change my mind about the errors and assertions (or `ensure`-ances :stuck_out_tongue:). WIth `anyhow` it definitely feels a bit less invasive. They kind of document some assumptions of the account state in the code.

Great job :rocket:!

driver/src/orderbook/streamed/state.rs
```diff
@@ -79,57 +81,47 @@ impl State {
 
     /// Apply an event to the state, modifying it.
     ///
-    /// In case of error no modifications take place.
+    /// Consumes and returns back `self` because it cannot be reused in case of error.
```
fleupold, line 84: That's a cool way of enforcing you don't work with the state in case of an error.

fleupold (approved):
Looks great! Love the simple control flow.

driver/src/orderbook/streamed/balance.rs
e00E: The contract always sets the deposit's batch id to `getCurrentBatchId` so I think it's correct.

nlordell: Yes definitely! I meant if you were applying past deposits and your current batch was in the present, but if the events always get applied with the correct “current batch ID” then my point is moot.

driver/src/orderbook/streamed/state.rs
```diff
@@ -44,7 +44,9 @@ impl State {
                 let token_id = self.tokens.get_id_by_address(*token_address)?;
                 Some((
                     (*user_id, token_id),
-                    balance.current_balance(batch_id).low_u128(),
+                    // TODO: Remove unwrap. Can fail while not all trades of a solution have been
+                    // received. Can fail if user's balance exceeds U256::max.
+                    balance.get_balance(batch_id).unwrap().low_u128(),
```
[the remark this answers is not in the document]
e00E: Currently `AccountState` is u128 which is why it is used here. I don't know why we aren't using U256. I'll change it to U256 here and then later code decide to convert to u128.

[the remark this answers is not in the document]
nlordell: We should probably change it in the rest of the codebase as well as a separate PR. I’ll create an issue so we don’t forget.

PR #736, opened 2020-04-18
Retrieve orderbook on last block of solving batch.

nlordell:
This PR introduces code to search for the last block of the solving batch and retrieve the orderbook on that batch. This makes so the open solver and EVM-scheduler solvers that start later than system time solvers don't potentially retrieve an invalid orderbook that contains solution trades.

### Test Plan

CI and let it run in staging for a bit to see what happens. We have tried this before, but ran into issues where we were not able to query state at a specific block. The theory is that the load balancer was connecting different requests to different nodes that had not yet seen the block.

If we run into this, we can try enforcing a keep-alive policy to ensure that these queries get done to the same node.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -77,6 +77,10 @@ pub trait StableXContract {
     /// Retrieve the time remaining in the batch.
     fn get_current_auction_remaining_time(&self) -> Result<Duration>;
 
+    /// Searches for the block number of the last block of the given batch. If
+    /// the batch has not yet been finalized, then `None` is returned.
+    fn get_last_block_for_batch(&self, batch_id: u32) -> Result<Option<u64>>;
+
     /// Retrieve one page of indexed auction data that is filtered on chain
     /// to only include orders valid at the given batchId.
     fn get_filtered_auction_data_paginated(
@@ -86,11 +90,8 @@ pub trait StableXContract {
         page_size: u16,
         previous_page_user: Address,
         previous_page_user_offset: u16,
+        block_number: Option<BlockNumber>,
     ) -> Result<FilteredOrderPage>;
-
-    /// Searches for the block number of the last block of the given batch. If
-    /// the batch has not yet been finalized, then `None` is returned.
-    fn get_last_block_for_batch(&self, batch_id: u32) -> Result<Option<u64>>;
 
     /// Retrieve one page of auction data.
     /// `block` is needed because the state of the smart contract could change
@@ -136,34 +137,6 @@ impl StableXContract for StableXContractImpl {
         Ok(Duration::from_secs(remaining_seconds.as_u64()))
     }
 
-    fn get_filtered_auction_data_paginated(
-        &self,
-        batch_index: U256,
-        token_whitelist: Vec<u16>,
-        page_size: u16,
-        previous_page_user: Address,
-        previous_page_user_offset: u16,
-    ) -> Result<FilteredOrderPage> {
-        let target_batch = batch_index.low_u32();
-        let mut builder = self.viewer.get_filtered_orders_paginated(
-            // Balances should be valid for the batch at which we are submitting (target batch + 1)
-            [target_batch, target_batch, target_batch + 1],
-            token_whitelist,
-            previous_page_user,
-            previous_page_user_offset,
-            page_size,
-        );
-        builder.m.tx.gas = None;
-        let (indexed_elements, has_next_page, next_page_user, next_page_user_offset) =
-            builder.call().wait()?;
-        Ok(FilteredOrderPage {
-            indexed_elements,
-            has_next_page,
-            next_page_user,
-            next_page_user_offset,
-        })
-    }
-
     fn get_last_block_for_batch(&self, batch_id: u32) -> Result<Option<u64>> {
         let web3 = self.instance.raw_instance().web3();
         let get_block = |block_number: BlockNumber| -> Result<_> {
@@ -192,6 +165,36 @@ impl StableXContract for StableXContractImpl {
         }
 
         Ok(block_number)
+    }
+
+    fn get_filtered_auction_data_paginated(
+        &self,
+        batch_index: U256,
+        token_whitelist: Vec<u16>,
+        page_size: u16,
+        previous_page_user: Address,
+        previous_page_user_offset: u16,
+        block_number: Option<BlockNumber>,
+    ) -> Result<FilteredOrderPage> {
+        let target_batch = batch_index.low_u32();
+        let mut builder = self.viewer.get_filtered_orders_paginated(
+            // Balances should be valid for the batch at which we are submitting (target batch + 1)
+            [target_batch, target_batch, target_batch + 1],
+            token_whitelist,
+            previous_page_user,
+            previous_page_user_offset,
+            page_size,
+        );
+        builder.block = block_number;
+        builder.m.tx.gas = None;
+        let (indexed_elements, has_next_page, next_page_user, next_page_user_offset) =
+            builder.call().wait()?;
+        Ok(FilteredOrderPage {
+            indexed_elements,
+            has_next_page,
+            next_page_user,
+            next_page_user_offset,
+        })
     }
 
     fn get_auction_data_paginated(
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -174,10 +174,14 @@ impl StableXContract for StableXContractImpl {
                 .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
         };
 
-        let mut block_number = None;
         let mut current_block = get_block(BlockNumber::Latest)?;
-        while (batch_id as u64) < current_block.timestamp.as_u64() / BATCH_DURATION {
-            let previous_block_number = current_block.number.ok_or_else(|| anyhow!(""))? - 1;
+        let mut block_number = None;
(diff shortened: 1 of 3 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -108,20 +114,45 @@ impl StableXContract for StableXContractImpl {
...
+        let get_block = |block_number: BlockNumber| -> Result<_> {
+            web3.eth()
+                .block(block_number.into())
+                .wait()?
+                .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
+        };
+
+        let mut current_block = get_block(BlockNumber::Latest)?;
+        let mut block_number = None;
+        while (batch_id as u64) < current_block.timestamp.as_u64() / BATCH_DURATION
```
fleupold, line 130: Can `current_block.timestamp.as_u64() / BATCH_DURATION` become a utility function since we probably use this logic elsewhere to.

driver/src/contracts/stablex_contract.rs
```diff
@@ -108,20 +114,45 @@ impl StableXContract for StableXContractImpl {
...
+            && current_block.number != Some(0.into())
+        {
+            let previous_block_number = current_block.number.ok_or_else(|| {
+                anyhow!("block {:?} has a missing block number", current_block.hash)
+            })? - 1;
```
fleupold, line 135: I feel like this is a classic interview question, and my hunch would be that the optimal strategy involves doubling the gap and then doing a binary search when we overshot the target block time.
Not sure this is necessary here though since we should always be pretty close to the needle.

driver/src/contracts/stablex_contract.rs
```diff
@@ -65,6 +66,10 @@ pub trait StableXContract {
     /// Retrieve the time remaining in the batch.
     fn get_current_auction_remaining_time(&self) -> Result<Duration>;
 
+    /// Searches for the block number of the last block of the given batch. If
+    /// the batch has not yet been finalized, then `None` is returned.
+    fn get_last_block_for_batch(&self, batch_id: u32) -> Result<Option<u64>>;
```
fleupold, line 71: I think what we want is `first_block_for_batch`, since in the last block of the previous batch all balances are still wrong (they reflect the balances for the solver in that batch). There might be pending withdraws/deposits that will mature in the next block and should be available to the solver.

This means that there will technically still be a race condition since in the first block of a solving process, there could already be a valid solution applied.

driver/src/contracts/stablex_contract.rs
```diff
@@ -108,20 +114,45 @@ impl StableXContract for StableXContractImpl {
...
+
+        let web3 = self.instance.raw_instance().web3();
+        let get_block = |block_number: BlockNumber| -> Result<_> {
+            web3.eth()
+                .block(block_number.into())
+                .wait()?
+                .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
+        };
+
+        let mut current_block = get_block(BlockNumber::Latest)?;
```
fleupold, line 128: Maybe we should start with pending here?

fleupold (commented):
Really nice PR.

I think what we need is the **first** block in the new batch (in order to get correct balances for the batch that is currently being solved). This means there is still a small chance for a race condition but I think that the simplicity of this proposal over the other one (fetching latest solution, and listening on invalidation events) outweighs the disadvantage.

Alternatively, we could use the `getOpenOrderBooksPaginated` method on the viewer contract which adjusts volumes for the fact that they should be targeted at the next batch (but this method hasn't really been battle tested and shown some issues in the price estimation service).

driver/src/contracts/stablex_contract.rs
nlordell: I think that's a great idea, I would implement it as a separate PR though.

driver/src/contracts/stablex_contract.rs
nlordell: Can we use the `getOpenOrderbook` method from the viewer contract then?

fleupold: >  Can we use the getOpenOrderbook method from the viewer contract then?

Yes, I might spend some time today to write a small script which compares the results of `getOpenOrderbookPaginated` on the last block of the previous batch with `getFinalizedOrderbookPaginated` on the first block of the next batch and `getEncodedOrderPaginated` (with client side filtering invalid orders) to get some more confidence in the correctness of these methods (given our issues with them last week).

fleupold: Script is running so we should have data soon. Note, that this also depends on https://github.com/gnosis/dex-contracts/pull/668 in order to reconstruct the orderIds that are required in `submitSolution`

driver/src/contracts/stablex_contract.rs
nlordell: I tried this and ran into some issues. So, from what I understood we want the entire orderbook to be retrieved at a specific block height. That means that we need an actual block number to retreive the orderbook at. If we use pending, it is possible that a solution gets submitted partway through the orderbook and modifies orders and balances that have not yet been queried and then get included once we reach that page (the current problem we have). Also, querying at block `latest + 1` doesn't seem to work.

driver/src/contracts/stablex_contract.rs
bh2smith: I’m not sure what exactly the interview problem coming from this is meant to be? I can only assume it has something to do with getting blocks.

nlordell: Its kind of doing a binary search on the block :smile:

nlordell:
Currently doesn't work with ganache since the balances aren't yet credited when we read the orderbook since we are reading in the previous batch. Should be fixed once we use the `getOpenOrdersPaginated` request.

driver/src/contracts/stablex_contract.rs
nlordell: Yeah, that was an oversight. If we don't use "pending" here, this method might return `None` when that last block in the previous batch is indeed "latest".

fleupold:
Updated the code to use the onchain filter who can correctly compute target balances for the batch we are trying to solve. Please take another look.

Ran it locally and shadows seem to match on rinkeby. After merging, I will enable shadow traffic on the staging standard solver.

driver/src/contracts/stablex_contract.rs
nlordell: @fleupold Where the awesome search strategy?! :stuck_out_tongue_closed_eyes:

driver/src/contracts/stablex_contract.rs
```diff
@@ -160,15 +202,14 @@ impl StableXContract for StableXContractImpl {
         page_size: u16,
         previous_page_user: Address,
         previous_page_user_offset: u16,
+        block_number: Option<BlockNumber>,
```
nlordell, line 205: maybe the change to this method should be reverted since we shouldn't use it anymore?

nlordell (commented):
Can't approve my own PR but, approved!

driver/src/contracts/stablex_contract.rs
fleupold: I didn't want to take the fun stuff away from you 😉 More seriously, this is probably something we want to test and thus requires some more abstraction which would go beyond the scope of this PR.

driver/src/contracts/stablex_contract.rs
fleupold: I feel the Reader should be in charge of setting the block (like in the method above). How about removing the .or(Some(BlockNumber::Pending)) from here and moving it to the PaginatedOrderbookReader so that the two APIs are consistent?

driver/src/contracts/stablex_contract.rs
nlordell: I'm curious about the reason we switched back to latest?

fleupold: There is an issue with web3-rs `getBlock` when specifying `BlockNumber::Pending`. 

>  Apr 22 18:28:21.080 ERRO [driver::orderbook::shadow_orderbook] encountered an error reading the orderbook with the shadow reader: Decoder error: Error("invalid type: null, expected a 0x-prefixed hex string with length of 512", line: 0, column: 0)

I will investigate more and file an issue in their repo. For now, I think Latest should work.

fleupold (approved):
I guess I can approve this PR.

fleupold:
I'm going to merge this to get some data for it today. @bh2smith @e00E feel free to add you review after it's merged and we can address in a follow up PR.

driver/src/contracts/stablex_contract.rs
bh2smith: I see, the binary search is the "way we should do it" but for now we will simply step backwards one at a time.

bh2smith (commented):
Looks good, but my only concern now is that raised by my own recent questions regarding how getting the order book from the last block of the previous batch would not come along with the correct balances. Is this still a concern, or do we plan to update the account state manually?

fleupold:
If you invoke use `getEncodedOrdersPaginated()` you receive the orders and balances for the batch that is currently being solved (`currentBatchId() - 1`). Therefore, if you invoke this method in the context of the last block of the previous batch you are actually getting orders and balances as of the batch before the one you are trying to solve.

With the onchain filtering methods (in particular `getFilteredOrdersPaginated`) you can provide a "target BatchID" and then all balances and orders will be for that target batch. Then you can actually query state in the context of the last block of the previous batch, by supplying the correct "future" target batchID.

PR #740, opened 2020-04-20
Event based orderbook: get events from node

e00E:
This is a draft for whoever finishes this implementation. Probably @bh2smith or myself.


e00E:
I'm going to split this up into reviewable parts immediately.

PR #741, opened 2020-04-20
[event based orderbook] Add event methods to StablexContract

e00E:


driver/src/contracts/stablex_contract.rs
```diff
@@ -189,6 +200,29 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
...
+        &self,
+    ) -> BoxFuture<'static, Result<Vec<Event<batch_exchange::Event>>, ExecutionError>> {
+        Box::pin(
+            self.instance
+                .all_events()
+                .from_block(ethcontract::BlockNumber::Earliest)
+                .to_block(ethcontract::BlockNumber::Latest)
+                .query(),
+        )
+    }
```
e00E, line 225: In the future we will probably implement a better way to get past events so that they are not all in returned in one call. But not sure whether we do that here or in ethcontract so I am using this for now.

nlordell: Yeah, its a tough problem. For me to get it to work with Infura, I had to pick a block range and just chain a bunch of calls to get these pages one a time.

driver/src/contracts/stablex_contract.rs
```diff
@@ -189,6 +200,29 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
 
         Ok(())
     }
+
+    fn stream_events(
+        &self,
+    ) -> BoxStream<'static, Result<Event<batch_exchange::Event>, ExecutionError>> {
+        Box::pin(
```
nlordell, line 207: nit: `futures::stream::StreamExt` has a `boxed()` extension method for this.

driver/src/contracts/stablex_contract.rs
```diff
@@ -189,6 +200,29 @@ impl<'a> StableXContract for StableXContractImpl<'a> {
...
+            self.instance
+                .all_events()
+                .from_block(BlockNumber::Earliest)
```
nlordell, line 210: Should this be `latest` since it doesn't retrieve past events anyways? I find it would better document what the stream is actually producing.

nlordell (approved):
Easy!

driver/src/contracts/stablex_contract.rs
fleupold: Interesting question if this is something that could be abstracted away in ethcontracts (maybe for now we implement the logic here, but could push it upstream later?).

How do the nodes behave if we query too much data? Do they just cap at 10k event or do they error out? Is this node-dependent or rather standard? We should probably address this issue before merging this PR.

fleupold (commented):
I think we should find a solution to getting the past_events in this PR as this will already be an issue on mainnet.

driver/src/contracts/stablex_contract.rs
nlordell: > How do the nodes behave if we query too much data? Do they just cap at 10k event or do they error out?

Infura errors out, not sure about our parity nodes.

bh2smith (approved):
Seems reasonable.

driver/src/contracts/stablex_contract.rs
nlordell: > this is something that could be abstracted away in ethcontracts 

I think that is where the implementation should live. I'm going to work on adding it there so it can be used here.

bh2smith (approved):
Everything appears to be ready for merge around here. I will be happy to put this major feature to ~~rest~~ work.

fleupold (approved):

driver/src/contracts/stablex_contract.rs
```diff
@@ -120,6 +125,14 @@ pub trait StableXContract {
...
         block_timeout: Option<usize>,
     ) -> Result<(), MethodError>;
+
+    fn past_events(
+        &self,
+    ) -> BoxFuture<'static, Result<Vec<Event<batch_exchange::Event>>, ExecutionError>>;
+
+    fn stream_events(
+        &self,
+    ) -> BoxStream<'static, Result<Event<batch_exchange::Event>, ExecutionError>>;
```
e00E, line 135: @nlordell You changed this to return a lifetime tied to self. Why? The futures we get from ethcontract are `'static` after all so we might as well show that in the interface.

nlordell: The owned events definitely are, but the `query_past_events_paginated` was written with `async` and I wasn't sure if the generated `Future` would have lifetimes tied to self.

e00E: But rust would prevent us from having dangling references, no?

nlordell: Yes, but as far as I know a future can still have a lifetime that is tied to one of its parameters (like self or some buffer it’s supposed to write to).

I just thought ‘static added some extra unneeded restrictions to the trait but in this case it really isn’t a problem

PR #742, opened 2020-04-20
[getOpenOrderbook 1] Split PaginatedAuctionDataReader

fleupold:
We are using the `PaginatedAuctionDataReader` abstraction to process pages that we fetch via `getEncodedOrdersPaginated`. I would like to reuse that abstraction for the open orderbook fetching which filters valid orders on-chain.

The pagination logic for this way of fetching is done differently (also on-chain). This PR therefore splits the functionality of PaginatedAuctionDataReader into two parts

1) AuctionDataReader which just applies pages and knows nothing about pagination
2) PaginatedAuctionDataReader which uses 1) for page application and on top of that keeps track of page info

This way we will be able to reuse component 1) in the next PR

I slightly changed the semantics of the component in so far as that instead of returning the elements inside the unfiltered page (which would not be available in the open orderbook fetching) we know make the pagination field optional. None will indicate that there is no next page.

### Test Plan

Adjusted unit tests and made sure they are still passing.

### DIFF
--- a/driver/src/orderbook/auction_data_reader.rs
+++ b/driver/src/orderbook/auction_data_reader.rs
@@ -95,7 +95,7 @@ impl AuctionDataReader {
         packed_auction_bytes
             .chunks(AUCTION_ELEMENT_WIDTH)
             .map(|chunk| {
-                let mut result = auction_element_from_chunk(chunk);
+                let mut result = auction_element_from_slice(chunk);
                 let order_counter = self
                     .user_order_counts
                     .entry(result.order.account_id)
@@ -158,7 +158,7 @@ impl PaginatedAuctionDataReader {
             return;
         }
 
-        let previous_page_user = auction_element_from_chunk(
+        let previous_page_user = auction_element_from_slice(
             &packed_auction_bytes
                 [packed_auction_bytes.len() - AUCTION_ELEMENT_WIDTH..packed_auction_bytes.len()],
         )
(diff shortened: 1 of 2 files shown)


e00E (approved):

driver/src/orderbook/auction_data_reader.rs
```diff
@@ -136,9 +95,7 @@ impl PaginatedAuctionDataReader {
         packed_auction_bytes
             .chunks(AUCTION_ELEMENT_WIDTH)
             .map(|chunk| {
-                let mut chunk_array = [0; AUCTION_ELEMENT_WIDTH];
-                chunk_array.copy_from_slice(chunk);
-                let mut result = StableXAuctionElement::from_bytes(&chunk_array);
+                let mut result = auction_element_from_chunk(chunk);
```
nlordell, line 98: naming nit: I would call this `..._from_slice` or `..._from_bytes` to be more consistent with Rust naming.

nlordell (approved):
Looks good to me.

driver/src/orderbook/auction_data_reader.rs
```diff
@@ -151,6 +108,89 @@ impl PaginatedAuctionDataReader {
...
+                previous_page_user: Address::zero(),
+                previous_page_user_offset: 0,
+            }),
+            page_size,
+        }
+    }
+
+    /// The pagination data used when reading data from the smart contract in
+    /// batches.
+    pub fn pagination(&self) -> Option<&Pagination> {
```
nlordell, line 147: uber-nano naming nit: I feel like `next_page` is a little more appropriate.

PR #743, opened 2020-04-20
[Shadow Orderbook] Implement orderbook diffing algorithm.

nlordell:
This PR impements a simple diffing algorithm for orderbooks that will eventually be used to determine if two orderbooks are equivalent or not.

Note that this code is currently unused.

### Test Plan

New unit test!


driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -0,0 +1,312 @@
...
+        match self {
+            OrderChange::Added(order) | OrderChange::Removed(order) => (order.account_id, order.id),
+            OrderChange::Modified { user, id, .. } => (*user, *id),
+        }
+    }
+}
+
+/// Values that can possibly differ between orders.
+#[derive(Debug, PartialEq)]
+struct OrderValues {
```
fleupold, line 153: I'd be fine reusing the Order struct since address and userId will also match, but you are right that this is a little cleaner.

It would be good to also have the validity information on these orders (to fetch race condition with cancellations) but I suppose this would involve more piping work as we currently don't populate the on the orders.

fleupold (approved):
Very nice! I would have probably just printed their json serialization and used an external diffing tool to find the difference, but this is much more sophisticated.

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// two.
+    ///
+    /// The diff is expressed in changes in the primary orderbook relative to
+    /// the orderbook read with the reader.
+    fn compare_to_reader(
+        primary_orderbook: &Orderbook,
+        reader: &dyn StableXOrderBookReading,
+        batch_id: u32,
+    ) -> Result<Self> {
+        let shadow = reader.get_auction_data(batch_id.into())?;
```
fleupold, line 35: I'm wondering if it would be cleaner to have the `Diff` struct only take care of diffing and pass it in the orderbooks already fetched. Not a huge difference, but maybe it encapsulates responsibilities in a more straightforward way.

driver/src/orderbook/shadow_orderbook.rs
nlordell: I wanted the type to capture that only those values can be different between orders (since user ID and order ID are used to identify them).

Yeah, we currently don’t provide validity information from the readers. I think it’s fairly easy to implement though, so I’ll add it as a separate PR.

driver/src/orderbook/shadow_orderbook.rs
nlordell: Oh yeah! That makes perfect sense. I’ll change it.

e00E (approved):

nlordell:
> I would have probably just printed their json serialization and used an external diffing tool to find the difference, but this is much more sophisticated.

I had considered that. One issue for balances, the absence of a balance is equivalent to a 0 balance so we would also have to ensure that all keys appear in both.

I can switch to shelling out to `diff` or some JSON specific diffing program if you prefer! I thought the implementation was simple enough that it could be included.

fleupold:
> I can switch to shelling out to diff or some JSON specific diffing program if you prefer! I thought the implementation was simple enough that it could be included.

This might have come across wrong, I also like this implementation and was positively surprised how simple a diffing algorithm can be!

driver/src/models/account_state.rs
```diff
@@ -9,6 +9,10 @@ impl AccountState {
     pub fn read_balance(&self, token_id: u16, account_id: Address) -> u128 {
         self.0.get(&(account_id, token_id)).cloned().unwrap_or(0)
     }
+
+    pub fn user_token_pairs(&self) -> impl Iterator<Item = (Address, u16)> + '_ {
+        self.0.iter().map(|(&pair, _)| pair)
+    }
```
bh2smith, line 15: Is there a test for this? I find this method so abstract that I am unable to determine what is should be returning. Would really help if I had an "expectation" example.

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+impl BalanceChange {
+    /// Compare a primary and shadow orderbook account state and return the
+    /// changed balances.
+    fn compare_account_state(primary: &AccountState, shadow: &AccountState) -> Vec<BalanceChange> {
+        let user_token_pairs = primary
+            .user_token_pairs()
+            .chain(shadow.user_token_pairs())
+            .collect::<HashSet<_>>();
```
bh2smith, line 53: Nice!

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -0,0 +1,295 @@
...
+                    shadow: OrderValues::from(shadow_order),
+                },
+                _ => continue,
+            };
+            changes.push(change);
+        }
+
+        for (_, remaining_shadow_order) in shadow_orders {
+            changes.push(OrderChange::Removed(remaining_shadow_order.clone()))
+        }
```
bh2smith, line 119: This is really great! Nice clean logic here.

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    /// Retrieves the user and order ID for the order change.
+    #[cfg(test)]
+    fn user_and_id(&self) -> (Address, u16) {
+        match self {
+            OrderChange::Added(order) | OrderChange::Removed(order) => (order.account_id, order.id),
+            OrderChange::Modified { user, id, .. } => (*user, *id),
+        }
+    }
+}
```
bh2smith, line 132: Is this a test or a helper method for a testing... Does the `cfg(test)` label mean that this method is implemented for `OrderChange` but can only be used from within a test? If so, very cool.

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -0,0 +1,295 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn account_state_diff() {
+        let addr = |i: u8| Address::repeat_byte(i);
```
bh2smith, line 160: my first accurately labeled UBERNIT: 

Any particular reason the bytes are chosen to be repeated? Is this the simplest method offered by Address to construct something quick? I just wonder because I guess this makes `addr(1) = 0x11111.11111`  rather than `0x0000...1` so if someone is expected to get the "1-Address", they would be confused to discover that it is not `addr(1)`

bh2smith (commented):
This is way cool!

driver/src/orderbook/shadow_orderbook.rs
nlordell: That is exactly what its for :smile:.

driver/src/orderbook/shadow_orderbook.rs
nlordell: It would make it `0x010101...01`. I had used `Address::from_low_u64_le(number)` but when seeing the `assert_eq` failure messages, it was very hard to keep track of the addresses since they were mostly 0's and with just 1 important number at the very end. It was especially hard when the 0's were split with a newline because of terminal width.

Maybe I can rename the closure to something more descriptive? But in general I do prefer using `repeat_byte` here as it makes the assertion failures easier to read.

bh2smith: Nah, this seems like pretty good rational to me!

PR #744, opened 2020-04-20
[Shadow Orderbook] Orderbook reading implementation

nlordell:
This PR implements `StableXOrderbookReading` for the "shadow" orderbook which works by sending a clone of the orderbook computed by the primary orderbook reader to a background thread that reads the orderbook at the same batch with the shadow orderbook and then comparing the two orderbooks and producing a reduced diff that is logged to the error console (for slack events :rotating_light:).

Note that this implementation is not yet integrated with the rest of the driver.

### Test Plan

 Not sure what the best way to test the threading code is... The diffing of orderbooks is already tested :smile:

### DIFF
--- a/driver/src/orderbook/shadow_orderbook.rs
+++ b/driver/src/orderbook/shadow_orderbook.rs
@@ -12,6 +12,7 @@ use crate::models::{AccountState, Order, TokenId};
 use anyhow::Result;
 use ethcontract::{Address, U256};
 use std::collections::{HashMap, HashSet};
+use std::fmt;
 use std::sync::mpsc::{self, Receiver, SyncSender};
 use std::thread::{self, JoinHandle};
 
@@ -90,10 +91,9 @@ fn background_shadow_reader(
     reader: &dyn StableXOrderBookReading,
     channel: Receiver<(u32, Orderbook)>,
 ) {
-    while let Ok((batch_id, orderbook)) = channel.recv() {
-        let diff = match Diff::compare_to_reader(&orderbook, reader, batch_id) {
-            Ok(diff) if diff.is_empty() => continue,
-            Ok(diff) => diff,
+    while let Ok((batch_id, primary_orderbook)) = channel.recv() {
+        let shadow_orderbook = match reader.get_auction_data(batch_id.into()) {
+            Ok(orderbook) => orderbook,
             Err(err) => {
                 log::error!(
                     "encountered an error reading the orderbook with the shadow reader: {:?}",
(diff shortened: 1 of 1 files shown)


driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -10,12 +10,140 @@
...
+        for balance_change in balance_changes {
+            log::error!(
+                "user {:?} token {} primary balance of {} different than shadow {}",
+                balance_change.user,
+                balance_change.token.0,
+                balance_change.primary,
+                balance_change.shadow,
+            );
+        }
+        for order_change in order_changes {
```
fleupold, line 116: Would it be better to implement Display on OrderChange and use that here?

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -10,12 +10,140 @@
...
+    pub fn new(
+        primary: &'a dyn StableXOrderBookReading,
+        shadow: Box<dyn StableXOrderBookReading + Send + 'static>,
+    ) -> Self {
+        // NOTE: Create a bounded channel with a 0-sized buffer, this makes it
+        //   if the primary orderbook is read and the shadow is still reading,
+        //   the diff for that specific orderbook is skipped.
+        let (shadow_channel_tx, shadow_channel_rx) = mpsc::sync_channel(0);
+        let shadow_thread =
+            thread::spawn(move || background_shadow_reader(&*shadow, shadow_channel_rx));
```
fleupold, line 41: Not necessary in this PR, but if I wanted to test this class, I would try to implement a small TaskQueue abstraction (maybe some lightweight implementation already exists) that can be initialized with thread/channels for actually running the program, but that can also have a fake synchronous implementation we it directly calls the lambda on `try_send`.

This would allow testing the logic of invoking the secondary reader once the primary completes without any thread/wait logic.

fleupold (approved):
🌚

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -10,12 +10,140 @@
...
+
+        ShadowedOrderbookReader {
+            primary,
+            shadow_thread,
+            shadow_channel: shadow_channel_tx,
+        }
+    }
+
+    /// Eplicitely stop the shadowed orderbook reader returning an error if the
```
e00E, line 50:
```suggestion
    /// Explicitely stop the shadowed orderbook reader returning an error if the
```

e00E (approved):

driver/src/orderbook/shadow_orderbook.rs
nlordell: Oh yeah! That's what Display is there for!

driver/src/orderbook/shadow_orderbook.rs
nlordell: >  I would try to implement a small TaskQueue abstraction 

Thats a neat idea, I'll give that a shot in a separate PR then.

bh2smith:
Why was this closed? I don't understand.

nlordell:
@bh2smith Oops! It looks like "merge-when-green" closed it when the branch it was merging to closed, and I didn't notice.
These changes should be included with #745 though.

PR #745, opened 2020-04-21
[Shadow Orderbook] Integrate the shadowed orderbook with the driver

### LINKED ISSUE #684: Reliably Fetch Account State

nlordell:
This PR makes the driver actually use the shadow orderbook. This means that for each batch, the orderbook will be queried twice and compared to one another.

This PR also changes the paginated orderbook to take ownership of its contract instance. This was to simplify the code around the lifetimes of the shadowed orderbook which must live in its own thread.

I had originally planed to use `crossbeam` scoped threads but that required the scope to be created in the `main` method which complicated things a bit and leaked some of the implementation details out of the shadowed orderbook.

### Test Plan

CI - this will make sure (at least on Rinkeby) that the orderbook being retrieved is the same at the beginning of the batch as it is in the middle of the batch. If this runs in staging it will also check this for mainnet. While, right now, this is not true in general (#684) it will be a useful test to see if #736 indeed fixes #684 once it is implemented.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -23,7 +23,10 @@ use crate::driver::{
 use crate::gas_station::GnosisSafeGasStation;
 use crate::http::HttpFactory;
 use crate::metrics::{HttpMetrics, MetricsServer, StableXMetrics};
-use crate::orderbook::{FilteredOrderbookReader, OrderbookFilter, PaginatedStableXOrderBookReader};
+use crate::orderbook::{
+    FilteredOrderbookReader, OrderbookFilter, PaginatedStableXOrderBookReader,
+    ShadowedOrderbookReader,
+};
 use crate::price_estimation::{PriceOracle, TokenData};
 use crate::price_finding::{Fee, SolverType};
 use crate::solution_submission::StableXSolutionSubmitter;
@@ -202,10 +205,15 @@ fn main() {
     let price_finder = price_finding::create_price_finder(fee, options.solver_type, price_oracle);
 
     // Create the orderbook reader.
-    let orderbook = PaginatedStableXOrderBookReader::new(&contract, options.auction_data_page_size);
-    info!("Orderbook filter: {:?}", options.orderbook_filter);
+    let primary_orderbook =
+        PaginatedStableXOrderBookReader::new(contract.clone(), options.auction_data_page_size);
+    let shadow_orderbook =
+        PaginatedStableXOrderBookReader::new(contract.clone(), options.auction_data_page_size);
+    let shadowed_orderbook = ShadowedOrderbookReader::new(&primary_orderbook, shadow_orderbook);
 
-    let filtered_orderbook = FilteredOrderbookReader::new(&orderbook, options.orderbook_filter);
+    info!("Orderbook filter: {:?}", options.orderbook_filter);
+    let filtered_orderbook =
+        FilteredOrderbookReader::new(&shadowed_orderbook, options.orderbook_filter);
 
     // Set up solution submitter.
     let solution_submitter = StableXSolutionSubmitter::new(&contract, &gas_station);
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -25,7 +25,7 @@ use crate::http::HttpFactory;
 use crate::metrics::{HttpMetrics, MetricsServer, StableXMetrics};
 use crate::orderbook::{
     FilteredOrderbookReader, OrderbookFilter, PaginatedStableXOrderBookReader,
-    ShadowedOrderbookReader,
+    ShadowedOrderbookReader, StableXOrderBookReading,
 };
 use crate::price_estimation::{PriceOracle, TokenData};
 use crate::price_finding::{Fee, SolverType};
@@ -162,6 +162,12 @@ struct Options {
         parse(try_from_str = duration_secs),
     )]
     price_source_update_interval: Duration,
+
+    /// Use a shadowed orderbook reader along side a primary reader so that the
+    /// queried data can be compared and produce log errors in case they
+    /// disagree.
+    #[structopt(long, env = "USE_SHADOWED_ORDERBOOK")]
+    use_shadowed_orderbook: bool,
 }
 
 fn main() {
@@ -195,8 +201,9 @@ fn main() {
     .unwrap();
 
     // Set up web3 and contract connection.
-    let contract =
-        StableXContractImpl::new(&web3, options.private_key.clone(), options.network_id).unwrap();
+    let contract = Arc::new(
+        StableXContractImpl::new(&web3, options.private_key.clone(), options.network_id).unwrap(),
+    );
     info!("Using contract at {:?}", contract.address());
     info!("Using account {:?}", contract.account());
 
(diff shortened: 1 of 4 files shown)


driver/src/main.rs
```diff
@@ -202,10 +205,15 @@ fn main() {
     let price_finder = price_finding::create_price_finder(fee, options.solver_type, price_oracle);
 
     // Create the orderbook reader.
-    let orderbook = PaginatedStableXOrderBookReader::new(&contract, options.auction_data_page_size);
-    info!("Orderbook filter: {:?}", options.orderbook_filter);
+    let primary_orderbook =
+        PaginatedStableXOrderBookReader::new(contract.clone(), options.auction_data_page_size);
```
fleupold, line 209: Instead of cloning the contract would it be better to use an RC here instead?

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -30,15 +30,15 @@ impl<'a> ShadowedOrderbookReader<'a> {
     /// Create a new instance of a shadowed orderbook reader that starts a
     /// background thread
     pub fn new(
-        primary: &'a dyn StableXOrderBookReading,
-        shadow: Box<dyn StableXOrderBookReading + Send + 'static>,
+        primary: &'a (dyn StableXOrderBookReading + Sync),
+        shadow: impl StableXOrderBookReading + Send + 'static,
```
fleupold, line 34: Would it be semantically simpler of both of these were passed in the same way (with ownership)?

fleupold (approved):

driver/src/main.rs
```diff
@@ -202,10 +205,15 @@ fn main() {
     let price_finder = price_finding::create_price_finder(fee, options.solver_type, price_oracle);
 
     // Create the orderbook reader.
-    let orderbook = PaginatedStableXOrderBookReader::new(&contract, options.auction_data_page_size);
-    info!("Orderbook filter: {:?}", options.orderbook_filter);
+    let primary_orderbook =
```
fleupold, line 208: Maybe we should have a shadow flag that we can pass in to the run command so we only run the shadow orderbook on one of our machines.

e00E (approved):

driver/src/main.rs
nlordell: I think we would have to use `Arc`, but that would work.

driver/src/orderbook/shadow_orderbook.rs
nlordell: I think the reference is fine. The only bit that was a bit of a problem is the `shadow` instance since it is run on a separate thread and so it either needs to be `'static` which is what we use here (since the type is owned) or use a scoped thread, then the shadow needs to live as long as the scope. So that means that scope needs to live in `main.rs` and leaks a bit of the implementation details.

driver/src/main.rs
nlordell: Actually, this is something that `graphprotocol` did a lot, they used `Arc` quite extensively and `Mutex`es for interior mutability. I think it definitely makes some code easier to write (no need to reason about lifetimes for example) and has a fairly low runtime cost.

Maybe we can use to use `Arc` in a some places where it makes sense and remove some of the `'a` references we carry around.

bh2smith: I'm a major fan of these types of simplifications. If only, I knew why/when clone vs RC vs Arc and why they make it easier.

driver/src/main.rs
```diff
@@ -192,8 +201,9 @@ fn main() {
     .unwrap();
 
     // Set up web3 and contract connection.
-    let contract =
-        StableXContractImpl::new(&web3, options.private_key.clone(), options.network_id).unwrap();
+    let contract = Arc::new(
+        StableXContractImpl::new(&web3, options.private_key.clone(), options.network_id).unwrap(),
+    );
```
bh2smith, line 206: I see that `Arc` is used as mentioned below in the comments about not cloning the contract, but the clone still occurs below in the same spot. Was this an oversight?

driver/src/main.rs
```diff
@@ -202,13 +212,28 @@ fn main() {
...
+    let primary_orderbook =
+        PaginatedStableXOrderBookReader::new(contract.clone(), options.auction_data_page_size);
+
+    // NOTE: Keep the shadowed orderbook around so it doesn't get dropped and we
+    //   can pass a reference to the filtered orderbook reader.
+    let unfiltered_orderbook: Box<dyn StableXOrderBookReading + Sync> = if options
+        .use_shadowed_orderbook
+    {
+        let shadow_orderbook =
+            PaginatedStableXOrderBookReader::new(contract.clone(), options.auction_data_page_size);
```
bh2smith, line 224: This order book seems identical to the primary order book (since they are both acquired via the same `PaginatedStableXOrderBookReader`). I thought that each order book was meant to be fetch in a different way so that comparison would serve a purpose.

nlordell: Yes, but we only have one for now (another one is coming in a separate PR). This does verify that the orderbook retrieval does not depend on when in the batch it happens.

driver/src/main.rs
nlordell: We clone the `Arc` pointer here instead of the contract itself. `Arc` is basically a shared reference and the `clone` here is saying give "me a reference"!

PR #746, opened 2020-04-21
[event based orderbook] Handle trades in balance

e00E:


driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,17 +2,43 @@ use super::*;
 use anyhow::{anyhow, ensure, Result};
 use serde::{Deserialize, Serialize};
 
+/// Balance change from a potential solution that might still get replaced by a better solution in
+/// the same batch.
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+struct Proceeds {
+    batch_id: BatchId,
+    // Two fields instead of one I256 because although unlikely it could be possible to overflow a
+    // I256 with multiple trades in one batch.
```
nlordell, line 11: Maybe we can consider using a `BigInt` library so we don't have to deal with math overflows in places where they are allowed.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,17 +2,43 @@ use super::*;
...
+    }
+}
+
 /// The balance of a token for a user.
 #[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
 pub struct Balance {
     balance: U256,
     deposit: Flux,
     withdraw: Flux,
+    proceeds: Proceeds,
```
nlordell, line 36: Just curious, what is the reason these can't be applied to the balance immediately? Is it because balances can temporarily drop below 0 during settlement?

nlordell (approved):
Looks very correct. I think we could have gotten away with a more naive implementation. That being said I think the it looks very clean and easy to follow.

**edit**: Looks like keeping track of the pending proceeds is indeed needed! Great stuff!

driver/src/orderbook/streamed/balance.rs
e00E: Yes. I think Felix asked the same question previously and I forgot to move the comment here. You can in the same batch buy a token and sell it for a balance that includes the new buy amount and is larger than the balance without the new buy. That is how I understood the smart contract code.

nlordell: Makes sense.

driver/src/orderbook/streamed/balance.rs
fleupold: I also find a single variable simpler, if possible. This way we wouldn't need the `getField` method and only have to do one addition in `balance_with_deposit_and_proceeds`.

e00E: That is a neat idea. If we used a BigInt everywhere (not just proceeds) for balances we could remove all of the overflow checks. Then we would only need to check whether the value really fits in a U256 when actually creating the account states.

Since we have this code now and with this PR the balance code should be done and its already reviewed we should keep this as is now and I will keep this as a todo for when the event based orderbook has been meged into master.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -2,17 +2,43 @@ use super::*;
...
+    decrease: U256,
+}
+
+enum TradeType {
+    Sell,
+    Buy,
+}
+
+impl Proceeds {
+    fn get_field(&mut self, operation: TradeType) -> &mut U256 {
```
fleupold, line 22: nit: maybe an update proceeds function that takes a closure `Fn(batchId, amount) -> (new_batch_id, new_amount)` would allow a similar level of code-reuse without the need for TradeType and two fields. But this is just my personal taste, no good technical reason.

fleupold (approved):
Looks good to me! Do we already have tests or are we planning to add unit test this logic?

e00E:
> Looks good to me! Do we already have tests or are we planning to add unit test this logic?

There will be tests that test trade events in `State` which tests the balances. I'll add more tests later and try to get 100% coverage then.

driver/src/orderbook/streamed/balance.rs
nlordell: I think the closure would fit nicely here.

e00E: I don't understand this idea. We still need `increase` and `decrease` because of the overflow (unless we change to BigInt which I want to do later). If I get rid of `TradeType` but still have `increase` and `decrease` then how would the update function know which field to pass to the closure.

nlordell: I imagined it like this: `self.sell_buy_internal(amount, batch_id, |proceeds| &mut proceeds.increase)`, but I see that Felix has a alternate signature and maybe something slightly different in mind.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -63,29 +89,91 @@ impl Balance {
...
+
+    pub fn revert_sell(&mut self, amount: u128, batch_id: BatchId) -> Result<()> {
+        self.revert_sell_buy_internal(amount, batch_id, TradeType::Sell)
+    }
+
+    pub fn revert_buy(&mut self, amount: u128, batch_id: BatchId) -> Result<()> {
+        self.revert_sell_buy_internal(amount, batch_id, TradeType::Buy)
+    }
+
+    fn sell_buy_internal(
```
bh2smith, line 116: nit: I would have preferred if this was `buy_sell_internal` since nobody, anywhere, has ever said "sell-buy".

driver/src/orderbook/streamed/balance.rs
```diff
@@ -63,29 +89,91 @@ impl Balance {
...
+                .checked_add(self.proceeds.increase)
+                .ok_or_else(|| anyhow!("math overflow"))?
+                .checked_sub(self.proceeds.decrease)
+                .ok_or_else(|| anyhow!("math underflow"))?;
+        }
         Ok(balance)
     }
 
-    fn apply_existing_deposit(&mut self, current_batch_id: BatchId) -> Result<()> {
+    fn apply_existing_deposit_and_proceeds(&mut self, current_batch_id: BatchId) -> Result<()> {
```
bh2smith, line 168: Is there no analogue method which "applied existing withdraw"? Or should withdraws be included here?

e00E: Withdraw Requests only get applied in the `withdraw` function. That function already existed so this diff doesn't show it.

bh2smith: In that case, I am unclear on why deposits are handled together with proceeds, while withdraws are handled on their own.

e00E: Deposits and proceeds can be applied whenever the batch id has increased because then they are final.
Withdraw requests on the other hand start being valid at some point at which we need to deduct them from the available balance (`get_balance`) but they only disappear when we see a Withdraw event.

bh2smith: Right, yes! Thanks very much for clearing that up for me. Do you think a description like this might be useful to have somewhere within the code comments?

e00E: I feel like the code is self describing in that way because I would just repeat what is already happening. But maybe I have worked with that code for so long that I don't  notice non obvious cases.

PR #747, opened 2020-04-21
[event based orderbook] Add helper methods to order

e00E:


driver/src/orderbook/streamed/order.rs
```diff
@@ -1,19 +1,106 @@
...
     pub valid_until: BatchId,
-    pub price_numerator: u128,
-    pub price_denominator: u128,
-    pub used_amount: u128,
+    pub valid_from: BatchId,
+    price_numerator: u128,
+    price_denominator: u128,
+    // Invariant: used_amount + pending_used_amount.amount <= price_denominator
+    used_amount: u128,
+    pending_used_amount: PendingUsedAmount,
```
nlordell, line 25: I don't understand why this is necessary, doesn't `TradeReversion` contain this data? I also thought that events were going to get applied in a way that only the relevant events (that is up until the batch that we are requesting the orderbook for) so this shouldn't be necessary for ensuring that the trade is not included in the orderbook right?

driver/src/orderbook/streamed/order.rs
```diff
@@ -1,19 +1,106 @@
...
+            // Cannot fail because of the invariant.
+            self.used_amount
+                .checked_add(self.pending_used_amount.amount)
+                .unwrap()
+        } else {
+            self.used_amount
+        }
+    }
+
+    pub fn get_remaining_amount(&self, batch_id: BatchId) -> u128 {
```
nlordell, line 71: I think if the account state is only ever constructed with events that are to be considered for a batch, then passing the batch ID here shouldn't be necessary.

nlordell:
I think the `PendingUsedAmount` adds some unneeded complexity. Maybe I missed why its needed though.

driver/src/orderbook/streamed/order.rs
```diff
@@ -1,19 +1,106 @@
...
     pub sell_token: TokenId,
-    pub valid_from: BatchId,
     pub valid_until: BatchId,
-    pub price_numerator: u128,
-    pub price_denominator: u128,
-    pub used_amount: u128,
+    pub valid_from: BatchId,
+    price_numerator: u128,
+    price_denominator: u128,
+    // Invariant: used_amount + pending_used_amount.amount <= price_denominator
```
fleupold, line 23: I assume we can enforce this invariant as soon as we start writing the pending & used_amount.

fleupold (approved):
Looks good to me!

driver/src/orderbook/streamed/order.rs
fleupold: I believe the reason is that for some orderbooks you don't want to take the pending proceeds into account (namely the one for which the proceed is pending in case we want to resolve it after a solution has already been submitted), but for others you do (e.g. the next batch for which we optimistically assume the current solution will stay the best one).

Thus we need to have a way even without a new event to conditionally deduct or incorporate pending proceeds.

driver/src/orderbook/streamed/order.rs
fleupold: If I remember correctly we dismissed the idea of instantiating the orderbook reader with a target batch and only processing the events that matter for that batch.
Otherwise the same would apply for balances as you would know upfront if a deposit will still be pending or already matured.

driver/src/orderbook/streamed/order.rs
nlordell: > in case we want to resolve the batch after another solver has already submitted a solution [...] but for other orderbooks you want to include them

I was under the impression that we were going to keep all the events in memory and then apply only the needed events to build the orderbook. In that case aren't the events for previous solution just not going to be applied? If we want to query the open orderbook I would imagine that just more events would be applied (like the current winning solution).

e00E: The way I understood what we decided and implemented it is that the "event manager" is completely agnostic about how events work. It applies all of them because it doesn't know which events are relevant. This is why this code needs to handle this as Felix said.

driver/src/orderbook/streamed/order.rs
e00E: In this version the used amount can't change so the invariant is enforced :) . It is going to change and be enforced in the `trade` and `trade_reversion` functions that I left unimplemented in this PR.

driver/src/orderbook/streamed/order.rs
nlordell: I see the comment below, I guess I slightly misunderstood what we ended up deciding. This is indeed needed then!.

driver/src/orderbook/streamed/order.rs
nlordell: Ahh! I guess I misunderstood a bit what we had decided. I had understood that we didn't want specialized logic for filtering events based on their data, but that we would still construct the orderbook only based on the events up to a certain block number (determined based on the batch ID that we were fetching for).

The only tiny nit I have with the API then is that calling `order.get_remaining_amount(some_batch_in_the_past)` doesn't actually return the amount at the batch that was specified but the current batch.

e00E: It returns the current remaining amount but to know whether pending is part of current it needs to know the batch id. Looking at this function in isolation I can see that it might be confusing but it's done like this throughout `State`.

nlordell: >  Looking at this function in isolation I can see that it might be confusing

I don't see a way around it though. I 100% OK with how it is right now.

driver/src/orderbook/streamed/state.rs
```diff
@@ -159,9 +143,16 @@ impl State {
...
-    fn order_deletion(&mut self, event: &OrderDeletion) -> Result<()> {
-        // Orders are allowed to be deleted multiple times.
-        self.orders.remove(&(event.owner, event.id));
+    fn order_deletion(&mut self, event: &OrderDeletion, block_batch_id: BatchId) -> Result<()> {
+        if let Some(order) = self.orders.get(&(event.owner, event.id)) {
+            ensure!(
+                !order.is_valid_in_batch(block_batch_id - 1),
+                "deleting valid order"
+            );
+            self.orders.remove(&(event.owner, event.id));
```
nlordell, line 152: I think order deletion also needs a pending state. This is because an order can be valid for batch X and deleted in batch X+1, but then we will just not have the order if we try to query the orderbook at batch X.

driver/src/orderbook/streamed/order.rs
nlordell: > what we decided is that the "event manager" is completely agnostic about how events work

Yes, I had understood that it just meant that it would not inspect data to decide if an event should be included or not.

driver/src/orderbook/streamed/state.rs
e00E: If we are solving batch X, then X+1 is the active batch. If the order was valid in X then it cannot get deleted in X+1 as per the `ensure` above.

nlordell: Oh yeah! It can only be deleted if its not valid in the solving batch!

nlordell (approved):
:+1: Sorry for the confusion! I had slightly misunderstood what we had decided on.

driver/src/orderbook/streamed/order.rs
```diff
@@ -1,19 +1,106 @@
...
+            .unwrap()
+    }
+
+    pub fn trade(&mut self, _used_amount: u128, _batch_id: BatchId) -> Result<()> {
+        unimplemented!();
+    }
+
+    pub fn revert_trade(&mut self, _used_amount: u128, _batch_id: BatchId) -> Result<()> {
+        unimplemented!();
+    }
```
bh2smith, line 84: I might suggest naming the parameters `sold_amount` (since there are two different types of "amounts") involved in an order. I don't believe it will lead to any confusion later when implementing things like `self.used_amount += sold_amount` etc...

PR #748, opened 2020-04-21
[getOpenOrderbook 2] Use onchain filter to fetch orders

fleupold:
Draft to get some early feedback. Still missing unit tests and proper wiring (not sure if I should make it configurable of if we want to get rid of the existing Orderbook reader, which is probably safer).

This PR introduces a new Reader implementation that uses the BatchExchangeViewer and in particular its methods that filter order on0chain (instead of on the client). The main benefit is that this methods allows to query balance for a future batch (e.g. the one that starts in the next block) and will thus allow us to fetch the orderbook from the last block of the previous batch rather than on the first block of the batch we are actually trying to solve

### Test Plan

Will add a unit test to the AuctionDataReader. Also running this locally and making sure that at least number of orders/users/tokens are correct. Would probably be cool to already use this in combination with the Diff tool and see that it runs smooth locally.

### DIFF
--- a/driver/src/contracts/stablex_auction_element.rs
+++ b/driver/src/contracts/stablex_auction_element.rs
@@ -27,7 +27,9 @@ impl StableXAuctionElement {
     /// serialized information.
     pub fn from_bytes(bytes: &[u8; AUCTION_ELEMENT_WIDTH]) -> Self {
         let mut indexed_bytes = [0u8; INDEXED_AUCTION_ELEMENT_WIDTH];
-        indexed_bytes.copy_from_slice(bytes);
+        for (index, bit) in bytes.iter().enumerate() {
+            indexed_bytes[index] = *bit;
+        }
         Self::from_indexed_bytes(&indexed_bytes)
     }
 
--- a/driver/src/contracts/stablex_auction_element.rs
+++ b/driver/src/contracts/stablex_auction_element.rs
@@ -159,6 +159,38 @@ pub mod tests {
     }
 
     #[test]
+    fn test_index_auction_element() {
+        let bytes: [u8; 114] = [
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, // user: 20 elements
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 3, // sellTokenBalance: 3, 32 elements
+            1, 2, // buyToken: 256+2,
+            1, 1, // sellToken: 256+1,
+            0, 0, 0, 2, // validFrom: 2
+            0, 0, 1, 5, // validUntil: 256+5
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, // priceNumerator: 258
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 3, // priceDenominator: 259
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, // remainingAmount: 2**8 + 1 = 257
+            0, 1, // order index
+        ];
+        let res = StableXAuctionElement::from_indexed_bytes(&bytes);
+        let auction_element = StableXAuctionElement {
+            valid_from: U256::from(2),
+            valid_until: U256::from(261),
+            sell_token_balance: 3,
+            order: Order {
+                id: 1,
+                account_id: Address::from_low_u64_be(1),
+                buy_token: 258,
+                sell_token: 257,
+                buy_amount: (258 * 257 + 258) / 259,
+                sell_amount: 257,
+            },
+        };
+        assert_eq!(res, auction_element);
+    }
+
+    #[test]
     #[should_panic]
     fn test_from_bytes_fails_on_hopefully_null() {
         StableXAuctionElement::from_bytes(&[1u8; 112]);
--- a/driver/src/contracts/stablex_auction_element.rs
+++ b/driver/src/contracts/stablex_auction_element.rs
@@ -55,7 +55,7 @@ impl StableXAuctionElement {
         let numerator = BigEndian::read_u128(&bytes[64..80]);
         let denominator = BigEndian::read_u128(&bytes[80..96]);
         let remaining = BigEndian::read_u128(&bytes[96..112]);
-        let id = u16::from_le_bytes([bytes[112], bytes[113]]);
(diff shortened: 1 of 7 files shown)


driver/src/contracts/stablex_auction_element.rs
```diff
@@ -5,6 +5,8 @@ use ethcontract::{Address, U256};
 use crate::util::CeiledDiv;
 
 pub const AUCTION_ELEMENT_WIDTH: usize = 112;
+// Indexed auction elements have the orderId appened at the end
```
nlordell, line 8: nit: I think this should just be a doc comment
```suggestion
/// Indexed auction elements have the orderId appened at the end
```

driver/src/contracts/stablex_contract.rs
```diff
@@ -56,6 +56,13 @@ impl StableXContractImpl {
     }
 }
 
+pub struct FilteredAuctionData {
```
nlordell, line 59: naming nit: Maybe include in the name somehow that this is a single page of auction data? I'm fine with how it is though.

driver/src/contracts/stablex_contract.rs
```diff
@@ -108,6 +125,38 @@ impl StableXContract for StableXContractImpl {
...
+        previous_page_user_offset: u16,
+    ) -> Result<FilteredAuctionData> {
+        let target_batch = batch_index.low_u32();
+        let mut builder = self.viewer.get_filtered_orders_paginated(
+            [target_batch, target_batch, target_batch],
+            vec![],
+            previous_page_user,
+            previous_page_user_offset,
+            page_size,
+        );
```
nlordell, line 142: All that sexy type-safety...

driver/src/contracts/stablex_contract.rs
```diff
@@ -108,6 +125,38 @@ impl StableXContract for StableXContractImpl {
...
+        builder.m.tx.gas = None;
+        builder
+            .call()
+            .wait()
+            .map(
```
nlordell, line 147: I'm curious why you prefer `map` and `map_err` to `?` and then using the return value in the function?

I tend to prefer `?` as it creates less "rightwards drift".

driver/src/contracts/stablex_contract.rs
```diff
@@ -65,6 +72,16 @@ pub trait StableXContract {
     /// Retrieve the time remaining in the batch.
     fn get_current_auction_remaining_time(&self) -> Result<Duration>;
 
+    /// Retrieve one page of indexed auction data that is filtered on chain
+    /// to only include orders valid at the given batchId.
+    fn get_filtered_auction_data_paginated(
```
nlordell, line 77: Should this method take a token filter as well? Or a `Option<Vec<TokenId>>` that defaults to the whitelist filter?

driver/src/orderbook/onchain_filtered_orderbook.rs
```diff
@@ -0,0 +1,44 @@
...
+use anyhow::Result;
+use ethcontract::{Address, U256};
+
+pub struct OnchainFilteredOrderBookReader<'a> {
+    contract: &'a (dyn StableXContract + Sync),
+    page_size: u16,
+}
+
+impl<'a> OnchainFilteredOrderBookReader<'a> {
+    pub fn new(contract: &'a (dyn StableXContract + Sync), page_size: u16) -> Self {
```
nlordell, line 16: Should this reader also use the `OrderbookFilter` type that gets passed into the `FilteredOrderbookReader` and then if the token filter is `TokenFilter::Whitelist` also pass it to the contract?

nlordell (approved):
Short and sweet. ~~Maybe add a unit test to the `OnchainFilteredOrderBookReader` implementation using a mocked contract instance just as a sanity check?~~

**edit**: I see in the PR description that you will add a test.

driver/src/contracts/stablex_contract.rs
fleupold: No reason, just the way it was done in this class before. Will change!

e00E (approved):

driver/src/contracts/stablex_auction_element.rs
```diff
@@ -24,6 +26,16 @@ impl StableXAuctionElement {
     /// Sets `id` to `0` because this information is not contained in the
     /// serialized information.
     pub fn from_bytes(bytes: &[u8; AUCTION_ELEMENT_WIDTH]) -> Self {
+        let mut indexed_bytes = [0u8; INDEXED_AUCTION_ELEMENT_WIDTH];
+        for (index, bit) in bytes.iter().enumerate() {
```
nlordell, line 30: nit: Can you use copy_from_slice here?
```suggestion
        indexed_bytes[..AUCTION_ELEMENT_WIDTH].copy_from_slice(&bytes)
```

driver/src/orderbook/auction_data_reader.rs
```diff
@@ -183,12 +185,58 @@ impl PaginatedAuctionDataReader {
     }
 }
 
+pub struct IndexedAuctionDataReader(AuctionDataReader);
```
nlordell, line 188: I think this abstraction is baller! It all fit super well together in the end :rocket:

nlordell (approved):
Awesome!

nlordell:
CI is failing on a clippy lint:
```
error: unneeded `return` statement
  --> driver/src/orderbook/onchain_filtered_orderbook.rs:54:9
   |
54 |         return Ok(reader.get_auction_data());
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove `return`: `Ok(reader.get_auction_data())`
   |
   = note: `-D clippy::needless-return` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
```

driver/src/contracts/stablex_auction_element.rs
fleupold: I wanted to but can't because of https://doc.rust-lang.org/std/primitive.slice.html#panics-17

> This function will panic if the two slices have different lengths.

nlordell: Yes, but that is why you first slice `indexed_bytes` to `AUCTION_ELEMENT_WIDTH` length, and then use `copy_from_slice`.

fleupold: I'm copying bytes into indexed_bytes, so I'm looking for a way to extend bytes with 2 zeros (actually making it larger not small). I agree that the current version doesn't feel syntactically as sweetest as it can be, but I don't think I can slice an array up rather than down can I?

nlordell: Sorry, I didn't explain it well :disappointed:. This is what I meant: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b0081941a120670bf8a018cc95087efd

fleupold: Ah cool, that's what I was missing. Let me adjust this in a separate PR.

PR #751, opened 2020-04-21
[event based orderbook] Handle trades in order

e00E:


driver/src/orderbook/streamed/order.rs
```diff
@@ -103,4 +128,24 @@ impl Order {
...
+
+    fn apply_pending_used_amount(&mut self, batch_id: BatchId) {
+        self.used_amount = self.get_used_amount(batch_id);
+        if self.pending_used_amount.batch_id < batch_id {
+            self.pending_used_amount.amount = 0;
+        }
+        // The invariant hasn't been changed because the total used amount stayed the same.
+    }
+
+    fn ensure_invariant(&self) -> Result<()> {
```
nlordell, line 140: naming nit: maybe state the invariant in the method name.

nlordell (approved):
Easy! All falling into place now

fleupold (approved):

bh2smith:
It appears I was too late to this big party and all the other PRs have been layered onto this. Considering there is no description to this PR I will have to guess what it was about and would like to raise one concern!

Since the title of this PR says handle trades "in order", I was wondering if we considered the fact that a single user can be involved in more than a single trade within one batch. More importantly a single user can acquire a token they did not previously have and sell that very same allotment in the same batch. Do we have this scenario covered with our solution here? In the smart contract we do all the additions followed by all the subtractions, which actually requires passing over all the trades twice.

e00E:
> It appears I was too late to this big party and all the other PRs have been layered onto this. Considering there is no description to this PR I will have to guess what it was about and would like to raise one concern!

Originally there was only one commit whose impact was very obvious so I didn't write a description. You can still see just that commit by clicking on `commits` and then on the first one.

> Since the title of this PR says "handle trades in order", I was wondering if we considered the fact that a single user can be involved in more than a single trade within one batch. More importantly a single user can acquire a token they did not previously have and sell that very same allotment in the same batch. Do we have this scenario covered with our solution here? In the smart contract we do all the additions followed by all the subtractions, which actually requires passing over all the trades twice.

This is handled in `balance.rs`. We keep track of sells and buys for a user for a token separately. This allows there to temporarily be a larger decrease of the balance than the balance was previously for the reason you mention: there can be another trade that we haven't observed yet that increases the balance

fleupold:
> Originally there was only one commit whose impact was very obvious so I didn't write a description.

I think even for the small PRs it's valuable to write a short description that puts the PR into context. 
Specifically when merging other separately reviewed PRs into this PR later it's nice to point out this plan in the description and maybe even include a link to the diff view that people should review (It's not obvious looking at the PR now).

Generally, I prefer to merge stacked PRs (master --> PR1 --> PR2 ... )  from the bottom up, e.g. merge PR1 into master first, then rebase PR2 onto master and merge, and so on (instead of folding everything in PR1 and merging this in one large commit). 
It requires each PR to be "in itself working" as in at least pass CI, and it emphasises an incremental development flow that focuses on bringing "early" versions into master rather than waiting for the big merge. It also avoids duplicating changesets/discussions over multiple PRs.

PR #752, opened 2020-04-21
[event based orderbook] Handle trades in State

e00E:


driver/src/orderbook/streamed/state.rs
```diff
@@ -155,6 +155,61 @@ impl State {
...
     }
+
+    fn apply_trade(&mut self, event: &Trade, block_batch_id: BatchId) -> Result<()> {
+        self.apply_trade_internal(
+            event.owner,
+            event.order_id,
+            |order| order.trade(event.executed_sell_amount, block_batch_id),
+            |balance| balance.sell(event.executed_sell_amount, block_batch_id),
+            |balance| balance.buy(event.executed_buy_amount, block_batch_id),
+        )
```
fleupold, line 166: nit: I think closures can be a nice way to reuse code. Here, I wonder if it would be more straightforward to have a method that returns the mutual references we need (order, sell, buy balance) and then do the apply/revert logic inline?

fleupold (approved):
Nice, I really like how simple the PRs are to review now. I'd still appreciate a small PR description putting this change into the bigger picture context and saying how this can/should be tested.

nlordell (approved):
Easy! Looks great!

driver/src/orderbook/streamed/state.rs
nlordell: I'm OK either way.

I actually find this to be visually pleasing and easy to read and scan through ("applying a trade for owner and order id where the order applies these trade values and the sell balance applies a sell and buy balance applies a buy").

I also think that returning a tuple with mutable references to the order and balances as you suggest would read in a similar way and would work as well.

My tiny ubernit small suggestion would be to name the parameters to the closures to include which one is the sell token balance and the buy token balance. The reason I suggest this is just so its super clear that these are two distinct token balances and not the same one.

e00E: I just noticed that I did name the parameters for revert but forgot to do it here. I thought of the "return mutable reference" way to do but slightly preferred this. The other might not actually work because we would be borrowing self mutably multiple times, not sure because I didn't try.

driver/src/orderbook/streamed/state.rs
```diff
@@ -155,6 +165,91 @@ impl State {
...
+    }
+
+    fn apply_trade_internal(
+        &mut self,
+        user_id: UserId,
+        order_id: OrderId,
+        order_fn: impl FnOnce(&mut Order) -> Result<()>,
+        sell_balance_fn: impl FnOnce(&mut Balance) -> Result<()>,
+        buy_balance_fn: impl FnOnce(&mut Balance) -> Result<()>,
```
bh2smith, line 199: I am not sure I understand the meaning of these functions as parameters here

driver/src/orderbook/streamed/state.rs
```diff
@@ -155,6 +165,91 @@ impl State {
...
+            .ok_or_else(|| anyhow!("solution without fee token"))?;
+        self.revert_last_solution(fee_token, block_batch_id);
+        self.last_solution.batch_id = block_batch_id;
+        self.last_solution.user_id = event.submitter;
+        self.last_solution.burnt_fees = event.burnt_fees;
+        self.balances
+            .entry((event.submitter, fee_token))
+            .or_default()
+            .solution_submission(event.burnt_fees, block_batch_id)
+    }
```
bh2smith, line 239: I've seen this already in a previously merged PR.

e00E: Not in a merged PR. We cancelled that one because it was too big iirc.

driver/src/orderbook/streamed/state.rs
e00E: We want to extract common code from trade and trade_reversion into this function. This is achieved by moving the action that is performed on the order and balances into closures. In `apply_trade` we call `trade` `sell` `buy` and in `apply_trade_reversion` we call `revert_trade` `revert_sell` `revert_buy`.

PR #753, opened 2020-04-21
[event based orderbook] Handle solution in State

e00E:


driver/src/orderbook/streamed/balance.rs
```diff
@@ -98,24 +98,34 @@ impl Balance {
...
+    }
+
+    pub fn solution_submission(&mut self, fee: U256, batch_id: BatchId) -> Result<()> {
+        // We are reusing the `Proceeds` machinery here because the extra balance from a solution
+        // submission behaves the same way.
+        self.sell_buy_internal(fee, batch_id, TradeType::Buy)
+    }
+
+    pub fn revert_solution_submission(&mut self, fee: U256, batch_id: BatchId) -> Result<()> {
+        self.revert_sell_buy_internal(fee, batch_id, TradeType::Buy)
```
fleupold, line 123: should this be of type sell?

driver/src/orderbook/streamed/balance.rs
```diff
@@ -98,24 +98,34 @@ impl Balance {
...
-        self.revert_sell_buy_internal(amount, batch_id, TradeType::Sell)
+        self.revert_sell_buy_internal(amount.into(), batch_id, TradeType::Sell)
     }
 
     pub fn revert_buy(&mut self, amount: u128, batch_id: BatchId) -> Result<()> {
-        self.revert_sell_buy_internal(amount, batch_id, TradeType::Buy)
+        self.revert_sell_buy_internal(amount.into(), batch_id, TradeType::Buy)
+    }
+
+    pub fn solution_submission(&mut self, fee: U256, batch_id: BatchId) -> Result<()> {
```
fleupold, line 116: Is there a benefit of having different methods for this if its logical equivalent to a buy/sell? Maybe we can find a name like credit/debit that capture the act of increasing/decreasing balance better.

driver/src/orderbook/streamed/state.rs
```diff
@@ -206,9 +209,37 @@ impl State {
...
+    fn revert_last_solution(&mut self, fee_token: TokenAddress, block_batch_id: BatchId) {
+        if self.last_solution.0 == block_batch_id {
+            // Neither unwrap can fail because we must have previously added the fee to the
+            // submitter in which case the balance must exist and be reversible.
+            self.balances
+                .get_mut(&(self.last_solution.1, fee_token))
+                .unwrap()
+                .revert_solution_submission(self.last_solution.2, block_batch_id)
+                .unwrap();
+            self.last_solution.2 = U256::zero();
```
fleupold, line 241: nit: I'd find a struct with named fields for last solution easier to read rather than having to look up .0,.1,.2.

fleupold (approved):
It's all coming together 👏

driver/src/orderbook/streamed/balance.rs
nlordell: I imagine its a buy as you are reverting receiving some amount. That's how I understood it at least.

driver/src/orderbook/streamed/state.rs
nlordell: Can you just reuse the `SolutionSubmission` event struct?

driver/src/orderbook/streamed/balance.rs
nlordell: As I see it doing it this way makes `State` a little simpler, where it becomes responsible for finding the `Balance`s and `Order`s that need to be modified and then forwarding the needed event data to those components. I think its nice that each event that can change a `Balance` has their own method which completely encapsulates all the required logic for that event.

It also allows you to look at `Balance` in terms of a stream of events. For example you can, independent of the logic in `State` create a unit test that applies these events in a sequence.

Also, from a quick glance at the code, `Trade`s and `TradeReversion`s have slightly different semantics with respect to applying the pending fluxes and proceeds so that information would be in `State` instead of `Balance`s.

nlordell (approved):
Wow! Home stretch!

nlordell:
On a sort of separate note, unrelated to this PR, I had a nice thought for modelling the getter methods in terms of batch IDs with respect to the `balances.get_balance(0)` not actually returning the balance at batch ID 0. The parameter their can be:
```
Batch {
    Solving,
    Future(std::num::NonZeroU32),
}
impl Batch {
    fn from_id(current_batch_id, batch_id) {
        ensure!(current_batch_id <= batch_id, "batch in the past");
        std::num::NonZeroU32::new(batch_id - current_batch_id)
            .map(Batch::Future)
            .unwrap_or(Batch::Solving)
    }
}
```

I'm still not 100% satisfied with this, but then it would make it impossible to try and get the balance in a past batch and misuse the API.

driver/src/orderbook/streamed/balance.rs
e00E: It should be a buy. The same way when reverting a buy we also use TradeType::Buy because it is going to subtract from the increase field.

driver/src/orderbook/streamed/balance.rs
e00E: I did it for the reasons Nick outlined. From state's perspective I find it a bit easier to have this method and it shows that you could change the implementation of it in balance without breaking anything.

driver/src/orderbook/streamed/state.rs
e00E: Agree it should be a struct. I made one instead of reusing so that we only store exactly what is needed.

driver/src/orderbook/streamed/state.rs
```diff
@@ -206,9 +216,39 @@ impl State {
...
+        let fee_token = self
+            .tokens
+            .get_address_by_id(0)
+            .ok_or_else(|| anyhow!("solution without fee token"))?;
+        self.revert_last_solution(fee_token, block_batch_id);
+        self.last_solution.batch_id = block_batch_id;
+        self.last_solution.user_id = event.submitter;
+        self.last_solution.burnt_fees = event.burnt_fees;
+        self.balances
+            .entry((event.submitter, fee_token))
```
bh2smith, line 236: Does this add a new entry to the balance "dict" if it doesn't already exist and do nothing otherwise?

e00E: With the line below `or_default` yes. It is like "get the element or create a default if it doesn't exist and then get that".

PR #755, opened 2020-04-22
[event based orderbook] Handle orderbook creation from State with regards to batch id better

### LINKED ISSUE #754: [event based orderbook] Nick's idea for better modelling getter methods in terms of batch id
On a sort of separate note, unrelated to this PR, I had a nice thought for modelling the getter methods in terms of batch IDs with respect to the `balances.get_balance(0)` not actually returning the balance at batch ID 0. The parameter there can be:
```
Batch {
    Solving,
    Future(std::num::NonZeroU32),
}
impl Batch {
    fn from_id(current_batch_id, batch_id) {
        ensure!(current_batch_id <= batch_id, "batch in the past");
        std::num::NonZeroU32::new(batch_id - current_batch_id)
            .map(Batch::Future)
            .unwrap_or(Batch::Solving)
    }
}
```

I'm still not 100% satisfied with this, but then it would make it impossible to try and get the balance in a past batch and misuse the API.

_Originally posted by @nlordell in https://github.com/gnosis/dex-services/pull/753#issuecomment-617545797_

e00E:
fixes #754

This prevents attempting to create the orderbook with a past batch with
is nonsensical as State does not actually support.
This prevents reading the orderbook for a future batch if that
orderbook would be inconsistent due to a partially applied solution.

I notice that the tests feels needlessly long. In the future I want to refactor the tests into more helper functions and macros to make them smaller and easier to read.

### DIFF
--- a/driver/src/orderbook/streamed/state.rs
+++ b/driver/src/orderbook/streamed/state.rs
@@ -59,7 +59,7 @@ impl State {
         let batch_id = match batch {
             Batch::Current => self.last_batch_id,
             Batch::Future(batch_id) => {
-                // We allow the bach ids being equal to prevent race conditions where the State gets
+                // We allow the batch ids being equal to prevent race conditions where the State gets
                 // a new event right before we want to get the orderbook.
                 ensure!(self.last_batch_id <= batch_id, "batch is in the past");
                 // TODO: in the future we might want to handle the case where
@@ -712,7 +712,6 @@ mod tests {
             .unwrap()
             .1;
         assert_eq!(balance, U256::zero());
-        assert!(state.orderbook_for_batch(Batch::Future(0)).is_err());
         let balance = state
             .orderbook_for_batch(Batch::Future(1))
             .unwrap()
@@ -767,9 +766,6 @@ mod tests {
             .collect::<HashMap<_, _>>();
         assert_eq!(balance.get(&(address(2), 0)), Some(&10.into()));
         assert_eq!(balance.get(&(address(2), 1)), Some(&10.into()));
-
-        // Error because the solution is not complete.
-        assert!(state.orderbook_for_batch(Batch::Future(2)).is_err());
 
         let event = SolutionSubmission {
             submitter: address(4),
--- a/driver/src/orderbook/streamed/state.rs
+++ b/driver/src/orderbook/streamed/state.rs
@@ -23,6 +23,18 @@ pub struct State {
     balances: HashMap<(UserId, TokenAddress), Balance>,
     tokens: Tokens,
     last_solution: LastSolution,
+    /// True when we have received some trades or trade reverts but not yet the final solution
+    /// submission that indiciates all trades have been received.
+    solution_partially_received: bool,
+    last_batch_id: BatchId,
+}
+
+#[derive(Debug)]
+pub enum Batch {
+    /// The current completed batch that can no longer change
+    Current,
+    /// A future potentially still changing batch if a new solution comes in
+    Future(BatchId),
 }
 
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
@@ -33,8 +45,31 @@ struct LastSolution {
 }
 
 impl State {
-    /// Can be used to create a `crate::models::AccountState`.
-    pub fn account_state(
+    /// Errors if State has only received a partial solution which would make the result
+    /// inconsistent.
+    /// Errors if the batch is `Future` but is not actually in the future.
+    /// Account balances that overflow a U256 are skipped.
+    pub fn orderbook(
+        &self,
+        batch: Batch,
+    ) -> Result<(
+        impl Iterator<Item = ((UserId, TokenId), U256)> + '_,
+        impl Iterator<Item = ModelOrder> + '_,
+    )> {
+        let batch_id = match batch {
+            Batch::Current => self.last_batch_id,
+            Batch::Future(batch_id) => {
+                ensure!(self.last_batch_id < batch_id, "batch is not in the future");
+                // This only needs to be ensured for future batches because the current batch cannot
+                // be inconsistent as it does not incorporate pending trades.
+                ensure!(!self.solution_partially_received, "partial solution");
+                batch_id
+            }
+        };
+        Ok((self.account_state(batch_id), self.orders(batch_id)))
+    }
+
+    fn account_state(
         &self,
         batch_id: BatchId,
     ) -> impl Iterator<Item = ((UserId, TokenId), U256)> + '_ {
@@ -46,17 +81,16 @@ impl State {
                 let token_id = self.tokens.get_id_by_address(*token_address)?;
                 Some((
                     (*user_id, token_id),
-                    // TODO: Remove unwrap.
                     // Can fail if user's balance exceeds U256::max.
                     // Can fail while not all trades of a solution have been received and batch_id
                     // is the next batch_id so that we assume that the current solution won't be be
                     // reverted.
-                    balance.get_balance(batch_id).unwrap(),
+                    balance.get_balance(batch_id).ok()?,
                 ))
             })
     }
 
-    pub fn orders(&self, batch_id: BatchId) -> impl Iterator<Item = ModelOrder> + '_ {
+    fn orders(&self, batch_id: BatchId) -> impl Iterator<Item = ModelOrder> + '_ {
         self.orders
             .iter()
             .filter(move |(_, order)| order.is_valid_in_batch(batch_id))
@@ -79,6 +113,8 @@ impl State {
     /// `block_batch_id` is the current batch based on the timestamp of the block that contains the
     ///  event.
     pub fn apply_event(mut self, event: &Event, block_batch_id: BatchId) -> Result<Self> {
+        ensure!(self.last_batch_id <= block_batch_id, "event in the past");
+        self.last_batch_id = block_batch_id;
         match event {
             Event::Deposit(event) => self.deposit(event, block_batch_id)?,
             Event::WithdrawRequest(event) => self.withdraw_request(event, block_batch_id)?,
@@ -198,6 +234,8 @@ impl State {
         sell_balance_fn: impl FnOnce(&mut Balance) -> Result<()>,
         buy_balance_fn: impl FnOnce(&mut Balance) -> Result<()>,
     ) -> Result<()> {
+        self.solution_partially_received = true;
+
         let order = self
             .orders
             .get_mut(&(user_id, order_id))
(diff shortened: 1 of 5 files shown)


driver/src/orderbook/streamed/state.rs
```diff
@@ -23,6 +23,18 @@ pub struct State {
     balances: HashMap<(UserId, TokenAddress), Balance>,
     tokens: Tokens,
     last_solution: LastSolution,
+    /// True when we have received some trades or trade reverts but not yet the final solution
+    /// submission that indiciates all trades have been received.
+    solution_partially_received: bool,
```
fleupold, line 28: Maybe this fits better directly on the LastSolution struct or could be expressed by making last_solution an option that is None when it's in the process of receiving (although I'm not sure that's very explicit).

driver/src/orderbook/streamed/state.rs
```diff
@@ -33,8 +45,31 @@ struct LastSolution {
...
+        &self,
+        batch: Batch,
+    ) -> Result<(
+        impl Iterator<Item = ((UserId, TokenId), U256)> + '_,
+        impl Iterator<Item = ModelOrder> + '_,
+    )> {
+        let batch_id = match batch {
+            Batch::Current => self.last_batch_id,
+            Batch::Future(batch_id) => {
+                ensure!(self.last_batch_id < batch_id, "batch is not in the future");
```
fleupold, line 62: with the strict inequality I'm a little worried about race conditions here, if we thought something was in a future, but happens to be `Current` now, I think that should be still fine.

fleupold (approved):
Nice 🍨

driver/src/orderbook/streamed/state.rs
```diff
@@ -33,8 +45,31 @@ struct LastSolution {
 }
 
 impl State {
-    /// Can be used to create a `crate::models::AccountState`.
-    pub fn account_state(
+    /// Errors if State has only received a partial solution which would make the result
+    /// inconsistent.
+    /// Errors if the batch is `Future` but is not actually in the future.
+    /// Account balances that overflow a U256 are skipped.
+    pub fn orderbook(
```
fleupold, line 52: uber naming nit: I think in the follow up PR it reads slightly nicer if this was called `get_orderbook` or even `get_orderbook_for_batch`.

driver/src/orderbook/streamed/state.rs
nlordell: This concerns me as it means there are times where the orderbook is unqueryable for future batches.

nlordell (approved):

driver/src/orderbook/streamed/state.rs
nlordell: Maybe the correct approach is to not include proceeds if a solution is pratially applied instead of erroring?

driver/src/orderbook/streamed/state.rs
nlordell: I like `orderbook_for_batch`. I'm fine with leaving out the `get_` is its pretty common in Rust for getters to simply be called what is being retrieved (such as `str::chars` instead of `str::get_chars`).

driver/src/orderbook/streamed/state.rs
e00E: I see your point. Since that is getting less straightforward and we are not actually going to solve optimistically yet I think we should restrict this to `Batch::Current` and revisit this later.

driver/src/orderbook/streamed/state.rs
```diff
@@ -33,8 +45,31 @@ struct LastSolution {
...
+        &self,
+        batch: Batch,
+    ) -> Result<(
+        impl Iterator<Item = ((UserId, TokenId), U256)> + '_,
+        impl Iterator<Item = ModelOrder> + '_,
```
bh2smith, line 57: Is this a tuple of results or either or?

driver/src/orderbook/streamed/state.rs
```diff
@@ -79,6 +113,8 @@ impl State {
     /// `block_batch_id` is the current batch based on the timestamp of the block that contains the
     ///  event.
     pub fn apply_event(mut self, event: &Event, block_batch_id: BatchId) -> Result<Self> {
+        ensure!(self.last_batch_id <= block_batch_id, "event in the past");
```
bh2smith, line 118: I see a lot of `ensure` happening... is this the same as assert? Are these directly relevant to this PR?

driver/src/orderbook/streamed/state.rs
```diff
@@ -198,6 +234,8 @@ impl State {
         sell_balance_fn: impl FnOnce(&mut Balance) -> Result<()>,
         buy_balance_fn: impl FnOnce(&mut Balance) -> Result<()>,
     ) -> Result<()> {
+        self.solution_partially_received = true;
```
bh2smith, line 239: What is the meaning and purpose of a partially received solution? I see it is set true here and then false again later. Is this flag only accounting for the time between here and a few lines down the road? If this is the case, does this flag act as a sort-of lock for some sort of threaded environment?

driver/src/orderbook/streamed/state.rs
e00E: Hmm but we do need the batch id in case the it has increased but we just didn't get an event for the next batch yet. Then I would still say we don't fix the optimistic solving case yet because it is more complicated but keep the code as is until we need optimistic solving or the event based orderbook has been merged and we have more time.

driver/src/orderbook/streamed/state.rs
```diff
@@ -652,4 +691,93 @@ mod tests {
             2
         );
     }
+
+    #[test]
+    fn orderbook_batch_id() {
+        let mut state = state_with_fee();
```
bh2smith, line 699: I was not aware of `state_with_fee` is this introduced here? I thought that constructors like this always took the form `State::new_with_fee`

driver/src/orderbook/streamed/state.rs
```diff
@@ -652,4 +691,93 @@ mod tests {
...
+        let event = Deposit {
+            user: address(3),
+            token: address(0),
+            amount: 1.into(),
+            batch_id: 0,
+        };
+        state = state.apply_event(&Event::Deposit(event), 0).unwrap();
+        let balance = state.orderbook(Batch::Current).unwrap().0.next().unwrap().1;
+        assert_eq!(balance, U256::zero());
+        assert!(state.orderbook(Batch::Future(0)).is_err());
```
bh2smith: Is this an error because there is no way 0 is in the future?

driver/src/orderbook/streamed/state.rs
```diff
@@ -652,4 +691,93 @@ mod tests {
...
+        let balance = state
+            .orderbook(Batch::Current)
+            .unwrap()
+            .0
+            .collect::<HashMap<_, _>>();
+        assert_eq!(balance.get(&(address(2), 0)), Some(&10.into()));
+        assert_eq!(balance.get(&(address(2), 1)), Some(&10.into()));
+
+        // Error because the solution is not complete.
+        assert!(state.orderbook(Batch::Future(2)).is_err());
```
bh2smith: Why is it not complete?

driver/src/orderbook/streamed/state.rs
```diff
@@ -652,4 +691,93 @@ mod tests {
...
+            .apply_event(&Event::SolutionSubmission(event), 1)
+            .unwrap();
+
+        let balance = state
+            .orderbook(Batch::Future(2))
+            .unwrap()
+            .0
+            .collect::<HashMap<_, _>>();
+        assert_eq!(balance.get(&(address(2), 0)), Some(&12.into()));
+        assert_eq!(balance.get(&(address(2), 1)), Some(&9.into()));
```
bh2smith, line 789: Nice!

bh2smith (approved):
Looks pretty great. Have a few inline questions and comments.

driver/src/orderbook/streamed/state.rs
e00E: Because the Result is on the outside it is a Result of a tuple. So either you have both elements of the tuple or an error.

driver/src/orderbook/streamed/state.rs
e00E: It is like an assert but intead of panicking it returns an error.

You can check where the ensures come from. It is at the top `use anyhow::{anyhow, bail, ensure, Result};`. The documentation for that is here https://docs.rs/anyhow/1.0.28/anyhow/macro.ensure.html . Or you can try `cargo doc --open` which generates the documentation locally and then searching for `ensure`.

driver/src/orderbook/streamed/state.rs
e00E: The intention is to notice if someone tries to get the orderbook from us while we have received only some trades of a solution but not all of them. In that case the orderbook wouldn't be correct.

driver/src/orderbook/streamed/state.rs
e00E: It is just a function that was already there from a previous test. If you open the whole file you can see it.

driver/src/orderbook/streamed/state.rs
e00E: It was an error (just changed it) because batch id 0 is not actually in the future and the code creating the orderbook checked that.

driver/src/orderbook/streamed/state.rs
e00E: Because we got Trade events but no SolutionSubmission event which means there could be more trades.

driver/src/orderbook/streamed/state.rs
nlordell: Do you mind opening an issue so we don't forget?

driver/src/orderbook/streamed/state.rs
bh2smith: Does the `SolutionSubmission` event contain any relevant information regarding how many trades we expect there to be?

bh2smith:
Looks ready for merger!

driver/src/orderbook/streamed/state.rs
e00E: No

fleupold:
> I notice that the tests feels needlessly long. In the future I want to refactor the tests into more helper functions and macros to make them smaller and easier to read.

Sounds good. Generally, I'd be careful of over-optimizing test code too much (in terms of trying to avoid duplicate code). It can quickly make the test very hard to understand. I generally like having all information that is relevant for a specific test inside the single function rather than having to jump through a bunch of common setup methods to see what the given state/expectations actually is/are. 
That being said, using some macros or the like to make them shorter is probably a cool idea.

PR #756, opened 2020-04-22
[event based orderbook] Implement StableXOrderBookReading for State

e00E:
I am going to add tests when this is fully implemented.

### DIFF
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -14,9 +14,8 @@ use std::collections::BTreeMap;
 // However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
 // list of all events based on which the state is built.
 
-/// The key by which events are sorted.
 #[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
-struct Key {
+struct EventSortKey {
     block_number: u64,
     /// Is included to differentiate events from the same block number but different blocks which
     /// can happen during reorgs.
@@ -32,7 +31,7 @@ struct Value {
 }
 
 pub struct Orderbook {
-    events: BTreeMap<Key, Value>,
+    events: BTreeMap<EventSortKey, Value>,
 }
 
 impl Orderbook {
@@ -50,7 +49,7 @@ impl Orderbook {
         block_timestamp: u64,
     ) {
         let batch_id = block_timestamp as BatchId / 300;
-        let key = Key {
+        let key = EventSortKey {
             block_number,
             block_hash,
             log_index,
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -1,0 +1,85 @@
+use super::*;
+use crate::{
+    contracts::stablex_contract::batch_exchange,
+    models::{AccountState, Order},
+    orderbook::StableXOrderBookReading,
+};
+use anyhow::Result;
+use ethcontract::{contract::EventData, H256, U256};
+use state::{Batch, State};
+use std::collections::BTreeMap;
+
+// Ethereum events (logs) can be both created and removed. Removals happen if the chain reorganizes
+// and ends up not including block that was previously thought to be part of the chain.
+// However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
+// list of all events based on which the state is built.
+
+/// The key by which events are sorted.
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+struct Key {
+    block_number: u64,
+    /// Is included to differentiate events from the same block number but different blocks which
+    /// can happen during reorgs.
+    block_hash: H256,
+    log_index: usize,
+}
+
+#[derive(Debug)]
+struct Value {
+    event: batch_exchange::Event,
+    /// The batch id is calculated based on the timestamp of the block.
+    batch_id: BatchId,
+}
+
+pub struct Orderbook {
+    events: BTreeMap<Key, Value>,
+}
+
+impl Orderbook {
+    pub fn new() -> Self {
+        // TODO: Set up a thread that applies past events and listens for new events.
+        unimplemented!();
+    }
+
+    fn handle_event_data(
+        &mut self,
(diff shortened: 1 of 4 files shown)


driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -0,0 +1,85 @@
...
+                state.apply_event(&value.event, value.batch_id)
+            })
+    }
+}
+
+impl StableXOrderBookReading for Orderbook {
+    fn get_auction_data(&self, _index: U256) -> Result<(AccountState, Vec<Order>)> {
+        // TODO: Handle future batch ids for when we want to do optimistic solving.
+        let state = self.create_state()?;
+        let (account_state, orders) = state.orderbook(Batch::Current)?;
```
fleupold, line 77: I think here it would be handy if Future(index) would also be valid if it's the current batch id.

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -0,0 +1,85 @@
...
+
+// Ethereum events (logs) can be both created and removed. Removals happen if the chain reorganizes
+// and ends up not including block that was previously thought to be part of the chain.
+// However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
+// list of all events based on which the state is built.
+
+/// The key by which events are sorted.
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+struct Key {
+    block_number: u64,
```
fleupold, line 20: nit: Is block_number really needed? Shoudn't hash and index be uniquely identifying an event?

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::collections::BTreeMap;
+
+// Ethereum events (logs) can be both created and removed. Removals happen if the chain reorganizes
+// and ends up not including block that was previously thought to be part of the chain.
+// However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
+// list of all events based on which the state is built.
+
+/// The key by which events are sorted.
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+struct Key {
```
fleupold, line 19: Not important, but if this was called `EventSortKey` or the like we wouldn't need the comment above. I think generally it's nice to replace comments with more explicit naming (as comments tend to not get updated when the code/logic changes).

fleupold (approved):
Nice, we are almost there!

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[derive(Debug)]
+struct Value {
+    event: batch_exchange::Event,
+    /// The batch id is calculated based on the timestamp of the block.
+    batch_id: BatchId,
+}
+
+pub struct Orderbook {
```
fleupold, line 34: Would it make sense to give this struct a prefixed name such as `EventBasedOrderbook`?

bh2smith (approved):

driver/src/orderbook/streamed/orderbook.rs
e00E: Not needed for identifying but needed for ordering.

driver/src/orderbook/streamed/orderbook.rs
e00E: I don't think so because it is already in the streamed module.

nlordell: In general I also tend to prefer shorter names if the module clearly states what its for. Outside of this module we can use `streamed::Orderbook` so its clear that its a streamed version of the orderbook.

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -0,0 +1,84 @@
...
+// and ends up not including block that was previously thought to be part of the chain.
+// However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
+// list of all events based on which the state is built.
+
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+struct EventSortKey {
+    block_number: u64,
+    /// Is included to differentiate events from the same block number but different blocks which
+    /// can happen during reorgs.
+    block_hash: H256,
```
nlordell, line 22: Should `log_index` come before `block_hash` in terms of ordering?

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -0,0 +1,84 @@
...
+        unimplemented!();
+    }
+
+    fn handle_event_data(
+        &mut self,
+        event_data: EventData<batch_exchange::Event>,
+        block_number: u64,
+        log_index: usize,
+        block_hash: H256,
+        block_timestamp: u64,
```
nlordell, line 49: This might be problematic as querying the node for all past time stamps can be really slow.

nlordell (approved):

driver/src/orderbook/streamed/orderbook.rs
e00E: I don't see a way around it. I have tested this on rinkeby by querying the node for all timestamps needed and caching the last timestamp to speed up multiple events from the same block.

driver/src/orderbook/streamed/orderbook.rs
e00E: It does not matter. The only time during which there can be multiple blocks of the same number is during reorg before all old events have been deleted. And in that case the resulting state isn't correct anyway.

nlordell: Actually in that case, having `block_hash` come first as it is makes a little more sense, as the events will be grouped together.

driver/src/orderbook/streamed/orderbook.rs
nlordell: The only way I see would be to handle past events slightly differently than future events, so for past events you apply all the events under the current batch ID (queried from the smart contract for example) and then for incoming events you can query the block timestamp for each (since they will be way more sparse).

Definitely a separate issue/PR though.

PR #757, opened 2020-04-22
[Easy] Ignore balances for orders that we filter out

fleupold:
When testing the shadow orderbook locally I'm getting a lot of difference of the form:

> ERRO [driver::orderbook::shadow_orderbook] user 0xaec3c8ed9516a206a4fd47ec77f026edd533cf17 token 4 primary balance of 1061889845 but shadow balance 0

Reason for this is that the onchain filtering orderbook does not serialize balances for sell tokens for which there are no valid orders, whereas the current implementation of AuctionDataReader takes all balances into account (regardless of whether they are filtered later).

The fix is simply to filter auction elements before serializing their balance.

I also added a print statement for us to know when a consistency check was successful.

### Test Plan

Adjusted unit test, running the shadowed orderbook and seeing them being consistent.

### DIFF
--- a/driver/src/orderbook/auction_data_reader.rs
+++ b/driver/src/orderbook/auction_data_reader.rs
@@ -47,17 +47,16 @@ impl AuctionDataReader {
     }
 
     fn apply_auction_data(&mut self, auction_elements: Vec<StableXAuctionElement>) {
-        self.apply_auction_elements_to_account_state(auction_elements.iter());
         // Workaround for borrow checker that would complain that `extend`
         // borrows self as mutable while at the same time the filter
         // closure borrows self as immutable because of using `self.index`.
         let index = self.index;
-        self.orders.extend(
-            auction_elements
-                .into_iter()
-                .filter(|x| x.in_auction(index) && x.order.sell_amount > 0)
-                .map(|auction_element| auction_element.order),
-        );
+        let auction_elements = auction_elements
+            .iter()
+            .filter(|x| x.in_auction(index) && x.order.sell_amount > 0);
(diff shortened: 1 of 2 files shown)


e00E (approved):

driver/src/orderbook/auction_data_reader.rs
```diff
@@ -47,17 +47,16 @@ impl AuctionDataReader {
...
-                .into_iter()
-                .filter(|x| x.in_auction(index) && x.order.sell_amount > 0)
-                .map(|auction_element| auction_element.order),
-        );
+        let auction_elements = auction_elements
+            .iter()
+            .filter(|x| x.in_auction(index) && x.order.sell_amount > 0);
+        self.apply_auction_elements_to_account_state(auction_elements.clone());
+        self.orders
+            .extend(auction_elements.map(|auction_element| auction_element.order.clone()));
```
nlordell, line 59: Should this be part of the diff logic instead?

nlordell (approved):

driver/src/orderbook/auction_data_reader.rs
bh2smith: Im not sure I can tell what the meaning of this change is. Seems like a reordering of the same logic.

bh2smith (approved):
After re-reading I see that the reordering of existing logic was the intention of the PR.

driver/src/orderbook/auction_data_reader.rs
fleupold: I think it's still valuable to not serialize it into the instance file since it's not needed.

PR #761, opened 2020-04-22
[event based orderbook] Implement updating of orderbook with events

e00E:
This implements the task that updates the orderbook with incoming events
from a stream. After this we still need to actually create the task with
the event stream from the contract which will happen in the next PR.

There will be tests after I see that we like this api.


driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -0,0 +1,23 @@
...
+
+/// Helper trait to make this functionality mockable for tests.
+pub trait BlockTimestamp {
+    fn block_timestamp(&self, block_hash: H256) -> BoxFuture<Result<u64>>;
+}
+
+/// During normal operation this is implemented by Web3.
+impl BlockTimestamp for Web3 {
+    fn block_timestamp(&self, block_hash: H256) -> BoxFuture<Result<u64>> {
+        let r = Compat01As03::new(self.eth().block(BlockId::Hash(block_hash)));
```
nlordell, line 14: nit: I prefer to use `FutureCompatExt` and then just call `.compat()`.

driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -0,0 +1,23 @@
...
+        let r = r.map(move |block| -> Result<u64> {
+            let block = block
+                .with_context(|| format!("failed to get block {}", block_hash))?
+                .with_context(|| format!("block {} does not exist", block_hash))?;
+            Ok(block.timestamp.low_u64())
+        });
+        r.boxed()
```
nlordell, line 21: nit: If the future is going to be boxed, I think the code would be easier to follow using `async { ... }.boxed()`. Then you don't need to use `map` and stuff.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,75 @@
...
+// This way `UpdatingOrderbookThread` is independently testable.
+
+/// Update the orderbook with events from the stream forever or until exit_indicator is dropped.
+///
+/// Returns Ok when exit_indicator is dropped.
+/// Returns Err if the stream ends.
+async fn update_with_events_forever(
+    orderbook: Arc<RwLock<Orderbook>>,
+    mut block_timestamp: impl BlockTimestamp,
+    exit_indicator: mpsc::Receiver<()>,
```
nlordell, line 23: nit: This should be a `oneshot::Receiver`.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,75 @@
...
+    past_events: impl Future<Output = Result<Vec<Event<batch_exchange::Event>>, ExecutionError>>,
```
nlordell, line 24: This might make more sense to be a stream since past events will come in pages. Alternatively the `ethcontract` implementation can collect it into a `Vec` for you as well.

nlordell: Now that I think about it, collecting in a `Vec` will be way easier to implement, so lets leave it as is.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,75 @@
...
+            exit_indicator.try_recv(),
+            Err(mpsc::TryRecvError::Disconnected)
+        )
+    };
+
+    for event in past_events.await? {
+        handle_event(&orderbook, &mut block_timestamp, event).await?;
+    }
+
+    while !should_exit() {
```
nlordell, line 38: This might cause issues if the `stream` never produces events then you will never exit. Alternatively you can use `select!` to choose what comes first between the `exit_indicator` and the `stream`'s next event. 

I would also suggest `select!`ing on the `past_events` as well in case an exit request comes in before the `past_events` query is done.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,75 @@
...
+                .handle_event_data(
+                    data,
+                    meta.block_number,
+                    meta.log_index,
+                    meta.block_hash,
+                    block_timestamp,
+                );
+            Ok(())
+        }
+        Event { meta: None, .. } => bail!("event without metadata"),
```
nlordell, line 73: I think everytime we see a `return Err(anyhow!(...))` in the codebase, we should convert it to `bail!`! Love it!

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: This is an example of select in action:
https://github.com/gnosis/dex-services/blob/1b2beca83f5dc7aeb1205249d74713dd753175bb/driver/src/orderbook/retrying_retrieval.rs#L158-L164

nlordell (approved):
So close to being done! Just some concerns with the exit logic in the events loop.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,75 @@
...
+    let should_exit = || {
+        matches!(
+            exit_indicator.try_recv(),
+            Err(mpsc::TryRecvError::Disconnected)
+        )
+    };
+
+    for event in past_events.await? {
+        handle_event(&orderbook, &mut block_timestamp, event).await?;
+    }
```
nlordell, line 36: I think you should lock the `RwLock` once instead of once per iteration here.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,75 @@
...
+// `pub struct UpdatingOrderbook` that also owns an `Arc<RwLock<Orderbook>>` and creates a
+// background thread running `UpdatingOrderbookThread::update_with_events`.
+// This way `UpdatingOrderbookThread` is independently testable.
+
+/// Update the orderbook with events from the stream forever or until exit_indicator is dropped.
+///
+/// Returns Ok when exit_indicator is dropped.
+/// Returns Err if the stream ends.
+async fn update_with_events_forever(
+    orderbook: Arc<RwLock<Orderbook>>,
```
nlordell, line 21: nit: You will probably get better performance from a `Mutex` here seeing as there is only one reader.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I don't mind either way. I used these types because that's what I prototyped in that other pr in `StablexContract`.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: Your async tips are very helpful. I forgot about select.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I was considering this but then if someone tries to get the orderbook it would block until we have handled all past events. On the other hand  you wouldn't get a useful orderbook anyway if we are still handling past events.

nlordell: Ahhhh, good point. I forgot that the block timestamp needs to be queried in the `handle_event`, yes then the lock should definitely happen once per event.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,82 @@
...
+/// Update the orderbook with events from the stream forever or until exit_indicator is dropped.
+///
+/// Returns Ok when exit_indicator is dropped.
+/// Returns Err if the stream ends.
+async fn update_with_events_forever(
+    orderbook: Arc<Mutex<Orderbook>>,
+    mut block_timestamp: impl BlockTimestamp,
+    exit_indicator: oneshot::Receiver<()>,
+    past_events: BoxFuture<'static, Result<Vec<Event<batch_exchange::Event>>, ExecutionError>>,
+    stream: BoxStream<'static, Result<Event<batch_exchange::Event>, ExecutionError>>,
```
nlordell, line 29: nit: I would make this method generic over the future and stream's lifetime.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,82 @@
...
+        handle_event(&orderbook, &mut block_timestamp, event).await?;
+    }
+
+    loop {
+        let event = select! {
+                event = stream.next() =>
+                    event.ok_or(anyhow!("stream ended"))??,
+                _ = exit_indicator => return Ok(()),
+        };
+        handle_event(&orderbook, &mut block_timestamp, event).await?;
```
nlordell, line 52: Thinking about this a bit more last night, you will probably have to select on all 3 in the loop (since `past_events` is fused it will only resolve once :smile:). That way the stream actually gets polled and `eth_newEventFilter` gets called so the filter gets setup on the node and we don't have a race condition where events come in between the past events being done and the stream getting polled for the first time.

nlordell (approved):
Looks really nice with the `select!`! Just one small concern about a possible race condition since just creating futures or streams actually doesn't do anything until they get polled for the first time.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I have changed this. Technically it still doesn't guarantee that `stream` gets polled first. Not sure if we can improve this. Maybe manually call poll once on the stream?

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -0,0 +1,86 @@
...
+    pin_mut!(past_events);
+    pin_mut!(stream);
+
+    loop {
+        // We select over everything together instead of for example the past events first then the
+        // stream to ensure that the stream gets polled at least once which it needs in order to
+        // create the corresponding filter on the node.
+        select_biased! {
+            _ = exit_indicator => return Ok(()),
+            event = stream.next() => {
```
fleupold, line 47: Don't we require all past events to be applied first?

fleupold (approved):

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: To the state yes. But the `Orderbook` struct used here will order the events internally.

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: This will only be an issue if the any of the futures resolve immediately which will never happen in our case.

driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -0,0 +1,23 @@
+use crate::contracts::Web3;
+use anyhow::{Context as _, Result};
+use ethcontract::{web3::types::BlockId, H256};
+use futures::{compat::Future01CompatExt as _, future::BoxFuture, FutureExt as _};
+
+/// Helper trait to make this functionality mockable for tests.
+pub trait BlockTimestamp {
```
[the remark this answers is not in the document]
nlordell: `BlockTimestampReading` is also nice as it is consistent with our other traits. This makes whenever we see `-ing` type, we immediately identify it as a trait and not a concrete type.

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: Also, since events are stored by key, duplicates are removed :smile:

driver/src/orderbook/streamed/block_timestamp.rs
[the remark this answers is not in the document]
e00E: I will include this name change in the Cache PR to not create a merge conflict.

PR #762, opened 2020-04-22
Min avg fee per trade

josojo:
This PR replaces the SolverType enum with a SolverConfig enum of structs. The long term value proposition is that we can store all solver configurations touching solely the solver in one place. Short term, it allows us to introduce a new variable min-avg-fee-per-trade.

Hopefully, the solver configurations are also easier to recognize/understand, if they are all provided in one json:
```
let json = r#"{
            "StandardSolver": { "min_avg_fee_per_order": 180 }
        }"#;
```
instead of several independent env variables.

Further refactorings could also import the time-limit variable into this SolverConfig structs.

---
testplan: unittests


e00E:
I prefer separate options that have similar names like `--solver-type open-solver --solver-time-limit 1 --sovler-min-avg-fee-per-order 180`. It doesn't feel right to config options as json.
We can still have this enum of struct in our code but I wouldn't change the config parsing.
On the other hand this is low importance to me so if you feel strongly that this is better then that's fine too.

README.md
```diff
@@ -197,9 +197,18 @@ OPTIONS:
...
+            Configuration for the solver. The possible solver types are:
+            'NaiveSolver' for the naive matching just two orders
+            'OpenSolver' for solutions on trade-pairs only
+            'StandardSolver' for solutions with ring-trades (mixed integer programming)
+
+            min_avg_fee_per_order allows to enforce economical viable solutions by
+            setting a minimum fee per order.
+
+            For example: {
+              "StandardSolver": { "min_avg_fee_per_order": 10000000000000000 }
```
bh2smith, line 210: Do we know what value we will be using for this? I guess what you have here is 0.1 OWL

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -24,51 +26,62 @@ impl Default for Fee {
...
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq)]
+pub enum SolverConfig {
     NaiveSolver,
-    StandardSolver,
-    FallbackSolver,
-    OpenSolver,
+    StandardSolver { min_avg_fee_per_order: u128 },
+    FallbackSolver { min_avg_fee_per_order: u128 },
+    OpenSolver { min_avg_fee_per_order: u128 },
 }
```
bh2smith, line 35: Nice Enum!

README.md
josojo: Yeah, we were thinking about 0.01 OWL, but it's up to adjustments/discussions/investigations

driver/src/price_finding/price_finder_interface.rs
fleupold: It's a bit weird that all solvers have the same filed (the fact that the NaiveSolver doesn't is probably just a missing feature inside the solver. Thus, isn't this new argument more similar to the already existing time limits, token info, etc.?

README.md
bh2smith: 0.01 OWL seems very low considering the transaction costs of solution submission.

driver/src/price_finding/price_finder_interface.rs
josojo: yeah, it's similar to time limits.
> Further refactorings could also import the time-limit variable into this SolverConfig structs.

I suggested to put it into the solver config if this style of a wholistic solver config is wanted.

The tokendata is mostly consumed by the solver, so I would not put it into the solver config.

josojo:
in this PR I did not want to hand down two parameters: time-limit and min_avg_fee_per_trade to all the functions.
Hence I created this solver_config enum to bundle these variables together.

But if you guys don't like it, I can create the solver_config struct outside of the enum.

josojo:
> We can still have this enum of struct in our code but I wouldn't change the config parsing.
If you guys agree, thumbs up please..

driver/src/price_finding/price_finder_interface.rs
fleupold: I would suggest to handle this new parameter like we handle the time limit and then propose the solver config enum with all the components that should go in there in a separate "refactoring" PR.

bh2smith:
> If you guys agree, thumbs up please..

I am indifferent to this opinion mostly because I am not currently aware of the pros and cons of the options here.

PR #763, opened 2020-04-22
[event based orderbook] Create the final self updating event based orderbook struct

e00E:
This is the struct that we reexport from the streamed module.


driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -12,10 +16,52 @@ use futures::{
...
+    ) -> Self {
+        let orderbook = Arc::new(Mutex::new(Orderbook::default()));
+        let orderbook_clone = orderbook.clone();
+        let (sender, receiver) = oneshot::channel();
+        // Create stream first to make sure we do not miss any events between it and past events.
+        // TODO: use the real functions once they are implemented
+        let stream = futures::stream::iter(vec![]).boxed(); // contract.stream_events();
+        let past_events = futures::future::ready(Ok(Vec::new())).boxed(); // contract.past_events();
+
+        std::thread::spawn(move || {
```
nlordell, line 40: Maybe it makes sense if this is done in two parts - first the past events need to be retrieved, and then the future events, otherwise it doesn't really make sense to get the orderbook while its still catching up on all the past events.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -12,10 +16,52 @@ use futures::{
...
+    pub fn new(
+        _contract: &impl StableXContract,
+        block_timestamp: impl BlockTimestamp + Send + 'static,
+    ) -> Self {
+        let orderbook = Arc::new(Mutex::new(Orderbook::default()));
+        let orderbook_clone = orderbook.clone();
+        let (sender, receiver) = oneshot::channel();
+        // Create stream first to make sure we do not miss any events between it and past events.
+        // TODO: use the real functions once they are implemented
+        let stream = futures::stream::iter(vec![]).boxed(); // contract.stream_events();
```
nlordell, line 37: Note that just creating the stream is not enough - it needs to be polled (and possibly `select!`ed with the `past_events` and `receiver` futures) so that it actually creates the filter on the node and starts polling it.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -12,10 +16,52 @@ use futures::{
...
+}
+
+impl UpdatingOrderbook {
+    pub fn new(
+        _contract: &impl StableXContract,
+        block_timestamp: impl BlockTimestamp + Send + 'static,
+    ) -> Self {
+        let orderbook = Arc::new(Mutex::new(Orderbook::default()));
+        let orderbook_clone = orderbook.clone();
+        let (sender, receiver) = oneshot::channel();
```
nlordell, line 34: naming nit: maybe call the sender and receiver based on what they are used for? Maybe `(exit_tx, exit_rx)`?

nlordell (approved):

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: Hmm and we need to make sure GetFilterChanges is called constantly and web3 probably has an internal buffer of events and stops calling that if the buffer fills up which could make the filter get deleted.
We could as you say select! it with the past events and forward the stream events to the orderbook too as it doesn't matter if they come in the wrong order.
Then with regards to your comment below there wouldn't be two parts but we could keep a flag "orderbook finished catching up" that is set when past events are handled and querying the orderbook before that fails with an error. Then we can also lock the mutex for all past events because flag would get checked before locking the mutex.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: Implemented something like this with a bool that gets set when orderbooks is caught up.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -11,19 +15,77 @@ use futures::{
...
+                orderbook_clone,
+                orderbook_ready_clone,
+                block_timestamp,
+                exit_rx,
+                past_events,
+                stream,
+            ));
+            if let Err(err) = result {
+                log::error!("event based orderbook failed: {}", err);
+            }
```
e00E, line 58: I am wondering about this. Let's say the node goes down for a short time. In this version the orderbook would fail and never update again but the program as a whole wouldn't exit. We should probably add a retry mechanism to recreate the orderbook from scratch if `update_with_events_forever` exits.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -11,19 +15,77 @@ use futures::{
...
+
+impl UpdatingOrderbook {
+    pub fn new(
+        _contract: &impl StableXContract,
+        block_timestamp: impl BlockTimestamp + Send + 'static,
+    ) -> Self {
+        let orderbook = Arc::new(Mutex::new(Orderbook::default()));
+        let orderbook_clone = orderbook.clone();
+        let orderbook_ready = Arc::new(AtomicBool::new(false));
+        let orderbook_ready_clone = orderbook_ready.clone();
```
fleupold, line 40: @nlordell wasn't there a syntactically nice way of doing the clone inside the closure to avoid ugly naming like this?

driver/src/orderbook/streamed/updating_orderbook.rs
fleupold: yes, we should do that. For now we could also panic, which would cause the pod to restart, but an intrinsic retry mechanism is probably good going forward.
I wonder if the stream itself could become a retrying stream (since this could avoid refetching the entire orderbook again).

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -11,19 +15,77 @@ use futures::{
...
-use std::sync::{Arc, Mutex};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc, Mutex,
+};
+
+/// An event based orderbook that automatically updates itself with new events from the contract.
+#[derive(Debug)]
+pub struct UpdatingOrderbook {
+    orderbook: Arc<Mutex<Orderbook>>,
```
fleupold, line 26: Nit: I wonder if this could be an Option which is None initially and then the thread sends backs the orderbook on a one shot channel as soon as it's ready. It's semantically the same, but avoids having a boolean which is implicitly coupled to another variable.

e00E: I like that. Will turn it into an option that is set by the thread (not sent over a channel because we already share this variable) after past events have been handled.

e00E: Didn't work out because then `update_with_events_forever` gets more complicated as it is possible that it gets an event from the stream first before past events.

The way you described would work but then we trade the atomic bool for a channel and channel code is more complicated than the bool in this case so I want to keep it as is.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I added an `abort`. Panic isn't enough because this happens in a thread.

PR #764, opened 2020-04-22
[event based orderbook] Implement a cache for the block timestamp

e00E:
This avoids having to query the node in the case where we receive
multiple events from the same block in a row.


driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -21,3 +21,37 @@ impl BlockTimestamp for Web3 {
...
+// BoxFuture even on cache hit and it makes it harder to update the cache after waiting for the
+// `self.block_timestamp.block_timestamp` future.
+impl<T: BlockTimestamp> BlockTimestampCache<T> {
+    pub async fn block_timestamp(&mut self, block_hash: H256) -> Result<u64> {
+        if self.hash != block_hash {
+            let timestamp = self.block_timestamp.block_timestamp(block_hash).await?;
+            self.hash = block_hash;
+            self.timestamp = timestamp;
+        }
+        Ok(self.timestamp)
```
nlordell, line 55: Shouldn't you just be able to box an async block like this:

```suggestion
        async move {
            if self.hash != block_hash {
                let timestamp = self.block_timestamp.block_timestamp(block_hash).await?;
                self.hash = block_hash;
                self.timestamp = timestamp;
            }
            Ok(self.timestamp)
        }.boxed()
```

driver/src/orderbook/streamed/mod.rs
```diff
@@ -17,4 +17,4 @@ type TokenId = u16;
 type BatchId = u32;
 
 pub use block_timestamp::BlockTimestamp;
-pub use updating_orderbook::UpdatingOrderbook as Orderbook;
+pub use updating_orderbook::UpdatingOrderbook as Orderbook;
```
nlordell, line 20: nit: missing line at end of file.

nlordell (approved):

nlordell:
It would be cool if we could get `BlockTimestampCache` to implement `BlockTimestamp` but I'm OK if it over-complicates things.

driver/src/orderbook/streamed/block_timestamp.rs
nlordell: Maybe the `move` is not necessary, but instinctively it is so that the future's lifetime is tied to `self`.

Also I noticed that the trait doesn't accept a mutable reference. Additionally, you might require to explicitely set lifetimes:
```
fn block_timestamp(&'a mut self, block_hash: H256) -> BoxFuture<'a, Result<u64>>;
```

e00E: Thank you. It worked with your first suggestion and I changed the trait to `&mut self`.

driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -21,3 +21,36 @@ impl BlockTimestamp for Web3 {
...
+    pub fn new(block_timestamp: T) -> Self {
+        Self {
+            block_timestamp,
+            hash: H256::zero(),
+            timestamp: 0,
+        }
+    }
+}
+
+impl<T: BlockTimestamp + Send> BlockTimestamp for BlockTimestampCache<T> {
```
nlordell, line 44: This is so cool! Layering abstractions! :tada:

driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -21,3 +21,36 @@ impl BlockTimestamp for Web3 {
         .boxed()
     }
 }
+
+/// A cache for the block timestamp which avoids having to query the node in the case where we
+/// receive multiple events from the same block in a row.
+#[derive(Debug)]
+pub struct BlockTimestampCache<T> {
+    block_timestamp: T,
```
fleupold, line 29: To me a `block_timestamp` is an already materialized concrete timestamp for a block not a component that allows querying the block_timestamp (given that T can be anything here it's not clear from the type either).
Can we maybe call this `block_timestamp_reader` or even more abstract `cache_miss_reader` or the like?

driver/src/orderbook/streamed/block_timestamp.rs
fleupold: Is there value in having no type specialization in the struct definition (and only adding it here). I think it hides a bit the purpose of this struct behind unused generalization.

driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -21,3 +21,36 @@ impl BlockTimestamp for Web3 {
...
+    hash: H256,
+    timestamp: u64,
+}
+
+impl<T> BlockTimestampCache<T> {
+    pub fn new(block_timestamp: T) -> Self {
+        Self {
+            block_timestamp,
+            hash: H256::zero(),
+            timestamp: 0,
```
fleupold, line 39: Is there on value in caching multiple blocks here instead of only the last one?

driver/src/orderbook/streamed/block_timestamp.rs
```diff
@@ -21,3 +21,36 @@ impl BlockTimestamp for Web3 {
...
+/// A cache for the block timestamp which avoids having to query the node in the case where we
+/// receive multiple events from the same block in a row.
+#[derive(Debug)]
+pub struct BlockTimestampCache<T> {
+    block_timestamp: T,
+    hash: H256,
+    timestamp: u64,
+}
+
+impl<T> BlockTimestampCache<T> {
```
fleupold, line 34: nit naming suggestion: MemozingBlockTimestamp[Reader]

fleupold (approved):
Looks good to me (some small naming suggestions). Have you considered using a more generalized library like https://docs.rs/cached/0.12.0/cached/ for this?

driver/src/orderbook/streamed/block_timestamp.rs
fleupold: or even just calling it `inner` would be helpful IMO

driver/src/orderbook/streamed/block_timestamp.rs
e00E: No because events nearly always come in in order.

e00E:
> Have you considered using a more generalized library like https://docs.rs/cached/0.12.0/cached/ for this?

That is overkill for such a simple struct.

driver/src/orderbook/streamed/block_timestamp.rs
e00E: I prefer requiring bounds only in the place where needed. That makes it more clear to me why they are required.

driver/src/orderbook/streamed/block_timestamp.rs
fleupold: When there is a reorg, we might get events again from a previous block, wouldn't we? And it's really not a lot of storage, is it? Are you worried about cache eviction logic?

fleupold:
> That is overkill for such a simple struct.

I don't have a strong preference but my thinking is that you are implementing something very standard (a memoizer for a function call). And doing this yourself while it's "easy", adds logic that needs to be maintained going forward. 
What exactly is the overkill? Is this library very large? Does it add runtime overhead? I would expect under the hood it works very similar to what you implemented.

driver/src/orderbook/streamed/block_timestamp.rs
e00E: This optimization really only does something for the past events because there we have a long list of events in order and need the timestamp for all of them.
New events from the stream don't occur that often so it won't help much there. Reorgs occur rarely and still contain a small number of events compared to the past events.

e00E:
This implementation is maybe 30 lines and used in one place. I can understand it quickly because it just caches the last value. I don't expect to have to maintain it because it wont change. With the dependency I feel that I have to read a lot more of its documentation to understand how to use it correctly .
It is not a clear line though. If we were already using this dependency or started caching many different things I would probably prefer the library.

PR #765, opened 2020-04-22
Do not run solver if time limit rounds to zero

e00E:
This does not make sense and the solver has an assertion that the limit
is greater than 0 if set.

I noticed this because we got a solver failure on rinkeby and eventually I found this line in kibana:

>  assert isinstance(timelimit, (int, float)) and timelimit > 0 or timelimit is None

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -154,9 +154,12 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
             Err(err) => return DriverResult::Retry(err),
         };
 
+        // Make sure the solver has at least some minimal time to run to have a chance for a
+        // solution. This also fixes an assert where the solver fails if the timelimit gets rounded
+        // to 0.
         let price_finding_time_limit = match deadline.checked_duration_since(Instant::now()) {
-            Some(time_limit) => time_limit,
-            None => {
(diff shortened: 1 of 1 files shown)


nlordell (approved):

e00E:
I changed this to move the logic `stablex_driver.rs` so that this is treated the same way as if the time limit was 0.

fleupold (approved):

driver/src/driver/stablex_driver.rs
```diff
@@ -154,9 +154,12 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
             Err(err) => return DriverResult::Retry(err),
         };
 
+        // Make sure the solver has at least some minimal time to run to have a chance for a
+        // solution. This also fixes an assert where the solver fails if the timelimit gets rounded
+        // to 0.
```
bh2smith, line 159: Where is the rounding being done?

bh2smith (approved):

driver/src/driver/stablex_driver.rs
e00E: https://github.com/gnosis/dex-services/blob/2f87ca88761d3e97aad566939d9b67f7ff91b9ae/driver/src/price_finding/optimization_price_finder.rs#L290

nlordell (approved):

PR #768, opened 2020-04-23
[Easy] Remove bulky debug log statements

bh2smith:
Each solver was previously logging the solver input and results as a json file every five minutes. This was causing the logs to grow very quickly. This removes them as they are accessible from other places anyway.


fleupold (approved):
✂️

e00E (approved):

bh2smith:
Do we need to restart the solver pods or will this all happen automagically?

PR #769, opened 2020-04-23
Simple solution for the new parameter min-avg-fee-per-order

josojo:
Replaces the "solver_config" solution from here:
https://github.com/gnosis/dex-services/pull/762

Just reads the single parameter MIN_AVG_FEE_PER_ORDER and stores it in the OptimisationFinder struct, in parallel to the solver_type

---
testplan: e2e tests

### DIFF
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -169,6 +169,7 @@ trait Io {
         result_folder: &str,
         solver_type: SolverType,
         time_limit: Duration,
+        min_avg_fee_per_order: u128,
     ) -> Result<()>;
     fn read_output(&self, result_folder: &str) -> std::io::Result<String>;
 }
@@ -178,6 +179,7 @@ pub struct OptimisationPriceFinder {
     fee: Option<Fee>,
     solver_type: SolverType,
     price_oracle: Box<dyn PriceEstimating + Sync>,
+    min_avg_fee_per_order: u128,
 }
 
 impl OptimisationPriceFinder {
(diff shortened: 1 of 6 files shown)


e00E (approved):

nlordell (approved):
Easy!

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -169,6 +169,7 @@ trait Io {
         result_folder: &str,
         solver_type: SolverType,
         time_limit: Duration,
+        min_avg_fee_per_order: u128,
```
fleupold, line 172: Since we have access to self in this method, do we need the extra argument? Can't we just use self. min_avg_fee_per_order inside the method?

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -97,6 +106,7 @@ pub fn execute_private_solver(
         .arg(format!("{}{}", "/app/", input_file.to_owned()))
         .arg(format!("--outputDir={}{}", "/app/", result_folder))
         .args(&["--solverTimeLimit", &time_limit])
+        .arg(format!("--minAvgFeePerOrder={}", min_avg_fee_per_order))
```
fleupold, line 109: How unfortunate that these arguments differ slightly across our solvers/.

fleupold (approved):
Thanks for making the change. I agree with your general sentiment that all these configs should be refactored in a nicer lass scattered way.

bh2smith (approved):

nlordell:
> Thanks for making the change. I agree with your general sentiment that all these configs should be refactored in a nicer lass scattered way.

One suggestion, maybe we can include all solving parameters as part of the solver input file? A special `config` section with all these parameters? That ensures we have a well-defined solver interface that is used by all our solvers.

driver/src/price_finding/optimization_price_finder.rs
josojo: self is referring to the DefaultIo or Io,  but not to the optimizaiton_price_finder itself

driver/src/price_finding/price_finder_interface.rs
josojo: The two solvers have different standards. Open-solver takes all parameters-like-this, while the standard solver takes all parametersLikeThis.

Hence, we would have to switch all parameters in one solver. Maybe we do that in the future.

PR #772, opened 2020-04-23
Use event based orderbook as primary orderbook

bh2smith:
This big exciting new PR takes all the challenging accomplishments from #678 and puts them to work by making the tremendous new Event Based Orderbook our driver's primary orderbook!

In brief, this PR simply set aside the existing primary order book `PaginatedOrderbook` to make room for the cool new `EventOrderbook`. 

Depends on a version bump of ethcontracts!

Test Plan: I would say that end to end tests should give us a pretty good idea of whether or not this is working. Note that acquiring all the past events could eventually become very slow... Not sure how this will affect the e2e testing.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -102,6 +102,10 @@ struct Options {
     #[structopt(long, env = "ORDERBOOK_FILTER", default_value = "{}")]
     orderbook_filter: OrderbookFilter,
 
+    /// Primary method for orderbook retrieval ("Paginated" or "OnchainFiltered")
+    #[structopt(long, env = "PRIMARY_ORDERBOOK", default_value = "Paginated")]
+    primary_orderbook: String,
+
     /// The private key used by the driver to sign transactions.
     #[structopt(short = "k", long, env = "PRIVATE_KEY", hide_env_values = true)]
     private_key: PrivateKey,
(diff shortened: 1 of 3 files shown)


nlordell:
@bh2smith Can you make it merge into the `even-orderbook-0` branch to make it easier to see the relevant diff?

driver/src/main.rs
```diff
@@ -212,8 +212,7 @@ fn main() {
     let price_finder = price_finding::create_price_finder(fee, options.solver_type, price_oracle);
 
     // Create the orderbook reader.
-    let primary_orderbook =
-        PaginatedStableXOrderBookReader::new(contract.clone(), options.auction_data_page_size);
+    let primary_orderbook = EventOrderbook::new(contract.as_ref(), web3);
```
nlordell, line 215: I think we want the `Options` struct to have a new option so we can configure at runtime which reader to use, either `OnchainOrderBookReader` shadowed by the `EventOrderbook` or the other way around.

It would be something similar to what we did with the  `scheduler` option and the `SchedulerKind` enum just above.

bh2smith:
@nlordell 

> @bh2smith Can you make it merge into the even-orderbook-0 branch to make it easier to see the relevant diff?

I was assuming this would be merged by now. I think that is the better option.

driver/src/main.rs
bh2smith: This is part of a different PR #773 unless you mean something different here.

bh2smith: Actually yes, this PR is now meant to be the second in a series of two. #773 introduces the configurability while this will simply add the new Event Orderbook to the configuration possibilities.

e00E:
It was merged but the merge included a squash whereas your branch doesn't have the commits squashed. What you can try is copying main.rs out (because that is the only file that changes), resetting this branch to master `git reset --hard origin/master` (careful this deletes all changes), then copying main back. Force push and finally change the base in github back to master.

nlordell:
> git reset --hard origin/master

Scary :fearful: ... I would try to rebase first:
- fetch latest master and update your branch to point to it: `git fetch origin master:master`
- rebase the latest commit onto master `git rebase --onto master HEAD~1 use-event-orderbook`

bh2smith:
NO worries, folks its all good now although I think this PR should probably come after #773

driver/src/orderbook/mod.rs
```diff
@@ -8,6 +8,7 @@ use self::auction_data_reader::PaginatedAuctionDataReader;
 pub use self::filtered_orderbook::{FilteredOrderbookReader, OrderbookFilter};
 pub use self::onchain_filtered_orderbook::OnchainFilteredOrderBookReader;
 pub use self::shadow_orderbook::ShadowedOrderbookReader;
+pub use self::streamed::Orderbook as EventOrderbook;
```
fleupold, line 11: naming nit: This sounds a bit like an orderbook full of events to me. Should we call it EventBasedOrderook or something like that (maybe there is a better option than that).

fleupold (commented):
Given this PR is only 5 lines, should it maybe be combined with #773?
Also, merging this by itself is a bit dangerous as thanks to autodeploy it would make the Event based orderbook the default on all our staging hosts.

driver/src/orderbook/mod.rs
bh2smith: I like `EventBasedOrderbook` but could also get into `EventStreamedOrderbook` if anyone is interested.

nlordell:
> Given this PR is only 5 lines, should it maybe be combined with #773?

Personally I like #773 as a standalone PR, as being able to configure the primary orderbook is valuable even without the event based orderbook being integrated and feels like a nice self-contained change.

bh2smith:
Ok this branch is now pointed at the `configurable_orderbook` branch and depends on having first merged #773. Both are ready for review.

nlordell (approved):
:sparkles:

fleupold (approved):
🍰

PR #773, opened 2020-04-23
Configurable Primary Orderbook

bh2smith:
This simple change introduces a `primary_orderbook` argument to the run time configuration in which the user can specify "Paginated" or "OnchainFiltered". 

Will eventually include "EventBased"

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -102,6 +102,10 @@ struct Options {
     #[structopt(long, env = "ORDERBOOK_FILTER", default_value = "{}")]
     orderbook_filter: OrderbookFilter,
 
+    /// Primary method for orderbook retrieval ("Paginated" or "OnchainFiltered")
+    #[structopt(long, env = "PRIMARY_ORDERBOOK", default_value = "Paginated")]
+    primary_orderbook: String,
+
     /// The private key used by the driver to sign transactions.
     #[structopt(short = "k", long, env = "PRIVATE_KEY", hide_env_values = true)]
     private_key: PrivateKey,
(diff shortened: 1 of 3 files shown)


driver/src/main.rs
```diff
@@ -102,6 +102,10 @@ struct Options {
     #[structopt(long, env = "ORDERBOOK_FILTER", default_value = "{}")]
     orderbook_filter: OrderbookFilter,
 
+    /// Primary method for orderbook retrieval ("Paginated" or "OnchainFiltered")
+    #[structopt(long, env = "PRIMARY_ORDERBOOK", default_value = "Paginated")]
+    primary_orderbook: String,
```
fleupold, line 107: Can you make this typesafe by doing something similar to what we have with `SOLVER_TYPE`. Basically creating an enum that implements FromStr. This way the logic of what cases are possible is colocated with the orderbook module and not inlined in this large main file.

nlordell (commented):
Looks good! I would just recommend the change that @fleupold suggested. You can take a look at how we did it with other enums like `SchedulerKind`: https://github.com/gnosis/dex-services/blob/d787830a6940a47c92af2ceb67081787a680ebb6/driver/src/driver/scheduler.rs#L73-L107

nlordell (approved):
Awesome! Squeaky clean PR :rocket:

driver/src/orderbook/paginated_orderbook.rs
```diff
@@ -0,0 +1,45 @@
...
+use super::StableXOrderBookReading;
+use anyhow::Result;
+use ethcontract::{BlockNumber, U256};
+use std::convert::TryInto;
+use std::sync::Arc;
+
+/// Implements the StableXOrderBookReading trait by using the underlying
+/// contract in a paginated way.
+/// This avoid hitting gas limits when the total amount of orders is large.
+pub struct PaginatedStableXOrderBookReader {
```
fleupold, line 14: Next time,  refactoring like this can maybe be its own PR

fleupold (approved):
Nice!

PR #776, opened 2020-04-24
[event based ordrebook] Fix revert of unlimited trade

e00E:
I noticed this yesterday when testing with rinkeby events but forgot to push it.


nlordell (approved):

bh2smith (commented):
What was the problem with the ordering of this code block before? "Fix revert of unlimited trade" is not very descriptive.

e00E:
I guess I fell into the "obvious" trap again. The `ensure` only makes sense if the order has a limited amount. It would fail if the order was unlimited. So we should only do the ensure after we know that the order is limited.

fleupold (approved):
I also didn't see at first, why this is necessary, but with your comment it makes sense.

PR #778, opened 2020-04-24
[Easy] Bump ethcontract 0.7.1

nlordell:
:point_up: 

This includes the new paginated event query which is needed for the event based orderbook. Note that 0.7.0 was skipped because of clippy lint errors in the generated code because of the newly released Rust 1.43.

### Test Plan

CI, only new stuff so existing code shouldn't be affected.


bh2smith (approved):
`is_good = True`

fleupold (approved):

PR #779, opened 2020-04-24
[easy] Revert bump of open-solver version

e00E:
The open solver on rinkeby is outputting invalid solutions which spams
the alert channel. Let's see if reverting the version bump fixes it.


bh2smith (approved):

e00E:
Merging this since it is `easy` and to get rid of the alerts and Felix recommend doing on Slack earlier.

PR #781, opened 2020-04-24
Do not use a stream for event based orderbook

e00E:
Using a error stream prevents use from using load balanced nodes because
if the node changes under us the new node will not have the log filter
that was set up by the stream.
Instead we always query all past logs of the last X blocks, delete the
previous events from this range and then apply the one.
This works even when the node switches and makes it easier to implement
retries on errors in the futures.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -128,6 +128,8 @@ pub trait StableXContract {
 
     fn past_events(
         &self,
+        from_block: BlockNumber,
+        to_block: BlockNumber,
     ) -> BoxFuture<'static, Result<Vec<Event<batch_exchange::Event>>, ExecutionError>>;
 
     fn stream_events(
@@ -290,11 +292,13 @@ impl StableXContract for StableXContractImpl {
 
     fn past_events(
         &self,
+        from_block: BlockNumber,
+        to_block: BlockNumber,
     ) -> BoxFuture<'static, Result<Vec<Event<batch_exchange::Event>>, ExecutionError>> {
         self.instance
             .all_events()
-            .from_block(ethcontract::BlockNumber::Earliest)
-            .to_block(ethcontract::BlockNumber::Latest)
+            .from_block(from_block)
+            .to_block(to_block)
             .query_past_events_paginated()
             .boxed()
     }
--- a/driver/src/orderbook/streamed/updating_orderbook.rs
+++ b/driver/src/orderbook/streamed/updating_orderbook.rs
@@ -9,16 +9,12 @@ use crate::{
 };
 use anyhow::{anyhow, bail, ensure, Result};
 use block_timestamp_reading::{BlockTimestampReading, CachedBlockTimestampReader};
-use ethcontract::{contract::Event, errors::ExecutionError, H256};
+use ethcontract::{contract::Event, BlockNumber, H256};
 use futures::{
-    channel::oneshot,
-    future::FutureExt,
-    pin_mut, select_biased,
-    stream::{Stream, StreamExt as _},
+    channel::oneshot, compat::Future01CompatExt as _, future::FutureExt, pin_mut, select_biased,
 };
 use orderbook::Orderbook;
 use std::collections::HashSet;
-use std::future::Future;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
     Arc, Mutex,
@@ -37,24 +33,21 @@ pub struct UpdatingOrderbook {
 }
 
 impl UpdatingOrderbook {
-    pub fn new(contract: &dyn StableXContract, web3: Web3) -> Self {
+    pub fn new(contract: Arc<dyn StableXContract + Send + Sync>, web3: Web3) -> Self {
         let orderbook = Arc::new(Mutex::new(Orderbook::default()));
         let orderbook_clone = orderbook.clone();
         let orderbook_ready = Arc::new(AtomicBool::new(false));
         let orderbook_ready_clone = orderbook_ready.clone();
         let (exit_tx, exit_rx) = oneshot::channel();
-        // Create stream first to make sure we do not miss any events between it and past events.
-        let stream = contract.stream_events();
-        let past_events = contract.past_events();
 
         std::thread::spawn(move || {
             let result = futures::executor::block_on(update_with_events_forever(
+                web3.clone(),
+                contract,
                 orderbook_clone,
                 orderbook_ready_clone,
                 CachedBlockTimestampReader::new(web3),
                 exit_rx,
-                past_events,
-                stream,
             ));
             if let Err(err) = result {
                 log::error!("event based orderbook failed: {:?}", err);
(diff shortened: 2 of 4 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -288,13 +290,15 @@ impl StableXContract for StableXContractImpl {
         Ok(())
     }
 
-    fn past_events(
+    fn events(
```
fleupold, line 293: nit: I'd still call it getPastEvents since the toBlock can not really be in the future, can it?

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -89,51 +83,71 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+    contract: Arc<dyn StableXContract>,
     orderbook: Arc<Mutex<Orderbook>>,
     orderbook_ready: Arc<AtomicBool>,
     mut block_timestamp_reader: impl BlockTimestampReading,
     exit_indicator: oneshot::Receiver<()>,
-    past_events: impl Future<Output = Result<Vec<Event<batch_exchange::Event>>, ExecutionError>>,
-    stream: impl Stream<Item = Result<Event<batch_exchange::Event>, ExecutionError>>,
 ) -> Result<()> {
-    // `select!` requires the futures to be fused...
+    const POLL_INTERVALL: Duration = Duration::from_secs(30);
```
fleupold, line 93: If we want to poll, we probably want to do this much more frequently (and something we can configure via an ENV variable). However, this made me think that we might just want to fetch the last "gap" whenever the orderbook is queried and not have a thread at all (since the atomic var is preventing us from accessing the orderbook before we processed everything anyways).

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -89,51 +83,71 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+    }
 
     loop {
-        // We select over everything together instead of for example the past events first then the
-        // stream to ensure that the stream gets polled at least once which it needs in order to
-        // create the corresponding filter on the node.
+        // TODO: This is not optimal because it means we might sleep for some time when exit
+        // indicator triggered. We should create a sleep future that expires after poll interval and
+        // join both together.
+        std::thread::sleep(POLL_INTERVALL);
```
fleupold, line 114: Maybe we should sleep at the end of the loop?

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -89,51 +83,71 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+        let current_block = web3.eth().block_number().compat().await?;
+        let from_block = current_block.saturating_sub(From::from(BLOCK_RANGE));
```
fleupold, line 116: I'm wondering what happens if the application of all state takes longer than 25 blocks (which on my machine is definitely the case). Would it make sense to have this be the min of current - BLOCK_RANGE and the maximum existing block number?

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: I think that might work nicely for the driver.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: Added a fix for this.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I like this idea because it get rid of the thread.

e00E: I want to implement that in a separate PR.

e00E: How do you think the initial update should work? Still in the background or should there be a blocking init method? I don't think  it should happen on first orderbook request because it would block for too long.

nlordell: Maybe it could happen at startup time before the scheduler gets started. This way once we start the scheduler and solving, the orderbook is ready.

The downside is that we may miss some batches when we restart, but that happens anyway now.

fleupold: At the moment we are seeing a lot of "Orderbook not yet ready for fetching" errors on startup, because we are blocking access to the orderbook until the first fetch is complete. 

I agree that we should not waste the initial waiting time (if we start between minute 3 and minute 5) and use it to fetch the orderbook, but I think it can happen on a blocking thread, since we cannot really do any processing before anyways.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -92,51 +85,67 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+    let exit_indicator = exit_indicator.fuse();
+    pin_mut!(exit_indicator);
+
+    let mut last_handled_block = 0u64;
     loop {
-        // We select over everything together instead of for example the past events first then the
-        // stream to ensure that the stream gets polled at least once which it needs in order to
-        // create the corresponding filter on the node.
+        let current_block = web3.eth().block_number().compat().await?;
+        let from_block = last_handled_block.saturating_sub(BLOCK_RANGE);
```
fleupold, line 107: 👍 Make sense. I was originally thinking to start from MIN(current - block_range, last_handled_block), but it's right that we might have missed a reorg that happened around last_handled_block.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -92,51 +85,67 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+        // join both together.
+        std::thread::sleep(POLL_INTERVALL);
     }
 }
 
+/// Apply a vector of events to the orderbook.
+async fn handle_events(
+    orderbook: &Mutex<Orderbook>,
+    block_timestamp_reader: &mut impl BlockTimestampReading,
+    events: Vec<Event<batch_exchange::Event>>,
```
fleupold, line 129: cool, now that there is one function for this, we probably want to prepare the timestamp cache here.

fleupold (approved):
Looks good to me! 💯

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -93,56 +86,78 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+                .ok_or_else(|| anyhow!("event without metadata: {:?}", event))?;
+            Ok(metadata.block_hash)
+        })
+        .collect::<Result<HashSet<H256>>>()?;
+    block_timestamp_reader.prepare_cache(block_hashes).await?;
+    // Locking here ensures that the orderbook is not observable after the events have been deleted
+    // but the new events not yet applied.
+    let mut orderbook = orderbook
+        .lock()
+        .map_err(|err| anyhow!("poison error: {}", err))?;
```
bh2smith, line 149: What is a poison error?

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -93,56 +86,78 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+    orderbook.delete_events_starting_at_block(delete_events_starting_at_block);
```
bh2smith, line 150: I find it awkward that we are calling a function whose name is the same as the variable we pass it.

Also, why are we "deleting events"?

driver/src/orderbook/streamed/updating_orderbook.rs
fleupold: https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning

driver/src/orderbook/streamed/updating_orderbook.rs
fleupold: Because we are refetching all events from `delete_events_starting_at_block` and thus we want to delete our prior version of these events (as they might have changed in a reorg).

For naming I'd suggest something like `refetch_cutoff_block` or the like, but I don't think it's important enough right now to justify another round trip of travis.

PR #782, opened 2020-04-24
Filter account states in event based orderbook

e00E:
By default the event based orderbook returns balances for users even
when the user does not have an order with that token. The other
orderbooks don't do that which confuses the shadowed orderbook diff.

### DIFF
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -65,16 +65,54 @@ impl Orderbook {
     }
 }
 
+fn filter_account_state(
+    account_states: impl Iterator<Item = ((UserId, TokenId), U256)>,
+    orders: &[Order],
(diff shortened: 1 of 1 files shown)


driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -65,16 +65,30 @@ impl Orderbook {
...
 
+fn filter_account_state(
+    account_states: impl Iterator<Item = ((UserId, TokenId), U256)>,
+    orders: &[Order],
+) -> AccountState {
+    let account_states = account_states
+        .filter(|((user, token), _)| {
+            orders.iter().any(|order| {
+                order.account_id == *user
+                    && (order.buy_token == *token || order.sell_token == *token)
```
fleupold, line 76: We are only expecting sellTokenBalances, so this check doesn't need to account for buy_tokens.

PR #784, opened 2020-04-25
[Easy] Don't log batches that ran out of time as successful

fleupold:
https://github.com/gnosis/dex-services/pull/656#discussion_r395529456 identified and explained the issue. 

We are currently already seeing a bunch of auctions in which we run out of time fetching the orderbook (cf. [kibana](https://logs.gnosisdev.com/goto/75c991c4167005acff9e14a8b72db20a)).

However this is not at all reflected in our top-level grafana metrics or alerts. Reason is that such batches are still recorded as being successfully processed.

This PR makes it so that we don't log a successful processing event in this case and adds some more docs to make the expectation of each event stage more explicit for future maintainers.

### Test Plan

Once merged, expect to see more "downtime" in the staging Grafana dashboard.

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -138,7 +138,7 @@ impl<'a> StableXDriverImpl<'a> {
 
         if !submitted {
             self.metrics
-                .autcion_processed_but_not_submitted(batch_to_solve);
+                .auction_processed_but_not_submitted(batch_to_solve);
         };
 
         Ok(())
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -137,7 +137,8 @@ impl<'a> StableXDriverImpl<'a> {
         };
 
         if !submitted {
-            self.metrics.autcion_processed_but_not_submitted(batch_to_solve);
+            self.metrics
+                .autcion_processed_but_not_submitted(batch_to_solve);
         };
 
         Ok(())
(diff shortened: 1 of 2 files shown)


driver/src/driver/stablex_driver.rs
```diff
@@ -137,7 +137,8 @@ impl<'a> StableXDriverImpl<'a> {
         };
 
         if !submitted {
-            self.metrics.auction_skipped(batch_to_solve);
+            self.metrics
+                .autcion_processed_but_not_submitted(batch_to_solve);
```
bh2smith, line 141: typo 
```suggestion
                .auction_processed_but_not_submitted(batch_to_solve);
```

driver/src/metrics/stablex_metrics.rs
```diff
@@ -258,12 +258,19 @@ trait InitializeableMetric: 'static + Sized + AsRef<str> {
...
+    /// A solution for the batch has been produced
     Solved,
+    /// The auction solution has been verified
     Verified,
+    /// The auction solution has been submitted
     Submitted,
-    Skipped,
+    /// Processing was successful, but solution was not submitted
+    /// (e.g. solution did not improve objective value)
+    SolutionNotSubmitted,
```
bh2smith, line 273: I see we lost Skipped here and introduced `SolutionNotSubmitted`. Is this also used for the case when orderbook took too long to fetch?

bh2smith (approved):
Looks good! Maybe fix the typo.

driver/src/metrics/stablex_metrics.rs
fleupold: No, I removed the line where we log this event in case the orderbook too long. For those case we now get an orderbook_fetched as a terminal event.

e00E (approved):

Issue #785, opened 2020-04-25
[Streamed Orderdbook] orderbook_for_batch returns orders/state for different batch


fleupold:
Surfaced by the alerts in our slack channel:

> 0x42b7e89374bceab213c699488f667c7a9144c28b token 7 primary balance of 1160400267 but shadow balance 1161400267

This happens when our e2e creates a new order and deposits. The order is picked up by the streamed orderbook, however the deposit is not.

The reason is that the `orderbook_for_batch` method returns orders that are valid for when `batch` is being solved, but it only returns account states that were valid in the batch before:

https://github.com/gnosis/dex-services/blob/4c32b625e840548aeafde4c3742edbbcce60373c/driver/src/orderbook/streamed/state.rs#L52-L71

Specifically, `self.orders` checks that the order is valid for the given batch, when `self. account_state` excludes all credits that have been made in that batch. Thus it effectively returns the balances of the end of `batch-1`.

`OrderbookReading` expects to be passed the batch_id that is currently being solved. At the moment we just forward this id to the streamed orderbook, which also violates the assumption that `self.last_batch_id <= batch_id` since we can already receive an event for the batch that is currently open for orders (batch_id +1), while still querying the batch before that is now being solved. Cf. this alert:

> Apr 25 10:51:11.131 ERRO [driver::driver::scheduler::system] Batch 5292705 failed with retryable error: batch is in the past

If we passed in `batch + 1` instead, we would satisfy the assert and get the correct balance, however we would only get the orders that are valid in the next batch.

PR #786, opened 2020-04-25
[Streamed Orderbook] Fix batchId off-by-one error

fleupold:
This is a ~pretty~ extremely hacky short-term fix for #785 with the goal to keep on getting more consistency results and elminate known issues.

To summarize, `orderbook_for_batch` currently returns the balances that are valid in `batch - 1` and the orders that are valid in `batch`. This PR changes it to also return the orders that are valid in `batch-1`. Then, we need to increment the batchID we get passed in from `OrderbookReading` as this is already the index of the batch we are trying to solve (and for which we want the orderbook).

I also renamed the variable in the OrderbookReading trait to make it more explicit which batch this ID is for.

I'm happy to make/accept a cleaner fix if you guys have an idea.

### Test Plan

1. The first bug was that we would get errors if an event is received in the first 30s of a batch (since this would advance last_batch_id and then result assertion errors that we are querying on a past batch id). I tested locally that this is no longer the case (https://rinkeby.etherscan.io/tx/0x7556fe7b1ddfcb51d1b72cbc67c3060bad9dc5cae3639cd540f74c363e8c4942)
2. When creating an order and a deposit in the same batch the order would get correctly included by the streamed orderbook, but the deposit would be ignored. I tested locally that this is no longer the case (tx from above + deposit in the same batch https://rinkeby.etherscan.io/tx/0x63f94ce7cd43aaf2295da66315fffc2dffd43612f1308ec5dd3725717bf6707c)

### DIFF
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -82,10 +82,15 @@ fn filter_account_state(
 }
 
 impl StableXOrderBookReading for Orderbook {
-    fn get_auction_data(&self, index: U256) -> Result<(AccountState, Vec<Order>)> {
+    fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)> {
         // TODO: Handle future batch ids for when we want to do optimistic solving.
         let state = self.create_state()?;
-        let (account_state, orders) = state.orderbook_for_batch(Batch::Future(index.low_u32()))?;
+        // `orderbook_for_batch` takes the index of the auction that is currently collecting orders and returns
+        // the orderbook for the batch index that is currently being solved. `get_auction_data` passed in the
+        // index for the auction that orders should be valid for (the one currently being solved). Thus we need
+        // to increment it.
+        let (account_state, orders) =
+            state.orderbook_for_batch(Batch::Future(batch_id_to_solve.low_u32() + 1))?;
         let orders = orders.collect::<Vec<_>>();
         let account_state = filter_account_state(account_state, &orders);
         Ok((account_state, orders))
--- a/driver/src/orderbook/streamed/state.rs
+++ b/driver/src/orderbook/streamed/state.rs
@@ -547,7 +547,7 @@ mod tests {
         };
         state = state.apply_event(&Event::OrderPlacement(event), 0).unwrap();
 
-        assert_eq!(state.orders(0).next(), None);
+        assert_eq!(state.orders(1).next(), None);
         let expected_orders = vec![ModelOrder {
             id: 0,
             account_id: address(2),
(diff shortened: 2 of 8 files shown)


driver/src/orderbook/streamed/state.rs
```diff
@@ -68,7 +68,7 @@ impl State {
                 batch_id
             }
         };
-        Ok((self.account_state(batch_id), self.orders(batch_id)))
+        Ok((self.account_state(batch_id), self.orders(batch_id - 1)))
```
fleupold, line 71: Here we make it so that `orderbook_for_batch` returns orders that are valid in the same batch as the balances it returns.

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -82,10 +82,15 @@ fn filter_account_state(
...
+    fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)> {
         // TODO: Handle future batch ids for when we want to do optimistic solving.
         let state = self.create_state()?;
-        let (account_state, orders) = state.orderbook_for_batch(Batch::Future(index.low_u32()))?;
+        // `orderbook_for_batch` takes the index of the auction that is currently collecting orders and returns
+        // the orderbook for the batch index that is currently being solved. `get_auction_data` passed in the
+        // index for the auction that orders should be valid for (the one currently being solved). Thus we need
+        // to increment it.
+        let (account_state, orders) =
+            state.orderbook_for_batch(Batch::Future(batch_id_to_solve.low_u32() + 1))?;
```
fleupold, line 93: Here we make it so that we query the state for the batch that is currently open (which returns data for the batch that is currently being solved)

fleupold (commented):
Highlighted the relevant lines. The rest is the renaming.

e00E (approved):

driver/src/orderbook/streamed/orderbook.rs
nlordell: I have a question, should this be 
```suggestion
            state.orderbook_for_batch(Batch::Future(batch_id_to_solve.low_u32()))?;
```

And then the change to `driver/src/orderbook/streamed/state.rs` be
```rust
Ok((self.account_state(batch_id + 1), self.orders(batch_id)))
```

From what I understand, the order needs to be valid in the previous batch (i.e. the batch to solve, i.e. the current batch - 1), and the balances at the end of the batch to solve (i.e. beginning of the current batch).

fleupold: This would be possible too, however there is some assertion logic in the orderbook that errors if we request a batch from the past. 
And since the batch we are passing here is the one that is currently being solved, the event based orderbook might have already seen an event that is from the batch currently collecting orders (index + 1) making this batch here a batch from the past and thus throwing.

driver/src/orderbook/streamed/state.rs
e00E: I see a problem now:

```
batch 10 (batch to solve 9):
there is a deposit, deposit batch is set to 10
there is a trade of an order, order has pending used amount set to batch 10.
```
We want orders that have to change `fn orders` so that it uses one batch_id for `order_is_valid_in_batch` and that batch_id + 1 for `as_model_order`.

bh2smith: So we subtract one from the batch ID here only to add it again in the internal function?

e00E (approved):

bh2smith (approved):

PR #787, opened 2020-04-25
Implement BatchTransport via Http

fleupold:
This PR implements the BatchTransport trait for our own HttpTransport implementation. Instead of sending a single JSON object containing the method, parms, id, etc, we send a list of multiple objects (potentially for different methods).

In the success case the result is a list of response objects one for each request.

### Test Plan

Using the new transport for fetching multiple block timestamps in a single request in the next PR.

### DIFF
--- a/driver/src/metrics/http_metrics.rs
+++ b/driver/src/metrics/http_metrics.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use ethcontract::jsonrpc::types::{Call, Request};
 use prometheus::{HistogramOpts, HistogramVec, Registry, DEFAULT_BUCKETS};
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


driver/src/transport.rs
```diff
@@ -109,6 +123,27 @@ impl Transport for HttpTransport {
     }
 
     fn send(&self, id: RequestId, request: Call) -> Self::Out {
-        self.0.clone().execute_rpc(id, request).boxed().compat()
+        self.0
+            .clone()
+            .execute_rpc(id, Request::Single(request))
```
fleupold, line 128: @nlordell can you remind me again why we are implementing our own transport instead of wrapping the web3-rs HTTP transport? Was it to have access to the raw response? If so, what about their error wrapping was insufficient for our purposes?

e00E (approved):

driver/src/transport.rs
nlordell: https://github.com/gnosis/dex-services/pull/543 , The main reason was to implement timeouts. Being able to log the raw response was a bonus as the error message was always sufficient to figure out the problem. For example if it failed to parse the response (for example when Ganache returns both a result and error), the error was just `"data did not match any variant of untagged enum Response"` and since the error is `.to_string()`-ed in the `web3::Error`, things like parse error position also get lost.

I think wrapping the web3 transport was presented as an option, but we decided to write our own transport as it was fairly simple and it wasn't clear if implementing timeouts correctly was as easy as it would require dealing with timers and dropping the HTTP transport future didn't actually cancel the HTTP request.

driver/src/metrics/http_metrics.rs
```diff
@@ -104,8 +105,24 @@ labels! {
...
         EthEstimateGas => "eth_estimate_gas",
         EthRpc => "eth_rpc",
+        EthBatchRPC => "eth_batch_rpc",
         Kraken => "kraken",
         Dexag => "dexag",
         GasStation => "gas_station",
     }
 }
+
+impl From<&Request> for HttpLabel {
```
nlordell, line 115: I like this!

nlordell (approved):

Issue #760, opened 2020-04-22
[event based orderbook] Avoid fetching block timestamp for past events


e00E:
The only way I see would be to handle past events slightly differently than future events, so for past events you apply all the events under the current batch ID (queried from the smart contract for example) and then for incoming events you can query the block timestamp for each (since they will be way more sparse).

Definitely a separate issue/PR though.

_Originally posted by @nlordell in https://github.com/gnosis/dex-services/pull/756#discussion_r413125454

e00E:
The way described above doesn't work with the level of assumption verification we currently have where we check in some places that the batch id makes sense.

fleupold:
The batchID is only really needed for Trades/Solutions/Reversions (which always come in the same block) and OrderCancellations. 

While we should probably add the batchId there in a future version of the smart contract, for now I think we would already do much better if we only queried the timestamps for this subset of events.

e00E:
Idea:
We make the batch id in State optional. If it is there we verify all assumptions. If it is not there we do not verify any assumption related to batch id. When we a get a `SolutionSubmission` without a previous reverted trade we know that the previous solution got accepted and that the batch id has incremented by at least one which in turn allows us to apply pending deposits, proceeds, order used amounts.

fleupold:
I like the idea!

Could we even directly apply all events that don't have a timestamp? We might have to keep the fee's stored somewhere but otherwise it would make sense to me to just directly add and subtract balances/used amounts since we would be guaranteed to not query state within one batch from any event without a timestamp.

However, the idea of simply omitting the assertions could be less intrusive and easier to implement. In that case, I'm all for it.

fleupold:
Since we will need historic timestamp for "what if" analysis and historic price estimations, we want to keep the timestamps of all events. Thus this task is no longer relevant.

PR #788, opened 2020-04-26
[Streamed Orderbook] Read block timestamps in batches

fleupold:
This PR improves the long cold start time when using the event based orderbook, by batching the requests for block timestamps. 
With the current business logic, we need to get the timestamp for each block that emitted a relevant event. On rinkeby alone this results in a few thousand requests that we send to the FullNode. By batching these calls into a few single requests we can significantly decrease roundtrip time and network overhead.

On my local machine (4G speed) I can now sync the event based orderbook on rinkeby in less than a minute (before it would take more than 20 miuntes) - **a speedup of 20x.**

Future changes to the business logic might make it so we don't have to fetch historic timestamps anymore (cf. discussion in #760), however even for our incremental fetching it can be beneficial to batch the timestamp requests.

My approach is to change the `MemoizingBlockTimestampReader` to contain a full map instead of just the last block's timestamp and furthermore make it "pre-heatable" by querying a bunch of block timestamps in advance. This way, it will instantly return the queried block hashes from memory (at least for now for the past events).

### Test Plan

run it using:

`cargo run -- --network-id=4 --node-url=https://blue-parity.rinkeby.gnosisdev.com --use-shadowed-orderbook true --primary_orderbook EventBased`

and see that we have "Finished applying past events" much earlier than before.

### DIFF
--- a/driver/src/orderbook/streamed/block_timestamp_reading.rs
+++ b/driver/src/orderbook/streamed/block_timestamp_reading.rs
@@ -1,11 +1,25 @@
 use crate::contracts::Web3;
+use crate::transport::HttpTransport;
 use anyhow::{Context as _, Result};
-use ethcontract::{web3::types::BlockId, H256};
+use ethcontract::web3::transports::Batch;
+use ethcontract::{
+    web3::types::{Block, BlockId},
+    H256,
+};
 use futures::{compat::Future01CompatExt as _, future::BoxFuture, FutureExt as _};
+use std::collections::{HashMap, HashSet};
+use std::iter::FromIterator;
 
 /// Helper trait to make this functionality mockable for tests.
 pub trait BlockTimestampReading {
     fn block_timestamp(&mut self, block_hash: H256) -> BoxFuture<Result<u64>>;
+}
+
+pub trait BlockTimestampBatchReading {
+    fn block_timestamps(
+        &mut self,
+        block_hashes: HashSet<H256>,
+    ) -> BoxFuture<Result<Vec<(H256, u64)>>>;
 }
 
 /// During normal operation this is implemented by Web3.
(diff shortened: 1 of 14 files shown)


driver/src/orderbook/streamed/block_timestamp_reading.rs
```diff
@@ -22,34 +36,83 @@ impl BlockTimestampReading for Web3 {
...
+            cache: HashMap::new(),
+        }
+    }
+
+    pub fn prepare_cache(&mut self, block_hashes: HashSet<H256>) -> BoxFuture<Result<()>> {
+        async move {
+            self.cache
+                .extend(self.inner.block_timestamps(block_hashes).await?);
+            Ok(())
         }
```
e00E, line 101: Here we are querying all timestamps from the node even those that are already in the cache. Doesn't matter in practice because it's only called for past_events where we don't have any timestamps.

e00E (approved):

nlordell (approved):
:guitar: :metal:

driver/src/orderbook/streamed/block_timestamp_reading.rs
fleupold: Very good point. Will make it so that it filters out the ones it already has.

driver/src/orderbook/streamed/block_timestamp_reading.rs
```diff
@@ -22,34 +36,83 @@ impl BlockTimestampReading for Web3 {
...
+                    .expect("blocks queried by hash should contain a hash"),
+                block.timestamp.low_u64(),
+            ))
+        })
+        .collect()
+}
+
+/// A cache for the block timestamp.
 #[derive(Debug)]
 pub struct MemoizingBlockTimestampReader<T> {
```
nlordell, line 83: naming nit: This should probably be called `CachedBlockTimestampReader` (I reviewed the original PR before the name change) as I would describe what this is doing as caching more than memoizing.

PR #789, opened 2020-04-26
[Easy] Log error chain or streamed orderbook exit

fleupold:
Three small changes to the way we exit the updating orderbook in case of an error:

1. Print the entire error stack trace not just the top level (by using the Debug print of anyhow)
2. Sleep for 1s before exiting to allow the logs to be flushed (I tried synchronously flushing them but that did not work)
3. Use process::exit instead of abort, as suggested by @nlordell (not feeling strongly about this)

### Test Plan

When working with the streamed orderbook and runnig into an error, make sure we can read the error reason and error chain.

### DIFF
--- a/driver/src/orderbook/streamed/updating_orderbook.rs
+++ b/driver/src/orderbook/streamed/updating_orderbook.rs
@@ -19,6 +19,7 @@ use std::sync::{
     atomic::{AtomicBool, Ordering},
     Arc, Mutex,
 };
+use std::{process, thread, time::Duration};
 
 /// An event based orderbook that automatically updates itself with new events from the contract.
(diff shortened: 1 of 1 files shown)


e00E (approved):

nlordell (approved):

nlordell:
> Use process::exit instead of abort

I think ultimately it doesn't make much of a difference. I'm just not used to seeing Rust code `abort` (I think even `panic!` doesn't cause an abort but an `exit(101)` under the hood). To the point that when you posted the abort error I immediately assumed it was an issue related to one of our C (transient) dependencies or some weird memory thing from compiler generated stack guards.

PR #795, opened 2020-04-27
Remove the background thread in event based orderbook

e00E:
Instead we do the initial update in `new` and future updates whenever
the orderbook is requested.
This makes the logic simpler and shorter because no threading is involved.

Test Plan:
I tested that it still creates the orderbook with the expected number of events on rinkeby and that getting the orderbook updates and doesn't fail.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -277,6 +277,11 @@ fn main() {
     let mut scheduler = options
         .scheduler
         .create(&*contract, &driver, scheduler_config);
+    // Retrieve the orderbook once because the event based orderbook needs to perform an expensive
+    // initialization step on first call.
+    filtered_orderbook
+        .get_auction_data(ethcontract::U256::zero())
+        .expect("orderbook initialization failed");
     scheduler.start();
 }
 
--- a/driver/src/orderbook/mod.rs
+++ b/driver/src/orderbook/mod.rs
@@ -62,7 +62,9 @@ impl OrderbookReaderKind {
                 auction_data_page_size,
                 orderbook_filter,
             )),
-            OrderbookReaderKind::EventBased => Box::new(EventBasedOrderbook::new(contract, web3)),
+            OrderbookReaderKind::EventBased => {
+                Box::new(EventBasedOrderbook::new(contract, web3).unwrap())
+            }
         }
     }
 }
--- a/driver/src/orderbook/mod.rs
+++ b/driver/src/orderbook/mod.rs
@@ -62,9 +62,7 @@ impl OrderbookReaderKind {
                 auction_data_page_size,
                 orderbook_filter,
             )),
-            OrderbookReaderKind::EventBased => {
-                Box::new(EventBasedOrderbook::new(contract, web3).unwrap())
-            }
+            OrderbookReaderKind::EventBased => Box::new(EventBasedOrderbook::new(contract, web3)),
         }
     }
 }
--- a/driver/src/orderbook/streamed/updating_orderbook.rs
+++ b/driver/src/orderbook/streamed/updating_orderbook.rs
@@ -35,8 +35,8 @@ struct Context {
 impl UpdatingOrderbook {
     /// Does not block on initializing the orderbook. This will happen in the first call to
     /// `get_auction_data` which can thus take a long time to complete.
-    pub fn new(contract: Arc<dyn StableXContract + Send + Sync>, web3: Web3) -> Result<Self> {
-        let result = Self {
+    pub fn new(contract: Arc<dyn StableXContract + Send + Sync>, web3: Web3) -> Self {
+        Self {
             contract,
             web3: web3.clone(),
             context: Mutex::new(Context {
@@ -44,8 +44,7 @@ impl UpdatingOrderbook {
                 last_handled_block: 0,
                 block_timestamp_reader: CachedBlockTimestampReader::new(web3),
             }),
-        };
-        Ok(result)
+        }
     }
 
     async fn update_with_events(&self, context: &mut Context) -> Result<()> {
--- a/driver/src/orderbook/streamed/updating_orderbook.rs
+++ b/driver/src/orderbook/streamed/updating_orderbook.rs
@@ -22,10 +22,10 @@ pub struct UpdatingOrderbook {
     /// We need a mutex because otherwise the struct wouldn't be Sync which is needed because we use
     /// the orderbook in multiple threads. The mutex is locked in `get_auction_data` while the
     /// orderbook is updated with new events.
-    behind_mutex: Mutex<BehindMutex>,
+    context: Mutex<Context>,
 }
 
-struct BehindMutex {
+struct Context {
     orderbook: Orderbook,
     last_handled_block: u64,
     block_timestamp_reader: CachedBlockTimestampReader<Web3>,
@@ -38,7 +38,7 @@ impl UpdatingOrderbook {
         let result = Self {
             contract,
             web3: web3.clone(),
-            behind_mutex: Mutex::new(BehindMutex {
+            context: Mutex::new(Context {
                 orderbook: Orderbook::default(),
                 last_handled_block: 0,
                 block_timestamp_reader: CachedBlockTimestampReader::new(web3),
@@ -47,21 +47,21 @@ impl UpdatingOrderbook {
 
         // Perform the initial update.
         {
-            let mut behind_mutex = result
-                .behind_mutex
+            let mut context = result
+                .context
                 .lock()
                 .map_err(|err| anyhow!("poison error: {}", err))?;
-            futures::executor::block_on(result.update_with_events(&mut behind_mutex))?;
+            futures::executor::block_on(result.update_with_events(&mut context))?;
         }
 
         Ok(result)
     }
 
-    async fn update_with_events(&self, behind_mutex: &mut BehindMutex) -> Result<()> {
+    async fn update_with_events(&self, context: &mut Context) -> Result<()> {
         const BLOCK_RANGE: u64 = 25;
         log::info!("Starting event based orderbook updating.");
         let current_block = self.web3.eth().block_number().compat().await?;
-        let from_block = behind_mutex.last_handled_block.saturating_sub(BLOCK_RANGE);
+        let from_block = context.last_handled_block.saturating_sub(BLOCK_RANGE);
         let to_block = BlockNumber::Number(current_block);
         log::info!(
             "The range is from block {} to block {}",
@@ -72,14 +72,14 @@ impl UpdatingOrderbook {
             .contract
             .past_events(BlockNumber::Number(from_block.into()), to_block)
             .await?;
-        self.handle_events(behind_mutex, events, from_block).await?;
-        behind_mutex.last_handled_block = current_block.as_u64();
+        self.handle_events(context, events, from_block).await?;
+        context.last_handled_block = current_block.as_u64();
         Ok(())
     }
 
     async fn handle_events(
         &self,
-        behind_mutex: &mut BehindMutex,
+        context: &mut Context,
         events: Vec<Event<batch_exchange::Event>>,
         delete_events_starting_at_block: u64,
     ) -> Result<()> {
@@ -94,15 +94,15 @@ impl UpdatingOrderbook {
                 Ok(metadata.block_hash)
             })
             .collect::<Result<HashSet<H256>>>()?;
-        behind_mutex
+        context
             .block_timestamp_reader
             .prepare_cache(block_hashes)
             .await?;
-        behind_mutex
+        context
             .orderbook
             .delete_events_starting_at_block(delete_events_starting_at_block);
         for event in events {
-            self.handle_event(behind_mutex, event).await?;
+            self.handle_event(context, event).await?;
         }
         log::info!("Finished applying events");
         Ok(())
@@ -111,7 +111,7 @@ impl UpdatingOrderbook {
     /// Apply a single event to the orderbook.
     async fn handle_event(
         &self,
-        behind_mutex: &mut BehindMutex,
+        context: &mut Context,
         event: Event<batch_exchange::Event>,
     ) -> Result<()> {
         match event {
@@ -119,11 +119,11 @@ impl UpdatingOrderbook {
                 data,
                 meta: Some(meta),
             } => {
-                let block_timestamp = behind_mutex
+                let block_timestamp = context
                     .block_timestamp_reader
                     .block_timestamp(meta.block_hash)
                     .await?;
-                behind_mutex.orderbook.handle_event_data(
+                context.orderbook.handle_event_data(
                     data,
                     meta.block_number,
                     meta.log_index,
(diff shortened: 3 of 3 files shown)


e00E:
This change rewrites the whole file so I suggest reading the new file instead of the diff.

driver/src/orderbook/mod.rs
```diff
@@ -62,7 +62,9 @@ impl OrderbookReaderKind {
                 auction_data_page_size,
                 orderbook_filter,
             )),
-            OrderbookReaderKind::EventBased => Box::new(EventBasedOrderbook::new(contract, web3)),
+            OrderbookReaderKind::EventBased => {
+                Box::new(EventBasedOrderbook::new(contract, web3).unwrap())
```
bh2smith, line 66: I suppose this was a mistake from the original PR.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,142 @@ use crate::{
...
 pub struct UpdatingOrderbook {
-    orderbook: Arc<Mutex<Orderbook>>,
-    // Indicates whether the background thread has caught up with past events at which point the
-    // orderbook is ready to be read.
-    orderbook_ready: Arc<AtomicBool>,
-    // When this struct is dropped this sender will be dropped which makes the updater thread stop.
-    _exit_tx: oneshot::Sender<()>,
+    contract: Arc<dyn StableXContract + Send + Sync>,
+    web3: Web3,
+    behind_mutex: Mutex<BehindMutex>,
```
bh2smith, line 22: I think we may have lost some commentary here.

bh2smith (commented):
This is a pretty big change.

driver/src/orderbook/mod.rs
e00E: No mistake in the original PR. The function signature of `new` has changed so that it returns a Result.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: Do you mean you would like there to be a comment explaining why this is needed? Sure, I'll add it.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,144 @@ use crate::{
...
 }
 
-impl UpdatingOrderbook {
-    pub fn new(contract: Arc<dyn StableXContract + Send + Sync>, web3: Web3) -> Self {
-        let orderbook = Arc::new(Mutex::new(Orderbook::default()));
-        let orderbook_clone = orderbook.clone();
-        let orderbook_ready = Arc::new(AtomicBool::new(false));
-        let orderbook_ready_clone = orderbook_ready.clone();
-        let (exit_tx, exit_rx) = oneshot::channel();
+struct BehindMutex {
```
fleupold, line 28: naming nit: Is this kind of naming standard? I personally find it a bit weird. Can we call it UpdatingOrderbookInner or something a like?

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,144 @@ use crate::{
...
+}
+
+impl StableXOrderBookReading for UpdatingOrderbook {
+    /// Blocks on updating the orderbook.
+    fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)> {
+        let mut behind_mutex = self
+            .behind_mutex
+            .lock()
+            .map_err(|err| anyhow!("poison error: {}", err))?;
+        futures::executor::block_on(self.update_with_events(&mut behind_mutex))?;
```
fleupold, line 147: Would it make sense to make `update_with_events` a synchronous method instead? Would we then still need the mutex?

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,144 @@ use crate::{
...
-            orderbook,
-            orderbook_ready,
-            _exit_tx: exit_tx,
+        // Perform the initial update.
+        {
+            let mut behind_mutex = result
+                .behind_mutex
+                .lock()
+                .map_err(|err| anyhow!("poison error: {}", err))?;
+            futures::executor::block_on(result.update_with_events(&mut behind_mutex))?;
```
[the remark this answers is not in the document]
e00E: This is part of the initialization for the struct so I find it sensible.
I do like the idea of having the constructor be an async function instead of forcing the blocking. Then we can decide elsewhere how to handle the Future. Let's do that in a separate PR.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: We always need the mutex because all orderbooks must be `Sync`. Instead of using the mutex here we could wrap the whole thing with a Mutex which would make this implementation slightly simpler.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I don't know what is standard for this. Imo BehindMutex is more meaningful than ...Inner but weak opinion.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,144 @@ use crate::{
...
-    let exit_indicator = exit_indicator.fuse();
-    pin_mut!(exit_indicator);
-
-    let mut last_handled_block = 0u64;
-    loop {
-        let current_block = web3.eth().block_number().compat().await?;
-        let from_block = last_handled_block.saturating_sub(BLOCK_RANGE);
+    async fn update_with_events(&self, behind_mutex: &mut BehindMutex) -> Result<()> {
+        const BLOCK_RANGE: u64 = 25;
+        log::info!("Starting event based orderbook updating.");
```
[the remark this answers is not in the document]
e00E: block range is a couple of lines down. I decided this should be separate prints in case `web3.eth.block_number` takes a long time.

driver/src/orderbook/streamed/updating_orderbook.rs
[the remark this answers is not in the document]
fleupold: I'm worried that in the current approach we will be stuck in the beginning of our main function for a minute or two while we are fetching the initial events. I'd really prefer if this happened already in the scheduler thread after all setup is done (otherwise we might panic after two minutes because some env variable is set incorrectly), rather than in main.

driver/src/orderbook/streamed/updating_orderbook.rs
fleupold: I might not understand the full implications of this, but my suggestion would be to synchronously wait on the "potentially long running" futures using our [util extension](https://github.com/gnosis/dex-services/blob/affcdbb27d20f1b64c0d01dc2d15af235311ec9a/driver/src/util.rs#L44-L46).

I see how `UpdatingOrderbook` itself must be Sync, but I don't understand why we need a Mutex for this (couldn't we use an arc option or the like)? Mutexes are a pretty heavy tool (opening the potential for deadlocks) which I'd like to avoid if not really necessary.

e00E: There *has* to be a mutex because `get_auction_data` can be called from multiple threads. Arc only guarantees that the data inside lives long enough but not that accessing it is safe.
The worry about deadlocks is misplaced because we can see that the mutex is only locked in one place.

driver/src/orderbook/streamed/updating_orderbook.rs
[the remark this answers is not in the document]
e00E: Looking at `main.rs` I see that we are already creating the price finder before the orderbook so it isn't blocked and the command line options have already been parsed. The only thing I see that can still fail is creating the AuctionTimingConfiguration which we can move in front of the orderbook.

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: I have usually seen it as `...Inner` or `...Context`.

driver/src/orderbook/streamed/updating_orderbook.rs
fleupold: I see now how we need the Mutex here in order to have interior mutability and Sync at the same time.

> The worry about deadlocks is misplaced because we can see that the mutex is only locked in one place.

Theoretically we could have a re-entrancy that causes the deadlock. In general, reasoning about concurrent code is extremely hard and commonly underestimated (it's easy for the initial author, but can easily evolve into a nightmare). That's why, if possible I'd like to avoid it.

driver/src/orderbook/streamed/updating_orderbook.rs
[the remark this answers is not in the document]
fleupold: I think generally we shouldn't have to think too much about the order in which we instantiate our components (other then explicit dependency satisfaction) and have to worry that some new functions might run for 2 minutes. 
I'd be fine with returning a promise and wait for it at the right time, but for now it seems the simplest to me to simply `get_auction_data()`in the scheduler to warm things up (this would also make it that you only access the inner Context in one place).

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: Correct me if I'm wrong, but a `Mutex` is also needed for interior mutability. Since we are sharing the orderbook across threads which potentially mutate (by adding new events for example) the orderbook, a `Mutex` is indeed needed. Just an `Arc` would allow you to access a reference to the underlying data but not a mutable reference which we need here.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,144 @@ use crate::{
...
-            orderbook,
-            orderbook_ready,
-            _exit_tx: exit_tx,
+        // Perform the initial update.
+        {
+            let mut context = result
+                .context
+                .lock()
+                .map_err(|err| anyhow!("poison error: {}", err))?;
+            futures::executor::block_on(result.update_with_events(&mut context))?;
```
fleupold, line 54: nit: can we use https://github.com/gnosis/dex-services/blob/affcdbb27d20f1b64c0d01dc2d15af235311ec9a/driver/src/util.rs#L44-L46

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,144 @@ use crate::{
...
+}
+
+impl StableXOrderBookReading for UpdatingOrderbook {
+    /// Blocks on updating the orderbook.
+    fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)> {
+        let mut context = self
+            .context
+            .lock()
+            .map_err(|err| anyhow!("poison error: {}", err))?;
+        futures::executor::block_on(self.update_with_events(&mut context))?;
```
fleupold, line 147: ditto

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -7,177 +7,144 @@ use crate::{
...
-    let exit_indicator = exit_indicator.fuse();
-    pin_mut!(exit_indicator);
-
-    let mut last_handled_block = 0u64;
-    loop {
-        let current_block = web3.eth().block_number().compat().await?;
-        let from_block = last_handled_block.saturating_sub(BLOCK_RANGE);
+    async fn update_with_events(&self, context: &mut Context) -> Result<()> {
+        const BLOCK_RANGE: u64 = 25;
+        log::info!("Updating event based orderbook.");
```
fleupold, line 62: uber nit: If you think it adds value to know we are querying the current block we can leave this (however we can also see this from the HTTP transport logs. Otherwise I'd say let's keep our logging to a few high signal lines, but really not important.

fleupold (approved):
Looks good, the only thing that is still bothering me is that we are freezing on startup for ~3 minutes because of doing long running work in new. I'd really prefer if this could happen lazily in the scheduler thread.

driver/src/main.rs
```diff
@@ -277,6 +277,11 @@ fn main() {
     let mut scheduler = options
         .scheduler
         .create(&*contract, &driver, scheduler_config);
+    // Retrieve the orderbook once because the event based orderbook needs to perform an expensive
+    // initialization step on first call.
+    filtered_orderbook
+        .get_auction_data(ethcontract::U256::zero())
+        .expect("orderbook initialization failed");
```
e00E, line 284: @fleupold What do you think of this? I don't see what advantage we get from doing this in the scheduler. It doesn't feel like there is a right place there because we have multiple schedulers.

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: > Theoretically we could have a re-entrancy that causes the deadlock. In general, reasoning about concurrent code is extremely hard 

This is a very true statement. Concurrent code is always hard especially when you didn't write it.

One solution to this is to have the orderbook updating running on its own thread and use message passing to perform updates and retrieve the current orderbook. That way all orderbook updates happen in one thread and "get orderbook" messages are sent over a channel to that thread.

nlordell:
> the only thing that is still bothering me is that we are freezing on startup for ~3 minutes because of doing long running work in new. I'd really prefer if this could happen lazily in the scheduler thread.

I thought about this a little and I didn't completely understand the value of doing it in the scheduler lazily.

driver/src/main.rs
fleupold: 👍 Yes, that works. It's pretty explicit that it can be a long running operation. I like that `::new` no longer returns a result.

The only slight drawback is that we can't start any of the other background operations before (like price fetching), but since they are fast it shouldn't be an issue.

nlordell: I think the `PriceOracle` is already started at this point and running in a background thread.

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: That being said, I am happy with the Mutex use here, it is fairly straightforward and its use isn't spread over multiple files.

fleupold:
> I thought about this a little and I didn't completely understand the value of doing it in the scheduler lazily since we have to wait for it to be done before we can do anything anyway.

Maybe my worries are a bit old-fashioned, but to me it's a smell to have a constructor that unexpectedly runs a long-blocking operation (especially in the bootstrapping phase of the service and given it's an optional optimization).
Most of our `unwraps` are when parsing the config, so that if we crash, we crash immediately. It would be annoying to crash only after 3 minutes because of a typo in `EvmShceduler`. In that case we would have also queried all this data for no reason.

But it's not a very strong argument and most related articles I found seem pretty old and focus around testability (e.g. http://misko.hevery.com/code-reviewers-guide/flaw-constructor-does-real-work/).

bh2smith (approved):

driver/src/orderbook/streamed/updating_orderbook.rs
fleupold: Agreed, I didn't immediately see the need for interior mutability and given this it seems like there is no real simple better option (I'd like to avoid more scattered threading).

PR #796, opened 2020-04-27
[Easy] Filter out orders without any remaining sell amount

nlordell:
This PR filters out orders that have no remaining sell amount in the streamed orderbook.

In the future we can also filter out orders whose sell amount is less than the minimum sell amount as well as orders that have no balance. I'm leaving it as a separate PR to get the quick fix in as that would require some unit test adjustments.

### Test Plan

Adjusted unit test to ensure empty order gets ignored.

### DIFF
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -65,11 +65,16 @@ impl Orderbook {
     }
 }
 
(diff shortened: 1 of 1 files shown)


fleupold (approved):
🚢 

Nice that there was even a unit test for it.

e00E (approved):

PR #805, opened 2020-04-28
Read/Write Orderbook to File

bh2smith:
For improved performance on driver restart, it would be nice if the Orderbook could be recovered from file rather than restreaming the entire ethereum event logs. In order to achieve this, we first require a a way to read and write orderbook to disk.

This PR implements exactly this, 

- `Orderbook::write_to_file(path)`
- `impl TryFrom<&Path> for Orderbook`
- `impl TryFrom<File> for Orderbook`
- `impl TryFrom<&[u8]> for Orderbook`

Along with a full cycle unit test!

### DIFF
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 use anyhow::Result;
 use ethcontract::{contract::EventData, H256, U256};
+use log::info;
 use ron;
 use serde::{Deserialize, Serialize};
 use state::{Batch, State};
@@ -101,7 +102,7 @@ impl Orderbook {
 }
 
 impl From<&[u8]> for Orderbook {
-    fn from(bytes: &[u8]) -> Orderbook {
+    fn from(bytes: &[u8]) -> Self {
         match ron::de::from_bytes(bytes) {
             Ok(x) => x,
             Err(e) => {
@@ -115,7 +116,7 @@ impl From<File> for Orderbook {
     fn from(mut file: File) -> Self {
         let mut contents = String::new();
         match file.read_to_string(&mut contents) {
-            Ok(value) => println!("Successfully read {} bytes", value),
+            Ok(value) => info!("Successfully loaded {} bytes from file", value),
             Err(error) => panic!("Failed to read file: {}", error),
         }
         Orderbook::from(contents.as_bytes())
@@ -208,7 +209,7 @@ mod tests {
     }
 
     #[test]
-    fn test_io_cycle() {
+    fn test_write_read_recover_full_cycle() {
         let event_key = EventSortKey {
             block_number: 0,
             block_hash: H256::zero(),
@@ -226,10 +227,13 @@ mod tests {
         events.insert(event_key, value);
         let initial_orderbook = super::Orderbook { events };
 
-        let test_path = Path::new("./my_test_orderbook.ron");
+        let test_path = Path::new("/tmp/my_test_orderbook.ron");
         initial_orderbook.write_to_file(test_path).unwrap();
 
         let recovered_orderbook = super::Orderbook::from(test_path);
         assert_eq!(initial_orderbook.events, recovered_orderbook.events);
+
+        // Cleanup the file created here.
+        assert!(fs::remove_file(test_path).is_ok());
     }
 }
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -12,7 +12,7 @@ use state::{Batch, State};
 use std::collections::BTreeMap;
 use std::fs;
 use std::fs::File;
-use std::io::Write;
+use std::io::{Read, Write};
 use std::path::Path;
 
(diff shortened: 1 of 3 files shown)


driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -145,4 +207,33 @@ mod tests {
         assert_eq!(orders.len(), 1);
         assert_eq!(orders[0].account_id, Address::zero());
     }
+
+    #[test]
+    fn test_write_read_recover_full_cycle() {
```
fleupold, line 211: Usually, we shouldn't be doing real I/O in a unit tests (we could argue this is a simple integration test, which maybe should be ignored by default like the KrakenAPI tests). 

Is there a simple way to mock the file system in memory in rust?

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -73,6 +99,39 @@ impl Orderbook {
...
+            .with_context(|| format!("Failed to read file: {:?}", file))?;
+        info!(
+            "Successfully loaded {} bytes from Orderbook file",
+            bytes_read
+        );
+        Orderbook::try_from(contents.as_bytes())
+    }
+}
+
+impl TryFrom<&Path> for Orderbook {
```
fleupold, line 125: This is really cool

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -64,6 +74,22 @@ impl Orderbook {
         });
     }
 
+    pub fn write_to_file(&self, path: impl AsRef<Path>) -> Result<()> {
+        // Write to tmp file until complete and then rename.
+        let mut temp_path = env::temp_dir();
+        temp_path.push("temp_orderbook.ron");
```
fleupold, line 79: ubernit: maybe make this name a bit more expressive (stablex_orderbook.temp)

fleupold (approved):
Looks great to me. Very nice usage of rust trades. One comment about the test, maybe there is an easy way to mock the file system so that can be a real unit test.

driver/src/orderbook/streamed/orderbook.rs
nlordell: Not that I'm aware :(

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -1,20 +1,29 @@
...
+use std::path::Path;
+use std::{env, fs};
 
 // Ethereum events (logs) can be both created and removed. Removals happen if the chain reorganizes
 // and ends up not including block that was previously thought to be part of the chain.
 // However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
 // list of all events based on which the state is built.
 
-#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+#[derive(Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
```
nlordell, line 26: Alphabetical order... You are a champ!

driver/src/orderbook/streamed/orderbook.rs
nlordell: Just a suggestion but there could be a second test that just serializes the orderbook to and from a `String` or `Vec<u8>` so that the file system isn't needed for the unit test, and then as @fleupold suggested keep this as a mini-integration test (like the Kraken API test) that is `#[ignore]`-d by default and can be manually run.

nlordell (approved):

driver/src/orderbook/streamed/orderbook.rs
bh2smith: Ya cool, I can ignore this test and make another one. Just wondering... if we `#[ignore]` the test... then what is it for or when is it executed?

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -64,6 +73,22 @@ impl Orderbook {
         });
     }
 
+    pub fn write_to_file(&self, path: impl AsRef<Path>) -> Result<()> {
+        // Write to tmp file until complete and then rename.
+        let mut temp_path = env::temp_dir();
```
e00E, line 78: Instead of using `temp_dir` we should use the same path but with something like `.temp` appended. The problem with `temp_dir` is that we don't know where this directory is and renaming a file will fail if the two paths are not on the same filesystem.

e00E (approved):

driver/src/orderbook/streamed/orderbook.rs
bh2smith: Great. Yes, I will do this. Although I tried this at first and got stuck not knowing how to concatenate strings... hopefully I can extend path or something.

e00E: https://doc.rust-lang.org/std/path/struct.Path.html#method.with_extension could work.

bh2smith: Managed to get the job done just like this:

```
let temp_path = path.as_ref().with_extension(".temp");
```

bh2smith:
All suggestions considered and repaired. Thanks for the great review!

Issue #806, opened 2020-04-28
Make all impls of OrderbookReading Send


fleupold:
The `ShadowedOrderbookReader` already requires the shadow orderbook to be `Send` so it can be used in a background thread.

This limits its use currently to only using a few implementations since for instance the FilteredOrderbookReader is not yet Send.

I'd suggest we make the Reader-chain take ownership of their inner readers (passing it via ARC instead of &) and make it so they also implement Send.

This will also avoid having to hold on to intermediate orderbooks in the main function like 

https://github.com/gnosis/dex-services/blob/efd6bbee7b0f8c42384d4e845993745df6672a48/driver/src/main.rs#L247-L249

PR #807, opened 2020-04-28
[Easy] Filter primary orderbook before passing it into shadow

fleupold:
This will allow us shadowing fallback solver traffic as well. There, we use a token whitelist and would thus get a lot of consistency errors since the event based order book doesn't enforce the whitelist internally whereas the onchain filter does (via the smart contract).

### Test Plan

Run locally on rinkeby with a filter. Make sure orderbooks are consistent.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -240,25 +240,24 @@ fn main() {
         web3,
     );
 
-    // NOTE: Keep the shadowed orderbook around so it doesn't get dropped and we
-    //   can pass a reference to the filtered orderbook reader.
-    let unfiltered_orderbook: Box<dyn StableXOrderBookReading + Sync> =
-        if options.use_shadowed_orderbook {
-            let shadow_orderbook = OnchainFilteredOrderBookReader::new(
-                contract.clone(),
-                options.auction_data_page_size,
-                &options.orderbook_filter,
-            );
-            let shadowed_orderbook =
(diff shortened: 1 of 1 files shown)


e00E (approved):

driver/src/main.rs
```diff
@@ -240,33 +240,32 @@ fn main() {
...
-
     info!("Orderbook filter: {:?}", options.orderbook_filter);
     let filtered_orderbook =
-        FilteredOrderbookReader::new(&*unfiltered_orderbook, options.orderbook_filter);
+        FilteredOrderbookReader::new(&*primary_orderbook, options.orderbook_filter.clone());
+
+    // NOTE: Keep the shadowed orderbook around so it doesn't get dropped and we
+    //   can pass a reference to the filtered orderbook reader.
+    let orderbook: Box<dyn StableXOrderBookReading + Sync> = if options.use_shadowed_orderbook {
+        let shadow_orderbook = OnchainFilteredOrderBookReader::new(
```
nlordell, line 250: If we start filtering accounts as well, we might get some more inconsistencies. Maybe we should additionally wrap the `shadow_orderbook` in a `FilteredOrderbookReader` as well.

nlordell (approved):

driver/src/main.rs
fleupold: Agree however this change would have been a bit more involved (cf. https://github.com/gnosis/dex-services/issues/806)

nlordell: Oh good point!

PR #808, opened 2020-04-28
Initialize orderbook early

### LINKED ISSUE #800: [event based orderbook] Enable early initialization
We had to revert the previous change because requested batch id 0 is a problem for the filtered orderbook because it uses `get_last_block_for_batch` which would take forever.

e00E:
The event based orderbook can take minutes to initialize. Instead of
implicitly doing this the first time the orderbook is requested we do it
on start up. This helps when we would be waiting to solve the next batch
anyway.

closes #800

### DIFF
--- a/driver/src/orderbook/mod.rs
+++ b/driver/src/orderbook/mod.rs
@@ -31,6 +31,11 @@ pub trait StableXOrderBookReading {
     /// # Arguments
     /// * `batch_id_to_solve` - the index for which returned orders should be valid
     fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)>;
+    /// Perform potential heavy initialization of the orderbook. If this fails or wasn't called
+    // the orderbook will initialize on first use of `get_auction_data`.
+    fn initialize(&self) -> Result<()> {
+        Ok(())
+    }
 }
 
 /// The different kinds of orderbook readers.
--- a/driver/src/orderbook/shadow_orderbook.rs
+++ b/driver/src/orderbook/shadow_orderbook.rs
@@ -59,9 +59,6 @@ impl<'a> StableXOrderBookReading for ShadowedOrderbookReader<'a> {
 
         Ok(orderbook)
     }
-    fn initialize(&self) -> Result<()> {
-        self.primary.initialize()
-    }
 }
 
 /// Background shadow thread that receives orders from the order channel,
@@ -74,9 +71,6 @@ fn background_shadow_reader(
     reader: &dyn StableXOrderBookReading,
     channel: Receiver<(u32, Orderbook)>,
 ) {
-    if let Err(err) = reader.initialize() {
-        log::error!("failed to initialize shadowed orderbook: {:?}", err);
-    }
     while let Ok((batch_id, primary_orderbook)) = channel.recv() {
         let shadow_orderbook = match reader.get_auction_data(batch_id.into()) {
             Ok(orderbook) => orderbook,
--- a/driver/src/orderbook/shadow_orderbook.rs
+++ b/driver/src/orderbook/shadow_orderbook.rs
@@ -59,6 +59,9 @@ impl<'a> StableXOrderBookReading for ShadowedOrderbookReader<'a> {
 
         Ok(orderbook)
     }
+    fn initialize(&self) -> Result<()> {
+        self.primary.initialize()
+    }
 }
 
 /// Background shadow thread that receives orders from the order channel,
@@ -71,6 +74,9 @@ fn background_shadow_reader(
     reader: &dyn StableXOrderBookReading,
     channel: Receiver<(u32, Orderbook)>,
 ) {
+    if let Err(err) = reader.initialize() {
+        log::error!("failed to initialize shadowed orderbook: {:?}", err);
+    }
     while let Ok((batch_id, primary_orderbook)) = channel.recv() {
         let shadow_orderbook = match reader.get_auction_data(batch_id.into()) {
             Ok(orderbook) => orderbook,
--- a/driver/src/orderbook/streamed/updating_orderbook.rs
+++ b/driver/src/orderbook/streamed/updating_orderbook.rs
@@ -192,6 +192,9 @@ impl StableXOrderBookReading for UpdatingOrderbook {
     /// Blocks on updating the orderbook. This can be expensive if `initialize` hasn't been called before.
     fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)> {
         self.do_with_context(|context| {
+            // In the case where initialize has not yet been called we update the orderbook again.
+            // This is useful because the initial update can take a long time so it is possible that
+            // the current block has changed since then.
             self.update_with_events(context).wait()?;
             context.orderbook.get_auction_data(batch_id_to_solve)
         })
--- a/driver/src/orderbook/streamed/updating_orderbook.rs
+++ b/driver/src/orderbook/streamed/updating_orderbook.rs
@@ -26,7 +26,8 @@ pub struct UpdatingOrderbook {
     /// We need a mutex because otherwise the struct wouldn't be Sync which is needed because we use
     /// the orderbook in multiple threads. The mutex is locked in `get_auction_data` while the
     /// orderbook is updated with new events.
-    context: Mutex<Context>,
+    /// None means that we have not yet been initialized.
+    context: Mutex<Option<Context>>,
     /// File path where orderbook is written to disk.
     filestore: Option<PathBuf>,
 }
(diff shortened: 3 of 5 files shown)


nlordell (approved):

bh2smith (approved):
Seems fine. I hope this concept doesn't conflict with the initialize from FileStore I am about to push.

driver/src/orderbook/mod.rs
```diff
@@ -29,6 +29,11 @@ pub trait StableXOrderBookReading {
     /// # Arguments
     /// * `batch_id_to_solve` - the index for which returned orders should be valid
     fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)>;
+    /// Perform potential heavy initialization of the orderbook. If this fails or wasn't called
+    // the orderbook will initialize on first use of `get_auction_data`.
+    fn initialize(&self) -> Result<()> {
```
fleupold, line 34: Should implementors of OrderbookReading rely on this method being called before the first invocation or is it more of a non-guaranteed optimization (like warming a cache)? In the latter case, we would probably want to call it something like prepare or so.

I don't feel strongly either way, but could we clarify the expected behavior in the docs?

driver/src/orderbook/shadow_orderbook.rs
```diff
@@ -71,6 +71,9 @@ fn background_shadow_reader(
     reader: &dyn StableXOrderBookReading,
     channel: Receiver<(u32, Orderbook)>,
 ) {
+    if let Err(err) = reader.initialize() {
```
fleupold, line 74: If initialize had an "optional" semantics it would be fine to not call it here.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -139,4 +139,11 @@ impl StableXOrderBookReading for UpdatingOrderbook {
         self.update_with_events(&mut context).wait()?;
         context.orderbook.get_auction_data(batch_id_to_solve)
     }
+    fn initialize(&self) -> Result<()> {
+        let mut context = self
+            .context
+            .lock()
+            .map_err(|err| anyhow!("poison error: {}", err))?;
+        self.update_with_events(&mut context).wait()
```
fleupold, line 147: ubernit: we could combine these 5 lines into a internal helper

fleupold (commented):
Shouldn't shadow_orderbook also implement initialize and forward it to the primary? Otherwise I don't understand how the event based orderbook receives the call (since it's wrapped in the shadowed one).

Can you please add a test plan that includes testing locally we actually initialize the event based orderbook early?

driver/src/orderbook/mod.rs
nlordell: I think the docs mention that it doesn't have to be called:
> `If this fails or wasn't called the orderbook will initialize on first use of get_auction_data.`

fleupold: 🙈

driver/src/orderbook/shadow_orderbook.rs
e00E: It's not needed but I we might as well do it.

e00E:
> Shouldn't shadow_orderbook also implement initialize and forward it to the primary? Otherwise I don't understand how the event based orderbook receives the call (since it's wrapped in the shadowed one).
> 
> Can you please add a test plan that includes testing locally we actually initialize the event based orderbook early?

The shadowed orderbook just takes a reference and we make sure to initialize the primary orderbook at the end of main which is the filtered orderbook which does forward the call.
But I agree I should have specified the test plan:
Run the driver with the orderbook set to `eventbased` and the shadowed orderbook once enabled, once disable and check that we the see the log message indicating that the event based orderbook is initializing before the scheduler starts.

e00E:
I see now. When I made this PR this was indeed true. But after rebasing master the code did not compile anymore because we changed the way we create `orderbook` in main. Now it is correct that we have to initialize through the shadowed orderbook.

e00E:
@fleupold

driver/src/orderbook/shadow_orderbook.rs
fleupold: An argument for not doing it would be that this would happen even before a signal to the channel was sent and thus likely happen at the same time (and thus consume node power) as our more critical primary orderbook fetching.

fleupold (approved):
Makes sense why it used to work. I think it's better the way it is now (when the shadow orderbook does it itself) as it's consistent with the way the other non-owning wrapping OrderbookReading implementations.

driver/src/orderbook/shadow_orderbook.rs
e00E: For our real configuration this line doesn't do anything anyway because the orderbook we use as the shadow does not do any initialization.
Conceptually it makes more sense to initialize here because we are in another thread after all and we will need this orderbook.

e00E:
I had to change the `UpdatingOrderbook` part of this PR to incorporate Bens' loading from file changes. For this I created a new `do_with_context` function that internally initializes which makes us only lock the mutex in exactly one place.

driver/src/orderbook/shadow_orderbook.rs
fleupold: I'm fine with this decision, but want to point out that the argument 

> For our real configuration this line doesn't do anything anyway

is very brittle and can easily change in the future without someone remembering to update this place (the shadow orderbook knows nothing about the concrete implementation of Reading).

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -166,14 +189,14 @@ impl UpdatingOrderbook {
...
+    /// Blocks on updating the orderbook. This can be expensive if `initialize` hasn't been called before.
     fn get_auction_data(&self, batch_id_to_solve: U256) -> Result<(AccountState, Vec<Order>)> {
-        let mut context = self
-            .context
-            .lock()
-            .map_err(|err| anyhow!("poison error: {}", err))?;
-        self.update_with_events(&mut context).wait()?;
-        context.orderbook.get_auction_data(batch_id_to_solve)
+        self.do_with_context(|context| {
+            self.update_with_events(context).wait()?;
```
fleupold, line 195: Now, if the optional `initialize` is not called, we `update_with_events` twice in a row (fetching at least 25 blocks). Not a big issue, but why don't we do the update_with_events unconditionally but only once inside `do_with_context` and maybe call it `do_with_updated_context` ?

driver/src/orderbook/shadow_orderbook.rs
e00E: That argument works the other way too. If we had an expensive to initialize orderbook here then it would be a problem to not initialize it early because then the orderbook comparison would start too late.
What I meant by "conceptually" is that if I read the orderbookreading trait and see this initialize method then I would definitely call it in this place in the shadow orderbook.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I didn't think of that when I wrote this code but after trying your idea out I realized that the side effect of the  current version that the orderbook is updated twice is actually a positive. The initial update with all past events can take a long time so it makes sense to do the second small update again in case there have been new blocks since then.
I'll add a comment to explain this and keep the behavior.

driver/src/orderbook/shadow_orderbook.rs
nlordell: > happen at the same time (and thus consume node power) as our more critical primary orderbook fetching.

Is this still a problem if our node is sitting behind a node balancer? I would be interested in seeing a quick (and un-scientific) benchmark of running an event orderbook shadowed by an event orderbook and compare start up times (just for curiosity's sake).

Also, does excessive querying affect the ability of a node to verify a solution and submit a transaction? If so I'd be more inclined to have a slower initialization then have potential problems querying the node or submitting a transaction when it actually matters.

> the orderbook comparison would start too late.

This shouldn't be a problem since we ignore new comparisons if the previous ones are taking too long. Arguably warming up as fast as possible (and not missing batches) has more value than missing a shadowed orderbook verification (which is just for piece of mind and has no impact to users)

> I read the orderbookreading trait and see this initialize method then I would definitely call it in this place in the shadow orderbook.

This also makes sense to me. To me calling `initialize` is also the "intuitive" way to implement this given the trait description.

nlordell (approved):
THought I had already approved but I didn't

driver/src/orderbook/shadow_orderbook.rs
fleupold: > I read the orderbookreading trait and see this initialize method then I would definitely call it in this place in the shadow orderbook.

> This also makes sense to me. To me calling initialize is also the "intuitive" way to implement this given the trait description.

To me the intuitive (semi-lazy) way would be to initialize the shadow when someone is calling initialize on the shadowing orderbook itself (if no-one wants to initialize the shadowing orderbook, why should it initialize its sub-orderbooks). 
I see that this might be a bit annoying to implement given the owner structure here.

nlordell: > initialize on the shadowing orderbook

That’s a very good point, I hadn't thought about it.  Sorry if I had misunderstood our previous point.

Maybe the “hidden” expensive background initialization is not the best idea.

> I see that this might be a bit annoying to implement given the owner structure here.

I think it would be fairly easy to do over the channel. It would send a `Request` enum that is either a `Initialize` or a `Verify`.

e00E: To keep this PR simple and let it accomplish the original goal which was just to initialize the event based orderbook I am going to remove any initialization from the shadow orderbook. We can rethink this later.

PR #809, opened 2020-04-28
Sync Event Based Orderbook from Disk.

bh2smith:
- Initialize Event Based Orderbook from disk in constructor, (Using default if file not available) 
- Write orderbook to file at the end of each update. 
- Include new configuration parameter `orderbook_filepath` with default value of `/var/db/stablex_orderbook.ron`

Looking for suggestions on how to test this.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -184,6 +184,13 @@ struct Options {
         parse(try_from_str)
     )]
     use_shadowed_orderbook: bool,
+
+    #[structopt(
+        long,
+        env = "ORDERBOOK_FILEPATH",
+        default_value = "/var/db/stablex_orderbook.ron"
+    )]
+    orderbook_filepath: String,
 }
 
 fn main() {
@@ -238,6 +245,7 @@ fn main() {
         options.auction_data_page_size,
         &options.orderbook_filter,
         web3,
+        options.orderbook_filepath,
     );
 
     // NOTE: Keep the shadowed orderbook around so it doesn't get dropped and we
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -37,6 +37,7 @@ use ethcontract::PrivateKey;
 use log::info;
 use prometheus::Registry;
 use std::num::ParseIntError;
+use std::path::PathBuf;
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
@@ -185,12 +186,8 @@ struct Options {
     )]
     use_shadowed_orderbook: bool,
 
-    #[structopt(
-        long,
-        env = "ORDERBOOK_FILEPATH",
-        default_value = "/var/db/stablex_orderbook.ron"
-    )]
-    orderbook_filepath: String,
+    #[structopt(long, env = "ORDERBOOK_FILE", parse(from_os_str))]
+    orderbook_file: Option<PathBuf>,
 }
 
 fn main() {
@@ -245,7 +242,7 @@ fn main() {
         options.auction_data_page_size,
         &options.orderbook_filter,
         web3,
-        options.orderbook_filepath,
+        options.orderbook_file,
     );
 
     // NOTE: Keep the shadowed orderbook around so it doesn't get dropped and we
--- a/driver/src/orderbook/streamed/updating_orderbook.rs
+++ b/driver/src/orderbook/streamed/updating_orderbook.rs
@@ -12,10 +12,11 @@ use anyhow::{anyhow, bail, Result};
 use block_timestamp_reading::{BlockTimestampReading, CachedBlockTimestampReader};
 use ethcontract::{contract::Event, BlockNumber, H256};
 use futures::compat::Future01CompatExt as _;
+use log::error;
 use orderbook::Orderbook;
 use std::collections::HashSet;
 use std::convert::TryFrom;
-use std::path::Path;
+use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 
 /// An event based orderbook that automatically updates itself with new events from the contract.
@@ -27,7 +28,7 @@ pub struct UpdatingOrderbook {
     /// orderbook is updated with new events.
     context: Mutex<Context>,
     /// File path where orderbook is written to disk.
-    filestore: String,
+    filestore: Option<PathBuf>,
 }
 
 struct Context {
@@ -42,10 +43,13 @@ impl UpdatingOrderbook {
     pub fn new(
         contract: Arc<dyn StableXContract + Send + Sync>,
         web3: Web3,
-        path_str: String,
+        path: Option<PathBuf>,
     ) -> Self {
         // Recover the orderbook from file (if possible, otherwise use default).
-        let orderbook = Orderbook::try_from(Path::new(&path_str)).unwrap_or_default();
+        let orderbook = match &path {
+            Some(path) => Orderbook::try_from(path.as_path()).unwrap_or_default(),
+            None => Orderbook::default(),
+        };
         let last_handled_block = orderbook.last_handled_block().unwrap_or(0);
         Self {
             contract,
@@ -55,7 +59,7 @@ impl UpdatingOrderbook {
                 last_handled_block,
                 block_timestamp_reader: CachedBlockTimestampReader::new(web3),
             }),
-            filestore: path_str,
+            filestore: path,
         }
     }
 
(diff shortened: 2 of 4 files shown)


driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -64,6 +76,7 @@ impl UpdatingOrderbook {
             .past_events(BlockNumber::Number(from_block.into()), to_block)
             .await?;
         self.handle_events(context, events, from_block).await?;
+        context.orderbook.write_to_file(&self.filestore)?;
```
e00E, line 79: It might be better log this error if it occurs but not make the whole function fail. So instead of `?` I suggest `if let Err(err) = ... { log error }`. 
This way if we for some reason cannot write to the file at least the orderbook itself still works, it just doesn't update.

driver/src/main.rs
```diff
@@ -184,6 +184,13 @@ struct Options {
         parse(try_from_str)
     )]
     use_shadowed_orderbook: bool,
+
+    #[structopt(
+        long,
+        env = "ORDERBOOK_FILEPATH",
+        default_value = "/var/db/stablex_orderbook.ron"
+    )]
+    orderbook_filepath: String,
```
e00E, line 193: Let's make this a `PathBuf`. I found this in the structop documentation:
```rust
    /// Input file
    #[structopt(parse(from_os_str))]
    input: PathBuf,
```

or maybe even better `Option<PathBuf>` where `None` indicates that we shouldn't persist the orderbook. That way we do not have to pick a default path.

Then `UpdatingOrderbook::new` would also take and store `Option<PathBuf>`.

e00E:
> Looking for suggestions on how to test this.

For a temporary local test you could change `main` so that it just creates the orderbook and queries it in a loop (remove the driver and scheduler parts). Then you can run the binary locally and see from the log messages whether it works and check the file on disk.

driver/src/main.rs
bh2smith: Just want to mention that I have not implemented TryFrom for `PathBuf`. So I hope I can find some way to convert PathBuf into Path.

e00E (approved):

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -35,15 +40,26 @@ struct Context {
...
+        contract: Arc<dyn StableXContract + Send + Sync>,
+        web3: Web3,
+        path: Option<PathBuf>,
+    ) -> Self {
+        // Recover the orderbook from file (if possible, otherwise use default).
+        let orderbook = match &path {
+            Some(path) => Orderbook::try_from(path.as_path()).unwrap_or_default(),
+            None => Orderbook::default(),
+        };
+        let last_handled_block = orderbook.last_handled_block().unwrap_or(0);
```
fleupold, line 53: I would do this also as part of the initialize method implemented in #808 (since it could also be a little bit long running and is kind of optional)

driver/src/main.rs
```diff
@@ -184,6 +185,9 @@ struct Options {
         parse(try_from_str)
     )]
     use_shadowed_orderbook: bool,
+
+    #[structopt(long, env = "ORDERBOOK_FILE", parse(from_os_str))]
+    orderbook_file: Option<PathBuf>,
```
fleupold, line 190: If later we would store more than one file we should probably just pass a "cache directory" here and let each components handle their file names internally. Maybe not needed now, I'm just starting to get worried about all these little one use case specific config values.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -64,6 +80,14 @@ impl UpdatingOrderbook {
             .past_events(BlockNumber::Number(from_block.into()), to_block)
             .await?;
         self.handle_events(context, events, from_block).await?;
+        if self.filestore.is_some() {
```
fleupold, line 83: nit: to avoid the unwrap below you can do

`if let Some(filestore) = self.filestore { ...`

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -64,6 +80,14 @@ impl UpdatingOrderbook {
...
+            if let Err(write_error) = context
+                .orderbook
+                .write_to_file(&self.filestore.as_ref().unwrap())
```
fleupold, line 86: What are your thoughts about not doing this on the critical path (but as a background task instead)? How long does writing the file to disk take? Probably not long compared to fetching the data, but it still feels a bit weird to do this synchronously when the rest of the system is desperately waiting for orderbook data.

Also, is there value in only writing if it's dirty (not in this PR).

fleupold (approved):

driver/src/main.rs
nlordell: +1 for the `Option<PathBuf>`.

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -35,15 +40,26 @@ struct Context {
...
     /// `get_auction_data` which can thus take a long time to complete.
-    pub fn new(contract: Arc<dyn StableXContract + Send + Sync>, web3: Web3) -> Self {
+    pub fn new(
+        contract: Arc<dyn StableXContract + Send + Sync>,
+        web3: Web3,
+        path: Option<PathBuf>,
+    ) -> Self {
+        // Recover the orderbook from file (if possible, otherwise use default).
+        let orderbook = match &path {
+            Some(path) => Orderbook::try_from(path.as_path()).unwrap_or_default(),
```
nlordell, line 50: Maybe we should log a warning or something if this fails but was expected to succeed?

nlordell (approved):
Looks good!

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: We can merge this first and I'll move it there in my PR.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: We don't know whether this is suspected to succeed. The first time you run you might not have a cache yet but still pass the path in order to create it. I don't mind the warning though as a way we can see that it works in the logs.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: I expect this write to take less than a second unless there is something weird going on with kubernetes. Still we can move it into a thread in the future but it let's not do it in this PR.

driver/src/main.rs
e00E: I prefer to introduce that if we actually have another thing we want to persist over pod recreations and not now.

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: Just spitballing but if `path` doesn't exist then `Ok(Orderbook::default())` can be returned instead of an error? That was we don't get false-positive warning logs when there is no cache.

driver/src/main.rs
bh2smith: Done.

driver/src/orderbook/streamed/updating_orderbook.rs
bh2smith: > Also, is there value in only writing if it's dirty (not in this PR).

Will create an issue here.. but if we define "dirty" as out of sync, then it would be consider always dirty at this point.

driver/src/orderbook/streamed/updating_orderbook.rs
bh2smith: So, if I understand the discussion here, we would like to include a warning that file was not found before using default and logging something different when we use default orderbook because there was no path? Or simply just log some warning that the orderbook at specified path was not there?

bh2smith: I have made the suggested changes here. Would be nice if I could get one last review of this particular chunk of code before merge! @e00E  and/or @nlordell

e00E (approved):

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -35,15 +40,40 @@ struct Context {
...
+                    info!("successfully recovered orderbook from path");
+                    orderbook
+                }
+                Err(error) => {
+                    // This will always happen when file does not exist (i.e. on first startup)
+                    // TODO: match on this error and don't warn when file doesn't exist.
+                    warn!(
+                        "Failed to construct orderbook from path (using default): {}",
+                        error
+                    );
```
nlordell, line 61:
```suggestion
                    if path.exists() {
                        warn!(
                            "Failed to construct orderbook from path (using default): {}",
                            error
                        );
                    }
```

bh2smith: SOrry @nlordell ... too late. Will make a new PR with this change.

bh2smith: I suppose this is cheeky way to determine that the failure was not because the file did not exist?

nlordell: It checks if the file exists before logging the warning.

bh2smith: Addressed in #817

PR #810, opened 2020-04-28
[FilteredOrderbook] Filter balance for which there are no sell orders

fleupold:
This fixes the alerts in the fallback solver we saw today about having balances in the primary orderbook but not in the shadow orderbook.

The on-chain filtered orderbook doesn't return orders (and thus also doesn't return balances) that are not between whitelisted tokens. Thus any balance for tokens >18 (and any balance for tokens that only have an order to a non-whitelisted one) are missing in this implementation.

The client side filtered orderbook also omits orders from such tokens, however it doesn't prune the account balance state. Thus, if we fetched orders using the event based orderbook that contained excluded tokens, their balances will still be reflected in the resulting instance file.

This PR moves the filtering logic that we were also using to exclude balances for which there are no orders at all (even including non-whitelisted tokens) into a common util crate for reuse.

### Test Plan

Added a unit test and made sure (this time on mainnet) that primary and staging orderbooks are consistent with with one another.

Note, there seems to be an issue with the BatchExchangeViewer when running with page size of 100 (@nlordell is working on a fix), however it works with a page size of 1000 which is what we are using in staging.

### DIFF
--- a/driver/src/orderbook/filtered_orderbook.rs
+++ b/driver/src/orderbook/filtered_orderbook.rs
@@ -83,20 +83,20 @@ impl<'a> StableXOrderBookReading for FilteredOrderbookReader<'a> {
                 })
                 .collect(),
         };
-        let user_filtered_orders = token_filtered_orders
-            .into_iter()
-            .filter(|o| {
-                if let Some(user_filter) = self.filter.users.get(&o.account_id) {
-                    match user_filter {
-                        UserOrderFilter::All => false,
-                        UserOrderFilter::OrderIds(ids) => !ids.contains(&o.id),
-                    }
-                } else {
-                    true
+        let user_filtered_orders = token_filtered_orders.into_iter().filter(|o| {
+            if let Some(user_filter) = self.filter.users.get(&o.account_id) {
+                match user_filter {
+                    UserOrderFilter::All => false,
+                    UserOrderFilter::OrderIds(ids) => !ids.contains(&o.id),
                 }
-            })
-            .collect();
-        Ok((state, user_filtered_orders))
+            } else {
+                true
+            }
+        });
+        Ok(util::filter_auction_data(
+            state.into_iter(),
+            user_filtered_orders,
+        ))
     }
 }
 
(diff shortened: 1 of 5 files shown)


driver/src/orderbook/filtered_orderbook.rs
```diff
@@ -83,20 +83,20 @@ impl<'a> StableXOrderBookReading for FilteredOrderbookReader<'a> {
...
                 }
-            })
-            .collect();
-        Ok((state, user_filtered_orders))
+            } else {
+                true
+            }
+        });
+        Ok(util::filter_auction_data(
+            state.into_iter(),
```
nlordell, line 97: nit: Doesn't the function takes `IntoIterator`? It think you shouldn't need to call it here if it does.
```suggestion
            state,
```

driver/src/orderbook/util.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Filters empty orders as and account balances for tokens for which there is
+/// not at least one sell order by a user
+pub fn filter_auction_data(
+    account_states: impl IntoIterator<Item = ((Address, u16), u128)>,
+    orders: impl IntoIterator<Item = Order>,
+) -> (AccountState, Vec<Order>) {
+    let orders = orders
+        .into_iter()
+        .filter(|order| order.sell_amount > 0)
```
nlordell, line 12: Nice! Common filtering code :smile:

nlordell (approved):
Great cleanup PR!

driver/src/orderbook/util.rs
```diff
@@ -0,0 +1,62 @@
+use crate::models::{AccountState, Order};
+use ethcontract::Address;
+
+/// Filters empty orders as and account balances for tokens for which there is
+/// not at least one sell order by a user
+pub fn filter_auction_data(
```
nlordell, line 6: One comment I just though of, We now how many "filtered" things going on in the orderbook:
- `FilteredOrderbookReader`
- `OnchainFilteredOrderbookReader`
- `filter_auction_data`

Maybe this method should be renamed just to avoid confusion. Something like `canonicalize_auction_data`? I'm not 100% sold on the name, maybe you can think of something better.

e00E (approved):

PR #812, opened 2020-04-28
[Easy] Always query orderbook at a pinned block number

### LINKED ISSUE #811: Make `get_last_block_for_batch` always return a concrete block number
https://github.com/gnosis/dex-services/blob/c3983eee2204748780015087adea42c940a767ee/driver/src/contracts/stablex_contract.rs#L164-L166

If the while condition is false in the first attempt, we return Block::Latest. If this is then used in a long running ordebook fetching process (e.g. using the OnChainFiltered orderbook) we query each page on latest, which might be changing while we paginate and thus yield incosistent results.

nlordell:
This PR modifies the `get_last_block_for_batch` method to return the "latest" block number instead of `None` when the batch isn't finalized.

Fixes #811 

### Test Plan

CI, Rinkeby e2e test will use this new logic.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -83,8 +83,9 @@ pub trait StableXContract {
     fn get_current_auction_remaining_time(&self) -> Result<Duration>;
 
     /// Searches for the block number of the last block of the given batch. If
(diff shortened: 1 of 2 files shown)


fleupold (approved):
:shipit:

e00E (approved):

PR #816, opened 2020-04-29
[easy] Add more tests to orderbook

e00E:


driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -209,4 +209,55 @@ mod tests {
...
+        // Add a couple of events in random order.
+        add_event(2, 0, 2);
+        add_event(4, 1, 1);
+        add_event(0, 0, 0);
+        add_event(5, 1, 2);
+        add_event(3, 1, 0);
+        add_event(1, 0, 1);
+        assert_eq!(orderbook.events.len(), 6);
+        // Check that the order is correct.
+        for (i, (_key, value)) in orderbook.events.iter().enumerate() {
```
fleupold, line 256: I find this test quite hard to understand. I have to wrap my had around this loop and dig into the implementation of the add_event closure to see deposit_id is used as batch_id. It's not very explicit that this is actually testing the sorting.

Can we maybe just list out the 6 value explicitly by calling .next() ?

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -209,4 +209,55 @@ mod tests {
...
+            match &value.event {
+                Event::Deposit(deposit) if deposit.batch_id == i as u32 => (),
```
fleupold, line 258: Can we explicitly assert what we are testing. In this case, I'd expect an `assert_eq!(deposit.batch_id, i)`

I find this is easier to spot and understand for the reader rather than thinking inversely in what cases we panic.

fleupold (approved):
Looks good, although I'd prefer if the test cases could be more linear and easily to understand (even if it involves not writing the most elegant code, I think that is fine in tests).

Also, would it be more rewarding to set up test coverage measuring before to see how we increase?

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -209,4 +209,55 @@ mod tests {
...
+            );
+        }
+        assert_eq!(orderbook.events.len(), 3);
+        orderbook.delete_events_starting_at_block(1);
+        assert_eq!(orderbook.events.len(), 1);
+        assert_eq!(orderbook.events.iter().next().unwrap().0.block_number, 0);
+    }
+
+    #[test]
+    fn events_get_sorted() {
```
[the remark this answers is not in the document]
e00E: It says what we are testing, that "events get sorted". *How* we are testing that events get sorted doesn't have to be in the name imo.

driver/src/orderbook/streamed/orderbook.rs
e00E: >  I'd prefer if the test cases could be more linear and easily to understand (even if it involves not writing the most elegant code, I think that is fine in tests).

I disagree. To me inlining and repeating everything makes the test harder to understand because there is more noise and I cannot be sure that all cases have been inlined the same way.

e00E:
>  I'd prefer if the test cases could be more linear and easily to understand (even if it involves not writing the most elegant code, I think that is fine in tests).

I disagree. To me inlining and repeating everything makes the test harder to understand because there is more noise and I cannot be sure that all cases have been inlined the same way.

driver/src/orderbook/streamed/orderbook.rs
e00E: >  I'd prefer if the test cases could be more linear and easily to understand (even if it involves not writing the most elegant code, I think that is fine in tests).

I disagree. To me inlining and repeating everything makes the test harder to understand because there is more noise and I cannot be sure that all cases have been inlined the same way.

e00E:
> Also, would it be more rewarding to set up test coverage measuring before to see how we increase?

Maybe? Test coverage looked more complicated and these tests are helpful regardless of how coverage changes.

driver/src/orderbook/streamed/orderbook.rs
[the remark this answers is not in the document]
nlordell: I actually prefer to not prefix test method names with `test_`. I prefer `test_*` functions and methods to be for testing utilities (like mock data or some helpers for setting up tests).

The reason I feel like the `test_` prefix is a bit redundant is that:
- They already have a the `#[test]` attribute to indicate they are a test.
- They are always in `mod tests` modules, so calling them would look like `tests::my_test_method` which already indicates it is a test.
- When you run tests locally they show up like this: `test util::tests::test_ceiled_div_u128 ... ok`, I feel like `test util::tests::ceiled_div_u128 ... ok` reads a little nicer.

To me, the method name should be a snake_cased version of the description you have in `it('...')` for mocha tests.

To be clear, I have no real strong opinion about this, so I'm OK either way, those are just my 2c.

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -209,4 +209,55 @@ mod tests {
         // Cleanup the file created here.
         assert!(fs::remove_file(test_path).is_ok());
     }
+
+    #[test]
+    fn delete_events_starting_at_block() {
+        let mut orderbook = Orderbook::default();
+        for block_number in 0..3 {
+            orderbook.handle_event_data(
+                EventData::Added(Event::Deposit(Deposit::default())),
```
bh2smith, line 218: I wasn't aware of the defaults on events! I would have used this in my recent test as well.

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -209,4 +209,55 @@ mod tests {
...
+                0,
+            );
+        };
+        // Add a couple of events in random order.
+        add_event(2, 0, 2);
+        add_event(4, 1, 1);
+        add_event(0, 0, 0);
+        add_event(5, 1, 2);
+        add_event(3, 1, 0);
+        add_event(1, 0, 1);
```
bh2smith, line 253: If the input parameters to add event were aligned with the way `EventSortKeys` were ordered, I would think that these values would be much easier to digest. I think this would be as easy as using 

`block_number, log_index, deposit_id` instead of `deposit_id, block_number, log_index`.

Just an idea.

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -209,4 +209,55 @@ mod tests {
...
+        assert_eq!(orderbook.events.len(), 6);
+        // Check that the order is correct.
+        for (i, (_key, value)) in orderbook.events.iter().enumerate() {
+            match &value.event {
```
nlordell, line 257: Just a suggestion, but we can use the new `matches!` macro here to be a little more succinct:
```
assert!(matches!(&value.event, Event::Deposit(deposit) if deposit.batch_id == i));
```

driver/src/orderbook/streamed/orderbook.rs
bh2smith: Also, it appears that `deposit_id` is only ever being used as `batch_id`. Does this mean that is should really just be `batch_id`?

bh2smith (approved):
Seems fine, the last test is a bit strange to read, but I like the closure.

driver/src/orderbook/streamed/orderbook.rs
nlordell: Maybe a nice macro with a DSL might be a good compromise? While I also find noise in tests extremely frustrating (since they make the tests so long and hard to navigate because they are so many lines of similar code), maybe a macro with a dash of sugar can make things a bit easier to read?
```
add_deposit_events! {
        deposit # 2 block 0 log 2,
        deposit # 4 block 1 log 1,
        // ...
}
```

I find macros that define compact DSLs are a good way to make tests more readable and expressive without having too much additional noise.

nlordell (approved):
More testing, YAY! I agree with @fleupold that getting some coverage stats might make this more rewarding as you see the coverage numbers going up and up :rocket:

driver/src/orderbook/streamed/orderbook.rs
nlordell: I think the test is using `batch_id` to just hold the expected order of the deposits.

driver/src/orderbook/streamed/orderbook.rs
fleupold: From a future maintainers perspective, I still don't fully understand what this is testing. 

We are checking that the batch_id is equal to the index of our loop. The connection between that and the thing that we are actually sorting by (block_number and event index) is super hidden in six randomly ordered  `add_event` with three times the same argument type combined with the detail that the deposit ID is reused as the batch_id (I like @nlordell's idea of using a macro that can name the types).

Moreover, isn't the fact that events are stored sorted in some way an implementation detail (it's not a promise we make via the interface). The class might as well store it as a Vec or HashMap. Therefore I'm not sure this should be tested. Maybe we want to test sorting `EventSortKey` directly but that doesn't require setting up an entire orderbook and events.

fleupold: > To me inlining and repeating everything makes the test harder to understand because there is more noise and I cannot be sure that all cases have been inlined the same way.

I do understand that long test cases are hard to read. However, unit tests should be testing logic and avoid adding more logic (which itself should be tested if it's non trivial). I agree it's a balance.

In this concrete example, writing 6 lines of the concrete expectation is in my mind much simpler to understand (worth 3 extra lines of code) than writing this enumerate loop.

e00E: At this point the discussion about whether we should use loops and functions in tests isn't fruitful. We disagree on something subjective and the arguments have been said.

> Moreover, isn't the fact that events are stored sorted in some way an implementation detail (it's not a promise we make via the interface). The class might as well store it as a Vec or HashMap. Therefore I'm not sure this should be tested. Maybe we want to test sorting EventSortKey directly but that doesn't require setting up an entire orderbook and events.

It is an implementation detail. There is a philosophy of testing only public things and never implementation details. I disagree with taking this to the extreme but in general it makes sense.
In this specific case the alternative is to instead pick events that result in a valid `State` and then get the orderbook for that state to check that the events have been applied in the correct order. I considered this but didn't like that it would make the test more tedious and verbose.

I am going to change the test to work that way.

bh2smith: What is DSL? Digital Second Line?

e00E: domain specific language

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -209,4 +209,119 @@ mod tests {
...
+            amount: 3.into(),
+        }));
+
+        let mut orderbook = Orderbook::default();
+        orderbook.handle_event_data(token_listing_1, 0, 1, H256::zero(), 0);
+        orderbook.handle_event_data(order_placement, 0, 2, H256::zero(), 0);
+        orderbook.handle_event_data(withdraw, 2, 0, H256::zero(), 300);
+        orderbook.handle_event_data(deposit, 0, 3, H256::zero(), 0);
+        orderbook.handle_event_data(withdraw_request, 1, 0, H256::zero(), 0);
+        orderbook.handle_event_data(token_listing_0, 0, 0, H256::zero(), 0);
```
fleupold, line 319: In order to make this test more minimal, how about simply placing an order before listing it and asserting we can create state without an error? Would that be assuming too much about the internal consistency checks?

e00E: I'm assuming you mean placing an order before listing the tokens. In that case like you said I thought it was assuming too much about the consistency check so I wanted to go with something measurable in the orderbook like a balance or an order used amount.

bh2smith:
I feel like this has been ready for quite a while. Is there something special we are waiting for here?

fleupold (commented):
Nothing from my side. We discussed today that the fact the test for checking the ordering of events is so long, could be a sign that our orderbook and state are too tightly coupled. 

Maybe in the future, the orderbook could expose an event Iterator that emits the events in order (and can be tested in isolation without having to worry about the logic of how events are applied).
Then we can test the State separately by creating it from such an iterator (which in the test could be a simple vector)

PR #822, opened 2020-04-30
Use alternative Serialization Type

### LINKED ISSUE #820: Can't serialize event based orderbook because serializer doesn't support u128
When I tested on rinkeby I noticed this message

> Failed to write to orderbook Custom message: u128 is not supported

Apparently ron doesn't support u128? I am not sure yet. You can reproduce this by changing the test `test_serialize_deserialize_orderbook` to use a Trade event instead of Deposit like so
```rust
  let event = Event::Trade(Trade::default());
```

bh2smith:
Turns out RON does not yet support `u128` so we will need a new serialization type. First commit includes more thorough test highlighting RON's the incapability. Failing travis should say this:

```
thread 'orderbook::streamed::orderbook::tests::test_serialize_deserialize_orderbook' panicked at 'called `Result::unwrap()` on an `Err` value: Message("u128 is not supported")'
```

Still need to find an alternative. It was suggested by @e00E that we look into using [bincode](https://github.com/servo/bincode). The file format is not human readable (which doesn't matter) but we would get smaller filesizes.

Closes #820

### DIFF
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -161,23 +161,37 @@ mod tests {
 
     #[test]
     fn test_serialize_deserialize_orderbook() {
-        let event_key = EventSortKey {
-            block_number: 0,
-            block_hash: H256::zero(),
-            log_index: 1,
-        };
-        let event = Event::Deposit(Deposit {
-            user: Address::from_low_u64_be(1),
-            token: Address::from_low_u64_be(2),
-            amount: 1.into(),
-            batch_id: 2,
-        });
-        let value = Value { event, batch_id: 0 };
+        let event_list: Vec<Event> = vec![
+            Event::Deposit(Deposit::default()),
+            Event::WithdrawRequest(WithdrawRequest::default()),
+            Event::Withdraw(Withdraw::default()),
+            Event::TokenListing(TokenListing::default()),
+            Event::OrderPlacement(OrderPlacement::default()),
+            Event::OrderCancellation(OrderCancellation::default()),
+            Event::OrderDeletion(OrderDeletion::default()),
+            Event::Trade(Trade::default()),
+            Event::TradeReversion(TradeReversion::default()),
+            Event::SolutionSubmission(SolutionSubmission::default()),
+        ];
 
-        let mut events = BTreeMap::new();
-        events.insert(event_key, value);
+        let events: BTreeMap<EventSortKey, Value> = event_list
+            .iter()
+            .enumerate()
+            .map(|(i, event)| {
+                (
+                    EventSortKey {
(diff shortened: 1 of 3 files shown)


fleupold:
I think @e00E was considering making a pull request to ron for adding support. What do you guys think is better considering that we are currently losing ~1 batch whenever we restart our hosts all at once (we could also move to a phased restart)

bh2smith:
> I think @e00E was considering making a pull request to ron for adding support.

PR was already made here https://github.com/ron-rs/ron/pull/219 and appears to have gotten some attention already. The fix is elegant and simple. I would really hope that we don't have to change our serialization type because this would essentially force us to use JSON and write our own Serializations for our objects.

fleupold:
Until this is merged, we could depend on the forked repo/branch that already contains this change (not sure if there is a risk of non-backwards compatible changes in the way serialization will work).

bh2smith:
> Until this is merged, we could depend on the forked repo/branch that already contains this change (not sure if there is a risk of non-backwards compatible changes in the way serialization will work).

I would be happy to do this. Its a pretty nice clean change anyway and shouldn't cause any problems.

e00E:
I don't want to depend on the branch that I used for the PR in RON because we might make backwards incompatible changes to that branch that would break our build here. We could make another fork ourselves but I'd rather use a different format where we can use the real version from crates.io .

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -161,26 +161,39 @@ mod tests {
...
+                )
+            })
+            .collect();
         let orderbook = Orderbook { events };
-
-        let serialized_orderbook =
-            ron::ser::to_string_pretty(&orderbook, ron::ser::PrettyConfig::default()).unwrap();
-        let deserialized_orderbook = Orderbook::try_from(serialized_orderbook.as_bytes()).unwrap();
+        let serialized_orderbook = bincode::serialize(&orderbook).unwrap();
+        let deserialized_orderbook = Orderbook::try_from(&serialized_orderbook[..]).unwrap();
```
fleupold, line 196: nit: using `expect` instead of `unwrap` with a nice description makes it slightly easier to debug an issue if the test ever failed because of that (it gives the maintainer a bit of context).

fleupold (approved):
Looks good to me! 👍

driver/src/orderbook/streamed/orderbook.rs
bh2smith: Yes I will fix this and merge.

bh2smith:
No that we no longer use RON, I will change the gnosis-staging filename to use something that does not end in `.ron`. Does anyone have a suggestion for the file extension (if any) that might be appropriate to use for this? Perhaps `.binary`

PR #847, opened 2020-05-04
Do not read file as string

e00E:
The file doesn't have been to be valid utf-8 anymore since we switched
to a binary format.

Test Plan:
Run main twice with an orderbook file path set and see that loading works the second time.

### DIFF
--- a/driver/src/orderbook/streamed/orderbook.rs
+++ b/driver/src/orderbook/streamed/orderbook.rs
@@ -113,15 +113,15 @@ impl TryFrom<File> for Orderbook {
     type Error = anyhow::Error;
 
     fn try_from(mut file: File) -> Result<Self> {
-        let mut contents = String::new();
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Should we maybe use the `File` as a `Reader` so we don't need to load the entire file in memory? https://docs.rs/bincode/1.2.1/bincode/fn.deserialize_from.html

bh2smith:
> Should we maybe use the File as a Reader so we don't need to load the entire file in memory? 

Seems like a reasonable choice, unless we plan to go back to RON sometime soon (I know they have bumped their version and included @e00E's fix). 

Also, would with me we scrap `TryFrom(File)` and use `TryFrom(Reader)`?

bh2smith (approved):

e00E:
We can absolutely do that, but in a separate PR.

PR #848, opened 2020-05-05
Allow specifying the internal solver used by the solver

### LINKED ISSUE #737: Add a gurobi solver config
The solver can now be invoked with gurobi (cf. https://github.com/gnosis/dex-solver/pull/205). The goal of this task is to make this option available as a top level solver configuration.

Before this can be deployed we also have to include the gurobi licence in our kubernetes config.

e00E:
This adds support for the gurobi solver.

I intentionally did not create an enum for the internal solver type
because then we would end up duplicating this from the solver. Instead
we just forward the argument.

https://github.com/gnosis/dex-solver/blob/2122d7aad044b14a2bf0a54698d8d3b4ed65d511/src/util/Parameters.py#L104
https://github.com/gnosis/dex-solver/blob/2122d7aad044b14a2bf0a54698d8d3b4ed65d511/src/util/enums/Solver.py#L7

closes #737 

### Test Plan
This change does not yet do anything. Once this and the corresponding gnosis staging PR is merged I will test that it works.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -73,6 +73,11 @@ struct Options {
     #[structopt(long, env = "SOLVER_TYPE", default_value = "naive-solver")]
     solver_type: SolverType,
 
+    /// Which optimization solver the solver should use internally. Is passed as
+    /// `--solver` to the solver.
+    #[structopt(long, env = "SOLVER_INTERNAL_SOLVER")]
+    solver_internal_solver: Option<String>,
+
     /// JSON encoded backup token information to provide to the solver.
     ///
     /// For example: '{
@@ -234,6 +239,7 @@ fn main() {
         options.solver_type,
         price_oracle,
         options.min_avg_fee_per_order,
+        options.solver_internal_solver,
     );
 
     // Create the orderbook reader.
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -30,7 +30,7 @@ use crate::orderbook::{
     ShadowedOrderbookReader, StableXOrderBookReading,
 };
 use crate::price_estimation::{PriceOracle, TokenData};
-use crate::price_finding::{Fee, SolverType};
+use crate::price_finding::{Fee, InternalOptimizer, SolverType};
 use crate::solution_submission::StableXSolutionSubmitter;
 
 use ethcontract::PrivateKey;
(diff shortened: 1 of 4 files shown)


driver/src/main.rs
```diff
@@ -73,6 +73,11 @@ struct Options {
     #[structopt(long, env = "SOLVER_TYPE", default_value = "naive-solver")]
     solver_type: SolverType,
 
+    /// Which optimization solver the solver should use internally. Is passed as
+    /// `--solver` to the solver.
```
fleupold, line 77: nit: can you document the default behavior (which is to use scip)?

driver/src/main.rs
```diff
@@ -73,6 +73,11 @@ struct Options {
...
+    #[structopt(long, env = "SOLVER_INTERNAL_SOLVER")]
+    solver_internal_solver: Option<String>,
```
fleupold, line 79: Would it make sense to validate that if the option is set, it's a valid choice (e.g. either gurobi or scip)?
I don't think our selection will change anytime soon and it might lead to weird unexpected behavior if there is a typo in the string (e.g. we think we use gurboi but fall back to scip or error when trying to solve the batch after one day of inactivity). Having an explicit type check at parsing time could be valuable imo.

driver/src/main.rs
```diff
@@ -73,6 +73,11 @@ struct Options {
     #[structopt(long, env = "SOLVER_TYPE", default_value = "naive-solver")]
     solver_type: SolverType,
 
+    /// Which optimization solver the solver should use internally. Is passed as
+    /// `--solver` to the solver.
+    #[structopt(long, env = "SOLVER_INTERNAL_SOLVER")]
```
fleupold, line 78: naming nit: the double solver reads a bit weird. How about calling it SOLVER_INTERNAL_OPTIMIZER ?

fleupold (approved):
Swift and sweet!

I'm sure @twalth3r is excited to see this...

driver/src/main.rs
e00E: I was debating this:

> I intentionally did not create an enum for the internal solver type
because then we would end up duplicating this from the solver. Instead
we just forward the argument.

I have no strong opinion either way. I'll see if someone else has an opinion on this and if not I'll add the enum.

driver/src/main.rs
e00E: It doesn't matter whether it reads weirdly, just that the name makes sense but I am fine with that name to so I'll change it.

driver/src/main.rs
nlordell: I think since we already validate other parameters, we should probably do it here as well. Also, I think that since we are handling a specific solver argument in this case we should probably limit its scope to known good values. It would also enable us to make decisions in the driver (if needed) based on the optimizer implementation.

I do see your point about the fact that is just passed to the solver directly. In that case maybe we can have a `solver_args` option that just forwards all specified parameters as is to the solver (kind of like a `-Z` flag for C/C++ compilers for specifying linker arguments). This might be useful if we need some "emergency" arguments on one of our deployments that the driver doesn't have support for.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -163,13 +163,14 @@ mod solver_input {
...
     fn write_input(&self, input_file: &str, input: &str) -> std::io::Result<()>;
-    fn run_solver(
+    fn run_solver<'a>(
         &self,
         input_file: &str,
         result_folder: &str,
         solver_type: SolverType,
         time_limit: Duration,
         min_avg_fee_per_order: u128,
+        internal_solver: Option<&'a str>,
```
nlordell, line 173: Can this be elided with `'_`?

nlordell (approved):

josojo (approved):

PR #860, opened 2020-05-12
[easy] Turn error into unwrap

### LINKED ISSUE #770: [event based orderbook] Add retry mechanism instead of aborting in UpdatingOrderbook

e00E:
We expect the error state to be impossible to reach so an unwrap is
cleaner.

Fixes #770 

### Test Plan
Run driver and see that orderbook still updates.


fleupold (approved):
Can we explain in the comment why we think there is "no way that the mutex can be poisoned"?

PR #862, opened 2020-05-12
Open-solver bump: better performance for min-avg-fee..

josojo:
better performance for min-avg-fee and better solution for edge cases


bh2smith (approved):
It seems there must have been a separate PR with the changes related to the "better performance". Could we link this here?

nlordell (approved):

fleupold:
For future PRs like this maybe you can link the release note (https://github.com/gnosis/dex-open-solver/releases/tag/v0.0.9 in this case) and make sure that when drafting a release it contains a relate-able message.

PR #863, opened 2020-05-13
Do not test rinkeby in CI

### LINKED ISSUE #845: Don't test rinkeby in CI
This makes CI slower and rinkeby can be down which then makes CI fail.

Imo it is cleaner to test only with ganache but in practice. On the other hand this could mean that we miss an error that the test would have shown but I don't remember the last time it happened. But I do remember the test failing for no reason a bunch of times.

e00E:
The rinkeby test is flaky because the network can be down. We decided to
no longer test as part of CI.
I have kept the configuration and test in the repository because there
can still be value in executing the test manually.

Fixes #845

### Test Plan
Check the CI output to see that we no longer do the test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,10 +47,6 @@ matrix:
         - docker-compose -f docker-compose.yml -f docker-compose.private-solver.yml up -d stablex
         - cargo test -p e2e ganache -- --nocapture
         - docker-compose logs
-        # StableX e2e Tests (Rinkeby)
-        - docker-compose down && docker-compose -f docker-compose.yml -f docker-compose.rinkeby.yml -f docker-compose.private-solver.yml up -d stablex
-        - cargo test -p e2e rinkeby -- --nocapture
-        - docker-compose logs
       deploy:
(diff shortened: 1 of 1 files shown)


josojo (approved):

nlordell (approved):

bh2smith (approved):
Fine by me. Do we still have the possibility to run rinkeby tests if we want to? Also interesting to see we are using docker-compose still even after eliminating the old snapp code.

fleupold (commented):
Can we make it so that we still run the private solver in the e2e tests somehow (e.g. on ganache in the other e2e tests). It would be nice to make sure that we can still invoke both the open and standard solver before allowing a merge.

e00E:
I'll do that in a separate PR.

e00E:
> Do we still have the possibility to run rinkeby tests if we want to?
Yes. You can still execute what the previous travis config was doing. Should also be explained in the e2e readme. This PR does'nt remove any code related to that.

PR #864, opened 2020-05-13
Use private solver instead of naive solver on travis

e00E:
Now that we have a public, open solver it is no longer important to test
that the naive solver works. This commit changes the ci test from the
naive solver to the private solver.
We might want to remove the naive solver code in the future or replace
it with something from pricegraph.

### Test Plan
Check the CI output to see that the standard solver is used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
         - docker-compose logs
         # Build and publish compact image with compiled binary
         - docker build --tag stablex-binary-private --build-arg SOLVER_BASE=163030813197.dkr.ecr.eu-central-1.amazonaws.com/dex-solver:v0.6.4 --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
-        # StableX e2e Tests (Ganache) - naive solver
+        # StableX e2e Tests (Ganache) - private solver
         - docker-compose down
         - ./scripts/setup_contracts.sh
         - docker-compose -f docker-compose.yml -f docker-compose.private-solver.yml up -d stablex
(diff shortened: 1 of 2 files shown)


josojo (commented):
When I incorporated the open solver, we said that either we should continue testing the naive solver or completely remove it from the codebase.

Since the open-solver is more and more mature, now my vote would be to remove the naive solver.

bh2smith:
My argument against removing the naive solver or even running our CI with this open solver instead is that we wanted something quick that would give us a valid end to end test that does not take too long to run. 

I am not strongly opposed to this, but just want to point out this fact that the naive solver is still the fastest solver we have.

e00E:
I'll make an issue for discussing whether we should keep the naive solver. Let's keep this PR only about changing the travis test to use the private solver instead of the naive solver.
I do not want to test *both* to keep the CI time low.

.travis.yml
```diff
@@ -41,7 +41,7 @@ matrix:
         - docker-compose logs
         # Build and publish compact image with compiled binary
         - docker build --tag stablex-binary-private --build-arg SOLVER_BASE=163030813197.dkr.ecr.eu-central-1.amazonaws.com/dex-solver:v0.6.4 --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
-        # StableX e2e Tests (Ganache) - naive solver
+        # StableX e2e Tests (Ganache) - private solver
```
bh2smith, line 44: Should this be "open solver" considering there is nothing private about it.

bh2smith (approved):

.travis.yml
e00E: This isn't the open solver, it's the "standard solver" which requires the private docker image.

fleupold (approved):
👍

PR #868, opened 2020-05-13
Always fail on overflow in ceiled_div

### LINKED ISSUE #861: test_ceiled_div_overflow_u128 fails on release mode
https://github.com/gnosis/dex-services/blob/320a118cea27cfd66ddddb62d304a5cd696b6df6/driver/src/util.rs#L96

This test expects to panic but it only panics in debug mode because rust only checks for integer overflow in debug builds. With `cargo test --release` the test fails because the function does not panic.

e00E:
The tests assumed that the math operation would always panic on
overflow. This is only true for debug builds. When running `cargo test
--release` the test would fail because it would not panic.

fixes #861

### Test Plan
Check that `cargo test` and `cargo test --release` work.

### DIFF
--- a/driver/src/util.rs
+++ b/driver/src/util.rs
@@ -2,20 +2,31 @@ use ethcontract::U256;
 use std::future::Future;
 
 pub trait CeiledDiv {
+    /// Panics on overflow.
     fn ceiled_div(&self, divisor: Self) -> Self;
 }
 
 impl CeiledDiv for u128 {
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Nice catch! Should we include some test that demonstrate this is the case or just continue to leave it untested?

e00E:
It is being tested it is just that the test behavior was different on debug vs release build.

nlordell (approved):

PR #873, opened 2020-05-14
[Pricegraph] Update pricegraph to use indexed orders

nlordell:
This PR updates the pricegraph to use indexed orders instead of having to keep track of the indices internally. This is done so that the internal order indices are correct regardless of whether the orders are pre-filtered.

### Test Plan

Updated unit tests


e00E (approved):
We should probably document in pricegraph what the encoding is, even if it is just a reference to dex-contracts.

PR #876, opened 2020-05-15
[Easy] Log options to single line

nlordell:
This PR changes the initial log message the prints the startup options used by the driver to be a single line. This is done to make it easier to find the startup logs, since our logging service splits each separate line into a separate log document, making it very hard to actually see what the command line arguments actually are.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -198,7 +198,7 @@ struct Options {
 fn main() {
     let options = Options::from_args();
     let (_, _guard) = logging::init(&options.log_filter);
-    info!("Starting driver with runtime options: {:#?}", options);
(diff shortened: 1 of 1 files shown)


e00E (approved):
I am okay with this because it is practical but conceptually we should fix logging and not change our output. Output with new lines makes more sense for humans reading stdout.

bh2smith (approved):

bh2smith:
Totally, they have become quite large!

nlordell:
> Output with new lines makes more sense for humans reading stdout.

Agreed.

> I am okay with this because it is practical but conceptually we should fix logging 

I'll ask devops.

nlordell:
CLosing as we can group the logs.

PR #884, opened 2020-05-18
Use 256 bit integer to represent balance instead of 128

e00E:
In the smart contract account balances are serialized as 256 bit
integers but our current implementation uses 128 bit integers instead
and asserts that balances are never larger than the maximum 128 bit
integer.
It is more correct to use the same number type which allows us to
represent all balances and gets rid of the panic.

### Test Plan
CI. Relevant code that decodes an auction element and serializes numbers to the solver is already being tested.

### DIFF
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -6,15 +6,18 @@ use crate::price_finding::price_finder_interface::{
 
 use anyhow::{anyhow, Context, Result};
 use chrono::Utc;
+use ethcontract::U256;
 use log::error;
 use serde::{Deserialize, Serialize};
 use serde_with::rust::display_fromstr;
 use std::collections::BTreeMap;
+use std::fmt::{Debug, Display};
 use std::fs::{create_dir_all, File};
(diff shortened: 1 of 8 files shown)


driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -25,7 +28,10 @@ use std::time::Duration;
 /// the serialization cannot be controlled.
 #[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Deserialize, Serialize)]
 #[serde(transparent)]
-pub struct Num(#[serde(with = "display_fromstr")] pub u128);
+pub struct Num<T>(#[serde(with = "display_fromstr")] pub T)
+where
+    T: Copy + Clone + Debug + Default + Eq + PartialEq + Display + FromStr,
```
nlordell, line 33: nit: built-in derive proc macros add appropriate trait bounds automatically:
```suggestion
    T: Display + FromStr,
```

nlordell (approved):
Awesome cleanup! Just one tiny nit.

bh2smith (approved):

PR #885, opened 2020-05-18
[Easy] Split up `install` section in CI for tests and coverage

e00E:
Not only are some sections not needed for the coverage test but they can
cause it to fail because clippy and rustfmt might not be available on
nightly.

### Test Plan
Check CI output to see that both jobs still work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,8 @@ dist: xenial
 language: rust
 
 install:
-  - rustup component add clippy rustfmt
-  - sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
-  - $(aws ecr get-login --no-include-email --region $AWS_REGION)
   - nvm install 10 && nvm alias default 10 && npm install -g yarn@latest && yarn --version
   - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack)
-  - ./scripts/setup_contracts.sh
 
 jobs:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -64,6 +62,9 @@ jobs:
 
     - name: Coverage
       rust: nightly
+      install:
+        - nvm install 10 && nvm alias default 10 && npm install -g yarn@latest && yarn --version
+        - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack)
```
nlordell, line 67: nit: Can this be part of the upper level `install` and then use `{before,after}_install` for the "Build and Local Testing" job specific steps (like installing clippy and rustfmt as well as doing the aws login)?

nlordell (approved):
:+1: , just one minor nit.

.travis.yml
```diff
@@ -9,12 +9,8 @@ dist: xenial
 language: rust
 
 install:
-  - rustup component add clippy rustfmt
-  - sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
-  - $(aws ecr get-login --no-include-email --region $AWS_REGION)
```
nlordell, line 14: Do we still need to remove the node-gyp directory?

nlordell (approved):
Thanks! Looks great

.travis.yml
e00E: Seemed like a one time thing that fixed itself after I deleted once and ran that through travis.

e00E: Still needed apparently. I want to investigate that in a separate PR.

.travis.yml
```diff
@@ -64,6 +65,9 @@ jobs:
 
     - name: Coverage
       rust: nightly
+      before_install:
+        # TODO vk: Not sure why this is needed but without it I get some errors related to gyp.
+        - rm -rf /home/travis/.cache/node-gyp
```
bh2smith, line 70: see this added here and removed immediately again in #888. Is this a temporary crutch?

PR #886, opened 2020-05-18
Use arbitrary size integers for event based orderbook

### LINKED ISSUE #750: [event based orderbook] Use BigInt for balances in event based orderbook
Nick has this idea. It would allow us to get rid of all (?) balance related overflow checks. Only when constructing the account states we would need to check that the number fits into a U256.
I want to look into this after the event based orderbook has been merged into master.
This BigInt library seems ok at first glance https://gitlab.com/tspiteri/rug .

e00E:
Currently we use 256 bit integers, like the smart contract to keep track
of account balances and how they change with trades.
This commit changes this to use arbitrary size integers. This has the
advantage of making the code simpler because we can get rid of some
error conditions and no longer need to keep track of positive and
negative balance changes separately.
This is also a disadvantage because the new code is less accurate: It
does not detect some invalid event chains that the old would have
detected.
Another disadvantage is that arbitrary sized integers by design allocate
and live on the heap which makes the new code slower. Although I do not
expect the change in speed to matter.

fixes #750 

### Test Plan
Existing unit tests test this code and I will test that the result still still matches the filtered orderbook.

### DIFF
--- a/driver/Cargo.toml
+++ b/driver/Cargo.toml
@@ -15,6 +15,7 @@ futures = { version = "0.3.5", features = ["compat"] }
 isahc = { version = "0.9.2", features = ["json"] }
 lazy_static = "1.4.0"
 log = "0.4.8"
+num = { version = "0.2", features = ["serde"] }
 prometheus = "0.9.0"
 rouille = "3.0.0"
 rustc-hex = "2.1.0"
--- a/driver/src/orderbook/streamed/balance.rs
+++ b/driver/src/orderbook/streamed/balance.rs
@@ -1,53 +1,25 @@
 use super::*;
 use anyhow::{anyhow, ensure, Result};
+use num::BigInt;
+use num::Zero as _;
 use serde::{Deserialize, Serialize};
 
-/// Balance change from a potential solution that might still get replaced by a better solution in
-/// the same batch.
-#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
-struct Proceeds {
-    batch_id: BatchId,
-    // Two fields instead of one I256 because although unlikely it could be possible to overflow a
-    // I256 with multiple trades in one batch.
-    increase: U256,
-    decrease: U256,
-}
-
-enum TradeType {
-    Sell,
-    Buy,
-}
-
-impl Proceeds {
-    fn get_field(&mut self, operation: TradeType) -> &mut U256 {
-        match operation {
-            TradeType::Sell => &mut self.decrease,
-            TradeType::Buy => &mut self.increase,
-        }
-    }
-}
-
 /// The balance of a token for a user.
-#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct Balance {
-    balance: U256,
+    balance: BigInt,
     deposit: Flux,
     withdraw: Flux,
-    proceeds: Proceeds,
+    proceeds: Flux,
 }
 
 impl Balance {
-    pub fn deposit(&mut self, amount: U256, batch_id: BatchId) -> Result<()> {
-        self.apply_existing_deposit_and_proceeds(batch_id)?;
+    pub fn deposit(&mut self, amount: U256, batch_id: BatchId) {
+        self.apply_existing_deposit_and_proceeds(batch_id);
         // Works like in the smart contract: If there is an existing deposit we override the
         // batch id and add to the amount.
         self.deposit.batch_id = batch_id;
-        self.deposit.amount = self
-            .deposit
-            .amount
-            .checked_add(amount)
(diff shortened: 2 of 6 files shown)


driver/Cargo.toml
```diff
@@ -15,6 +15,8 @@ futures = { version = "0.3.5", features = ["compat"] }
 isahc = { version = "0.9.2", features = ["json"] }
 lazy_static = "1.4.0"
 log = "0.4.8"
+num-bigint = { version = "0.2", features = ["serde"] }
+num-traits = "0.2"
```
nlordell, line 19: ubernit: I would just use the `num` crate, its fairly small and won't add much compilation time. It would also make dependabot's life easier at it has a hard time upgrading crates whose versions are tied together and I think will likely require manual intervention.

nlordell (approved):
Love it!

> This is also a disadvantage [...] Another disadvantage [...]

These are valid points. If you think we shouldn't merge this, then we can close the issue and keep the code as is.

e00E:
Overall it's an improvement and I do want to include it.

driver/src/orderbook/streamed/balance.rs
```diff
@@ -60,137 +32,124 @@ impl Balance {
         // withdraw request because the smart contract should have emitted a withdraw event for the
         // previous one first.
         ensure!(
-            self.withdraw.batch_id >= current_batch_id || self.withdraw.amount == U256::zero(),
+            self.withdraw.batch_id >= current_batch_id || self.withdraw.amount.is_zero(),
             "new withdraw request before clearing of previous withdraw request"
         );
         self.withdraw.batch_id = batch_id;
-        self.withdraw.amount = amount;
+        self.withdraw.amount = bigint_u256::u256_to_bigint(amount);
```
bh2smith, line 39: Wouldn't it be nice to implement from for this so that we can... `BigInt::from(amount)`

driver/src/orderbook/streamed/balance.rs
```diff
@@ -60,137 +32,124 @@ impl Balance {
...
+            result += deposit
         }
-        if self.proceeds.batch_id < current_batch_id {
-            balance = balance
-                .checked_add(self.proceeds.increase)
-                .ok_or_else(|| anyhow!("math overflow"))?
-                .checked_sub(self.proceeds.decrease)
-                .ok_or_else(|| anyhow!("math underflow"))?;
+        if let Some(proceed) = self.proceeds.amount(current_batch_id) {
+            result += proceed
```
bh2smith, line 118: I can see the benefit of this cleanup... Actually almost seems like it would have been easier to do it this way in the first place.

driver/src/orderbook/streamed/bigint_u256.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub fn u256_to_biguint(n: U256) -> BigUint {
+    let mut bytes = [0u8; 256 / 8];
+    n.to_little_endian(&mut bytes);
+    BigUint::from_bytes_le(&bytes)
+}
+
+pub fn u256_to_bigint(n: U256) -> BigInt {
+    BigInt::from_biguint(Sign::Plus, u256_to_biguint(n))
+}
```
bh2smith, line 21: Could any of these be a `From` or `TryFrom` implementation? Is there some reason you have decided against this?

driver/src/orderbook/streamed/state.rs
```diff
@@ -646,24 +647,54 @@ mod tests {
...
+            owner: address(3),
             order_id: 0,
-            executed_sell_amount: 4,
-            executed_buy_amount: 3,
+            executed_sell_amount: 2,
+            executed_buy_amount: 1,
             ..Default::default()
         };
-        state = state.apply_event(&Event::Trade(event), 1).unwrap();
+        state = state.apply_event(&Event::TradeReversion(event), 1).unwrap();
```
bh2smith, line 689: This seems like a strange thing to change. Was there something wrong with the test before?

bh2smith (approved):
Love this! Really nice improvement! Just wondering about `impl TryFrom<X> for Y` for the `BigInt <-> U256` and one little curious change in the unit test `Trade -> TradeReversion`.

driver/src/orderbook/streamed/state.rs
e00E: Before the test had trade reversions but no second solution submission which doesn't make sense. So I changed the test to revert and submit a new solution which increases coverage because it tests that the fee from the previous solution gets reverted properly.

driver/src/orderbook/streamed/bigint_u256.rs
e00E: It's not possible because in Rust either the trait or the type on which we implement the trait has to come from this module. In this case U256 and BigInt come from other crates and From comes from the standard library.

Issue #871, opened 2020-05-14
Update travis ubuntu and node version


e00E:
The current ubuntu version used is Xenial but we can switch to Bionic.
The current node version is 10 but we can switch to 12.

nlordell:
> The current ubuntu version used is Xenial but we can switch to Bionic.

I would just use the default OS, and we should probably explicitly set the NodeJS version to LTS.

e00E:
I don't see why we wouldn't use the never ubuntu release on principle. I expect in general for things to work better on the newer release. The only reason we are using Xenial after all is because that is what travis happens to have picked for us.

fleupold:
We used to build the rust binary inside travis and then copy it into the docker environment. Therefore our docker distribution and travis distribution need to be compatible (create the same binaries) which is why we pinned it., if I remember correctly

nlordell:
>  which is why we pinned it

As far as I can tell it is currently not pinned to a distribution.

e00E:
> We used to build the rust binary inside travis and then copy it into the docker environment.

This sounds like you expect that we no longer do this and instead rebuild in the container. This isn't the case from what I see in the build logs. We are still copying the binary over.

Here is the travis documentation for the build images:
https://docs.travis-ci.com/user/reference/xenial/
https://docs.travis-ci.com/user/reference/bionic/

There isn't a huge difference but I noticed that bionic comes with node 12 preinstalled so we wouldn't have to download it during build anymore.

e00E:
@fleupold Tagging so you see the above message.

fleupold:
My message was worded poorly, I was expecting we still do that today. 
I think the reason for running the tests natively in travis and not inside docker was so that it is a little less indirect and thus potentially easier to debug (also the docker image can really be minimal and doesn't have to install any rust or testing stuff).

PR #888, opened 2020-05-18
Update nodejs version to 12

e00E:
#871 

dex-contracts works with a newer version so we should switch.

Removes caching of node-gyp because it is causing problems on CI. We can
reintroduce it later again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,9 @@ dist: xenial
 language: rust
 
 install:
-  - nvm install 10 && nvm alias default 10 && npm install -g yarn@latest && yarn --version
+  - nvm install 12 && nvm alias default 12 && npm install -g yarn@latest && yarn --version
   - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack)
+  - ./scripts/setup_contracts.sh
 
 jobs:
   fast_finish: true
@@ -22,7 +23,6 @@ jobs:
       rust: stable
       cache:
         directories:
-          - $HOME/.cache/node-gyp
           - $HOME/.cache/yarn
           - $HOME/.cargo/bin
       before_cache:
@@ -33,7 +33,6 @@ jobs:
         - sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
         - $(aws ecr get-login --no-include-email --region $AWS_REGION)
       script:
-        - ./scripts/setup_contracts.sh
         - cargo build --all-targets --locked
         # Unit Tests and Linting
         - cargo test
@@ -65,9 +64,6 @@ jobs:
 
     - name: Coverage
       rust: nightly
-      before_install:
-        # TODO vk: Not sure why this is needed but without it I get some errors related to gyp.
-        - rm -rf /home/travis/.cache/node-gyp
       script:
         - curl --location https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
         # These flags are recommended by https://github.com/mozilla/grcov#grcov-with-travis .


nlordell:
We should use nodejs 12 LTS. There is a build error, but it is for an optional dependency and doesn't cause `yarn` to fail.

e00E:
Thanks, updated to 12.

bh2smith (approved):

bh2smith:
I approve the changes pending that the build passes. This will be an awesome improvement!

e00E:
Node-gyp is still causing problems. I am removing its caching here because I am not sure if we are doing that correctly. We can reintroduce that later again after understanding why it is failing better.

.travis.yml
```diff
@@ -9,8 +9,9 @@ dist: xenial
 language: rust
 
 install:
-  - nvm install 10 && nvm alias default 10 && npm install -g yarn@latest && yarn --version
-  - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack)
+  - nvm install 12 && nvm alias default 12 && npm install -g yarn@latest && yarn --version
+  # Yarn exits with non zero exit code because an optional dependency does not compile.
+  - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack || true)
```
bh2smith, line 14: This seems pretty intense... What if the install actually fails?

e00E: Then the `script` step which builds and tests our code is going to fail because it can't find the generated contract json artifacts.

bh2smith: Excellent reasoning!

nlordell: Weird.. for me yarn exits with code 0.

nlordell:
> Node-gyp is still causing build problems on CI. I am removing its caching here because I am not sure if we are doing that correctly. We can reintroduce that later again after understanding why it is failing better.

I think at some point we had a `before_cache` rule that fixed this.

.travis.yml
nlordell, line 14: This shouldn't be necessary... I just tested locally and:
```
$ yarn
...
gyp ERR! cwd /var/home/nlordell/Developer/dex-contracts/node_modules/scrypt
gyp ERR! node -v v12.16.2
gyp ERR! node-gyp -v v5.1.0
gyp ERR! not ok"
Done in 32.74s.
$ echo $?
0
```

.travis.yml
```diff
@@ -22,7 +23,6 @@ jobs:
       rust: stable
       cache:
         directories:
-          - $HOME/.cache/node-gyp
```
nlordell, line 25: So if I'm not mistaken, we started caching `node-gyp` because we would get intermittent failures where it wasn't properly initialized.

I imagine the issues with caching node-gyp are related to changing the NodeJS version?

I'll dig up the PR

nlordell: Here it is: https://github.com/gnosis/dex-services/pull/500

nlordell: Now that I think of it, maybe this was just an issue with NodeJS 10 that is fixed now. I'm all for removing this line for now and if the intermittent failures come back, investigate again.

.travis.yml
e00E: This doesn't seem to be the case on CI. I just tested it again and it definitely fails there without `||
true`.

nlordell: Hmm...

nlordell: Ah I figured out the issue! This is because the submodule version does not include the following commit, which is required for NodeJS 12 support:
https://github.com/gnosis/dex-contracts/commit/052ba1bc4f9c5467970ba671aca5f5e013cd6f9d

nlordell: Maybe we should first update the submodule to `v0.3.0` and then merge this PR. Otherwise I imagine it will not work with NodeJS 12.

e00E:
Working without `|| true` after the dex-contract upgrade. I noticed that coverage didn't work so I added a one line fix for that too: moving ./setup-scripts.sh back to to the global install section.

nlordell (approved):

PR #889, opened 2020-05-18
Test splitting order into two trades in the same solution

fedgiac:
Tests if there are any issues in case the same order is used for multiple trades in the same solution. No changes to the code are needed to make the test work.

### Test Plan
```
cargo test orderbook_solution_touching_same_order_twice
```

### DIFF
--- a/driver/src/orderbook/streamed/state.rs
+++ b/driver/src/orderbook/streamed/state.rs
@@ -787,4 +787,106 @@ mod tests {
         assert_eq!(balance.get(&(address(2), 0)), Some(&12.into()));
         assert_eq!(balance.get(&(address(2), 1)), Some(&9.into()));
     }
+
+    #[test]
+    fn orderbook_solution_touching_same_order_twice() {
+        let mut state = state_with_fee();
+        let event = TokenListing {
+            token: address(1),
+            id: 1,
+        };
+        state = state.apply_event(&Event::TokenListing(event), 0).unwrap();
+
+        let event = Deposit {
+            user: address(2),
+            token: address(0),
+            amount: 20.into(),
+            batch_id: 0,
+        };
+        state = state.apply_event(&Event::Deposit(event), 0).unwrap();
+        let event = Deposit {
+            user: address(3),
+            token: address(1),
+            amount: 20.into(),
+            batch_id: 0,
+        };
+        state = state.apply_event(&Event::Deposit(event), 0).unwrap();
+
+        let event = OrderPlacement {
+            owner: address(2),
+            index: 0,
+            buy_token: 1,
+            sell_token: 0,
+            valid_from: 0,
+            valid_until: 10,
+            price_numerator: 10,
+            price_denominator: 10,
+        };
+        state = state.apply_event(&Event::OrderPlacement(event), 0).unwrap();
+        let mut event = OrderPlacement {
+            owner: address(3),
+            index: 0,
+            buy_token: 0,
+            sell_token: 1,
+            valid_from: 0,
+            valid_until: 10,
+            price_numerator: 5,
+            price_denominator: 5,
+        };
+        state = state
+            .apply_event(&Event::OrderPlacement(event.clone()), 0)
+            .unwrap();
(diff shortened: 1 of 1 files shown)


bh2smith:
Really nice test plan!

driver/src/orderbook/streamed/state.rs
```diff
@@ -787,4 +787,106 @@ mod tests {
         assert_eq!(balance.get(&(address(2), 0)), Some(&12.into()));
         assert_eq!(balance.get(&(address(2), 1)), Some(&9.into()));
     }
+
+    #[test]
+    fn orderbook_solution_touching_same_order_twice() {
+        let mut state = state_with_fee();
```
bh2smith, line 793: Could we not initialize the state as we want it without having to apply all the token listings and deposits?

driver/src/orderbook/streamed/state.rs
```diff
@@ -787,4 +787,106 @@ mod tests {
...
+        state = state.apply_event(&Event::Trade(event), 1).unwrap();
+        let mut event = Trade {
+            owner: address(3),
+            order_id: 0,
+            executed_sell_amount: 4,
+            executed_buy_amount: 4,
+            ..Default::default()
+        };
+        state = state.apply_event(&Event::Trade(event.clone()), 1).unwrap();
+        event.order_id = 1;
```
bh2smith, line 860: What is the purpose of this? Such an order doesn't already exist with this index.

driver/src/orderbook/streamed/state.rs
```diff
@@ -787,4 +787,106 @@ mod tests {
...
+
+        let balance = state
+            .orderbook_for_batch(Batch::Current)
+            .unwrap()
+            .0
+            .collect::<HashMap<_, _>>();
+        assert_eq!(balance.get(&(address(2), 0)), Some(&20.into()));
+        assert_eq!(balance.get(&(address(2), 1)), Some(&0.into()));
+        assert_eq!(balance.get(&(address(3), 0)), Some(&0.into()));
+        assert_eq!(balance.get(&(address(3), 1)), Some(&20.into()));
```
bh2smith, line 871: ~~These balances are confusing me. Both users deposited 20 of one of the tokens. Then I see executed sell and buy  amounts are 4, occurring twice for each order. This tells me that only 8 of token 1 was swapped for token 0. Not sure what the fee is, but I expect the balances to be 12 and 8.~~

I see, this is a pre-acution settlement balance confirmation. Could we include a comment here saying that.

driver/src/orderbook/streamed/state.rs
```diff
@@ -787,4 +787,106 @@ mod tests {
...
+            sell_token: 1,
+            valid_from: 0,
+            valid_until: 10,
+            price_numerator: 5,
+            price_denominator: 5,
+        };
+        state = state
+            .apply_event(&Event::OrderPlacement(event.clone()), 0)
+            .unwrap();
+        event.index = 1;
```
fleupold, line 839: Why do we need two order placements here? It feels like one counter order should be enough.

bh2smith: How can you match an order without a counter order?

fedgiac: Yes, a single order would have been enough. I went for two orders because:
1. they serve as a contrast to what is tested in this PR (a single order split into two should be like two independent orders)
2. I see it slightly more likely to fail when the same order is matched against two different orders than when matched twice against the same order.

driver/src/orderbook/streamed/state.rs
fedgiac: This is the second order we are discussing [here](https://github.com/gnosis/dex-services/pull/889/files/b5bda99d59276eaf7156ba0721997d018b3595b0#r429781008).

bh2smith:
While we are commenting on this already merged PR again, I would like to re-voice my question that we could dramatically simplify this test by only having the two relevant trade events (that is, initializing a non-trivial state instance without having to make deposits etc...).

PR #891, opened 2020-05-18
Use python3-pip instead of python2

e00E:
I noticed warnings in CI about python2 hitting end of life. Following
the amazon documentation this commit changes the way we install the aws
sdk to python3. Also removes sudo from the pip command because it is not
needed.

### Test Plan
CI


bh2smith (approved):
Love this!

nlordell (approved):

PR #892, opened 2020-05-18
Refactor CI cargo commands

e00E:
* Turn the deprecated --all into --workspace.
* Move cargo fmt and cargo clippy before the build step to catch errors
earlier.

### Test Plan
CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,11 @@ jobs:
         - sudo apt-get update && sudo apt-get install -y python3-pip python3-setuptools && pip3 install --upgrade --user awscli
         - $(aws ecr get-login --no-include-email --region $AWS_REGION)
       script:
-        - cargo build --all-targets --locked
+        - cargo fmt --all -- --check
+        - cargo clippy --locked --workspace --all-targets --all-features -- -D warnings
+        - cargo build --locked --workspace --all-targets
         # Unit Tests and Linting
         - cargo test
-        - cargo clippy --all --all-targets -- -D warnings
-        - cargo fmt --all -- --check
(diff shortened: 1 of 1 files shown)


bh2smith (approved):

.travis.yml
```diff
@@ -27,16 +27,18 @@ jobs:
...
         - rustup component add clippy rustfmt
         - sudo apt-get update && sudo apt-get install -y python-pip && sudo pip install awscli
         - $(aws ecr get-login --no-include-email --region $AWS_REGION)
       script:
-        - cargo build --all-targets --locked
+        - cargo build --workspace --all-targets --locked
         # Unit Tests and Linting
         - cargo test
-        - cargo clippy --all --all-targets -- -D warnings
+        - cargo clippy --workspace --all-targets -- -D warnings
```
nlordell, line 41: is the `-D warnings` still needed here now that they are set with `RUSTFLAGS`?

nlordell (approved):
:tada: ! Just one small question.

nlordell:
Sorry! I hit the `Udate branch` by accident :frowning_face:

.travis.yml
e00E: I just double checked the clippy documentation https://github.com/rust-lang/rust-clippy#travis-ci :
> Note that adding -D warnings will cause your build to fail if any warnings are found in your code. That includes warnings found by rustc (e.g. dead_code, etc.).

That means we don't need the RUSTFLAG at all. I am going to remove it and move the clippy command above the build command because clippy is faster than build which catches errors earlier.

PR #893, opened 2020-05-19
Update ubuntu to bionic from xenial

### LINKED ISSUE #871: Update travis ubuntu and node version

e00E:
This is the newest LTS release travis supports. It comes with node 12
preinstalled.

fixes #871

### Test Plan
CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ notifications:
 if: (branch = master) OR (type = pull_request) OR (tag IS present)
 
 os: linux
-dist: xenial
+dist: bionic
 language: rust
 
 install:


nlordell:
I restarted the build as it seemed to fail when retrieving Ubuntu packages (hopefully intermittent error).

From what I remember, we used to build the driver binary and then copy it into a Docker image that already has the solver set up. We may run into libc incompatibilities if the solver Docker image is based on a different distribution than the Travis image.

e00E:
Looks the solver is also using bionic anyway https://github.com/gnosis/dex-open-solver/blob/35ea2920487795097b30c2339eef47029edff8ef/docker/Dockerfile#L1 .

nlordell (approved):

PR #895, opened 2020-05-19
Ignore orders with 0 price numerator or denominator

### LINKED ISSUE #894: pricegraph NaN panic
We saw this in the logs and I was able to reproduce it with fuzzing (PR for that soon).

Add this test to orderbook.rs : 

```rust
    #[test]
    fn nan_failure() {
        use crate::encoding::*;
        use primitive_types::{H160, U256};

        let elements = vec![
            Element {
                user: H160::zero(),
                balance: U256::from_dec_str(
                    "12715861048125203197891841262599926257840766446203714759525741717161676636160",
                )
                .unwrap(),
                pair: TokenPair { buy: 0, sell: 1 },
                valid: Validity { from: 0, to: 0 },
                price: Price {
                    numerator: 37218383890677207382744681053558142492,
                    denominator: 43657760336300720852568604,
                },
                remaining_sell_amount: 37363768194016619614905451585067810816,
                id: 0,
            },
            Element {
                user: H160::zero(),
                balance: U256::from_dec_str(
                    "12664765281704264833469460898684845727998501522769485522719856179532372122652",
                )
                .unwrap(),
                pair: TokenPair { buy: 0, sell: 1 },
                valid: Validity { from: 0, to: 0 },
                price: Price {
                    numerator: 0,
                    denominator: 0,
                },
                remaining_sell_amount: 13438235513639339688651946631354122240,
                id: 1,
            },
        ];

        let _ = Orderbook::from_elements(elements);
    }
```

> panicked at 'orders cannot have NaN prices'

nlordell:
This PR filters out order with invalid price fractions.

Closes #894

### Test Plan

Added new unit test.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -57,10 +57,7 @@ impl Orderbook {
         let mut orders = OrderCollector::default();
         let mut users = UserMap::default();
 
-        for element in elements
-            .into_iter()
-            .filter(should_include_auction_element)
-        {
+        for element in elements.into_iter().filter(should_include_auction_element) {
             let TokenPair { buy, sell } = element.pair;
             max_token = cmp::max(max_token, cmp::max(buy, sell));
             users.entry(element.user).or_default().set_balance(&element);
@@ -439,10 +436,10 @@ fn should_include_auction_element(element: &Element) -> bool {
     const MIN_AMOUNT_U256: U256 = U256([MIN_AMOUNT as _, 0, 0, 0]);
 
     let is_dust_order =
-        || element.remaining_sell_amount < MIN_AMOUNT_U128 || element.balance < MIN_AMOUNT_U256;
-    let has_valid_price = || element.price.numerator != 0 && element.price.denominator != 0;
+        element.remaining_sell_amount < MIN_AMOUNT_U128 || element.balance < MIN_AMOUNT_U256;
+    let has_valid_price = element.price.numerator != 0 && element.price.denominator != 0;
 
-    !is_dust_order() && has_valid_price()
+    !is_dust_order && has_valid_price
 }
 
 /// An error indicating that an operation over a path failed because of a
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -59,7 +59,7 @@ impl Orderbook {
 
         for element in elements
             .into_iter()
-            .filter(|e| should_include_auction_element(e))
+            .filter(should_include_auction_element)
         {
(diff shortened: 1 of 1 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -57,7 +57,10 @@ impl Orderbook {
         let mut orders = OrderCollector::default();
         let mut users = UserMap::default();
 
-        for element in elements.into_iter().filter(|e| !is_dust_order(e)) {
+        for element in elements
+            .into_iter()
+            .filter(|e| should_include_auction_element(e))
```
e00E, line 62: I predict clippy will complain that this should be

```suggestion
            .filter(should_include_auction_element)
```

e00E (approved):

pricegraph/src/orderbook.rs
```diff
@@ -425,12 +428,21 @@ fn format_path(path: &[NodeIndex]) -> String {
...
+///   than the minimum amount that the exchange allows for trades
+/// - They have a `0` price numerator or denominator
+fn should_include_auction_element(element: &Element) -> bool {
     const MIN_AMOUNT_U128: u128 = MIN_AMOUNT as _;
     const MIN_AMOUNT_U256: U256 = U256([MIN_AMOUNT as _, 0, 0, 0]);
 
-    element.remaining_sell_amount < MIN_AMOUNT_U128 || element.balance < MIN_AMOUNT_U256
+    let is_dust_order =
+        || element.remaining_sell_amount < MIN_AMOUNT_U128 || element.balance < MIN_AMOUNT_U256;
+    let has_valid_price = || element.price.numerator != 0 && element.price.denominator != 0;
```
e00E, line 443:
```suggestion
    let has_valid_price = element.price.numerator != 0 && element.price.denominator != 0;
```

pricegraph/src/orderbook.rs
nlordell: :see_no_evil: Nice catch.

pricegraph/src/orderbook.rs
nlordell: I tried to be cleaver and use closures to take advantage of short cutting but that doesn't really make things clearer. I'll remove it.

e00E: I missed the closure and function call below. I thought you had copy pasted this from the statement above. In this case the operation is so cheap that I also prefer the non closure version.

nlordell: Just checked in play ground and the compiler optimizes the let bindings away so it makes no difference and the closures only add confusion.

PR #896, opened 2020-05-19
Fuzzy tests

nlordell:
This PR introduces a new test module for integration tests discovered by the new fuzzy tester @e00E is working on. Tests that get discovered by the fuzzer can be placed into the new module.

### Test Plan

It is the test plan.

### DIFF
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -7,5 +7,5 @@ mod orderbook;
 #[path = "../data/mod.rs"]
 mod data;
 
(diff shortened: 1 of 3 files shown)


e00E (commented):
I don't think we need to include every fuzzer discovered test just for the sake of it. In this case I prefer keeping the test you manually wrote because it is shorter and there is no extra value added by the full fuzzer test.

nlordell:
It would just "prove" that the handwritten test is indeed equivalent to the fuzzer test, but I'm OK with removing this for now.

PR #897, opened 2020-05-19
Fuzz pricegraph

e00E:
Allow fuzzing pricegraph using https://github.com/rust-fuzz/cargo-fuzz . For people that don't know how fuzzing works I recommend reading the fuzzing book linked in that repo.
The fuzzer runs your code with automatically generated inputs and tries to find panics. This can surface unintentional panics like the NaN panic we just fixed. This works well with use of of `debug_assert` to check invariants.

I have split the PR up into 3 commits that can be read individually.

### Test Plan
Run the fuzz commands from the readme  `cargo +nightly fuzz run orderbook`. CI.

### DIFF
--- a/pricegraph/fuzz/.gitignore
+++ b/pricegraph/fuzz/.gitignore
@@ -1,0 +1,2 @@
+corpus
+artifacts
--- a/pricegraph/fuzz/Cargo.toml
+++ b/pricegraph/fuzz/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "pricegraph-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.3"
+pricegraph = { path = "..", features = ["arbitrary"] }
+
+[[bin]]
+name = "element_read_all"
+path = "fuzz_targets/element_read_all.rs"
+
+[[bin]]
+name = "orderbook"
+path = "fuzz_targets/orderbook.rs"
--- a/pricegraph/src/encoding.rs
+++ b/pricegraph/src/encoding.rs
@@ -21,6 +21,7 @@ pub type UserId = H160;
 
 /// A struct representing a buy/sell token pair.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct TokenPair {
     /// The buy token.
     pub buy: TokenId,
@@ -29,7 +30,8 @@ pub struct TokenPair {
 }
 
 /// A struct representing the validity of an order.
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Validity {
     /// The batch starting from which the order is valid.
     pub from: BatchId,
@@ -38,7 +40,8 @@ pub struct Validity {
 }
 
 /// A price expressed as a fraction of buy and sell amounts.
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Price {
     /// The price numerator, or the buy amount.
     pub numerator: u128,
@@ -47,7 +50,7 @@ pub struct Price {
 }
 
 /// An orderbook element that is retrieved from the smart contract.
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct Element {
     /// The user that placed the order.
     pub user: UserId,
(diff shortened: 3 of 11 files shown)


pricegraph/fuzz/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+
```
nlordell, line 1: nit: extra newline

pricegraph/fuzz/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "element_read_all"
+path = "fuzz_targets/element_read_all.rs"
+
+[[bin]]
+name = "orderbook"
+path = "fuzz_targets/orderbook.rs"
```
nlordell, line 26: nit: missing newline

pricegraph/src/lib.rs
```diff
@@ -7,5 +7,5 @@ mod orderbook;
 #[path = "../data/mod.rs"]
 mod data;
 
-pub use encoding::{TokenId, TokenPair};
+pub use encoding::{Element, TokenId, TokenPair};
```
nlordell, line 10: nit: We should probably export `Price` and `Validity` as well as they are needed for constructing `Element`.

nlordell (approved):
Really cool!

nlordell:
Should `pricegraph/fuzz` be added to the workspace?

e00E:
I am not sure. Cargo.toml got generated by `cargo fuzz init` and I kept it this way. What is the difference in practice? I think if it is not part of the workspace then it won't get built by CI so for that reason we should probably include it.

nlordell:
>  What is the difference in practice?

Shared lock file and target directory.

>  think if it is not part of the workspace then it won't get built by CI

I think its a good idea to build on CI, this ensures that nothing is broken with the fuzzer code as the price graph changes. Also, cargo only build default members, unless `--workspace` flag is provided.

pricegraph/fuzz/.gitignore
```diff
@@ -0,0 +1,4 @@
+
+target
```
nlordell, line 2: This should no longer be needed now that it is part of the workspace.

pricegraph/src/lib.rs
```diff
@@ -7,5 +7,5 @@ mod orderbook;
 #[path = "../data/mod.rs"]
 mod data;
 
-pub use encoding::{TokenId, TokenPair};
+pub use encoding::{BatchId, Element, OrderId, Price, TokenId, TokenPair, UserId, Validity};
```
nlordell, line 10: This might be an appropriate place for `::*`? I'm OK either way.

pricegraph/fuzz/.gitignore
```diff
@@ -0,0 +1,2 @@
+corpus
+artifacts
```
bh2smith, line 2:
```suggestion
artifacts

```

e00E: I just accepted it for the Cargo.toml but what is the point of newline at end of file? Neither for toml nor for .gitignore does it have any meaning.

bh2smith (approved):
I like this "derive arbitrary"! Very cool.

pricegraph/fuzz/.gitignore
nlordell: It’s for version control. Specifically, if you add a new element to the got ignore file, it will show up as 1 line changed and 1 line added. While this is not a huge deal it can interfere with automatic merges (in case, for example, the last element of the got ignore was modified in a separate PR).

fleupold (commented):
Really cool initiative! Are we planning on running this periodically as part of an automated system or should we run it locally when changing code in the pricegraph repo?

pricegraph/src/encoding.rs
```diff
@@ -121,6 +124,69 @@ impl Element {
...
 
+#[cfg(feature = "arbitrary")]
+mod abitrary_impl {
+    use super::*;
+    use arbitrary::{Arbitrary, Result, Unstructured};
+
+    // We want Element to implement Arbitrary but cannot derive it because Element contains H160
+    // and U256 which do not implement arbitrary and come from foreign crates. Instead of
+    // implementing Arbitrary manually or introducing wrappers for H160 and U256 we create an
+    // equivalent to Element which can derive Arbitrary and forward all methods.
```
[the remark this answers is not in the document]
bh2smith: Yea cool! but how will we remember to do it!? Is there a TODO here? or an issue somewhere?

pricegraph/src/encoding.rs
```diff
@@ -121,6 +124,69 @@ impl Element {
...
+        balance: [u8; 32],
+        pair: TokenPair,
+        valid: Validity,
+        price: Price,
+        remaining_sell_amount: u128,
+        id: OrderId,
+    }
+
+    impl ArbitraryElement {
+        fn from_element(e: &Element) -> Self {
```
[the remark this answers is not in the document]
e00E: This is only needed locally in one place. I find this cleaner as there is no need to involve the conversion machinery.
If it was for public, exported types then I'd use those traits.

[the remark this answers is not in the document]
fleupold: Out of curiosity, is there an overhead for using the conversion traits?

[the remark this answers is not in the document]
e00E: I don't expect there to be any runtime overhead but probably compile time overhead and mental overhead.

PR #899, opened 2020-05-20
Improve event based orderbook tests with macros

e00E:
This makes the tests more readable by making the event definitions
shorter and resembling human language more. We still show exactly the
same information as before when the events are created just in a
different format.

In another commit I am going to try turning the tests into a more
declarative style which would get ride of passing `state` to the macro.

### Test Plan
CI

### DIFF
--- a/driver/src/orderbook/streamed/state.rs
+++ b/driver/src/orderbook/streamed/state.rs
@@ -323,17 +323,122 @@ mod tests {
     use super::*;
     use crate::models::AccountState;
 
+    macro_rules! apply_event {
+        (to $state:ident for batch $batch:expr; TokenListing token $token:expr) => {
+            let event = TokenListing {
+                token: address($token),
+                id: $token,
+            };
+            $state = $state
+                .apply_event(&Event::TokenListing(event), $batch)
+                .unwrap();
+        };
+        (to $state:ident for batch $batch:expr; Deposit token $token:expr, to user $user:expr, amount $amount:expr) => {
+            let event = Deposit {
+                user: address($user),
+                token: address($token),
+                amount: U256::from($amount),
+                batch_id: $batch,
+            };
+            $state = $state.apply_event(&Event::Deposit(event), $batch).unwrap();
+        };
+        (to $state:ident for batch $batch:expr; WithdrawRequest from user $user:expr, of token $token:expr, amount $amount:expr, valid after batch $valid_from:expr) => {
+            let event = WithdrawRequest {
+                user: address($user),
+                token: address($token),
+                amount: U256::from($amount),
+                batch_id: $valid_from,
+            };
+            $state = $state
+                .apply_event(&Event::WithdrawRequest(event), $batch)
+                .unwrap();
+        };
+        (to $state:ident for batch $batch:expr; Withdraw from user $user:expr, of token $token:expr, amount $amount:expr) => {
+            let event = Withdraw {
+                user: address($user),
+                token: address($token),
+                amount: U256::from($amount),
+            };
+            $state = $state.apply_event(&Event::Withdraw(event), $batch).unwrap();
+        };
+        (to $state:ident for batch $batch:expr; OrderPlacement number $index:expr, from user $user:expr, selling $sell_amount:expr, of token $sell_token:expr, for at least $buy_amount:expr, of token $buy_token:expr, for batch interval [$valid_from:expr, $valid_until:expr]) => {
+            let event = OrderPlacement {
+                owner: address($user),
+                index: $index,
+                buy_token: $buy_token,
+                sell_token: $sell_token,
+                valid_from: $valid_from,
+                valid_until: $valid_until,
+                price_numerator: $sell_amount,
+                price_denominator: $buy_amount,
+            };
+            $state = $state
+                .apply_event(&Event::OrderPlacement(event), $batch)
+                .unwrap();
+        };
+        (to $state:ident for batch $batch:expr; Trade order number $index:expr, from user $user:expr, selling $sell_amount:expr, for $buy_amount:expr) => {
+            let event = Trade {
+                owner: address($user),
+                order_id: $index,
+                executed_sell_amount: $sell_amount,
+                executed_buy_amount: $buy_amount,
+                ..Default::default()
+            };
+            $state = $state.apply_event(&Event::Trade(event), $batch).unwrap();
+        };
+        (to $state:ident for batch $batch:expr; TradeReversion order number $index:expr, from user $user:expr, selling $sell_amount:expr, for $buy_amount:expr) => {
+            let event = TradeReversion {
+                owner: address($user),
+                order_id: $index,
+                executed_sell_amount: $sell_amount,
+                executed_buy_amount: $buy_amount,
+                ..Default::default()
+            };
+            $state = $state
+                .apply_event(&Event::TradeReversion(event), $batch)
+                .unwrap();
+        };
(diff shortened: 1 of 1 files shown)


driver/src/orderbook/streamed/state.rs
```diff
@@ -323,17 +323,122 @@ mod tests {
...
     }
 
     fn state_with_fee() -> State {
         let mut state = State::default();
-        let event = TokenListing {
-            token: address(0),
-            id: 0,
-        };
-        state = state.apply_event(&Event::TokenListing(event), 0).unwrap();
+        apply_event!(to state in batch 0, list token 0);
```
bh2smith, line 441: This is pretty strange to look at. I think mostly because `list token` is two separate words and the word `to` and `in` are they even anything?

bh2smith (commented):
The macros are cute and interesting, but I don't feel like this is something that makes the codebase easier to use. Feels to me like something people would learn to use and then forget what is actually happening. I like that unit tests show me how to construct instances of the structure in question. This PR would remove what I like.

Could use more convincing here. Is there anywhere left in the codebase where an event is handcrafted?

driver/src/orderbook/streamed/state.rs
e00E: This is strange if you're used to functions because it is a macro. I intentionally tried to pick something that reads more like a human language. The meaning of the words come entirely from the macro so the syntax here can be nearly anythign.

e00E:
> Is there anywhere left in the codebase where an event is handcrafted?

Yes, for example in orderbook.rs . But regardless this is not the goal of the tests. The event types, like any struct tell you how to create them by reading their definition or checking the docs (`cargo doc --open`). But we're testing `State` here and not the events.

I agree that this change is controversial. It is an experiment and in my opinion an improvement. The macro itself is readable and straightforward if you have encountered macros before. I tried to avoid using "magic" to help understanding of what is happening. All of the arguments to the macro get transformed into the structures in a clear way.

We increase complexity by using the macro but the tests themselves end up more readable. I like how much shorter they are and that they show me the relevant information with less noise than long struct definitions. The idea for this change came to me out of annoyance with the tests before.

I'm not overly attached to this change so we should try to find out whether this is worth it or not.

bh2smith:
> I agree that this change is controversial. It is an experiment and in my opinion an improvement. The macro itself is readable and straightforward if you have encountered macros before. I tried to avoid using "magic" to help understanding of what is happening. All of the arguments to the macro get transformed into the structures in a clear way.

> We increase complexity by using the macro but the tests themselves end up more readable. I like how much shorter they are and that they show me the relevant information with less noise than long struct definitions. The idea for this change came to me out of annoyance with the tests before.

Very nice presentation and argument for the changes. I am no swayed in favour. Thanks for going the extra mile to make these tests more readable. We should probably leave this change open a while for other to see it as well.

driver/src/orderbook/streamed/state.rs
```diff
@@ -323,17 +323,122 @@ mod tests {
     use super::*;
     use crate::models::AccountState;
 
+    macro_rules! apply_event {
+        (to $state:ident in batch $batch:expr, list token $token:expr) => {
```
nlordell, line 327: nit: 
```suggestion
        (to $state:ident for batch $batch:expr, list token $token:expr) => {
```

nlordell, line 327: ubernit: 
```suggestion
        (to $state:ident for batch $batch:expr; TokenListing token $token:expr) => {
```
- I would `s/in batch/for batch/` as it sounds a little nicer for me.
- I slightly prefer a semi-colon separating the batch value and the event information, this is just so there is a clearer separation between the state and batch information and the event parameters. Also since commas are used elsewhere in the macro, it makes it easier at a quick glance to separate the two parts.
- I would also slightly prefer to use the event name instead of a verb. For me this just removes some of the cognitive load in mapping event keywords to actual events.

nlordell (approved):
:+1: This is really nice and makes the tests much easier to read! Also I really like how it just adds some sugar for applying events without hiding important information.

driver/src/orderbook/streamed/state.rs
```diff
@@ -323,17 +323,122 @@ mod tests {
...
+                ..Default::default()
+            };
+            $state = $state
+                .apply_event(&Event::SolutionSubmission(event), $batch)
+                .unwrap();
+        };
+    }
+
+    macro_rules! assert_balance {
+        (in $state:ident batch $batch:expr, user $user:expr, has token $token:expr, balance $balance:expr) => {
```
nlordell, line 414: ubernit:
```suggestion
        (in $state:ident for batch $batch:expr; user $user:expr, has token $token:expr, balance $balance:expr) => {
```
- I think `for` here makes the sentence read nicer
- the `;` is just for consistency with the `apply_event!` macro... although I'm not sure it makes it better... (just a something to try :))

driver/src/orderbook/streamed/state.rs
```diff
@@ -323,17 +323,122 @@ mod tests {
...
+            assert_eq!(
+                account_state(&$state, $batch).read_balance($token, address($user)),
+                U256::from($balance)
+            );
+        };
+    }
+
+    macro_rules! assert_used_amount {
+        (in $state:ident batch $batch:expr, of order number $index:expr, from user $user:expr, is $amount:expr) => {
```
nlordell, line 423: ubernit:
```suggestion
        (in $state:ident for batch $batch:expr; of order number $index:expr, from user $user:expr, is $amount:expr) => {
```

Same as above

nlordell:
> I don't feel like this is something that makes the codebase easier to use. 

To me the macros are made to make testing easier, not to make using the actual orderbook types easier. The point is now when you read the test:
```
        apply_event!(to state in batch 0, deposit token 0, to user 1, amount 2);
            user: address(1),
        apply_event!(to state in batch 0, withdraw request from user 1, of token 0, amount 2, valid after batch 0);
            token: address(0),
        apply_event!(to state in batch 1, withdraw from user 1, of token 0, amount 1);
            amount: 2.into(),
        assert_balance!(in state batch 1, user 1, has token 0, balance 1);
```

It makes it very clear the events that are being simulated and the expected account state that is we are supposed to end up with. I find the macros are really just there for shortening how the test is written and not hiding any details about the events themselves.

driver/src/orderbook/streamed/state.rs
bh2smith: I like these suggestions, especially the last one!

bh2smith (approved):

fleupold (approved):
I also like how this change improves readability. It's a shame that autocomplete doesn't help you fill out the macros (whereas for methods it would tell you the arguments and order). I feel like in practice we will have to copy a bunch of premade statements around and change to our desired parameters.

I'd slightly prefer if the macros read more like a sentence,e.g.

```
assert_balance!(user 3 has 1 of token 0 in state for batch 1);
```
which might make remembering the order a bit more intuitive (but that's subjective). Not sure if its worth rewriting things.

e00E:
I prefer having the `in state for batch` part at the front and both are valid english sentences so I am going to keep it like this.
We can't get rid of `,` and `;` because the macro system needs them to separate expressions.

nlordell:
> We can't get rid of , and ; because the macro system needs them to separate expressions.

If we restrict the macro to only accept literals or a single token (`:tt`) in some places where it makes sense, then you might be able to remove some of them.

On the other hand, I also like starting with `in state for batch` for consistency with the `apply_event` macro.

Issue #843, opened 2020-05-04
Make more things async


e00E:
Currently we have some mix of blocking code and code that returns futures. It would be cleaner if everything was async (returned futures).

Changing this is somewhat annoying because Rust does not yet support async in traits so all the functions have to return `BoxFuture`.

nlordell:
I'd be up to making this change in `ethcontract` as well.

e00E:
What is left related to this:
Make the top level async: main.rs and the schedulers.
For this we'll want an executor probably tokio or asyncstd to use their task abstraction. I prefer async-std.
When we have that we can change threaded price source to use a Task instead of a background thread.

nlordell:
>  I prefer async-std.

Just curious about your reasons for this (not that I disagree).

e00E:
The main reason is that when I tested tokio and async-std, async-std futures were able to run in all executors (async-std, futures::executor::block_on, tokio) but tokio futures only ran in tokio (panic otherwise).
I can't formulate it exactly but reading docs, blogs, github issues made me like async-std.

nlordell:
> but tokio futures only ran in tokio

Yes, I think it's because tokio futures need to be driven by a tokio reactor (which gets started by the tokio executor).

Ultimately, `async-std` is more about providing an out-of-the box async version of the `std` library which is think is more what we need for this project. I imagine that the price estimator might benefit from tokio for the HTTP service.

PR #902, opened 2020-05-22
Make stablex_contract async

e00E:
One step for #843 .

The diff becomes shorter when hiding whitespace changes.

### Test Plan
CI

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -4,19 +4,21 @@
 use crate::{
     contracts,
     models::{ExecutedOrder, Solution},
-    util::FutureWaitExt,
 };
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Error, Result};
 use ethcontract::{
     contract::Event,
     errors::{ExecutionError, MethodError},
+    prelude::Web3,
     transaction::{confirm::ConfirmParams, GasPrice, ResolveCondition},
-    web3::{futures::Future as _, types::Block},
-    Address, BlockNumber, PrivateKey, U256,
+    transport::DynTransport,
+    web3::types::Block,
+    Address, BlockNumber, PrivateKey, H256, U256,
 };
 use futures::{
-    future::{BoxFuture, FutureExt},
-    stream::{BoxStream, StreamExt},
+    compat::Future01CompatExt as _,
+    future::{BoxFuture, FutureExt as _},
+    stream::{BoxStream, StreamExt as _},
 };
 use lazy_static::lazy_static;
 #[cfg(test)]
@@ -40,11 +42,11 @@ pub struct StableXContractImpl {
 }
 
 impl StableXContractImpl {
-    pub fn new(web3: &contracts::Web3, key: PrivateKey, network_id: u64) -> Result<Self> {
+    pub async fn new(web3: &contracts::Web3, key: PrivateKey, network_id: u64) -> Result<Self> {
         let defaults = contracts::method_defaults(key, network_id)?;
 
-        let viewer = BatchExchangeViewer::deployed(&web3).wait()?;
-        let mut instance = BatchExchange::deployed(&web3).wait()?;
+        let viewer = BatchExchangeViewer::deployed(&web3).await?;
+        let mut instance = BatchExchange::deployed(&web3).await?;
         *instance.defaults_mut() = defaults;
 
         Ok(StableXContractImpl { instance, viewer })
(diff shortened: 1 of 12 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -4,17 +4,19 @@
...
 use ethcontract::{
     contract::Event,
     errors::{ExecutionError, MethodError},
+    prelude::Web3,
     transaction::{confirm::ConfirmParams, GasPrice, ResolveCondition},
-    web3::{futures::Future as _, types::Block},
-    Address, BlockNumber, PrivateKey, U256,
+    transport::DynTransport,
+    web3::types::Block,
+    Address, BlockNumber, PrivateKey, H256, U256,
```
nlordell, line 16: I think I like this new way of importing more and more... it makes it easier to see what comes from what crate at a glance.

driver/src/contracts/stablex_contract.rs
```diff
@@ -77,15 +79,15 @@ pub struct FilteredOrderPage {
 pub trait StableXContract {
     /// Retrieve the current batch ID that is accepting orders. Note that this
     /// is always `1` greater than the batch ID that is accepting solutions.
-    fn get_current_auction_index(&self) -> Result<u32>;
+    fn get_current_auction_index(&self) -> BoxFuture<'static, Result<u32>>;
```
nlordell, line 82: nit: Can this be '_? The reason I say this is that it makes the bounds less restrictive and is more resistant to changes in the implementation (which may end up borrowing self over the course of the `Future`.

driver/src/contracts/stablex_contract.rs
```diff
@@ -139,40 +141,38 @@ pub trait StableXContract {
 }
 
 impl StableXContract for StableXContractImpl {
-    fn get_current_auction_index(&self) -> Result<u32> {
-        let auction_index = self.instance.get_current_batch_id().call().wait()?;
-        Ok(auction_index)
+    fn get_current_auction_index(&self) -> BoxFuture<'static, Result<u32>> {
+        let future = self.instance.get_current_batch_id().call();
```
nlordell, line 145: Is the `let future = ...` outside the `async` block because of the `'static` lifetime?

driver/src/contracts/stablex_contract.rs
```diff
@@ -139,40 +141,38 @@ pub trait StableXContract {
...
+        async move { future.await.map_err(Error::from) }.boxed()
```
nlordell, line 146: Is there a reason you prefer this over [`TryFutureExt::map_err`](https://docs.rs/futures/0.3.5/futures/future/trait.TryFutureExt.html#method.map_err)?

nlordell (commented):
I think I would prefer not making the resulting futures be `BoxFuture<'static, _>`. This makes the traits require that the `ethcontract` implementations not require the contract to be borrowed for the future. While this is currently the case, it will most probably not be as it moves to `async/await`.

driver/src/contracts/stablex_contract.rs
```diff
@@ -40,11 +42,11 @@ pub struct StableXContractImpl {
...
 
 impl StableXContractImpl {
-    pub fn new(web3: &contracts::Web3, key: PrivateKey, network_id: u64) -> Result<Self> {
+    pub async fn new(web3: &contracts::Web3, key: PrivateKey, network_id: u64) -> Result<Self> {
         let defaults = contracts::method_defaults(key, network_id)?;
 
-        let viewer = BatchExchangeViewer::deployed(&web3).wait()?;
-        let mut instance = BatchExchange::deployed(&web3).wait()?;
+        let viewer = BatchExchangeViewer::deployed(&web3).await?;
+        let mut instance = BatchExchange::deployed(&web3).await?;
```
bh2smith, line 49: So I see that here we changed `wait()` to `await` but lower down we are putting `.wait()` on things? Is this intended?

driver/src/contracts/stablex_contract.rs
e00E: It can't be `'_` (doesn't compile in the trait) but it can be `fn foo<'a>(&'a self) -> BoxFuture<'a, Result<...>>` which makes the most sense to me.

driver/src/contracts/stablex_contract.rs
e00E: I definitely can't put it in the async block because of compilation errors. Probably because of 'static and will go away by changing to &'a self.

driver/src/contracts/stablex_contract.rs
e00E: No. Just didn't know about that.

driver/src/contracts/stablex_contract.rs
e00E: There is no `wait()` in this file. If you mean other files then it is intended everywhere where the function containing that code is not itself async. For example in tests it will stay this way and in other places we are going to push the async forward in following PRs.
If there is ever a `wait()` inside an `async fn` then that is a mistake.

driver/src/contracts/stablex_contract.rs
nlordell: That is what I hoped the elided `'_` would mean, but I guess it doesn't work in this particular case.

driver/src/contracts/stablex_contract.rs
e00E: Seeing as I want to have an async block anyway to make it look like this is an async function I am keeping it like this.

nlordell (approved):
Awesome! :tada:

bh2smith (approved):

Issue #758, opened 2020-04-22
[event based orderbook] Correct approach for partial solutions


e00E:
This is needed when we want to start solving optimistically.

Maybe the correct approach is to not include proceeds if a solution is pratially applied instead of erroring?

_Originally posted by @nlordell in https://github.com/gnosis/dex-services/pull/755_

https://github.com/gnosis/dex-services/pull/903/files#r429848905

nlordell:
I think this is no longer an issue. Since we now use `eth_getLogs` to retrieve events, solutions emit all events in a single block, and we apply all events atomically, I don't think we will ever be in a situation where a solution will be partially applied.

I think the TODO just needs to be cleaned up, and maybe assert that we are never in a partially applied solution at the end of applying events.

PR #903, opened 2020-05-22
Make StableXOrderBookReading trait async

e00E:
One step for #843 .

### Test Plan
CI. Tested locally that event based orderbook doesn't deadlock.

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -3,6 +3,7 @@ use crate::models::{account_state::AccountState, order::Order, Solution};
 use crate::orderbook::StableXOrderBookReading;
 use crate::price_finding::PriceFinding;
 use crate::solution_submission::{SolutionSubmissionError, StableXSolutionSubmitting};
+use crate::util::FutureWaitExt as _;
 use anyhow::{Error, Result};
 use ethcontract::U256;
 use log::{info, warn};
@@ -43,7 +44,10 @@ impl<'a> StableXDriverImpl<'a> {
     }
 
     fn get_orderbook(&self, batch_to_solve: U256) -> Result<(AccountState, Vec<Order>)> {
-        let get_auction_data_result = self.orderbook_reader.get_auction_data(batch_to_solve);
+        let get_auction_data_result = self
+            .orderbook_reader
+            .get_auction_data(batch_to_solve)
+            .wait();
         self.metrics
             .auction_orders_fetched(batch_to_solve, &get_auction_data_result);
         get_auction_data_result
@@ -188,6 +192,7 @@ mod tests {
     use crate::solution_submission::MockStableXSolutionSubmitting;
     use crate::util::test_util::map_from_slice;
     use anyhow::anyhow;
+    use futures::future::FutureExt as _;
     use mockall::predicate::*;
     use std::thread;
 
@@ -232,7 +237,7 @@ mod tests {
             .with(eq(batch))
             .return_once({
                 let result = (state.clone(), orders.clone());
-                |_| Ok(result)
+                |_| async { Ok(result) }.boxed()
             });
 
         submitter
@@ -269,7 +274,7 @@ mod tests {
 
         reader
             .expect_get_auction_data()
-            .returning(|_| Err(anyhow!("Error")));
+            .returning(|_| async { Err(anyhow!("Error")) }.boxed());
 
         let driver = StableXDriverImpl::new(&pf, &reader, &submitter, &metrics);
 
@@ -294,7 +299,7 @@ mod tests {
             .with(eq(batch))
             .return_once({
                 let result = (state, orders);
-                |_| Ok(result)
+                |_| async { Ok(result) }.boxed()
             });
 
         submitter
@@ -331,7 +336,7 @@ mod tests {
         reader
             .expect_get_auction_data()
             .with(eq(batch))
-            .return_once(move |_| Ok((state, orders)));
+            .return_once(move |_| async { Ok((state, orders)) }.boxed());
 
         let driver = StableXDriverImpl::new(&pf, &reader, &submitter, &metrics);
         assert!(driver.run(batch, time_limit).is_ok());
@@ -355,7 +360,7 @@ mod tests {
             .with(eq(batch))
             .return_once({
                 let result = (state.clone(), orders.clone());
-                |_| Ok(result)
+                |_| async { Ok(result) }.boxed()
             });
 
         let solution = Solution::trivial();
@@ -387,7 +392,7 @@ mod tests {
             .with(eq(batch))
             .return_once({
                 let result = (state.clone(), orders.clone());
-                |_| Ok(result)
+                |_| async { Ok(result) }.boxed()
             });
 
         submitter
@@ -433,7 +438,7 @@ mod tests {
             .with(eq(batch))
             .return_once({
                 let result = (state.clone(), orders.clone());
-                |_| Ok(result)
+                |_| async { Ok(result) }.boxed()
             });
 
         submitter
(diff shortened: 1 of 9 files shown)


driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -77,28 +79,23 @@ impl UpdatingOrderbook {
...
-    fn do_with_context<T>(&self, callback: impl FnOnce(&mut Context) -> Result<T>) -> Result<T> {
-        // We unwrap because there is no way the mutex can be poisoned. If the mutex somehow was
-        // poisoned anyway against our expectation then panicking is correct because we cannot
-        // recover from this.
-        // The mutex cannot be poisoned because `do_with_context` is the only function accessing
-        // the mutex and is only called within this file. The `callback`s we use do not panic.
-        let mut context_guard = self.context.lock().expect("poison error");
+    /// Ensure that the orderbook has been initialized and updated.
+    /// Post condition on success: self.context is Some(_).
+    async fn update_context(&self) -> Result<()> {
```
e00E, line 84: Commenting here to highlight that this file had a logic change. We moved to an async Mutex and I had to change this function to not use a callback because I couldn't get the lifetimes to work out in `get_auction_data` below.
The callback had the advantage that the types made it clear that Context had been initialized which is now explained in this function.

driver/src/driver/stablex_driver.rs
```diff
@@ -3,6 +3,7 @@ use crate::models::{account_state::AccountState, order::Order, Solution};
 use crate::orderbook::StableXOrderBookReading;
 use crate::price_finding::PriceFinding;
 use crate::solution_submission::{SolutionSubmissionError, StableXSolutionSubmitting};
+use crate::util::FutureWaitExt as _;
```
nlordell, line 6: Ahh... I keep on forgetting to use this...

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -334,7 +342,7 @@ mod tests {
         orderbook.handle_event_data(withdraw_request, 1, 0, H256::zero(), 0);
         orderbook.handle_event_data(token_listing_0, 0, 0, H256::zero(), 0);
 
-        let auction_data = orderbook.get_auction_data(2.into()).unwrap();
+        let auction_data = orderbook.get_auction_data(2.into()).wait().unwrap();
```
nlordell, line 345: Maybe for tests we should use [`FutureExt::now_or_never`](https://docs.rs/futures/0.3.5/futures/future/trait.FutureExt.html#method.now_or_never)?

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell, line 84: Is there a reason we no longer use `do_with_context`? I thought it was a nice abstraction. The function signature would be a bit clunky though:
```
async fn do_with_context<T, F, Fut>(&self, callback: F) -> Result<T>
where
    F: FnOnce(&mut Context) -> Fut,
    Fut: Future<Output = Result<T>>,
{
```

nlordell (approved):
Looks great! Only comment is about refactoring `do_with_context`, I thought it was a nice abstraction.

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: **edit**: I see your comment now.

nlordell: Hmm, this is a tough one. It looks like if you declare a `Fut` generic as I did above, you get compile errors because `Fut` needs to live as long as the `do_with_context` call, but it can only live as long as `&mut Context` parameter to the function. Usually this is solved with higher rank trait lifetime bounds, but that only works for the `F` type :cry: 

One way to work around this is require the callback to be a boxed future:
```rust
    async fn do_with_context<T, F>(&self, callback: F) -> Result<T>
    where
        F: for<'ctx> FnOnce(&'ctx mut Context) -> BoxFuture<'ctx, Result<T>>,
    {
        let mut context_guard = self.context.lock().await;
        match context_guard.as_mut() {
            Some(context) => {
                self.update_with_events(context).await?;
                callback(context).await
            }
            None => {
                let mut context = Context {
                    orderbook: Orderbook::default(),
                    last_handled_block: 0,
                    block_timestamp_reader: CachedBlockTimestampReader::new(self.web3.clone()),
                };
                self.load_orderbook_from_file(&mut context);
                self.update_with_events(&mut context).await?;
                let result = callback(&mut context).await;
                *context_guard = Some(context);
                result
            }
        }
    }

```

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -77,28 +79,23 @@ impl UpdatingOrderbook {
...
+        let mut context_guard = self.context.lock().await;
         let context_optional: &mut Option<Context> = &mut context_guard;
         match context_optional {
```
nlordell, line 87: ubernit:
```suggestion
        match context_guard.as_mut() {
```

driver/src/orderbook/streamed/updating_orderbook.rs
nlordell: @e00E Not sure if you saw this comment

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: Doesn't work because we're writing to `*context_optional` below.

nlordell: `MutexGuard` implements `DerefMut` só you can write to it directly.

driver/src/orderbook/streamed/updating_orderbook.rs
e00E: Thanks. I am using this now.

driver/src/orderbook/streamed/orderbook.rs
```diff
@@ -135,17 +136,23 @@ impl TryFrom<&Path> for Orderbook {
...
-        let (account_state, orders) =
-            state.orderbook_for_batch(Batch::Future(batch_id_to_solve.low_u32() + 1))?;
-        let (account_state, orders) = util::normalize_auction_data(account_state, orders);
-        Ok((account_state, orders))
+    fn get_auction_data<'a>(
+        &'a self,
+        batch_id_to_solve: U256,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>> {
+        async move {
+            // TODO: Handle future batch ids for when we want to do optimistic solving.
```
bh2smith, line 144: Could we create an issue to go along with this TODO!

driver/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -192,16 +194,15 @@ impl UpdatingOrderbook {
...
+        &'a self,
+        batch_id_to_solve: U256,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>> {
+        self.do_with_context(move |context| context.orderbook.get_auction_data(batch_id_to_solve))
+            .boxed()
     }
-    fn initialize(&self) -> Result<()> {
-        self.do_with_context(|_| Ok(()))
+    fn initialize<'a>(&'a self) -> BoxFuture<'a, Result<()>> {
+        // self.update_context().boxed()
```
bh2smith, line 205: Is this commented bit meant to be here?
```suggestion
```

bh2smith (approved):
This must have been tricky! Great work!

driver/src/orderbook/streamed/orderbook.rs
e00E: Added a link to #758

PR #904, opened 2020-05-22
More efficient searching for last block of batch

### LINKED ISSUE #832: Find last block in batch with incrementing step size
Currently we search the last block of a batch by decrementing the block index one at a time.

https://github.com/gnosis/dex-services/blob/6f794300512e6332e1a80b4ad3ad6700f89fccf1/driver/src/contracts/stablex_contract.rs#L170-L173

A more optimal solution (and fun to implement) would probably be something like: https://www.geeksforgeeks.org/find-position-element-sorted-array-infinite-numbers/

fedgiac:
Fixes #832.

Implements a faster algorithm to determined the last block of a batch based on binary search. I'd be interested in some early comments, code is ready but untested.

### Test Plan
Will be provided when PR is ready.

### DIFF
--- a/driver/src/contracts/stablex_contract.rs
+++ b/driver/src/contracts/stablex_contract.rs
@@ -162,17 +162,63 @@ impl StableXContract for StableXContractImpl {
                 .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
         };
 
-        let mut current_block = get_block(BlockNumber::Latest)?;
-        let mut block_number = current_block
+        struct Bounds {
+            lower: u64,
+            upper: u64,
+        }
+        impl Bounds {
+            fn diff(&self) -> u64 {
+                self.upper - self.lower
+            }
+            fn mid(&self) -> u64 {
+                self.diff() / 2
+            }
+        }
+
+        let get_bounds_for_binary_search =
+            |current_block: u64, target_batch_id: u32| -> Result<_> {
+                let mut step = 1;
+                let mut lower_bound = current_block;
+                let mut upper_bound = current_block;
+                let mut lower_block = get_block(lower_bound.into())?;
+                while target_batch_id > get_block_batch_id(&lower_block) {
+                    upper_bound = lower_bound;
+                    if step >= lower_bound {
+                        lower_bound = 0;
+                        break;
+                    } else {
+                        lower_bound -= step;
+                        lower_block = get_block(lower_bound.into())?;
+                    }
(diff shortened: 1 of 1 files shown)


driver/src/contracts/stablex_contract.rs
```diff
@@ -162,17 +162,63 @@ impl StableXContract for StableXContractImpl {
                 .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
         };
 
-        let mut current_block = get_block(BlockNumber::Latest)?;
-        let mut block_number = current_block
+        struct Bounds {
```
fleupold, line 165: Can we use https://doc.rust-lang.org/std/ops/struct.Range.html ?

fedgiac: It has everything I need except for `.mid()`, but I think it would be a bit of a overhead, especially since I introduced this struct only for readability.

driver/src/contracts/stablex_contract.rs
```diff
@@ -162,17 +162,63 @@ impl StableXContract for StableXContractImpl {
...
+        impl Bounds {
+            fn diff(&self) -> u64 {
+                self.upper - self.lower
+            }
+            fn mid(&self) -> u64 {
+                self.diff() / 2
+            }
+        }
+
+        let get_bounds_for_binary_search =
```
bh2smith, line 178: Maybe this is personal preference, but I would call this `binary_search_bounds` (or anything without `get_` for that matter.

driver/src/contracts/stablex_contract.rs
nlordell: I prefer having a `Bounds` struct as well, especially since `Range` is intended as an iterator (it is, in fact, what you create with `1..100`).

driver/src/contracts/stablex_contract.rs
```diff
@@ -162,17 +162,63 @@ impl StableXContract for StableXContractImpl {
...
+                    }
+                    step *= 2;
+                }
+                Ok(Bounds {
+                    lower: lower_bound,
+                    upper: upper_bound,
+                })
+            };
+
+        let get_last_block_for_batch_within_bounds =
```
nlordell, line 201: Interesting use of a closure! Not used to seeing it in Rust this but it works nicely here.

bh2smith:
Why didn't you just change the code here in this existing branch?

PR #908, opened 2020-05-25
Only enable logging for driver

e00E:
Currently we enable info logging globally which includes code from our
dependencies. I noticed this becaus new log messages related to
isahc started appearing ("send_async").
This commit changes our default logging to enable only the driver
module. I do not think we intentionally enabled logging from our
dependencies before.

For our deployment this will be a separate change because we override the logging options there.

### Test Plan
Before this change:
Running the driver prints out "send_async" log messages.
After this change:
Running the driver does not print out "send_async" log messages but still contains other log messages.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -55,7 +55,7 @@ struct Options {
     /// The log filter to use.
     ///
     /// This follows the `slog-envlogger` syntax (e.g. 'info,driver=debug').
-    #[structopt(long, env = "DFUSION_LOG", default_value = "info")]
+    #[structopt(long, env = "DFUSION_LOG", default_value = "warn,driver=info")]
     log_filter: String,
 
     /// The Ethereum node URL to connect to. Make sure that the node allows for


bh2smith (approved):
OMG YES! Does this mean we would no longer be seeing the `ethcontract` request logs? I am very happy and somewhat concerned about this at the same time. 

While I would prefer to only see driver logs and not all the RPC calls, maybe we should include a few additional logs when we are about to send or receive some RPC requests. Just wondering if we might lose too much...

e00E:
The RPC logs are created in the driver code itself so their visibility does not change. If we wanted to hide the rpc logs we would set the log level for `driver` to `info` instead of `debug` for our deployments.

bh2smith:
> hide the rpc logs we would set the log level for driver to info instead of debug for our deployments.

Should we consider using debug only in `dev`and info for `prod` or are we already doing this?

e00E:
Both of them are using debug. Iirc we discussed this before with the result that it can be helpful to see the full transport messages. You can always filter them out in kibana again.

nlordell (approved):
I like it, maybe we should enable warning logs from dependencies just in case?

e00E:
Changed to included warnings for other modules.

PR #910, opened 2020-05-26
Make dexag api async

e00E:
One method was not returning a Future.

#843 

### Test Plan
Ci

### DIFF
--- a/driver/src/price_estimation/dexag.rs
+++ b/driver/src/price_estimation/dexag.rs
@@ -3,6 +3,7 @@ mod api;
 use super::{PriceSource, Token};
 use crate::http::HttpFactory;
 use crate::models::TokenId;
+use crate::util::FutureWaitExt;
 use anyhow::{anyhow, Context, Result};
 use api::{DexagApi, DexagHttpApi};
 use futures::future::{self, BoxFuture};
@@ -43,7 +44,7 @@ where
     }
 
     fn create_api_tokens(&self) -> Result<ApiTokens> {
-        let tokens = self.api.get_token_list()?;
+        let tokens = self.api.get_token_list().wait()?;
         let mut tokens: HashMap<String, api::Token> = tokens
             .into_iter()
             .map(|token| (token.symbol.to_uppercase(), token))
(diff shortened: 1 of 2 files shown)


driver/src/price_estimation/dexag.rs
```diff
@@ -177,38 +183,39 @@ mod tests {
...
-                address: None,
-            },
-            super::api::Token {
-                name: String::new(),
-                symbol: "USDC".to_string(),
-                address: None,
-            },
-        ];
+        lazy_static! {
+            static ref API_TOKENS: [super::api::Token; 3] = [
```
nlordell, line 187: nit: This doesn't have to be `lazy_static` right? Is this because of lifetime issues with the boxed futures? I feel like scoping the `MockDexagApi` to a smaller lifetime and then using a borrow *should* work (but I may be wrong).

Shouldn't `Rc` (or `Arc` if it needs to be `Send`) also work here?

nlordell (approved):
Looks good! My only nit is with the `lazy_static` test data - I feel like there should be a way around lifetime issues without needing it.

driver/src/price_estimation/dexag.rs
e00E: This is about lines like `returning(|| async { Ok(API_TOKENS.to_vec()) }.boxed())`.
This closure's lifetime has to be 'static because of a mockall limitiation. Previously this was accomplished by cloning the api tokens and moving into the closure. This is still possible with the move to async but I find this way cleaner. Otherwise it looks like
```rust
        let a = api_tokens.clone();
        api.expect_get_token_list().returning(move || {
            let a = a.clone();
            async move { Ok(a.to_vec()) }.boxed()
        });
```

nlordell: Yuck! `lazy_static` makes a lot of sense then. Thanks for clarifying!

driver/src/price_estimation/dexag/api.rs
```diff
@@ -55,12 +55,15 @@ impl DexagHttpApi {
...
-        url.set_path("token-list-full");
-        self.client
-            .get_json(url.to_string(), HttpLabel::Dexag)
-            .context("failed to parse token list json from dexag response")
+    fn get_token_list<'a>(&'a self) -> BoxFuture<'a, Result<Vec<Token>>> {
+        async move {
+            let mut url = self.base_url.clone();
+            url.set_path("token-list-full");
+            self.client
+                .get_json(url.to_string(), HttpLabel::Dexag)
```
fleupold, line 63: Should we use `get_json_async().await?` here so that we can yield while waiting on the response?

fleupold (approved):
Looks good to me! 🚀

driver/src/price_estimation/dexag/api.rs
nlordell: Definitely! I missed that.

One thing we should do, once all the HTTP components are migrated, is remove the blocking requests so that we ensure that they aren't accidentally being used anywhere.

e00E: Ah I missed that too. I should have grepped for this string.

PR #911, opened 2020-05-26
Make kraken api async

e00E:
#843 

### Test Plan
CI and `cargo test online_kraken_api -- --ignored --nocapture`.

### DIFF
--- a/driver/src/price_estimation/kraken.rs
+++ b/driver/src/price_estimation/kraken.rs
@@ -6,6 +6,7 @@ use self::api::{Asset, AssetPair, KrakenApi, KrakenHttpApi};
 use super::{PriceSource, Token};
 use crate::http::HttpFactory;
 use crate::models::TokenId;
+use crate::util::FutureWaitExt as _;
 use anyhow::{anyhow, Context, Result};
 use std::collections::HashMap;
 
@@ -42,8 +43,8 @@ where
         //   complicated is determining when the cache needs to be invalidated
         //   as new assets get added to Kraken.
 
-        let assets = self.api.assets()?;
-        let asset_pairs = self.api.asset_pairs()?;
+        let assets = self.api.assets().wait()?;
+        let asset_pairs = self.api.asset_pairs().wait()?;
 
         let usd =
             find_asset("USD", &assets).ok_or_else(|| anyhow!("unable to locate USD asset"))?;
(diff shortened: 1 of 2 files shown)


driver/src/price_estimation/kraken/api.rs
```diff
@@ -46,35 +50,50 @@ impl KrakenHttpApi {
...
+        let url = format!("{}/Ticker?pair={}", self.base_url, pairs.join(","));
+        async move {
+            self.client
+                .get_json_async::<_, KrakenResult<_>>(url, HttpLabel::Kraken)
+                .await
+                .context("failed to parse ticker JSON")?
+                .into_result()
+        }
+        .boxed()
     }
```
e00E, line 97: The reason for having two async blocks in one function is that this allows us to make the resulting Future's lifetime only depend on `self`.

nlordell: Can't you you just make `pairs: &'a [&'a str]`?

nlordell: Alternatively, you could make `url: Option<String>` and then have it be `None` when pairs are empty. That allows you to move the `if` statement into the `async` block and not need two of them.

nlordell (approved):
Nice! I think the `ticker` implementation is a little awkward though.

driver/src/price_estimation/kraken.rs
```diff
@@ -42,8 +43,8 @@ where
         //   complicated is determining when the cache needs to be invalidated
         //   as new assets get added to Kraken.
 
-        let assets = self.api.assets()?;
-        let asset_pairs = self.api.asset_pairs()?;
+        let assets = self.api.assets().wait()?;
+        let asset_pairs = self.api.asset_pairs().wait()?;
```
fleupold, line 47: nit: would it be hard to join these to have them run in parallel (I realize this wasn't the behavior before but could be a nice and simple improvement)

fleupold (approved):

driver/src/price_estimation/kraken/api.rs
e00E: > Can't you you just make `pairs: &'a [&'a str]`? I feel like the two `async` blocks are a little awkward.

That needlessly restricts the lifetimes of `pairs` to be the same as self which means it gets tied to the resulting Future.

I changed it to use your second suggestion.

driver/src/price_estimation/kraken.rs
e00E: It would be easy. I thought about this too but didn't want to change the behavior yet. I will change it when making price estimation async.

PR #914, opened 2020-05-26
Make solution submission async

e00E:
#843 

### Test Plan
CI

### DIFF
--- a/driver/src/solution_submission.rs
+++ b/driver/src/solution_submission.rs
@@ -3,16 +3,15 @@
 use crate::contracts::stablex_contract::StableXContract;
 use crate::gas_station::GasPriceEstimating;
 use crate::models::Solution;
-use crate::util::FutureWaitExt as _;
 
 use anyhow::{Error, Result};
 use ethcontract::errors::{ExecutionError, MethodError};
 use ethcontract::web3::types::TransactionReceipt;
 use ethcontract::U256;
+use futures::future::{BoxFuture, FutureExt as _};
 use log::info;
 #[cfg(test)]
 use mockall::automock;
-use std::thread;
 use std::time::Duration;
 use thiserror::Error;
 
@@ -33,11 +32,11 @@ pub trait StableXSolutionSubmitting {
     /// * `batch_index` - the auction for which this solutions should be evaluated
     /// * `orders` - the list of orders for which this solution is applicable
     /// * `solution` - the solution to be evaluated
-    fn get_solution_objective_value(
-        &self,
+    fn get_solution_objective_value<'a>(
+        &'a self,
         batch_index: U256,
         solution: Solution,
-    ) -> Result<U256, SolutionSubmissionError>;
+    ) -> BoxFuture<'a, Result<U256, SolutionSubmissionError>>;
 
     /// Submits the provided solution and returns the result of the submission
     ///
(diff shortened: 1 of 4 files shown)


driver/src/solution_submission.rs
```diff
@@ -103,33 +103,57 @@ impl<'a> StableXSolutionSubmitter<'a> {
...
         }
     }
+
+    /// Turn a method error from a solution submission into a SolutionSubmissionError.
+    async fn make_error(
+        &self,
+        batch_index: U256,
+        solution: Solution,
+        err: MethodError,
+    ) -> SolutionSubmissionError {
```
e00E, line 113: This was hard to read in the original function and very hard to convert to async inline so I refactored it into a separate function.

nlordell: :+1: I like the helper function as well

nlordell (approved):
So much async!

driver/src/solution_submission.rs
```diff
@@ -103,33 +103,57 @@ impl<'a> StableXSolutionSubmitter<'a> {
...
-            info!("Solved batch is not yet accepting solutions, waiting for next batch.");
-            thread::sleep(POLL_TIMEOUT);
-        }
+    ) -> BoxFuture<Result<U256, SolutionSubmissionError>> {
+        async move {
+            // NOTE: Compare with `>=` as the exchange's current batch index is the
+            //   one accepting orders and does not yet accept solutions.
+            while batch_index.as_u32() >= self.contract.get_current_auction_index().await? {
+                info!("Solved batch is not yet accepting solutions, waiting for next batch.");
+                thread::sleep(POLL_TIMEOUT);
```
fleupold, line 140: Should we use https://docs.rs/async-std/1.1.0/async_std/task/fn.sleep.html here to not block on the sleep?

fleupold (approved):

driver/src/solution_submission.rs
e00E: Eventually we need to decide on a runtime (probably asyncstd or tokio) and then we can use their sleep function. I was hoping to do that after this PR. In general it is wrong to do blocking IO or sleeping like this in an async function. In this case it is still fine because we're not yet running the future on any executor.
But this is a pitfall so I should have mentioned this somewhere.

Edit: I *thought* we couldn't use this function if the future isn't running on async-std's executor because that is what I remembered from tokio, but maybe I'm wrong? I'll have to check this.

driver/src/solution_submission.rs
```diff
@@ -138,35 +162,17 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
-                        .wait()
-                    {
-                        Ok(_) => None,
-                        Err(e) => Some(SolutionSubmissionError::from(e)),
-                    }
-                })
-                .unwrap_or_else(|| SolutionSubmissionError::Unexpected(err.into()))
-        })
-        .map(|_| ())
+        .or_else(move |err| async move { Err(self.make_error(batch_index, solution, err).await) })
```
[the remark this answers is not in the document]
bh2smith: I was looking around this PR and thinking "I wonder how retry_with_gas_price_increase was handled". Stumbled upon this and might have said the same thing as has already been said. However, in my case I wasn't sure what to expect when I got here. 

Is it safe to assume that the entire content of this error mapping has been offloaded into the cool new `make_error` function? I was kind of under the impression that this particular error was one of a kind and might not generalize well. Do we expect that `make_error` (containing this logic) will be reusable... because currently it is only used here.

bh2smith (approved):
Looks good, just have a question about the re-usability of `make_error`.

driver/src/solution_submission.rs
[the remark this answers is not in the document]
e00E: It is not reusable but it also doesn't have to be reusable. I am splitting a part that can stand alone out of a bigger function.

driver/src/solution_submission.rs
e00E: I decided to use the https://docs.rs/futures-timer/3.0.2/futures_timer/ crate for now. It is executor agnostic and internally used by async-std.

driver/src/solution_submission.rs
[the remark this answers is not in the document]
e00E: > ubernit: I would find an `async` block with either a `match` on the `retry_with_gas_price_increase` a little easier to understand. 100% OK with how it currently is though.

I don't agree. I feel that if I had written in that way then I would have gotten the comment that `or_else` is specialized method for this and I should use it. I have changed it anyway because everyone else prefers the other way.

PR #915, opened 2020-05-26
[Pricegraph] Add script for encoding instance JSON files

nlordell:
This PR adds a new script to encode instance JSON files used as input to the solver so that they may be used to generate pricegraph test data.

### Test Plan

Includes new orderbook generated from the instance file for batch 5301531


pricegraph/data/convert.js
```diff
@@ -0,0 +1,39 @@
...
+	const a = order.accountID.substr(2).padStart(40, "0");
+	const b = hex(BigInt(balance || 0), 32);
+	const bt = hex(Number(order.buyToken.substr(1)), 2);
+	const st = hex(Number(order.sellToken.substr(1)), 2);
+	const vf = hex(0x00000000, 4);
+	const vu = hex(0xffffffff, 4);
+	const ba = hex(BigInt(order.buyAmount), 16);
+	const sa = hex(BigInt(order.sellAmount), 16);
+	const id = hex(Number(order.orderID), 2);
+	return `${a} ${b} ${bt} ${st} ${vf} ${vu} ${ba} ${sa} ${sa} ${id}\n`
```
fleupold, line 15: nit: I'm not a fan of abbreviating variable names but I guess here it's fine.

pricegraph/data/convert.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+async function readJson(path) {
+	const json = await fs.readFile(path, "utf8");
+	return JSON.parse(json);
+}
+
+async function main(args) {
+	if (args.length != 2) {
```
fleupold, line 24: nit: Is the second argument really needed? It could just transform the xyz.json into xyz.hex?

fleupold (approved):
Nice and useful script. Feel free to ignore the comments as they are super nitty...

pricegraph/data/convert.js
nlordell: I use the batch ID to create multiple `orderbook-{batch}.hex` files for test data for the price graph. The hope was to collect "problematic" instances to be able to create regression tests for them targeting those specific scenarios.

We currently have two orderbooks so far:
- orderbook-5298183.hex
- orderbook-5301531.hex

Additionally, the `data/mod.rs` file, expects the test orderbooks to be named in this way: https://github.com/gnosis/dex-services/blob/943ba01b22631584c2263413e585118612bf6d4b/pricegraph/data/mod.rs#L25

Since this script is really meant to be for collecting these test cases, I think it makes sense to require this parameter so it produces an encoded orderbook file ready for consumption.

e00E (approved):
For future proofing I'd prefer if this script was written in Rust.

fleupold:
This seems to have broken travis: https://travis-ci.com/github/gnosis/dex-services/jobs/340420996

PR #917, opened 2020-05-26
[Orderbook] Check solution is not partially applied

### LINKED ISSUE #758: [event based orderbook] Correct approach for partial solutions

nlordell:
Fixes #758 

This PR adds a new error when retrieving an orderbook with a partially applied solution. This used to be an issue when events were emited from an event filter (since they come one by one). Now events get applied block per block, so it would be an error if an orderbook is in a state with a partially applied solution. 

### Test Plan

Updated unit test that verifies error.

### DIFF
--- a/driver/src/orderbook/streamed/state.rs
+++ b/driver/src/orderbook/streamed/state.rs
@@ -653,8 +653,7 @@ mod tests {
         );
         apply_event!(to state for batch 1; Trade order number 0, from user 2, selling 1, for 2);
 
-        assert_balance!(in state for batch 1; user 2, has token 0, balance 10);
-        assert_balance!(in state for batch 1; user 2, has token 1, balance 10);
+        assert!(state.orderbook_for_batch(Batch::Future(1)).is_err());
 
         apply_event!(to state for batch 1; SolutionSubmission from user 4, with fee 42);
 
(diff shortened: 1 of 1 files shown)


fleupold (approved):
👍

e00E (approved):
At first I wasn't sure we should rely on this the way you phrased it in the issue becaus it depends on how the node works and it depends on how the higher level after State, UpdatingOrderbook works. If it had a different way to lock then maybe this could still happen.

But if we're just returning an error like you do in this PR then even if this unlikely case happens we'll just try fetching the orderbook again so I think it's fine.

nlordell:
> At first I wasn't sure we should rely on this the way you phrased it in the issue becaus it depends on how the node works and it depends on how the higher level after State

At first I had moved the `State` creation into a `from_elements` method and asserted after folding the events into a `State`.

I ended up just settling for this as it allows states to be partially applied, but prohibits reading from partially applied states, by returning an error.

PR #918, opened 2020-05-27
Print errors with more detail

e00E:
As the anyhow documentation states, printing an error using `{}` only
shows the outermost error. By using `{:?}` we get the most information:
the full error chain and a backtrace.

### Test Plan
Only affects logging.

### DIFF
--- a/driver/src/driver/scheduler/evm.rs
+++ b/driver/src/driver/scheduler/evm.rs
@@ -118,7 +118,7 @@ impl Scheduler for EvmScheduler<'_> {
     fn start(&mut self) -> ! {
         loop {
             if let Err(err) = self.step() {
-                error!("EVM scheduler error: {}", err);
+                error!("EVM scheduler error: {:?}", err);
             }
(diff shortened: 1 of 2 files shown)


fleupold (approved):

driver/src/transport.rs
```diff
@@ -63,7 +63,7 @@ impl HttpTransportInner {
             .post_raw_json_async(&self.url, request, label)
             .await
             .map_err(|err| {
-                warn!("[id:{}] returned an error: '{}'", id, err.to_string());
+                warn!("[id:{}] returned an error: '{:?}'", id, err);
```
nlordell, line 66: nit: This error should get propagated, and printed in full at a higher level, I think here it should be ok to stay with `Display` instead of `Debug`.

nlordell (approved):
Awesome! Welcome change.

driver/src/transport.rs
e00E: With that reasoning we don't need to print it at all but I'll change it back.

PR #919, opened 2020-05-27
[Easy] Ignore ethcontract files for coverage

fleupold:
Looking at our current coverage, we are tracking the autogenerated ethcontract files which we won't test and thus should not conisder in our coverage report. This PR ignores the builds folder from the coverage report achieving this.

I also made coverage work locally on MacOS which required updating the `core-foundation` package (cf. https://github.com/rust-lang/rust/issues/63047)

### Test Plan

If not already done `rustup nightly` && `cargo install grcov`

Then
```
env CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings" cargo +nightly test

grcov --branch --ignore-not-existing --llvm --ignore "target/debug/build/**" target/debug/ --output-type html --source-dir . --service-job-id 1 --service-name foo
```

Open the html/coverage report and see the code generated files no longer covered.


Cargo.toml
```diff
@@ -13,3 +13,8 @@ default-members = [
 
 [profile.release]
 opt-level = "s"
+
+[patch.crates-io]
+# Needed to make coverage work on MacOS (https://github.com/servo/core-foundation-rs/pull/357)
+core-foundation = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
+core-foundation-sys = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
```
bh2smith, line 20: Dang. Unfortunate, but awesome that something exists!

bh2smith (approved):
Looks really cool! I'm excited to see the report!

.travis.yml
```diff
@@ -73,6 +73,6 @@ jobs:
         # introduce some new warnings and when testing one such warning appeared in the ethcontract
         # generated contract code which caused the whole file to be printed as part of the warning
         # message which made the build fail (probably travis related).
-        - env CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings" cargo test
-        - ./grcov --branch --ignore-not-existing --llvm --ignore "/*" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-pro --service-job-id $TRAVIS_JOB_ID
+        - env CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings" cargo +nightly test
```
e00E, line 76: The nightly in `cargo +nightly test` shouldn't be necessary but also doesn't hurt. Just mentioning.

Cargo.toml
e00E, line 20: This is a bit scary because it could silently break something or we might not notice that is no longer necessary. On the other hand if I understand it correctly this library is only used on macos anyway.

e00E (approved):

Cargo.toml
nlordell: I think, because we commit our lock file (which includes git revision and versions for every dependency), I don't think it can really break "silently". We can, however, run into issues if we try to update dependencies for example.

nlordell (approved):
Nice!

.travis.yml
fleupold: I added it so that I can easily copy paste the command locally where I have multiple envs installed.

Cargo.toml
fleupold: If there is a way to only patch the macOS target we could do that, but I wasn't able to find a way.

PR #920, opened 2020-05-27
[Travis] Don't run testsuite on master

### LINKED ISSUE #801: Should we only run the deploy CI step on master and omit the tests
We already run the tests in the state that master would be in in the PRs so running them again adds no useful information. The only step we need for master is running the deploy script.
This would speed up the time it takes to got from PR to deployed docker image significantly.
I don't know if this is possible with travis (probably is), just had this idea while waiting for my hot fix to be deployed.

fleupold:
Fixes #801 

Now that we require our PR branches to be up to date with master we don't have to run all lints/tests again on master.

This PR splits out the deploy part into a separate stage which only runs on master and tagged branches. The test staget will only run on PR branches.

Note that we still run test coverage on master, to allow having an up to date value on the main readme page (instead of having to find the most recent PR and reading the coverage report from there). I still would like to have coverage on each PR so that we can get an idea how our PRs impact this metric.

### Test Plan
- Run CI on this PR, expect to see tests and coverage to run
- On merge expect only coverage and deploy to run.


e00E (approved):

.travis.yml
```diff
@@ -7,11 +7,13 @@ if: (branch = master) OR (type = pull_request) OR (tag IS present)
 os: linux
 dist: bionic
 language: rust
+env:
+  - OPEN_SOLVER_VERSION=v0.0.10
+  - PRIVATE_SOLVER_VERSION=v0.7.0
```
nlordell, line 12: I think these should be `global:`? I'm a Travis noob so I'm not sure.

nlordell (approved):
Nice improvement!

bh2smith (approved):
Looks good. It might be worth mentioning that sometimes, when "updating with master" via the github browser interface, I've noticed that it sometimes does not trigger a rebuild and/or also uses existing cache from the last build. While the cache issue isn't relevant here, I think that the sometimes not rebuilding might give us a false sense on confidence that everything is actually working and allow the merge. Not sure if this is something I can reproduce, but we may want to this possibility in mind.

fleupold:
> Looks good. It might be worth mentioning that sometimes, when "updating with master" via the github browser interface, I've noticed that it sometimes does not trigger a rebuild and/or also uses existing cache from the last build.

Can you share an example where this happened (link to the last travis build and merge commit which didn't cause a new travis build)?

Issue #916, opened 2020-05-26
[Pricegraph] Order amounts and user balance underflows


nlordell:
Currently due to rounding errors, it is possible for an order amount or user balance to underflow because of rounding errors when sell token has a drastically different number of decimals than the other tokens being traded (such when selling USDC and going through tokens with the usual 18 decimals).

fleupold:
What happens in this case? Do we panic or just provide incoherent data?

e00E:
Just found this with fuzzing (had to fix a bug in fuzzing first, PR incoming):

```rust
    #[test]
    fn fuzz_error() {
        use crate::encoding::*;
        use primitive_types::{H160, U256};

        let mut orderbook = Orderbook::from_elements(vec![Element {
            user: H160::zero(),
            balance: U256::from_dec_str(
                "3532883802906008782403269293875209110100384173848976041480599189852671",
            )
            .unwrap(),
            pair: TokenPair { buy: 0, sell: 0 },
            valid: Validity {
                from: 285741312,
                to: 180095676,
            },
            price: Price {
                numerator: 13294906614391990988372451468773477386,
                denominator: 327042228921422829026657257798164547592,
            },
            remaining_sell_amount: 83798276971421254262445676335662107162,
            id: 2056,
        }]);
        orderbook.reduce_overlapping_orders();
    }
```

> panicked at 'remaining amount underflow for order 0x0000…0000-2056', pricegraph/src/orderbook.rs:364:13

Is this the same thing?

nlordell:
> Is this the same thing?

Yes

fleupold:
Unrelated to this issue but is there a way for us to integrate the fuzzing into some automated process? It feels like right now we rely on individuals running it after we think there might be an issue.

nlordell:
> What happens in this case? Do we panic or just provide incoherent data?

In release mode (used by the Wasm module), nothing, the result might include rounding errors. Since (from what I remember) we apply a small "buffer" to these values in the front-end, at the end of the day, it shouldn't really matter.

In debug mode (used for tests and fuzzing) we panic.

e00E:
> Unrelated to this issue but is there a way for us to integrate the fuzzing into some automated process? It feels like right now we rely on individuals running it after we think there might be an issue.

In my opinion fuzzing should be ran manually. Fuzzing doesn't prove the absence of bugs and you can run it for as long as possible. I wouldn't put in on travis because that is not what travis is made for and travis' cpu is slow.

[...]

PR #923, opened 2020-05-27
Improve pricegraph fuzzing

e00E:
In addition to fuzzing the building of the orderbook this commit enables
fuzzing of the other public methods.

### Test Plan
Fuzzing still works. This quickly finds the panic #916 currently so it isn't very useful until that is fixed.

### DIFF
--- a/pricegraph/fuzz/fuzz_targets/orderbook.rs
+++ b/pricegraph/fuzz/fuzz_targets/orderbook.rs
@@ -1,9 +1,0 @@
-#![no_main]
-use libfuzzer_sys::fuzz_target;
-use pricegraph::{Element, Orderbook};
-
-// Fuzz creation and usage of Orderbook.
-
-fuzz_target!(|elements: Vec<Element>| {
-    let _ = Orderbook::from_elements(elements);
-});
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2228,6 +2228,7 @@ dependencies = [
 name = "pricegraph-fuzz"
 version = "0.0.0"
 dependencies = [
+ "arbitrary",
  "libfuzzer-sys",
  "pricegraph",
 ]
--- a/pricegraph/fuzz/Cargo.toml
+++ b/pricegraph/fuzz/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2018"
 cargo-fuzz = true
 
 [dependencies]
+arbitrary = { version = "0.4", features = ["derive"] }
 libfuzzer-sys = "0.3"
 pricegraph = { path = "..", features = ["arbitrary"] }
 
@@ -17,5 +18,5 @@ name = "element_read_all"
 path = "fuzz_targets/element_read_all.rs"
 
 [[bin]]
-name = "orderbook"
-path = "fuzz_targets/orderbook.rs"
+name = "pricegraph"
+path = "fuzz_targets/pricegraph.rs"
--- a/pricegraph/fuzz/fuzz_targets/pricegraph.rs
+++ b/pricegraph/fuzz/fuzz_targets/pricegraph.rs
@@ -1,0 +1,62 @@
+#![no_main]
+use arbitrary::Arbitrary;
+use libfuzzer_sys::fuzz_target;
+use pricegraph::{Element, Pricegraph, TokenId, TokenPair};
+use std::iter::once;
+
+// Limit the maximum token id that is allowed to appear in the generated orders. Without this we can
+// make orderbook creation too slow if one order contains a large buy or sell token id.
+const MAX_TOKEN_ID: u16 = 16;
+
+// Fuzz creation and usage of Orderbook.
+
+#[derive(Arbitrary, Debug)]
+enum Operation {
+    OrderForSellAmount {
+        pair: TokenPair,
+        sell_amount: f64,
+    },
+    TransitiveOrderbook {
+        base: TokenId,
+        quote: TokenId,
+        spread: Option<f64>,
+    },
+}
+
+#[derive(Arbitrary, Debug)]
+struct Arguments {
+    elements: Vec<Element>,
+    operation: Operation,
+}
+
+fn largest_token_id(elements: &[Element]) -> Option<TokenId> {
+    elements
+        .iter()
+        .flat_map(|e| once(e.pair.buy).chain(once(e.pair.sell)))
+        .max()
+}
+
+fuzz_target!(|arguments: Arguments| {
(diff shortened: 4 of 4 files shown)


pricegraph/fuzz/fuzz_targets/orderbook.rs
```diff
@@ -1,9 +1,50 @@
...
+        .max()
+}
+
+fuzz_target!(|arguments: Arguments| {
+    if largest_token_id(&arguments.elements).unwrap_or(0) > MAX_TOKEN_ID {
+        return;
+    }
+
+    let mut orderbook = Orderbook::from_elements(arguments.elements);
+    match arguments.operation {
```
fleupold, line 39: Not a big deal, but could we instead of introducing the new operation type to test all operations in serial? I see that you want to fuzz the pair and volume, but I don't think we need the Operation enum (it adds bloat to the otherwise very small and easy to understand fuzzing code).

I could understand if multiple function executions were fuzzed in arbitrary order on the same instance (testing interdependency of e.g. first reducing and then filling a market order and vice versa).

fleupold (approved):

pricegraph/fuzz/fuzz_targets/orderbook.rs
e00E: I prefer it this way.
Testing operations in serial introduces dependencies between them where the order can matter. I suspect it would make fuzzing slower and harder to find relevant branches for the fuzzer.
I wouldn't call this bloat or hard to understand either. Of course it is slightly more complex than doing it in serial but it has a specific purpose.

nlordell: Hmm, I'm kind of on the fence about this one - I would actually kind of prefer to have two targets here:
1. `fill_market_order`
2. `is_overlapping -> reduce_ovelapping_orders -> assert!(!is_overlapping)`

But maybe I'm suggesting this because I know too much about the actual implementation.

One thing to note is that `is_overlapping` takes `&self`, and since we have no interior mutability, in theory it shouldn't affect operations following it.

On the otherhand its kind of nice to be super thorough.

e00E: If 2. is a post condition we might want to add it as a debug_assert to `reduce_overlapping_orders` instead. In general if you can think of more debug_asserts, they are helpful for the fuzzer.

nlordell: Created PR #926

bh2smith (approved):

nlordell:
Now that there is a new `Pricegraph` API, maybe that should be the fuzz target and not `Orderbook`.

e00E:
I agree we should only fuzz `Pricegraph` unless there would be some conditions hard or impossible to hit otherwise which I don't think is the case.
I updated the PR to do this.

pricegraph/fuzz/fuzz_targets/orderbook.rs
```diff
@@ -1,9 +1,62 @@
...
+// make orderbook creation too slow if one order contains a large buy or sell token id.
+const MAX_TOKEN_ID: u16 = 16;
 
 // Fuzz creation and usage of Orderbook.
 
-fuzz_target!(|elements: Vec<Element>| {
-    let _ = Orderbook::from_elements(elements);
+#[derive(Arbitrary, Debug)]
+enum Operation {
+    EstimateExchangeRate {
```
nlordell, line 15: There is also `estimate_buy_amount`.

pricegraph/fuzz/fuzz_targets/orderbook.rs
```diff
@@ -1,9 +1,62 @@
...
+        .max()
+}
+
+fuzz_target!(|arguments: Arguments| {
+    if largest_token_id(&arguments.elements).unwrap_or(0) > MAX_TOKEN_ID {
+        return;
+    }
+
+    let pricegraph = Pricegraph::new(arguments.elements);
+    match arguments.operation {
```
nlordell, line 45: Is this necessary now that `Pricegraph` is immutable? That means that order of operations no longer has any impact?

pricegraph/fuzz/fuzz_targets/orderbook.rs
e00E: Do you mean `order_for_sell_amount`? I haven't included it because it just calls `estimate_exchange_rate` internally.

pricegraph/fuzz/fuzz_targets/orderbook.rs
e00E: It's not necessary in the sense that the previous operation might make some cases for later operations impossible but I still like this for clarity. And it probably still helps the fuzzer isolate cases.

nlordell: Personally I would find it clearer to have separate fuzz targets for the two methods.

pricegraph/fuzz/fuzz_targets/orderbook.rs
nlordell: In that case, we should use `order_for_sell_amount` to get more coverage no?

pricegraph/fuzz/fuzz_targets/orderbook.rs
nlordell: Are there any concerns about not being able to fuzz each method individually?

nlordell: > And it probably still helps the fuzzer isolate cases.

**Disclaimer**: I know pretty much nothing about how fuzzers work

This surprises me. Intuitively, I would imagine fuzzers would perform better with a narrower search space.

e00E: Separate targets are probably even better, yes. What I meant before was doing all the operations in a row. I'll try out the separate targets.

nlordell: > What I meant before was doing all the operations in a row.

Oh yes, in that case I definitely agree.

e00E: My conclusion for separate targets is that I would like to keep this as a single target because a significant code path is shared in the creation of the orderbook and that we would duplicate the `largest_token_id` filter (of course, could probably put it into a module).
I am going to keep it this way and we can reevalute if we add more fuzzing.

nlordell (approved):
Sounds OK to me.

Issue #833, opened 2020-05-04
[event based orderbook] orderbook_for_batch Batch::Current not always correct


fleupold:
https://github.com/gnosis/dex-services/blob/6f794300512e6332e1a80b4ad3ad6700f89fccf1/driver/src/orderbook/streamed/state.rs#L52-L60

This method currently expects to be passed a batch number and returns the orderbook that is valid for batch-1. I think it would be nice to replicate the general `StableXOrderBookReading` trait's logic which returns the orderbook for the batchID it is passed in (rather than - 1).

Also the `Current` variant of the `Batch` enum cannot always be correctly computed as it is based on the last `batchID` for which we have seen an event. This assumes liveness of the contract and would return wrong results if for > 5 minutes there were no events (e.g. no new orders or deposits). We would then assume the last batch with events to be current, although this might no longer be true.

PR #924, opened 2020-05-27
[RFC] Refactor the semantics of batchID in event based orderbook

fleupold:
Continues #833 

This PR tries to make the semantics of the variable batchID more explicit in the streamed orderbook.

The problem originates in the fact that events come in with a batchID corresponding to the batch that is currently collecting orders (e.g. Deposit events that come in when batch `t` is collecting orders are valid when batch `t+1` is starting to collect orders). 
Consumers are most likely interested in getting the state to solve a certain batch (completed or still in progress). For this they need to include events that were received while this batch was collecting orders. In the example above, if we try to solve batch t, we need to include the deposit (as it is valid) and thus specify all events until t+1.

This is true for balances but not order validity as those explicitly expressed the batchIDs for which this they should be considered in the payload (and not via timestamp metadata on the event occurred).

In this PR I changed the consumer facing method to accept the batchID that is supposed to be solved and renamed some methods to make they semantics more explicit.
I don't think this goes far enough and am looking for feedback/suggestions on how we can make the behaviour easy to understand when we revisit this code in 6 months.

Arguably we could also make the State expose its data in terms of "at the beginning of a batch that is currently being collected", and keep the +1 in the StablexOrderbookReading implementations.

Let's discuss what's most intuitive on this PR.

### Test Plan

Updated unit tests.

### DIFF
--- a/driver/src/orderbook/streamed/balance.rs
+++ b/driver/src/orderbook/streamed/balance.rs
@@ -53,7 +53,7 @@ impl Balance {
         Ok(())
     }
 
-    pub fn get_balance(&self, batch_id: BatchId) -> BigInt {
+    pub fn get_balance_at_beginning_of_batch(&self, batch_id: BatchId) -> BigInt {
         let balance = self.balance_with_deposit_and_proceeds(batch_id);
         // Withdraw requests can be for amounts larger than balance.
         match self.withdraw.amount(batch_id) {
@@ -160,7 +160,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn get_balance() {
+    fn get_balance_at_beginning_of_batch() {
         let overdrawn_request = Balance {
             balance: BigInt::from(2),
             deposit: Flux::default(),
@@ -170,9 +170,18 @@ mod tests {
             },
             proceeds: Flux::default(),
         };
-        assert_eq!(overdrawn_request.get_balance(0), BigInt::from(2));
-        assert_eq!(overdrawn_request.get_balance(1), BigInt::from(2));
-        assert_eq!(overdrawn_request.get_balance(2), BigInt::zero());
+        assert_eq!(
+            overdrawn_request.get_balance_at_beginning_of_batch(0),
+            BigInt::from(2)
+        );
+        assert_eq!(
+            overdrawn_request.get_balance_at_beginning_of_batch(1),
+            BigInt::from(2)
+        );
+        assert_eq!(
+            overdrawn_request.get_balance_at_beginning_of_batch(2),
+            BigInt::zero()
+        );
 
         let standard_request = Balance {
             balance: BigInt::from(2),
@@ -183,9 +192,18 @@ mod tests {
             },
             proceeds: Flux::default(),
         };
-        assert_eq!(standard_request.get_balance(0), BigInt::from(2));
-        assert_eq!(standard_request.get_balance(1), BigInt::from(2));
-        assert_eq!(standard_request.get_balance(2), BigInt::from(1));
+        assert_eq!(
+            standard_request.get_balance_at_beginning_of_batch(0),
+            BigInt::from(2)
+        );
+        assert_eq!(
+            standard_request.get_balance_at_beginning_of_batch(1),
+            BigInt::from(2)
+        );
+        assert_eq!(
+            standard_request.get_balance_at_beginning_of_batch(2),
+            BigInt::from(1)
+        );
 
         let no_request = Balance {
             balance: BigInt::from(1),
@@ -193,8 +211,17 @@ mod tests {
             withdraw: Flux::default(),
             proceeds: Flux::default(),
         };
-        assert_eq!(no_request.get_balance(0), BigInt::from(1));
-        assert_eq!(no_request.get_balance(1), BigInt::from(1));
-        assert_eq!(no_request.get_balance(2), BigInt::from(1));
+        assert_eq!(
+            no_request.get_balance_at_beginning_of_batch(0),
+            BigInt::from(1)
+        );
+        assert_eq!(
+            no_request.get_balance_at_beginning_of_batch(1),
+            BigInt::from(1)
+        );
+        assert_eq!(
+            no_request.get_balance_at_beginning_of_batch(2),
+            BigInt::from(1)
+        );
     }
 }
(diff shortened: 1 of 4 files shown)


nlordell:
Does RFC stand for "Request For Change"?

fleupold:
More [Request For Comments](https://en.wikipedia.org/wiki/Request_for_Comments)

driver/src/orderbook/streamed/balance.rs
```diff
@@ -53,7 +53,7 @@ impl Balance {
         Ok(())
     }
 
-    pub fn get_balance(&self, batch_id: BatchId) -> BigInt {
+    pub fn get_balance_at_beginning_of_batch(&self, batch_id: BatchId) -> BigInt {
```
nlordell, line 56: Nice, I think the method renames helps clarify exactly what balance is being returned

driver/src/orderbook/streamed/order.rs
```diff
@@ -117,7 +117,9 @@ impl Order {
         let (buy_amount, sell_amount) = stablex_auction_element::compute_buy_sell_amounts(
             self.price_numerator,
             self.price_denominator,
-            self.get_remaining_amount(batch_id),
+            // Remaining amount needs to incorporate trades that happened while collecting orders
+            // for the batch we are trying to solve. Thus query balance as of beginning of the next.
+            self.get_remaining_amount_at_beginning_of_batch(batch_id + 1),
```
nlordell, line 122: Is this being unit tested? I don't see one in this file.

driver/src/orderbook/streamed/state.rs
```diff
@@ -73,11 +75,12 @@ impl State {
...
                     (*user_id, token_id),
+                    // We fetch the balance including all trades/deposits that happened while collecting orders
+                    // for the batch we are trying to solve (balance at the beginning of the next batch).
                     // Can fail if user's balance exceeds U256::max.
-                    // Can fail while not all trades of a solution have been received and batch_id
-                    // is the next batch_id so that we assume that the current solution won't be be
-                    // reverted.
-                    bigint_u256::bigint_to_u256(&balance.get_balance(batch_id))?,
+                    bigint_u256::bigint_to_u256(
+                        &balance.get_balance_at_beginning_of_batch(batch_id + 1),
```
nlordell, line 82: Won't this include fluxes (like withdraw requests) that only mature in the next batch? Basically, is `end_of_batch(batch_id) <=> beginning_of_batch(batch_id + 1)`.

If I understand correctly, we do this to ensure that solutions and trades that just happened are included in the next batch?

Sorry, I should know this but I always get confused.

nlordell (approved):
I like the method renames as they make the their semantics immediately apparent. Just have some questions about the batch ID logic that always confuses me :confounded:

nlordell:
> The problem originates in the fact that events come in with a batchID corresponding to the batch that is currently collecting orders (e.g. Deposit events that come in when batch t is collecting orders are valid when batch t+1 is starting to collect orders).

I was under the impression that the balance state was already storing the maturing batches for deposits and withdrawals. These off by 1 things are super confusing!

e00E:
Let's first understand how batch id currently works in State.

State works like the smart contract. batch ids are used the same way as in the smart contract. The current batch id refers to what the smart contract would return in `EpochTokenLocker::getCurrentBatchId`, aka the batch that is collecting orders, not the batch that is being solved.
`Balance::get_balance` is like `EpochTokenLocker::getBalance`. Thus, if you want to solve batch N, you ask for the balance in N+1, because if N+1 is the current batch then N is being solved.
`Order::get_used_amount` works the same way.
This is why we have in `impl StableXOrderBookReading for Orderbook` `state.orderbook_for_batch(batch_id_to_solve.low_u32() + 1)`.
`State::orders` does not have a smart contract equivalent because the smart contract does not check order validity when encoding orders but you can roughly see it with how `BatchExchange::checkOrderValidity` is called. It gets batchBeingSolved aka N-1. This is why we have in `State::orders` `.filter(move |(_, order)| order.is_valid_in_batch(batch_id - 1))`.

I like the name change to include "at_beginning_of_batch" because it signifies that this value (balance or order remaining amount) will not change. With the old style of reading the orderboko we ran into this problem with the smart contract, that if we read after a solution has been submitted the balances and remaining amounts are actually wrong.

Shifting the plus and minus ones around for the batch id mostly equivalent to what we currently have but I slightly prefer having batch id always refer to getCurrentBatchId which means we have that `batch_id_to_solve + 1` because StableXOrderBookReading has the batch id to solve as the parameter.

fleupold:
> Maybe it would be nice to have a unit test that would fail with the previous implementation and that now succeeds to better understand the exact case this is trying to fix?

It's not fixing a bug but rather trying to make the +1 logic more understandable and explicit.

In the current version, I agree with @e00E that the `+1` should probably happen outside of the `State` (in the StableXOrderbookReading implementation). I'd be even more convinced if order validity was also expressed in terms of "batch currently collecting orders", to get rid of this hybrid between the two semantics (but I'm not sure it's possible).

My initial idea was to take a more consumer focussed approach to the semantics. As a consumer I'd argue you always want to get the orderbook as if you were solving a certain batchID (even if it's only given all the orders/deposits we have so far).
Would it make sense to rewrite the logic inside `State` to store balances and remaining amounts with the "batch currently to solve" semantics in mind (and do the conversion when applying events)?

For now I'll move the +1 back out and keep the explicit renaming and we can take a look at that.

nlordell:
>  It's not fixing a bug but rather trying to make the +1 logic more understandable and explicit.

Oh sorry I had misunderstood!

>  keep the explicit renaming and we can take a look at that.

Yes! It was a nice addition.

WRT to where the +1 should be, I think I agree with @e00E that if we keep it the same as the smart contract it will make it easier to compare to the reference implementation that way. I also agree that its confusing that there are two meanings to a "batch ID" (solving vs collecting). Maybe this is something we can capture as an issue for a v2 of the smart contract?

e00E (approved):

e00E:
> Would it make sense to rewrite the logic inside State to store balances and remaining amounts with the "batch currently to solve" semantics in mind (and do the conversion when applying events)?

I suspect this would be more confusing because it works more differently from the smart contract and afterall we're trying to work with it's events. It feels more natural to have State work like this and then have the adaptation for solving on the next level like we have when we implement the StableXOrderBookReading trait with the `+ 1`.
But it is hard to tell without trying it out. But I also think the current version is fine so I'm not sure if it's worth it to experiment more here.

PR #925, opened 2020-05-27
Make OrderBookReading implementations Send

### LINKED ISSUE #806: Make all impls of OrderbookReading Send

fleupold:
Closes #806

See issue for more context. This PR makes it so that the orderbook layering chain owns each inner orderbook. This allows making all implementations Send which on the other hand will make them more plug and playable (e.g. ShadowOrderbook requires Send for shadow)

### Test Plan
rustc

### DIFF
--- a/driver/src/driver/stablex_driver.rs
+++ b/driver/src/driver/stablex_driver.rs
@@ -23,7 +23,7 @@ pub trait StableXDriver {
 
 pub struct StableXDriverImpl<'a> {
     price_finder: &'a (dyn PriceFinding + Sync),
-    orderbook_reader: &'a (dyn StableXOrderBookReading + Sync),
+    orderbook_reader: &'a (dyn StableXOrderBookReading),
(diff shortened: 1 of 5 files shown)


e00E (approved):

nlordell (approved):
Nice, I think using trait objects here makes the code a little cleaner!

PR #927, opened 2020-05-28
move `Arbitrary` implementation to the fuzz crate

nlordell:
I noticed this when building services, that even if `fuzz` crate is not being built, the `abitrary` feature is enabled for pricegraph. This is because a single `Cargo.lock` file only has one set of features per dependency.

This PR just moves the `Arbitrary` implementation into the `fuzz` crate so it is no longer a dependency of `pricegraph`.

### Test Plan

Run the fuzzer as per the README instructions - see magic happen.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1948,7 +1948,6 @@ dependencies = [
 name = "pricegraph"
 version = "0.1.0"
 dependencies = [
- "arbitrary",
  "assert_approx_eq",
  "criterion",
  "data-encoding",
@@ -1962,6 +1961,7 @@ dependencies = [
 name = "pricegraph-fuzz"
 version = "0.0.0"
 dependencies = [
+ "arbitrary",
  "libfuzzer-sys",
  "pricegraph",
 ]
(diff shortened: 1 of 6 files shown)


e00E:
I find having arbitrary as a feature in a library crate conceptually more correct so I am hesitant to change this because of a limitation of the way we put crates into the same workspaces.
If we had put pricegraph in its own repository or not in the workspace we wouldn't have this problem.

fleupold:
> I noticed this when building services, that even if fuzz crate is not being built, the abitrary feature is enabled for pricegraph.

How exactly does this manifest itself? When running clean cargo build, I don't see the print statement "Compiling abritrary" (only when I run the fuzzing command).

Also, when I try to put a line like this:

```rust
crate::encoding::abitrary_impl::ArbitraryElement::from_element(&element);
```

into the `pricegraph` crate code somewhere I get a compile error when running: `cargo build` but succeed when calling `cargo build --features arbitrary`

nlordell:
Sorry! I had configured my builds like so, which was enabling `arbirtrary`.
```
nnoremap <silent> <Leader>cc :!cargo clippy --workspace --all-features --all-targets<CR>
```

I guess we can close this!

PR #928, opened 2020-05-28
[Coverage] re-ignore / folder

fleupold:
#919 actually decreased coverage (https://coveralls.io/github/gnosis/dex-services) because we stopped ignoring the `/` directory. This doesn't generate any code when testing locally, however on travis there is a significant amount of cached cargo code in this directory which doesn't have any test coverage and thus distorts our report.

This PR re-ignores this directory

### Test Plan
See test coverage increasing on coveralls after this is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,8 @@ jobs:
         # generated contract code which caused the whole file to be printed as part of the warning
         # message which made the build fail (probably travis related).
         - env CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings" cargo +nightly test
-        - ./grcov --branch --ignore-not-existing --llvm --ignore "target/debug/build/** --ignore "/*" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-pro --service-job-id $TRAVIS_JOB_ID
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Maybe add a comment so we don't accidentally remove it again?

.travis.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         # generated contract code which caused the whole file to be printed as part of the warning
         # message which made the build fail (probably travis related).
         - env CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings" cargo +nightly test
-        - ./grcov --branch --ignore-not-existing --llvm --ignore "target/debug/build/**" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-pro --service-job-id $TRAVIS_JOB_ID
+        - ./grcov --branch --ignore-not-existing --llvm --ignore "target/debug/build/** --ignore "/*" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-pro --service-job-id $TRAVIS_JOB_ID
```
e00E, line 75: `--ignore "target/debug/build/** --ignore "/*"`
I think the `"` are wrong. Should probably be
`--ignore "target/debug/build/**" --ignore "/*"`

e00E (approved):

.travis.yml
fleupold: 🦅

PR #929, opened 2020-05-28
[Easy] Remove null prices during deserialization

### LINKED ISSUE #450: Remove 0 prices from PriceMap
@bh2smith suggested this in https://github.com/gnosis/dex-services/pull/443#discussion_r367938770 :

> Actually I was hoping we could actually leave zero prices out of the price map and adjust the logic in a couple other places to infer zero as a default when the prices are not included.
We would then no longer need the "TokenNotFound" error in objective function evaluation.

nlordell:
Fixes #450 

This PR removes `null` and `"0"` prices from the solver output.

### Test Plan

Adjusted unit test.


e00E (approved):

fleupold (approved):

bh2smith:
I am very happy about this change. In fact this was actually fixed in both repos!

https://github.com/gnosis/dex-solver/pull/245

So long, null prices!

PR #930, opened 2020-05-28
Make PriceSource async

e00E:
#843 

Recommend viewing diff without whitespaces.

### Test Plan
CI
`cargo test online_dexag_client -- --ignored --nocapture`
`cargo test online_kraken_prices -- --ignored --nocapture`

### DIFF
--- a/driver/src/price_estimation.rs
+++ b/driver/src/price_estimation.rs
@@ -11,8 +11,11 @@ mod threaded_price_source;
 pub use self::data::TokenData;
 use self::dexag::DexagClient;
 use self::kraken::KrakenClient;
-use crate::http::HttpFactory;
-use crate::models::{Order, TokenId, TokenInfo};
+use crate::{
+    http::HttpFactory,
+    models::{Order, TokenId, TokenInfo},
+    util::FutureWaitExt as _,
+};
 use anyhow::Result;
 use average_price_source::AveragePriceSource;
 use log::warn;
@@ -114,7 +117,7 @@ impl PriceOracle {
             return HashMap::new();
         }
 
-        match self.source.get_prices(tokens) {
+        match self.source.get_prices(tokens).wait() {
             Ok(prices) => prices,
             Err(err) => {
                 warn!("failed to retrieve token prices: {}", err);
@@ -154,6 +157,7 @@ mod tests {
     use super::data::TokenBaseInfo;
     use super::*;
     use anyhow::anyhow;
+    use futures::FutureExt as _;
     use price_source::MockPriceSource;
 
     #[test]
@@ -172,9 +176,12 @@ mod tests {
                 tokens == [Token::new(1, "WETH", 18), Token::new(2, "USDT", 6)]
             })
             .returning(|_| {
-                Ok(hash_map! {
-                    TokenId(2) => 1_000_000_000_000_000_000,
-                })
+                async {
+                    Ok(hash_map! {
+                        TokenId(2) => 1_000_000_000_000_000_000,
+                    })
+                }
+                .boxed()
             });
 
         let oracle = PriceOracle::with_source(tokens, source);
@@ -206,7 +213,7 @@ mod tests {
         let mut source = MockPriceSource::new();
         source
             .expect_get_prices()
-            .returning(|_| Err(anyhow!("error")));
+            .returning(|_| async { Err(anyhow!("error")) }.boxed());
 
         let oracle = PriceOracle::with_source(tokens, source);
         let prices = oracle.get_token_prices(&[Order::for_token_pair(0, 1)]);
@@ -235,7 +242,9 @@ mod tests {
         });
 
         let mut source = MockPriceSource::new();
-        source.expect_get_prices().returning(|_| Ok(HashMap::new()));
+        source
+            .expect_get_prices()
+            .returning(|_| async { Ok(HashMap::new()) }.boxed());
 
         let oracle = PriceOracle::with_source(tokens, source);
         let prices = oracle.get_token_prices(&[Order::for_token_pair(0, 7)]);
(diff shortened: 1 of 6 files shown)


driver/src/price_estimation/threaded_price_source.rs
```diff
@@ -1,12 +1,16 @@
 use super::{price_source::PriceSource, Token};
```
e00E, line 1: Later, when we have an executor, we can refactor ThreadedPriceSource into something simpler that doesn't use threads.

driver/src/price_estimation/average_price_source.rs
```diff
@@ -1,41 +1,44 @@
...
-            let handle_1 = s.spawn(|_| self.source_1.lock().unwrap().get_prices(tokens));
-            (handle_0.join().unwrap(), handle_1.join().unwrap())
-        })
-        .unwrap();
-        match prices {
-            (Ok(p0), Ok(p1)) => Ok(average_prices(p0, p1)),
-            (Ok(p), Err(e)) | (Err(e), Ok(p)) => {
-                log::warn!("one price source failed: {}", e);
-                Ok(p)
+impl<T0: PriceSource + Send + Sync, T1: PriceSource + Send + Sync> PriceSource
```
nlordell, line 19: ubernit: maybe we can use a `where` clause here to make things a little more readable?

driver/src/price_estimation/average_price_source.rs
```diff
@@ -1,41 +1,44 @@
...
+    for AveragePriceSource<T0, T1>
+{
+    fn get_prices<'a>(
+        &'a self,
+        tokens: &'a [Token],
+    ) -> BoxFuture<'a, Result<HashMap<TokenId, u128>>> {
+        async move {
+            let prices = future::join(
```
nlordell, line 27: Any reason you prefer `future::join(...).await` over `futures::join!(...)`?

driver/src/price_estimation/dexag.rs
```diff
@@ -341,7 +352,7 @@ mod tests {
 
         let client = DexagClient::new(&HttpFactory::default()).unwrap();
         let before = Instant::now();
-        let prices = client.get_prices(tokens).unwrap();
+        let prices = client.get_prices(tokens).wait().unwrap();
```
nlordell, line 355: Does `now_or_never` not work in this case?

nlordell (approved):
LGTM, but maybe DexAg and Kraken price source to async/await should have been in their own PRs? It adds a little noise to some of the intricacies around mocking price sources for example.

driver/src/price_estimation/average_price_source.rs
e00E: In general if a function can accomplish the same thing I prefer the function. In this case I'm not sure what the difference even is internally, except that the macro always does `.await` internally. https://github.com/rust-lang/futures-rs/issues/2127

driver/src/price_estimation/dexag.rs
e00E: This is the online dexag client so the future is not going to be guaranteed ready.

driver/src/price_estimation/average_price_source.rs
nlordell: I think it also selects `join`, or `join3`, `join4`, etc. based on the number of arguments. So if you were to add a new future to join, then you would just have to add an additional parameter.

driver/src/price_estimation/dexag.rs
nlordell: :see_no_evil:

driver/src/price_estimation.rs
```diff
@@ -261,10 +270,13 @@ mod tests {
             .expect_get_prices()
             .withf(|tokens| tokens == [Token::new(2, "USDT", 6)])
             .returning(|_| {
-                Ok(hash_map! {
-                    TokenId(1) => 1_000_000_000_000_000_000,
-                    TokenId(2) => 1_000_000_000_000_000_000,
-                })
+                async {
```
fleupold, line 273: nit: Given how often we write `async { ... }.boxed()` in the tests, maybe worth while creating a macro for this?

driver/src/price_estimation/average_price_source.rs
```diff
@@ -1,41 +1,46 @@
...
 use crate::models::TokenId;
 use anyhow::{anyhow, Result};
-use crossbeam_utils::thread;
+use futures::future::{self, BoxFuture, FutureExt as _};
 use std::collections::HashMap;
-use std::sync::Mutex;
 
 /// Combines two prices into one average.
 pub struct AveragePriceSource<T0, T1> {
-    source_0: Mutex<T0>,
```
fleupold, line 10: Kill it with 🔥 🎉

driver/src/price_estimation/price_source.rs
```diff
@@ -57,20 +58,45 @@ impl Token {
...
 
 impl PriceSource for NoopPriceSource {
-    fn get_prices(&self, _: &[Token]) -> Result<HashMap<TokenId, u128>> {
-        Ok(HashMap::new())
+    fn get_prices(&self, _: &[Token]) -> BoxFuture<Result<HashMap<TokenId, u128>>> {
+        async { Ok(HashMap::new()) }.boxed()
     }
 }
+
+// We would like to tag `PriceSource` with `mockall::automock` but mockall does not support the
```
fleupold, line 81: Would it be easier for future maintenance to pass the tokens by value rather than reference?

driver/src/price_estimation/threaded_price_source.rs
fleupold: 👍 It already looks like we got rid of some threading logic in the PR.

fleupold (approved):
Agree that this PR became a little bit larger and had a bunch of non-trivial rewiring going on. But looks great!

driver/src/price_estimation/price_source.rs
e00E: Reference is more correct because the value does not need to be owned. I dislike changing our real code because of limitations in the mocking framework. The mockall developer made it sound like this is something that would be fixed in the future at which we no longer need the workaround.

driver/src/price_estimation.rs
e00E: We could put a `async_boxed!` macro in utils maybe but not in this PR.

driver/src/price_estimation/price_source.rs
e00E: I merged it already but after thinking about it again maybe the owned version is better because after all it is simpler and not in a place where performance matters. I might change it to that in another PR.

fleupold: Thanks, I'd appreciate that since this workaround is quite hacky and the developer didn't make it sound like the fix would happen in our lifetime 😉

PR #934, opened 2020-05-29
Introducing the 'best-ring-solver'

josojo:
In order to improve the robustness of our solver fleet, we want to battle best the best-ring solver, which can unblock the standard-solver, in case it finds only solutions with negative utilities.

This PR introduces best-ring-solver.

This is only a very simple introduction, I have not forgotten your requests for a simple solver interface, but this one should first be discussed here https://github.com/gnosis/dex-solver/issues/259
### Test Plan
CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ language: rust
 env:
   global:
     - OPEN_SOLVER_VERSION=v0.0.10
-    - PRIVATE_SOLVER_VERSION=v0.7.0
+    - PRIVATE_SOLVER_VERSION=v0.7.1
 
 install:
   - nvm install 12 && nvm alias default 12 && npm install -g yarn@latest && yarn --version
@@ -57,10 +57,16 @@ jobs:
         - docker-compose -f docker-compose.yml -f docker-compose.private-solver.yml up -d stablex
         - cargo test -p e2e ganache -- --nocapture
         - docker-compose logs
+        # StableX e2e Tests (Ganache) - private best-ring solver
+        - docker-compose down
+        - ./scripts/setup_contracts.sh
+        - docker-compose -f docker-compose.yml -f docker-compose.best-ring-solver.yml up -d stablex
+        - cargo test -p e2e ganache -- --nocapture
+        - docker-compose logs
 
     - name: Coverage
       rust: nightly
-      before_script: 
+      before_script:
         - ./scripts/setup_contracts.sh
       script:
         - curl --location https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -84,14 +90,13 @@ jobs:
...
-          tags: true
-
+        - provider: script
+          script: ./docker/deploy.sh $TRAVIS_BRANCH
+          on:
+            branch: master
+        - provider: script
+          script: ./docker/deploy.sh $TRAVIS_TAG
+          on:
+            tags: true
```
nlordell, line 102: Nice! I like the formatting improvements

README.md
```diff
@@ -202,8 +202,12 @@ OPTIONS:
...
             [default: 0]
         --solver-type <solver-type>
-            Which style of solver to use. Can be one of: 'NAIVE' for the naive solver; 'MIP' for mixed integer
-            programming solver; 'NLP' for non-linear programming solver [env: SOLVER_TYPE=]  [default: naive-solver]
+            Which style of solver to use. Can be one of: 'naive-solver' for the naive solver;
+            'standard-solver' for mixed integer programming solver;
+            'fallback-solver' for a more conservative solver than the standard solver;
+            'best-ring-solver' for a solver searching only for the best ring;
+            'open-solver' for the open-source solver 
+            [env: SOLVER_TYPE=]  [default: naive-solver]
```
nlordell, line 210: Could you also update the help in `driver/src/main.rs`?

driver/src/main.rs
```diff
@@ -68,9 +68,12 @@ struct Options {
     #[structopt(short = "i", long, env = "NETWORK_ID")]
     network_id: u64,
 
-    /// Which style of solver to use. Can be one of: 'NAIVE' for the naive
-    /// solver; 'MIP' for mixed integer programming solver; 'NLP' for non-linear
-    /// programming solver.
+    /// Which style of solver to use. Can be one of:
```
fleupold, line 71: ubernit: Not sure this comment is valuable here as it might easily be forgotten when a new flavor is added. I think the SolverType itself is documentation enough about the different solver types.

.travis.yml
```diff
@@ -57,10 +57,16 @@ jobs:
         - docker-compose -f docker-compose.yml -f docker-compose.private-solver.yml up -d stablex
         - cargo test -p e2e ganache -- --nocapture
         - docker-compose logs
+        # StableX e2e Tests (Ganache) - private best-ring solver
```
fleupold, line 60: Just to play devils advocate, why do we need to e2e test this solver flavor (we didn't test e.g. the fallback solver)?

driver/src/main.rs
nlordell: Personally, I think these comments should stay since they are used to generate commandline help when running `driver --help`. But it definitely can be abridged as you mentioned to not exhaustively include all the types.

.travis.yml
josojo: To catch errors between the interface of the driver and the optimizer. While the fallback solver was mostly a restriction of data (and - at the beginning - different price consumption), this is a substantially different codebase with new preprocessing.

PR #935, opened 2020-05-29
removing fallback solver

josojo:
Clean up:
The fallback solver does not require a specific enum value, as there are no driver implications for it. The only difference for the fallback solver compared to the standard solver is the restriction of the token set configured via white-lists.

The PR need modifications in our Kubernetes deployment.

### Test Plan
CI


e00E (approved):

README.md
```diff
@@ -204,7 +204,6 @@ OPTIONS:
         --solver-type <solver-type>
             Which style of solver to use. Can be one of: 'naive-solver' for the naive solver;
             'standard-solver' for mixed integer programming solver;
-            'fallback-solver' for a more conservative solver than the standard solver;
```
fleupold, line 207: I think this comment needs to also be updated in `main.rs` (unless we remove it there completely)

fleupold (approved):

README.md
nlordell: nit: this section should stay in sync with the output from `cargo run -p driver -- --help`.

nlordell (approved):

README.md
fleupold: It would be very fancy to add a travis check that greps for the output of `cargo run` in out README and fails if it's not in sync.

bh2smith (approved):
I notice that the title say removing fallback, but the code also includes the best ring. No big deal, but from the title and description alone I feel tricked.

nlordell:
> but the code also includes the best ring.

I think that was added in a separate PR...

PR #937, opened 2020-05-29
Remove mockall workaround

e00E:
Instead of using the workaround we change the trait to require and owned
value even this is technically not needed. This is an advantage because
the workaround is hacky and performance of cloning the tokens doesn't
matter.

### Test Plan
CI

### DIFF
--- a/driver/src/price_estimation.rs
+++ b/driver/src/price_estimation.rs
@@ -112,7 +112,7 @@ impl PriceOracle {
     }
 
     /// Gets price estimates for some tokens
-    fn get_prices(&self, tokens: &[Token]) -> HashMap<TokenId, u128> {
+    fn get_prices(&self, tokens: Vec<Token>) -> HashMap<TokenId, u128> {
         if tokens.is_empty() {
             return HashMap::new();
         }
@@ -130,7 +130,7 @@ impl PriceOracle {
 impl PriceEstimating for PriceOracle {
     fn get_token_prices(&self, orders: &[Order]) -> Tokens {
         let (tokens_to_price, unpriced_token_ids) = self.split_order_tokens(orders);
-        let prices = self.get_prices(&tokens_to_price);
+        let prices = self.get_prices(tokens_to_price.clone());
 
         tokens_to_price
             .into_iter()
@@ -268,7 +268,7 @@ mod tests {
         let mut source = MockPriceSource::new();
         source
             .expect_get_prices()
-            .withf(|tokens| tokens == [Token::new(2, "USDT", 6)])
+            .withf(|tokens| tokens.as_slice() == &[Token::new(2, "USDT", 6)])
             .returning(|_| {
                 async {
                     Ok(hash_map! {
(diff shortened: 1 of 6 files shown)


e00E:
We should discuss this a bit.
There is a limitation in mockall that does not allow us to use this signature with the second reference of the same lifetime after `&self`.
We can either use the previous workaround (see diff) or turn the reference into an owned value. What should we do?

I dislike changing our code because of a limitation in the testing framework. Based on our code the parameter *should* be a reference, not an owned value. 
I also recognize that using an owned value is easier to understand than the workaround.

nlordell:
> I dislike changing our code because of a limitation in the testing framework. Based on our code the parameter should be a reference, not an owned value.

I agree with both these statements.

> I also recognize that using an owned value is easier to understand than the workaround.

I also agree with this one.

I think this is one of those borderline cases where neither option is great. As someone new coming into the codebase they will either ask:
- Why is this by value when by reference makes more sense?
- What is this black magic `PriceSource_`?

Ultimately I ever so slightly prefer the work around, as it is clearly marked as a work around and links to the issue explaining why it is needed, even if how it works is arguably more complicated.

In the end, I don't feel strongly either way.

driver/src/price_estimation.rs
```diff
@@ -268,7 +268,7 @@ mod tests {
         let mut source = MockPriceSource::new();
         source
             .expect_get_prices()
-            .withf(|tokens| tokens == [Token::new(2, "USDT", 6)])
+            .withf(|tokens| tokens.as_slice() == &[Token::new(2, "USDT", 6)])
```
nlordell, line 271: ubernit: 
```suggestion
            .withf(|tokens| tokens == vec![Token::new(2, "USDT", 6)])
```

e00E:
Another argument to consider is that changing the signature will likely lead to other code assuming the new signature even though it was just intended for mockall. Then later maybe mockall fixes this issue or we switch mocking frameworks and it will be harder to change back to reference.
With this in mind I also prefer the original workaround.

fleupold:
I'm personally more bothered by the cognitive overhead of this wrapped price source mock, than by seeing an unnecessary clone, but if you guys feel differently that's fine too.

I do see the argument that in the future some purist might start spending and afternoon getting rid of the "pass by value" just to get to the same point where we are right now which would be a waste of time (Either way the issue should be documented).

An alternative could be to implement the mocked price source manually (given that it only has one method this might be simple enough). But that's also code to maintain and the wrapper might be simpler than a custom mock implementation.

nlordell:
> I'm personally more bothered by the cognitive overhead of this wrapped price source mock, than by seeing an unnecessary clone, but if you guys feel differently that's fine too.

For me, its just that the workaround (which may look weird), is very clear in that it is a workaround, and why it is needed, while the `clone()` is a bit more implicit and harder to understand why its needed (although, arguably much easier to understand what it is actually doing).

I'm also OK either way, they both have their pros and cons.

bh2smith (approved):

e00E:
Seems like we're keeping the workaround to me. Closing.

PR #938, opened 2020-05-29
[Easy] Reduce solver logging output

nlordell:
We ran into issues with the logging service that the solver was outputting a lot which was being grouped into a single extremely large log document and caused issues.

This PR addresses this by reducing the logging level for the solver as suggested here https://github.com/gnosis/dex-solver/issues/257

### Test Plan

CI (to make sure the option is accepted), next time we get a solver error ensure the log message is not so large. I'm open to suggestions on how to improve this test plan.


e00E (approved):

bh2smith (approved):

fleupold:
> I'm open to suggestions on how to improve this test plan

Maybe a lot of hassle, but you could get the offending instance from S3 and change the services code to pass that hardcoded instance file instead of the freshly generated current batch. Then the execution should fail again (assuming no major version update) and we should see the log.

fleupold (approved):

nlordell:
> Maybe a lot of hassle, but you could get the offending instance from S3 and change the services code to pass that hardcoded instance file instead of the freshly generated current batch.

I couldn't get the solver working. In an effort to make sure we don't run into logging issues again, I'll merge this and keep on trying to get things to run in the mean time. I'll open up a follow up PR in case I run into issues.

PR #942, opened 2020-06-02
Refactor Io Trait

e00E:
It is cleaner to have only one function that runs the solver. Splitting
it up into 3 methods has no benefit because they always have to be
called in this order.

### Test Plan
CI


fleupold (approved):
🌻

PR #944, opened 2020-06-02
Move folder creation into Io trait

e00E:
The trait is supposed to encapsulate io so it makes sense for it to
handle folder creation as well.

### Test Plan
CI


nlordell (approved):
Great change!

bh2smith (approved):

PR #945, opened 2020-06-02
Make PriceFinding async

e00E:
I decided to use the "blocking" crate to handle the blocking io for
running the solver. I prefer this over using asyncstd or tokio because
* It makes the code stay executor agnostic and use a more light weight
dependency for the same task.
* It creates less churn in our code because we do not need to change the
Io trait to async.
* The `Command` module we use to run the solver does not yet have an
async-std equivalent.

#843 

### Test Plan
CI

### DIFF
--- a/driver/Cargo.toml
+++ b/driver/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 anyhow = "1"
 bincode = "1.2.1"
+blocking = "0.4.4"
 byteorder = "1.3.4"
 chrono = "0.4.11"
 crossbeam-utils = "0.7"
--- a/driver/src/price_finding/optimization_price_finder.rs
+++ b/driver/src/price_finding/optimization_price_finder.rs
@@ -3,11 +3,11 @@ use crate::price_estimation::PriceEstimating;
 use crate::price_finding::price_finder_interface::{
     Fee, InternalOptimizer, PriceFinding, SolverType,
 };
-use crate::util::FutureWaitExt as _;
 
 use anyhow::{anyhow, Context, Result};
 use chrono::Utc;
 use ethcontract::U256;
+use futures::future::{BoxFuture, FutureExt as _};
 use log::error;
 use serde::{Deserialize, Serialize};
 use serde_with::rust::display_fromstr;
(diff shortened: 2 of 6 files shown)


driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -237,54 +238,65 @@ fn deserialize_result(result: String) -> Result<models::Solution> {
...
-
-        let input_folder = format!("instances/{}", &date);
-        let input_file = format!(
-            "{}/instance_{}_{}.json",
-            &input_folder,
-            &batch_id,
-            &now.to_rfc3339()
-        );
+    ) -> BoxFuture<'a, Result<models::Solution>> {
+        let price_oracle = &*self.price_oracle;
```
e00E, line 247: If we do `self.price_oracle.get_token_prices` as it used to be but in the async block then the compiler complains that `self.price_oracle` isn't `Send`.

driver/src/price_finding/optimization_price_finder.rs
```diff
@@ -642,6 +654,7 @@ pub mod tests {
                 &AccountState::with_balance_for(&orders),
                 Duration::from_secs(180)
             )
+            .wait()
```
e00E, line 657: Cannot be `now_or_never` because `blocking::unblock` is not immediately ready.

driver/src/price_finding/price_finder_interface.rs
```diff
@@ -157,12 +156,41 @@ pub fn execute_private_solver(
...
         time_limit: Duration,
-    ) -> Result<models::Solution, Error>;
+    ) -> BoxFuture<'a, Result<models::Solution, Error>>;
 }
+
+// We would like to tag `PriceFinding` with `mockall::automock` but mockall does not support the
+// lifetime bounds on `tokens`: https://github.com/asomers/mockall/issues/134 . As a workaround
+// we create a similar trait with simpler lifetimes on which mockall works.
+#[cfg(test)]
+mod mock {
```
e00E, line 172: This is the same workaround as before. I think it is the last instance where we need it.

driver/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 anyhow = "1"
 bincode = "1.2.1"
+blocking = "0.4.4"
```
nlordell, line 10: I think this was a really nice decision!

nlordell (approved):
Great PR!

driver/src/price_finding/optimization_price_finder.rs
bh2smith: Thanks for letting me know.

bh2smith (approved):

fleupold (approved):
Very nice!

Issue #948, opened 2020-06-03
Implement service that exposes price estimation functionality


fleupold:
We basically want to move https://github.com/gnosis/dex-price-estimator to native rust. We should discuss if the code should live in this repo or get its own.

e00E:
This is related to the pricegraph. Could someone reiterate the arguments for including it in the main repository?

I prefer having price-estimator-rust in it's own repository because it's an individual binary that is code wise unrelated to the driver. Pricegraph could be it's own repository too if is going to be used by both price-esimator-rust and dex-services (to calculate prices for the solver). We don't have to publish it to crates io either, it could be be a git dependency.
I prefer this because it keeps the git repository and history of separate projects separate.

e00E:
Discussed in meeting. Will keep in monorepo for now based on experience with dex-contracts.

e00E:
@gnosis/dex-services 

We need to decide how to reuse the event based orderbook in the new rust price estimator. Ideas:

1. Add the price estimator as another binary target to driver.
2. Add a library target to driver which exports all modules.
3. Add a library crate to the workspace into which the event based orderbook moves.

1 is little work but not clean because the two binaries do very different things. I don't like this long term.
2 is little work and cleaner than 1 because the binaries are separated. It would be cleaner to not export everything in the library when we only need the event based orderbook but that is more work.
3 is more work but the cleanest approach. It involves disentangling other driver dependencies of the event based orderbook like the StableXOrderBookReadingTrait.

I like 2 the most because it is practical right now. For 3 it is unclear how much benefit we would get out of this. We could do 3 later if we want to make another separate binary that also uses the event based orderbook.

nlordell:
We can also do a 2/3 hybrid where we create a new `gnop-core` that is exactly the `driver` `lib` target from #951 and then have a `driver` binary and `price-estimator`as separate crates.

e00E:
I like your suggestion the most. Changed the PR to that.

nlordell:
I had some thoughts about how we can structure the price estimator service for **MAXIMUM THROUGHPUT** :racehorse: :car: :rocket:

We could have `n = num_cpu` fibers, each with their own price estimator component (which owns the `pricegraph::Orderbook` for estimating prices).

The HTTP service would send a message to a bounded shared channel requesting a price estimation or transitive orderbook, along with a one shot channel for the result. The fibers would then take the request, process it, and then send the response back:

```rust
enum Request {
    PriceEstimate {
[...]

PR #951, opened 2020-06-03
Split driver into library "core" and binary

e00E:
This allows reusing the library parts in the price-estimator. Once we
have this change we can move modules between core, driver,
price-estimator to the place where they make the most sense.

#948 

# Test Plan
Check that CI still works.


bh2smith:
This seems like a fairly large change just to share some base components between places. Can we comment here on what is actually left in the driver and how this core components were selected. This kinda seems like something that could be changed as needed rather than all at once.

e00E:
Everything except main.rs has moved into the library. In that sense this is a small change. It's just that git shows it as a large number of moved files but the number of lines changed is low.

Starting the core library small with only what is needed for the rust price estimator ends up being a far bigger change than this because even just the event based orderbook depends on most of your modules.

core/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "core"
+version = "0.1.0"
+authors = ["Alexander Herrmann <alex@gnosis.io>", "Ben Smith <ben@gnosis.io>", "Felix Leupold <felix@gnosis.io>"]
```
nlordell, line 4: Maybe now is a good time to come up with the correct "authors" field? Maybe "Gnosis Protocol Developers"?

core/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+#![recursion_limit = "256"]
```
nlordell, line 1: Just curious, what macro expansion made this needed?

nlordell (approved):
Looks great!

nlordell:
> This kinda seems like something that could be changed as needed rather than all at once.

We discussed this in the issue: https://github.com/gnosis/dex-services/issues/948. @bh2smith From your :rocket: emoji I understood that you agreed with this suggestion.

core/Cargo.toml
e00E: Do we even need authors? We don't publish these crates.

core/src/lib.rs
e00E: I kept it because it was there originally. Iirc this was introduced when we were using the select! macro for the event based orderbook. I doubt it's needed anymore.

nlordell: Oh my bad, I thought it was added here!

nlordell: I would just leave it as is then.

core/Cargo.toml
nlordell: I’m ok either way

nlordell: I’m ok either way

bh2smith: Seems like a good time to update the authors!

bh2smith:
Yes, its all good now! I put the 🚀after commenting here.

bh2smith (approved):

e00E:
FYI I accidentally merged "turn U256 batch id into u32" into here instead of into master separately but I didn't feel it was worth the work to revert that.

PR #953, opened 2020-06-04
Generalized Average Price Source

### LINKED ISSUE #838: Generalize; AveragePriceSource to accept n Sources
It seems the average price source was created to take the average over 2 price sources. This seems as if it is begging to be generalized to arbitrarily many.

bh2smith:
Fixes #838

This PR takes the perviously hardcoded two price source struct and expands to allow a vector of prices sources. We handle the necessary generalized form of averaging prices by implementing a "Lossless merger" of HashMap collections. This means, in particular, that if one token in found on one price source, but no others, an average can still be made from those available. 

### Test Plan
```
cargo test lossless_merge
```
Tests the merge of a vector of hash maps into an array of the initial values. 

```
cargo test average_prices
```
Demonstrate how the averaging technique is expected to play out in all possible scenarios.


e00E (approved):

driver/src/price_estimation/average_price_source.rs
```diff
@@ -2,89 +2,116 @@ use super::{PriceSource, Token};
...
+    let complete_key_set: HashSet<_> = map_collection.iter().map(|m| m.keys()).flatten().collect();
+    let mut gathered_maps: HashMap<T, Vec<U>> = HashMap::new();
+    for key in &complete_key_set {
+        let mut place_holder = vec![];
+        for map in &map_collection {
+            if map.get(key).is_some() {
+                let value = map.get(key).unwrap();
+                place_holder.push(*value);
+            }
+        }
```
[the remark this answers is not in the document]
bh2smith: I suppose I could go even further here and eliminate the loop over key_set... actually this suggestion looks like it is not doing what I expect.. These inserts will overwrite the values as they are inserted...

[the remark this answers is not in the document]
bh2smith: I'm sorry, but there is something wrong here.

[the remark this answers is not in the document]
bh2smith: Sorry, it looks fine. misread... Will put it back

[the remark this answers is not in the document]
bh2smith: actually there was a line that should have been removed and a missing bracket. Have replaced your suggestion and will make sure you get the street cred!

driver/src/price_estimation/average_price_source.rs
```diff
@@ -2,89 +2,113 @@ use super::{PriceSource, Token};
...
-    prices_0: HashMap<TokenId, u128>,
-    mut prices_1: HashMap<TokenId, u128>,
-) -> HashMap<TokenId, u128> {
-    for (token_id, price_1) in prices_0 {
-        prices_1
-            .entry(token_id)
-            .and_modify(|price| *price = (*price + price_1) / 2)
-            .or_insert(price_1);
+/// Lossless merger of a collection of maps puts all available values into a list for each available key
+pub fn lossless_merge<T: Copy + Eq + Hash + PartialEq, U: Clone + Copy>(
```
nlordell, line 49: ubernit: I would make this not `pub` since it isn't used outside the module.

driver/src/util.rs
```diff
@@ -1,6 +1,5 @@
 use ethcontract::U256;
 use std::future::Future;
-
```
nlordell, line 3: nit: the newline there was nice.

driver/src/price_estimation/average_price_source.rs
nlordell, line 49: ubernit: Because you are passing in the `map_collection` by value, you should be able to implement this without `U: Clone + Copy`. It would involve using `into_iter` instead of `iter` to get an interator on owned `U`s.

Since we are only using this for `u128`s here, this isn't an issue. But it would prevent this method from being used by types where `U: !Copy`.

nlordell (approved):
Nice! Looks real good! Just a couple of tiny nits (that you can safely ignore).

driver/src/price_estimation/average_price_source.rs
bh2smith: So I tried to remove the Clone and Copy from U, but then had to remove a copied inside the filter map and now somehow wind up with something like 

```
a value of type `std::vec::Vec<U>` cannot be built from an iterator over elements of type `&U`
```

driver/src/price_estimation/average_price_source.rs
```diff
@@ -2,89 +2,113 @@ use super::{PriceSource, Token};
...
-    prices_0: HashMap<TokenId, u128>,
-    mut prices_1: HashMap<TokenId, u128>,
-) -> HashMap<TokenId, u128> {
-    for (token_id, price_1) in prices_0 {
-        prices_1
-            .entry(token_id)
-            .and_modify(|price| *price = (*price + price_1) / 2)
-            .or_insert(price_1);
+/// Lossless merger of a collection of maps puts all available values into a list for each available key
+fn lossless_merge<T: Copy + Eq + Hash + PartialEq, U: Copy + Clone>(
```
fleupold, line 49: Nice use of generics here.

fleupold (approved):
🍠

driver/src/price_estimation/average_price_source.rs
nlordell: You might have to use `into_iter` instead of `iter` and `remove` instead of `get`, since you need to transfer ownership.

driver/src/price_estimation/average_price_source.rs
bh2smith: Thanks, and with a little assistance from @nlordell, it has become even more generic!

PR #954, opened 2020-06-04
Turn batch id into u32 from U256

e00E:
In the contract this is a u32 so there is no reason to use U256.

### Test Plan
CI.


bh2smith (approved):
Totally, nice catch!

nlordell (approved):
Great! This one had been bothering me for a while.

josojo (approved):
Nice, thanks!

Issue #947, opened 2020-06-03
[Pricegraph] Implement Orderbook "view"


fleupold:
We are currently using the js based "transitive orderbook" logic to visualize the orderbook in the frontend (cf. https://mesadev.eth.link/#/book)

The problem with this is that it takes up to 10s to load. The graph based system is already superior for price estimates. The goal of this issue is to also use this approach to display the orderbook (list of pairs of price and available volume at that price).

bh2smith:
Could potentially limit ourselves to a specific spread for improved accuracy and speed.

Limiting the number of hops wouldn't improve performance (using Bellman-Ford algorithm). However, there are some other algorithms (adaptations of BF - #875) that might be advantageous.

nlordell:
# Transient Orderbooks

I'm doing some investigation around using a graph approach to calculating transient orderbooks. Unfortunately it isn't very obvious to me what the "correct" and "expected" mean for transient orderbooks, specifically when two sides of the transitive orderbook share orders. Consider the following order graph (assuming 0 fees) where nodes represent tokens and directed edges are orders starting at the sell token with a weight of sell amount over buy amount, so: `SELL ---sellAmt:buyAmt---> BUY`.

![image](https://user-images.githubusercontent.com/4210206/83733422-cc63b300-a64d-11ea-82c3-5474bf1965e5.png)

What is special about this orderbook is that the transient orders `A -> C -> D -> B` (with a total price of 1:2) and `B -> C -> D -> A` (also with a total price of 1:2) both require the same order `C -> D`. What does this mean when calculating the transient orderbook? Assuming sufficient balances:

Is this the orderbook?
```
amount
  ^
  |
2 |----+                                       +----
  |####|                                       |0000
  |####|                                       |0000
  |####|                                       |0000
1 |####+---------+                   +---------+0000
  |##############|                   |00000000000000
  |##############|                   |00000000000000
  |##############|                   |00000000000000
0 +----+---------+---------+---------+---------+----> price
       0.25      0.5       1         2         4
```

The issue is that this is a bit misleading. If we place an `A -> B` order for `2:1`, then the orderbook would become:
```
amount
  ^
  |
3 |----+
  |####|
  |####|
  |####|
2 |####+---------+                             +----
  |##############|                             |0000
  |##############|                             |0000
  |##############|                             |0000
1 |##############+-------------------+---------+0000
  |##################################|00000000000000
  |##################################|00000000000000
  |##################################|00000000000000
[...]

PR #955, opened 2020-06-04
[Pricegraph] Compute overlapping transitive orderbook

nlordell:
First step towards #947 

This PR adds a new method to the pricegraph orderbook for reducing overlapping orders in a single market (i.e. base/quote token pair) and returning the transitive orderbook for those orders.

### Test Plan

New unit test for `find_cycle` (as it was modified to allow specifying a cycle origin) and new unit test that computes the transient orderbook for overlapping orders in a specific auction.


pricegraph/src/orderbook.rs
```diff
@@ -140,6 +141,69 @@ impl Orderbook {
...
+                    "failed to fill path along detected negative cycle {}",
+                    format_path(&path),
+                )
+            });
+        }
+
+        overlap
+    }
+
+    /// Computes the equivalent transient order along a path.
```
e00E, line 190:
```suggestion
    /// Computes the equivalent transitive order along a path.
```
I have infected you with this word :) .

e00E (approved):

pricegraph/src/orderbook.rs
nlordell: :rofl:

pricegraph/src/graph/path.rs
```diff
@@ -2,33 +2,54 @@
...
 
-    // NOTE: `current` is now guaranteed to be on the cycle, so just walk
-    // backwards until we reach `current` again.
-    let start = current;
-    let mut path = Vec::with_capacity(predecessor.len());
+    // NOTE: Allocate the cycle vector with enough capacity for the negative
+    // cycle plus one. This extra capacity is used by the orderbook graph to
+    // create a circular path equivalent to the negative cycle.
+    let len = step + 1 - visited[cursor.index()];
+    let mut path = Vec::with_capacity(len + 1);
```
fleupold, line 37: nit: Isn't it a bit hacky that we overallocate the vector for use in a specific case. Should this method really now how it's going to be used?

pricegraph/src/orderbook.rs
```diff
@@ -140,6 +141,69 @@ impl Orderbook {
...
+            bellman_ford::search(&self.projection, quote)
+        {
+            let path = {
+                let mut cycle = path::find_cycle(&predecessors, node, Some(quote))
+                    .expect("negative cycle not found after being detected");
+                cycle.push(cycle[0]);
+                cycle
+            };
+
+            if path[0] == quote {
```
fleupold, line 164: ubernit: We now paths is not empty but it's technically not checked anywhere. Maybe using first here is "safer"

pricegraph/src/orderbook.rs
```diff
@@ -140,6 +141,69 @@ impl Orderbook {
...
+                if let Some(base_index) = path.iter().position(|node| *node == base) {
```
fleupold, line 165: nit: could maybe use `path.split` here?

pricegraph/src/orderbook.rs
```diff
@@ -140,6 +141,69 @@ impl Orderbook {
...
+    /// Returns `None` if the path does not exist.
+    fn find_transitive_order(&self, path: &[NodeIndex]) -> Option<TransitiveOrder> {
+        let (capacity, price) = self.find_path_capacity_and_price(path)?;
+
+        // NOTE: We now have the capacity and price for this transitive order
+        // which needs to be converted to a buy and sell amount. We have:
+        // - `price = FEE_FACTOR * buy_amount / sell_amount`
+        // - `capacity = sell_amount * price`
+        // Solving for `buy_amount` and `sell_amount`, we get:
+        let buy = capacity / FEE_FACTOR;
```
fleupold, line 201: Just to double check, the fees for each sub-order of the path is already considered in the original orders, right?

fleupold (approved):
🚀👨‍🔬

pricegraph/src/graph/path.rs
nlordell: Yeah, I can refactor this bit.

pricegraph/src/orderbook.rs
nlordell: Unfortunately, we need slightly different semantics, as we want the `base` token to be in both sub-paths.

pricegraph/src/orderbook.rs
nlordell: I like using `first` better here as well. It relies on less implicit assumptions about the path, and the compiler should be smart enough to optimize this away anyways.

pricegraph/src/lib.rs
```diff
@@ -9,3 +9,35 @@ mod data;
...
+///
+/// A transitive order is defined as the transitive combination of multiple
+/// orders into a single equivalent order. For example consider the following
+/// two orders:
+/// - *A*: buying 1_000_000 token 1 selling 2_000_000 token 2
+/// - *B*: buying 4_000_000 token 2 selling 1_000_000 token 3
+///
+/// We can define a transitive order *C* buying 1_000_000 token 1 selling
+/// 500_000 token 3 by combining *A* and *B*. Note that the sell amount of token
+/// 3 is limited by the token 2 capacity for this transitive order.
```
bh2smith, line 34: Awesome description and example to go along with it. Is there any benefit to having so many zeros here? Looks like we could factor out 10^5 here...

pricegraph/src/orderbook.rs
nlordell: They are considered in the prices since they are calculated with `price = FEE_FACTOR * buy_amount / sell_amount`. So that means that if the transitive order "hops" over 3 orders (`A -> B -> C -> D`, each arrow being an order), for example, it would introduce fees of `FEE_FACTOR ** 3` (one for each hop). The last one we deduct, since orders implicitly include fees - so in our example, we would end up with `FEE_FACTOR ** 2`, which from what I gather is what we want.

bh2smith (approved):
Excellent progress here! really happy to see such well documented path finding algorithms.

pricegraph/src/orderbook.rs
```diff
@@ -606,6 +670,51 @@ mod tests {
...
+
+        let overlap = orderbook.reduce_overlapping_transitive_orderbook(1, 2);
+
+        // Transitive order `2 -> 3 -> 1`
+        assert_approx_eq!(overlap.asks[0].buy, 500_000.0);
+        assert_approx_eq!(overlap.asks[0].sell, 500_000.0 / FEE_FACTOR);
+
+        // Transitive order `1 -> 2`
+        assert_approx_eq!(overlap.bids[0].buy, 1_000_000.0);
+        assert_approx_eq!(overlap.bids[0].sell, 2_000_000.0);
```
nlordell, line 715: @fleupold Regarding your question about the fees, the example should *hopefully* help clarify:
- The transitive order of `1 -> 2` is just equal to the order, so buying `1_000_000` and selling `2_000_000`
- The transitive order for `2 -> 3 -> 1` combines two orders, so the fee needs to be added once to the final price (to account for lost fees needed for going over an additional order).

pricegraph/src/lib.rs
nlordell: I kept it there because of things like the minimum traded amount. However, I can just express them in base units (so write it like `1.0`, etc.).

pricegraph/src/graph/path.rs
nlordell: I modified `find_path` to now return an actual path along the cycle, i.e. the first and last items are guaranteed to be the same.

I feel like this makes things less hacky since we are no longer over-allocating the vector, but instead just changed the semantics of the method. Note that `find_cycle` was only ever used this way, so it got rid of some `cycle.push(cycle[0])` which was already a bit ugly in my opinion.

PR #956, opened 2020-06-05
Refactor Order to include all order information

e00E:
Previously we had split up the order information into
StableXAuctionElement which contained sell_token_balance, valid_from,
valid_until and Order which contained the modified numerator.
Now Order contains all order specific information: valid_from,
valid_until and the original numerator, denominator and remaining
amount.

This is a cleaner approach as we propagate the original order
information for as long as possible. Only when the order is passed to
the solver we calculate the new numerator and denominator so that there
is no remaining amount.

While working on this I noticed that Order derived a couple of unused
traits which I have removed.

The motivation for this change is that I ran into it when connecting the pricegraph with the event based orderbook for the rust price estimator. This change is probably not required because we can fake the order validity and use the transformed numerator and denominator but I feel that this is an improvement to the code base overall.

### Test Plan
CI. Double check that I haven't messed up the numerator, denominator, remaining transformation anywhere.


core/src/contracts/stablex_auction_element.rs
```diff
@@ -44,62 +40,29 @@ impl StableXAuctionElement {
...
+                denominator,
+                remaining,
+                valid_from,
+                valid_until,
             },
         }
     }
 }
 
-pub fn compute_buy_sell_amounts(
```
e00E, line 64: This function and its testes have been moved into `models/order.rs`.

core/src/price_finding/optimization_price_finder.rs
```diff
@@ -147,12 +147,13 @@ mod solver_input {
 
     impl From<&'_ models::Order> for Order {
         fn from(order: &models::Order) -> Self {
+            let (buy_amount, sell_amount) = order.compute_buy_sell_amounts();
```
e00E, line 150: !

core/src/price_finding/optimization_price_finder.rs
```diff
@@ -147,12 +147,13 @@ mod solver_input {
...
             Order {
                 account_id: order.account_id,
                 sell_token: TokenId(order.sell_token),
                 buy_token: TokenId(order.buy_token),
-                sell_amount: Num(order.sell_amount),
-                buy_amount: Num(order.buy_amount),
+                sell_amount: Num(sell_amount),
+                buy_amount: Num(buy_amount),
```
e00E, line 156: !

core/src/contracts/stablex_auction_element.rs
```diff
@@ -44,62 +40,29 @@ impl StableXAuctionElement {
...
                 account_id,
                 buy_token,
                 sell_token,
-                buy_amount,
-                sell_amount,
+                numerator,
+                denominator,
+                remaining,
+                valid_from,
+                valid_until,
```
e00E, line 54: !

core/src/contracts/stablex_auction_element.rs
```diff
@@ -44,62 +40,29 @@ impl StableXAuctionElement {
...
-        0
-    };
-    (buy_amount, remaining)
-}
-
 #[cfg(test)]
 pub mod tests {
     use super::*;
 
     fn emptyish_auction_element() -> StableXAuctionElement {
```
nlordell, line 64: nit: I would argue we could just remove `emptyish_auction_element` in favour of `Default::default()`.

core/src/orderbook/auction_data_reader.rs
```diff
@@ -53,7 +53,7 @@ impl AuctionDataReader {
         let index = self.index;
         let auction_elements = auction_elements
             .iter()
-            .filter(|x| x.in_auction(index) && x.order.sell_amount > 0);
+            .filter(|x| x.in_auction(index) && x.order.remaining > 0);
```
nlordell, line 56: Does the check also require the sell token balance to be `> 0`? I don't remember if the balance was somehow factored into the buy/sell amount calculations.

core/src/models/order.rs
```diff
@@ -21,10 +23,34 @@ impl Order {
...
-            sell_amount: 1_000_000_000_000_000_000,
+            numerator: 1_000_000_000_000_000_000,
+            denominator: 1_000_000_000_000_000_000,
+            remaining: 1_000_000_000_000_000_000,
+            valid_from: 0,
+            valid_until: 0,
         }
     }
+
+    pub fn compute_buy_sell_amounts(&self) -> (u128, u128) {
```
nlordell, line 34: I like that this is a method on `Order` and not a function now!

nlordell (approved):
Very welcome change! I really like this cleanup.

core/src/orderbook/auction_data_reader.rs
fleupold: This is a bit subtle, but we need to include orders even if they have sellTokenBalance of 0. Let's say the order is from token A->B. The solver could trade another order (C->A) of that user in the same batch to increase their sellTokenBalance of A. In that case we want the first order to still be in the instance so that the new proceeds of token A could potentially be traded right away into B.

core/src/models/order.rs
```diff
@@ -1,15 +1,17 @@
...
 pub struct Order {
     pub id: u16,
     pub account_id: Address,
     pub buy_token: u16,
     pub sell_token: u16,
-    pub buy_amount: u128,
-    pub sell_amount: u128,
+    pub numerator: u128,
+    pub denominator: u128,
+    pub remaining: u128,
```
fleupold, line 12: nit: Can we call this explicitly `remaining_sell[_amount]`?

core/src/models/order.rs
fleupold: Agree. Less important than on the struct field, but it could be nice to also rename this to be `compute_remaining_buy_sell_amounts`

fleupold (approved):
Awesome cleanup 🎇

bh2smith (approved):
Nice changes here. I am always confused by the numerator and denominator though because I can't recall which is the buy amount and which is the sell amount. I would really like if there was a comment somewhere that I could refer to when I forget (which is always).

PR #957, opened 2020-06-05
Set rust version to 1.43.0 temporarily

e00E:
The new rust version introduced more warnings or lints that make master
fail in the ethcontract generated code. https://travis-ci.com/github/gnosis/dex-services/jobs/343998701
Let's temporarily use the previous version.

Btw even though master failed I didn't get an email.

### Test Plan
CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ jobs:
   include:
     - name: "Build and Local Testing"
       if: type = pull_request
-      rust: stable
+      rust: 1.43.0
       cache:
         directories:
           - $HOME/.cache/yarn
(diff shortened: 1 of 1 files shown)


nlordell:
Are you sure this is the issue? We have a bunch of builds that succeeded with 1.44.0, like this one: https://travis-ci.com/github/gnosis/dex-services/jobs/344030434#L192

e00E:
You're right, but then what made CI fail. I had assumed travis had switched to the new rust version between the PR appearing green and the master deploy happening but as you say it was already on the new version.

nlordell:
Also, I just build locally with 1.44 and had no warnings.

bh2smith (approved):

e00E:
I also tried cargo clean, then build again and no warning.

e00E:
Could it be that the travis deploy that only runs on master (not PRs) didn't have the rust version specified.

PR #959, opened 2020-06-05
Add price estimator scaffolding

e00E:
This is the initial commit for the rust price estimator. It sets up the
smart contract but does not yet create an orderbook or handle the http
requests we want.

For the http server libraries I have looked at:
* https://github.com/http-rs/tide
* https://github.com/hyperium/hyper
* https://github.com/seanmonstar/warp
* https://github.com/tomaka/rouille

tide is based on async std. I did not choose it because the readme says
that is not ready for production.

rouille is the http server we use for metrics in the driver. It does not
use async which makes it slower than the async frameworks (see benchmark
in their readme). This speed difference is probably not important for
our task but we might as well use an async framework from the get go.

warp is based on hyper. It easier to use because hyper is too low
level. Both are linked in the tokio readme.

### Test Plan
You can curl against the server but there are no tests yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,9 +3,10 @@ members = [
     "core",
     "driver",
     "e2e",
+    "price-estimator",
     "pricegraph",
-    "pricegraph/wasm",
     "pricegraph/fuzz",
+    "pricegraph/wasm",
 ]
 default-members = [
     "driver",
--- a/price-estimator/src/main.rs
+++ b/price-estimator/src/main.rs
@@ -1,0 +1,96 @@
+use core::{
+    contracts::{stablex_contract::StableXContractImpl, web3_provider},
+    http::HttpFactory,
+    metrics::{HttpMetrics, MetricsServer},
+    util::FutureWaitExt as _,
+};
+use ethcontract::PrivateKey;
+use prometheus::Registry;
+use std::{num::ParseIntError, path::PathBuf, sync::Arc, thread, time::Duration};
+use structopt::StructOpt;
+use tokio::runtime;
+use url::Url;
+use warp::Filter;
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "price estimator", rename_all = "kebab")]
+struct Options {
+    /// The log filter to use.
+    ///
+    /// This follows the envlogger syntax (e.g. 'info,driver=debug').
+    #[structopt(
+        long,
+        env = "LOG_FILTER",
+        default_value = "warn,driver=info,price_estimator=info"
+    )]
+    log_filter: String,
+
+    /// The Ethereum node URL to connect to. Make sure that the node allows for
+    /// queries without a gas limit to be able to fetch the orderbook.
+    #[structopt(short, long, env = "NODE_URL")]
+    node_url: Url,
+
+    /// The timeout in seconds of web3 JSON RPC calls.
+    #[structopt(
+        long,
+        env = "TIMEOUT",
+        default_value = "10",
+        parse(try_from_str = duration_secs),
+    )]
+    timeout: Duration,
+
+    #[structopt(long, env = "ORDERBOOK_FILE", parse(from_os_str))]
+    orderbook_file: Option<PathBuf>,
+
+    #[structopt(
+        long,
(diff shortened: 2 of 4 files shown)


price-estimator/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+
+fn main() {
+    let options = Options::from_args();
+    env_logger::init();
+    log::info!(
+        "Starting price estimator with runtime options: {:#?}",
+        options
+    );
+
+    let driver_http_metrics = setup_driver_metrics();
```
bh2smith, line 62: Is this really "driver" metrics or is this just `http_metrics`? Looks like we may have recycled a term from the driver code here.

price-estimator/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+    // TODO: create event based orderbook
+    // TODO: handle http requests
+
+    let mut runtime = runtime::Builder::new()
+        .threaded_scheduler()
+        .enable_all()
+        .build()
+        .unwrap();
+
+    runtime.block_on(warp::serve(warp::any().map(|| "")).run(([127, 0, 0, 1], 8080)));
```
bh2smith, line 82: I see we are mapping to an empty string here. Is this a hack or does it mean something?

bh2smith (approved):
Looks pretty good so far! Thanks for keeping it short and easy to digest!

price-estimator/src/main.rs
e00E: I called it driver because presumably we will also have price estimator metrics but I wouldn't worry about the name yet.

price-estimator/src/main.rs
e00E: It means that every http requests is responded with an empty string. The code that handles the real requests comes later.

price-estimator/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+    log::info!(
+        "Starting price estimator with runtime options: {:#?}",
+        options
+    );
+
+    let driver_http_metrics = setup_driver_metrics();
+    let http_factory = HttpFactory::new(options.timeout, driver_http_metrics);
+    let web3 = web3_provider(&http_factory, options.node_url.as_str(), options.timeout).unwrap();
+    // The private key is not actually used but StableXContractImpl requires it.
+    let private_key = PrivateKey::from_raw([1u8; 32]).unwrap();
```
nlordell, line 66: ubernit: Doesn't `PrivateKey` have a `Default` implementation.

**edit**: It does not! It probably should, and it should use the "one key", which is [0,...0, 1], the first valid private key in secp256k1.

price-estimator/src/main.rs
nlordell: I think it should have written out "The name is Price, Price Estimator". :stuck_out_tongue:

Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
     "pricegraph",
     "pricegraph/wasm",
     "pricegraph/fuzz",
+    "price-estimator",
```
nlordell, line 9: ubernit: `price-` comes before `priceg` when sorting alphabetically.

nlordell (approved):
Very nice starting point. I mostly agree with your reasoning except for "the speed difference is probably not important for our task". I think we should be careful about performance for this particular service since it is customer facing and speedy responses go a long way towards happier users.

I do agree with your choice of web framework given the ones you look at. Did you take a look at [actix-web](https://github.com/actix/actix-web)? I've heard some good things about it as well (note, I am 100% happy with the decision made here - just thought you might be interested because of your own curiosity :smile:)

Either way, we should try and structure the application in a way were we detach the service implementation (i.e. how the price estimation and transitive orderbook computation is done) and the HTTP framework.

e00E:
> Did you take a look at [actix-web](https://github.com/actix/actix-web)?

Not in detail. With actix I remembered the unsafe problems it had and it seems like it is heavier / doing more work than we need.

price-estimator/src/main.rs
fleupold: Should private_key be optional in ethcontracts in case you only want to read from a contract?

PR #961, opened 2020-06-08
Price Estimator: Set up orderbook

e00E:
Add an Orderbook to the price estimator and set up one task to update
and another task to serve http requests.
To handle requests the get_reduced_orderbook method on Orderbook is
going to be used and the resulting pricegraph::Orderbook will be moved
into the function that handles the request.

### Test Plan
CI but not really applicable yet.

### DIFF
--- a/price-estimator/src/main.rs
+++ b/price-estimator/src/main.rs
@@ -1,14 +1,18 @@
+mod orderbook;
+
 use core::{
     contracts::{stablex_contract::StableXContractImpl, web3_provider},
     http::HttpFactory,
     metrics::{HttpMetrics, MetricsServer},
+    orderbook::EventBasedOrderbook,
+    orderbook::StableXOrderBookReading,
     util::FutureWaitExt as _,
 };
 use ethcontract::PrivateKey;
 use prometheus::Registry;
 use std::{num::ParseIntError, path::PathBuf, sync::Arc, thread, time::Duration};
 use structopt::StructOpt;
-use tokio::runtime;
+use tokio::{runtime, time};
 use url::Url;
 use warp::Filter;
 
@@ -51,6 +55,8 @@ struct Options {
     orderbook_update_interval: Duration,
 }
 
+type Orderbook = orderbook::Orderbook<EventBasedOrderbook>;
+
 fn main() {
     let options = Options::from_args();
     env_logger::init();
@@ -64,14 +70,15 @@ fn main() {
     let web3 = web3_provider(&http_factory, options.node_url.as_str(), options.timeout).unwrap();
     // The private key is not actually used but StableXContractImpl requires it.
     let private_key = PrivateKey::from_raw([1u8; 32]).unwrap();
-    let _contract = Arc::new(
+    let contract = Arc::new(
         StableXContractImpl::new(&web3, private_key, 0)
             .wait()
             .unwrap(),
     );
 
-    // TODO: create event based orderbook
-    // TODO: handle http requests
+    let orderbook = EventBasedOrderbook::new(contract, web3, options.orderbook_file);
+    orderbook.initialize().wait().unwrap();
+    let orderbook = Arc::new(Orderbook::new(orderbook));
 
     let mut runtime = runtime::Builder::new()
         .threaded_scheduler()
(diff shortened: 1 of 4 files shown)


price-estimator/src/main.rs
```diff
@@ -51,6 +55,8 @@ struct Options {
     orderbook_update_interval: Duration,
 }
 
+type Orderbook = orderbook::Orderbook<EventBasedOrderbook>;
+
 fn main() {
```
nlordell, line 60: Maybe this should be `async main` now (because we have futures in the main method now)? You could use `tokio::main` macro for that.

price-estimator/src/orderbook.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ethcontract::{Address, U256};
+    use std::collections::HashMap;
+
+    #[test]
+    fn order_to_element_() {
```
nlordell, line 97: ubernit: Not of fan of the `method_` naming, maybe the test name could describe what the test is doing like:
`order_is_converted_to_auction_element`.

price-estimator/src/main.rs
```diff
@@ -64,22 +70,45 @@ fn main() {
...
             .unwrap(),
     );
 
-    // TODO: create event based orderbook
-    // TODO: handle http requests
+    let orderbook = EventBasedOrderbook::new(contract, web3, options.orderbook_file);
+    orderbook.initialize().wait().unwrap();
+    let orderbook = Arc::new(Orderbook::new(orderbook));
 
     let mut runtime = runtime::Builder::new()
```
nlordell, line 83: *edit*: I see that we have specific runtime configuration here. Maybe we should just call an `async run`. I think `tokio` provides `spawn` methods for spawning fibers in an `async` context.

price-estimator/src/orderbook.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub fn new(orderbook_reading: T) -> Self {
+        Self {
+            orderbook_reading,
+            reduced_pricegraph_orderbook: RwLock::new(pricegraph::Orderbook::from_elements(
+                std::iter::empty(),
+            )),
+        }
+    }
+
+    pub async fn _get_reduced_orderbook(&self) -> pricegraph::Orderbook {
```
nlordell, line 26: Ah, cleaver, so you don't need an `#[allow(dead_code)]`

nlordell (approved):
Yay! :tada:

price-estimator/src/main.rs
e00E: As you mentioned below, the reason I didn't use tokio::main is that we configure the runtim. I wasn't sure how tokio configures it automatically when don't do that. At the moment I prefer being explicit in this case.

price-estimator/src/main.rs
```diff
@@ -64,22 +70,45 @@ fn main() {
...
         .build()
         .unwrap();
 
-    runtime.block_on(warp::serve(warp::any().map(|| "")).run(([127, 0, 0, 1], 8080)));
+    let orderbook_task = runtime.spawn(update_orderbook_forever(
+        orderbook,
+        options.orderbook_update_interval,
+    ));
+
+    // TODO: handle http requests
```
bh2smith, line 94: Is this to be done here in this PR or another?

price-estimator/src/orderbook.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl<T: StableXOrderBookReading> Orderbook<T> {
+    /// Recreate the pricegraph orderbook.
+    pub async fn update(&self) -> anyhow::Result<()> {
+        let (account_state, orders) = self
+            .orderbook_reading
+            .get_auction_data(current_batch_id() as u32)
+            .await?;
+
+        // TODO: Move this cpu heavy computation out of the async function using spawn_blocking.
```
bh2smith, line 39: Cool idea!

price-estimator/src/orderbook.rs
bh2smith: Or, would it be a reasonable compromise to use `test_method_name`? I believe this is a common convention.

bh2smith (approved):

price-estimator/src/main.rs
e00E: Another

price-estimator/src/orderbook.rs
e00E: The reason I use this name is that we are testing that whole function. I don't see how another name is helpful. I would use a more detailed name if we had more test cases like `convert_order_large_number`. Likewise prepending `test_` doesn't add any information because it is already clear that this is a test.

nlordell: > The reason I use this name is that we are testing that whole function. I don't see how another name is helpful. I would use a more detailed name if we had more test cases like `convert_order_large_number`.

My main argument is that it is not necessarily testing the whole function, but a single aspect of it. Additional tests (like `convert_order_large_number`) will be testing other aspects of this function, and if and when they are added, the original test is no longer testing the whole function and is no longer accurately named. This could happen if some refactoring or new code introduces new branches that need to be tested and deserve their own unit tests. As the code get modified, the original goal of the unit test might become obscured. Furthermore, a refactor might split the method into smaller pieces, in which case this unit test will be for a new method.

That being said, for this method I think the way it currently is is fine. It really was an **uber**nit and I don't think it necessarily needs to change.

> Likewise prepending test_ doesn't add any information because it is already clear that this is a test.

I also agree with this, especially since the function is already in a tests module so it shows up as `tests::method_name` and have a clear `#[test]` attribute.

PR #963, opened 2020-06-08
Price Estimator: Set up handling of estimated_buy_amount requests

e00E:
This is the first version that works and can be tested with curl.
In another PR I will refactor the estimate_price function into it's own
module and make it use the same json format as the nodejs price
estimator.

### Test Plan
Ci and `curl`ing localhost. I have an event based orderbook file that I keep out of the repository for fast start ups without having to query old events.


price-estimator/src/main.rs
```diff
@@ -1,34 +1,25 @@
...
-    /// The log filter to use.
-    ///
-    /// This follows the envlogger syntax (e.g. 'info,driver=debug').
-    #[structopt(
-        long,
-        env = "LOG_FILTER",
-        default_value = "warn,driver=info,price_estimator=info"
-    )]
-    log_filter: String,
-
```
e00E, line 31: This wasn't being used because we use env_logger directly without going through slog.

price-estimator/Cargo.toml
```diff
@@ -8,11 +8,13 @@ anyhow = "1.0"
 core = { path = "../core" }
 env_logger = "0.7"
 ethcontract = "0.7"
+futures = { version = "0.3" }
```
nlordell, line 11: nit: 
```suggestion
futures = "0.3"
```

price-estimator/src/filter.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut split = s.split('-');
+        let mut next_token_id = || -> Result<u16, ParseTokenPairError> {
+            let token_string = split
+                .next()
+                .ok_or(ParseTokenPairError::WrongNumberOfTokens)?;
+            token_string.parse().map_err(From::from)
+        };
+        let buy_token_id = next_token_id()?;
+        let sell_token_id = next_token_id()?;
```
nlordell, line 46: I think this method accepts `1-2-3` as a `TokenPair` which is incorrect.

price-estimator/src/filter.rs
```diff
@@ -0,0 +1,71 @@
...
+            .path("/markets/0-65535/estimated-buy-amount/1.5")
+            .filter(&estimated_buy_amount())
+            .now_or_never()
+            .unwrap()
+            .unwrap();
+        assert_eq!(token_pair.buy_token_id, 0);
+        assert_eq!(token_pair.sell_token_id, 65535);
+        assert!((volume - 1.5).abs() < 0.001);
+    }
+}
```
nlordell, line 71: nit: Could you add some unit tests to check the error conditions for the filter?

nlordell (approved):
Wow! Coming together very quickly!

price-estimator/src/filter.rs
```diff
@@ -0,0 +1,119 @@
...
+        let mut split = s.split('-');
+        let mut next_token_id = || -> Result<u16, ParseTokenPairError> {
+            let token_string = split
+                .next()
+                .ok_or(ParseTokenPairError::WrongNumberOfTokens)?;
+            token_string.parse().map_err(From::from)
+        };
+        let buy_token_id = next_token_id()?;
+        let sell_token_id = next_token_id()?;
+        if next_token_id().is_ok() {
```
nlordell, line 47: I don't think this is enough as `1-2-abc` will be accepted.

price-estimator/src/filter.rs
```diff
@@ -0,0 +1,119 @@
+use std::num::ParseIntError;
+use warp::Filter;
+
+/// Validate a request of the form
+/// `/markets/<baseTokenId>-<quoteTokenId>/estimated-buy-amount/<sellAmountInQuoteToken>`
+/// and extract the url parameters.
+pub fn estimated_buy_amount(
+) -> impl Filter<Extract = (TokenPair, f64), Error = warp::Rejection> + Copy {
+    warp::path!("markets" / TokenPair / "estimated-buy-amount" / f64)
```
nlordell, line 9: The amount is in atoms, so it should be a `u128` I think

price-estimator/src/filter.rs
e00E: How important is it to reject all wrong queries? I feel we shouldn't do more work to reject them but I also think that in general it is better to be more stingy with the api.

nlordell: Personally, I think its better to be more restrictive with accepted input then less restrictive. It also helps avoid things like breaking changes.

price-estimator/src/main.rs
```diff
@@ -87,13 +79,25 @@ fn main() {
...
 
-    // TODO: handle http requests
-    let serve_task = runtime.spawn(warp::serve(warp::any().map(|| "")).run(([127, 0, 0, 1], 8080)));
+    let filter = filter::estimated_buy_amount().and_then(
+        move |token_pair: TokenPair, sell_amount_in_quote| {
+            let orderbook = orderbook.clone();
+            async move {
+                // TODO: format the response the way the nodejs price estimator did.
+                let orderbook = orderbook.get_reduced_orderbook().await;
+                let result = estimate_price(token_pair, sell_amount_in_quote, orderbook);
```
e00E, line 92: We still need to calculate the buy amount after getting the price but I'll do that in the next PR.

price-estimator/src/filter.rs
e00E: Should be fixed now.

nlordell: Just curious, any reason why you didn't want to use regular expressions? I feel like they would have been a very nice fit here.

e00E: Felt like overkill to depend on regex for finding two numbers separated by `-`.

price-estimator/src/main.rs
```diff
@@ -87,13 +79,25 @@ fn main() {
...
+    let filter = filter::estimated_buy_amount().and_then(
+        move |token_pair: TokenPair, sell_amount_in_quote| {
+            let orderbook = orderbook.clone();
+            async move {
+                // TODO: format the response the way the nodejs price estimator did.
+                let orderbook = orderbook.get_reduced_orderbook().await;
+                let result = estimate_price(token_pair, sell_amount_in_quote as f64, orderbook);
+                // The compiler cannot infer the error type because we never return an error.
+                Result::<_, Rejection>::Ok(format!("{}", result.unwrap_or_default()))
+            }
```
fleupold, line 95: Would be nice if this could be its own async function, but I assume this is what you are planning to do in the next PR anyways?

fleupold (approved):
Very nice!

> I have an event based orderbook file that I keep out of the repository for fast start ups without having to query old events.

Would you mind sharing this vial slack or attached to this PR?

price-estimator/src/main.rs
e00E: Next PR is up. Let's comment on it there if it should still be changed.

price-estimator/src/main.rs
bh2smith: Seems reasonable.

PR #968, opened 2020-06-09
[Pricegraph] Fix fill order at price

nlordell:
This PR fixes an issue with the `fill_order_at_price` method that I noticed when adding benchmarks. The specified limit price was actually being used as if it were `sell_amount / buy_amount`. The limit price is now correctly inverted.

:see_no_evil: 

### Test Plan

Adjusted unit test, added new benchmark:
```
Orderbook::fill_order_at_price(reduced)/200                                                                           
                        time:   [56.977 us 57.390 us 57.714 us]
Orderbook::fill_order_at_price(reduced)/190                                                                           
                        time:   [189.15 us 192.99 us 196.01 us]
Orderbook::fill_order_at_price(reduced)/180                                                                           
                        time:   [335.67 us 342.30 us 350.79 us]
Orderbook::fill_order_at_price(reduced)/150                                                                            
                        time:   [710.69 us 745.72 us 774.40 us]
Orderbook::fill_order_at_price(reduced)/100                                                                            
                        time:   [804.67 us 838.13 us 879.61 us]
```

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -357,7 +357,7 @@ impl Orderbook {
         // the specified token pair, but since we are finding transitive orders
         // in the opposite direction to match this order, calculate the maximum
         // price that still respects the provided limit price.
-        let max_price = 1.0 / (limit_price * FEE_FACTOR);
+        let max_price = (1.0 / limit_price) / FEE_FACTOR;
 
         let mut total_volume = 0.0;
         let mut predecessors = self.reduced_shortest_paths(sell);
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -346,14 +346,18 @@ impl Orderbook {
     ///
     /// Note that the limit price is expressed as an exchange limit price, i.e.
     /// with implicitely included fees.
-    pub fn fill_order_at_price(&mut self, pair: TokenPair, actual_limit_price: f64) -> f64 {
+    pub fn fill_order_at_price(&mut self, pair: TokenPair, limit_price: f64) -> f64 {
         if !self.is_token_pair_valid(pair) {
             return 0.0;
         }
         self.update_projection_graph();
 
         let (sell, buy) = (node_index(pair.sell), node_index(pair.buy));
-        let effective_limit_price = actual_limit_price / FEE_FACTOR;
+        // NOTE: The limit price is expressed in `buy_amount / sell_amount` for
+        // the specified token pair, but since we are finding transitive orders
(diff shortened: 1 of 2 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -347,14 +347,18 @@ impl Orderbook {
...
         }
         self.update_projection_graph();
 
         let (sell, buy) = (node_index(pair.sell), node_index(pair.buy));
-        let effective_limit_price = actual_limit_price / FEE_FACTOR;
+        // NOTE: The limit price is expressed in `buy_amount / sell_amount` for
+        // the specified token pair, but since we are finding transitive orders
+        // in the opposite direction to match this order, calculate the maximum
+        // price that still respects the provided limit price.
+        let max_price = 1.0 / (limit_price * FEE_FACTOR);
```
bh2smith, line 361: Happens to the best of us! I was actually wondering about this line before but neglected to comment. Good place to introduce a wee comment

bh2smith (approved):

pricegraph/src/orderbook.rs
fleupold, line 361: If I understand correctly, we multiply with the fee factor (1.001) to tighten the limit price and thus allow for the fee of the counter order. Is this right? Maybe it's worth documenting this. I first read this as we are deducting the fee (but we are actually doubling it).

fleupold (approved):

pricegraph/src/orderbook.rs
nlordell: This is because there are two mesurements for prices, exchange prices (that have implicit fees) and graph prices (called effective prices in the code) that have explicit fees.

By this I mean, to get the buy amount amount for an exchange price:
```
buy_amount = sell_amount * (exchange_price / FEE_FACTOR)
```

While to get the buy amount from an effective price:
```
buy_amount = sell_amount * effective_price
```

Whats confusing is how I wrote this equation. We want to convert the inverted price to an effective price, so it would probably be more clear if written:
```
(1.0 / limit_price) / FEE_FACTOR;
```

> Maybe it's worth documenting this.

This should already be in the doc comments:
```rust
    /// Note that the limit price is expressed as an exchange limit price, i.e.
    /// with implicitely included fees.
```

Maybe I can elaborate a bit more.

> I first read this as we are deducting the fee (but we are actually doubling it).

I don't think this doubles the fees, just adds it once. The subtlety is that we are converting the inverted price from an exchange price to a effective price.

nlordell: >  Good place to introduce a wee comment

Do you think the `// NOTE:` is not enough?

PR #973, opened 2020-06-09
Efficient last block for batch bis

### LINKED ISSUE #832: Find last block in batch with incrementing step size

fedgiac:
Fixes #832.

Implements a faster algorithm to determine the last block of a batch based on binary search.
This is a rewrite of PR #904. The logic is mostly unchanged but the structure, based on @fleupold's comments, is very different. There is now
- a trait `BatchIdRetrieving` implemented for `web3` that yields information on batch ids.
- a generic function implementing binary search for the last block of a batch that relies on the implementation of `BatchIdRetriever`.

`BatchIdRetriever` is not tested. The function implementing binary search is tested over a mock structure implementing `BatchIdRetriever`.

Compared to the older method, it makes more web3 calls for 2 and 4 steps back.
```
before:  1 2 3 4 5  6  7  8  9 10 11 12 13 14 15 16...
now:     1 3 3 5 5  5  5  7  7  7  7  7  7  7  7  9...
diff:    0 1 0 1 0 -1 -2 -1 -2 -3 -4 -5 -6 -7 -8 -7... 
```
Afterwards it's much better, for searching in the first `2^n` blocks it takes at most `2n+1` web3 queries.

### Test Plan
```
cd core
cargo test contracts::stablex_contract::search_batches::tests::incremental_binary_search
```

### DIFF
--- a/core/src/contracts/stablex_contract.rs
+++ b/core/src/contracts/stablex_contract.rs
@@ -366,6 +366,7 @@ async fn get_block(
         .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
 }
 
+#[cfg_attr(test, automock)]
 trait BatchIdGetting {
     fn batch_id_from_block<'a>(&'a self, block_number: BlockNumber) -> BoxFuture<'a, Result<u32>>;
 
@@ -514,47 +515,78 @@ pub mod tests {
 
     #[test]
     fn incremental_binary_search() {
-        struct MockBlockchain {
-            batch_id: Vec<u32>,
-        }
-        impl BatchIdGetting for MockBlockchain {
-            fn batch_id_from_block<'a>(
-                &'a self,
-                block_number: BlockNumber,
-            ) -> BoxFuture<'a, Result<u32>> {
-                async move {
-                    Ok(self.batch_id[if let BlockNumber::Number(n) = block_number {
+        //                                   2        5     7     9  10
+        let batch_ids: Vec<u32> = vec![1, 1, 1, 2, 2, 2, 3, 3, 5, 5, 6];
+        let mut mock_batch_id_getting = MockBatchIdGetting::new();
+        mock_batch_id_getting
+            .expect_batch_id_from_block()
+            .withf(|block_number: &BlockNumber| match block_number {
+                BlockNumber::Number(_) => true,
+                _ => false,
+            })
+            .returning({
+                let batch_ids = batch_ids.clone();
+                move |block_number: BlockNumber| {
+                    let result = batch_ids[if let BlockNumber::Number(n) = block_number {
                         n
                     } else {
-                        panic!("Not necessary");
+                        panic!("Not implemented");
                     }
-                    .as_u64() as usize])
+                    .as_u64() as usize];
+                    async move { Ok(result) }.boxed()
                 }
-                .boxed()
-            }
+            });
 
-            fn current_batch_id_and_block_number<'a>(
-                &'a self,
-            ) -> BoxFuture<'a, Result<(u32, u64)>> {
-                async move {
-                    let latest_block = self.batch_id.len() as u64 - 1;
-                    Ok((self.batch_id[latest_block as usize], latest_block))
-                }
-                .boxed()
-            }
-        }
+        mock_batch_id_getting
+            .expect_current_batch_id_and_block_number()
+            .returning(move || {
+                let latest_block = batch_ids.len() as u64 - 1;
+                let latest_batch_id = batch_ids[latest_block as usize];
+                async move { Ok((latest_batch_id, latest_block)) }.boxed()
+            });
 
-        let blockchain = MockBlockchain {
-            //                   2        5     7     9  10
-            batch_id: vec![1, 1, 1, 2, 2, 2, 3, 3, 5, 5, 6],
-        };
-        assert_eq!(get_last_block_for_batch(&blockchain, 1).wait().unwrap(), 2);
-        assert_eq!(get_last_block_for_batch(&blockchain, 2).wait().unwrap(), 5);
-        assert_eq!(get_last_block_for_batch(&blockchain, 3).wait().unwrap(), 7);
-        assert_eq!(get_last_block_for_batch(&blockchain, 4).wait().unwrap(), 7); // note: no error if batch does not exist
-        assert_eq!(get_last_block_for_batch(&blockchain, 5).wait().unwrap(), 9);
-        assert_eq!(get_last_block_for_batch(&blockchain, 6).wait().unwrap(), 10);
-        assert_eq!(get_last_block_for_batch(&blockchain, 7).wait().unwrap(), 10);
+        assert_eq!(
+            get_last_block_for_batch(&mock_batch_id_getting, 1)
+                .wait()
+                .unwrap(),
+            2
+        );
+        assert_eq!(
+            get_last_block_for_batch(&mock_batch_id_getting, 2)
+                .wait()
+                .unwrap(),
+            5
+        );
+        assert_eq!(
+            get_last_block_for_batch(&mock_batch_id_getting, 3)
+                .wait()
+                .unwrap(),
+            7
+        );
+        assert_eq!(
+            get_last_block_for_batch(&mock_batch_id_getting, 4)
+                .wait()
+                .unwrap(),
+            7
+        ); // note: no error if batch does not exist
+        assert_eq!(
+            get_last_block_for_batch(&mock_batch_id_getting, 5)
+                .wait()
+                .unwrap(),
+            9
+        );
+        assert_eq!(
+            get_last_block_for_batch(&mock_batch_id_getting, 6)
+                .wait()
+                .unwrap(),
+            10
+        );
+        assert_eq!(
+            get_last_block_for_batch(&mock_batch_id_getting, 7)
+                .wait()
+                .unwrap(),
+            10
+        );
         // note: returns last batch for batches in the future
     }
 }
--- a/core/src/contracts/stablex_contract.rs
+++ b/core/src/contracts/stablex_contract.rs
@@ -4,6 +4,7 @@
 use crate::{
     contracts,
     models::{ExecutedOrder, Solution},
+    util::FutureWaitExt,
 };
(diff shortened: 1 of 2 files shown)


core/src/contracts/stablex_contract.rs
```diff
@@ -377,6 +367,95 @@ async fn get_block(
...
+                    anyhow!("latest block {:?} has no block number", current_block.hash)
+                })?
+                .as_u64();
+            Ok((batch_id, block_number))
+        }
+        .boxed()
+    }
+}
+
+trait IncrementalBinarySearch<T: BatchIdGetter = Self> {
```
e00E, line 401:
```suggestion
trait IncrementalBinarySearch {
```

e00E: This does not seem needed.

bh2smith: Is this trait the meaning behind the acronym "BIS" I see in the PR title?

core/src/contracts/stablex_contract.rs
```diff
@@ -377,6 +367,95 @@ async fn get_block(
...
+                self.current_batch_id_and_block_number().await?;
+
+            // find lower bound for binary search
+            let mut step = 1_u64;
+            let mut bounds = Bounds {
+                lower: current_block_number,
+                upper: current_block_number,
+            };
+            let mut lower_batch_id = current_batch_id;
+            while batch_id < lower_batch_id {
```
bh2smith, line 431: Not entirely sure, but I wonder if this might be a prime example of a use case for a `while ... let` statement. I say this because the lower batch id isn't used outside the scope of this loop.

core/src/contracts/stablex_contract.rs
```diff
@@ -437,4 +516,49 @@ pub mod tests {
...
+                }
+                .boxed()
+            }
+        }
+
+        let blockchain = MockBlockchain {
+            //                   2        5     7     9  10
+            batch_id: vec![1, 1, 1, 2, 2, 2, 3, 3, 5, 5, 6],
+        };
+        assert_eq!(blockchain.get_last_block_for_batch(1).wait().unwrap(), 2);
```
e00E, line 556: These `wait()`s could be replaced by `.now_or_never().unwrap()`. This ensures that the future completes immediately which is the case in the test as no network request happens.

core/src/contracts/stablex_contract.rs
```diff
@@ -437,4 +516,49 @@ pub mod tests {
...
+        assert_eq!(blockchain.get_last_block_for_batch(2).wait().unwrap(), 5);
+        assert_eq!(blockchain.get_last_block_for_batch(3).wait().unwrap(), 7);
+        assert_eq!(blockchain.get_last_block_for_batch(4).wait().unwrap(), 7); // note: no error if batch does not exist
```
bh2smith, line 559: The inline comments surely make the line go over max width. `cargo fmt` might have something to say about this.

core/src/contracts/stablex_contract.rs
e00E: I don't think this works. while let is when you have something like an Option that you can match, but in this case the variable and the while condition are separate things.

core/src/contracts/stablex_contract.rs
```diff
@@ -377,6 +367,95 @@ async fn get_block(
         .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
 }
 
+trait BatchIdGetter {
```
e00E, line 370: I would like to put the new traits and the test into their own file but depends on what others think.

fleupold: +1 for own file (`stablex_contract.rs` is huge as it is). Also, I personally prefer calling traits with ["gerunds"](https://en.wikipedia.org/wiki/Gerund) (`BatchIdGetting` in this case). This is because the trait itself doesn't actually perform the action but it describes what implementations of it can do (read: "this is the trait for getting batch ids").

core/src/contracts/stablex_contract.rs
```diff
@@ -377,6 +367,95 @@ async fn get_block(
...
+    fn batch_id_from_block<'a>(&'a self, block_number: BlockNumber) -> BoxFuture<'a, Result<u32>>;
+
+    fn current_batch_id_and_block_number<'a>(&'a self) -> BoxFuture<'a, Result<(u32, u64)>>;
+}
+
+impl BatchIdGetter for Web3<DynTransport> {
+    fn batch_id_from_block<'a>(&'a self, block_number: BlockNumber) -> BoxFuture<'a, Result<u32>> {
+        async move {
+            let current_block = get_block(&self, block_number).await?;
```
fleupold, line 379: If it makes factoring things easier get_block could probably become part of this trait as well (technically it doesn't really relate to BatchIdGetting though).

core/src/contracts/stablex_contract.rs
```diff
@@ -437,4 +516,49 @@ pub mod tests {
             (expected_prices, expected_token_ids)
         );
     }
+
+    #[test]
+    fn incremental_binary_search() {
+        struct MockBlockchain {
```
fleupold, line 522: Instead of implementing you own mock, you should be able to "automock" the trait like here:

https://github.com/gnosis/dex-services/blob/f1a0b703acb7512b767644415e12a2c6d1597b8e/core/src/contracts/stablex_contract.rs#L78-L79

and then configure the autogenerated mock with whatever you want to return like here:

https://github.com/gnosis/dex-services/blob/f1a0b703acb7512b767644415e12a2c6d1597b8e/core/src/solution_submission.rs#L263-L268

I see that here you are returning different values depending on the input. This should be also possible in automock as you have access tot he argument and can just look it up in a Mapping or list.

core/src/contracts/stablex_contract.rs
nlordell: I also think this is a good idea, it keeps the logic, tests, etc. associated with the new trait in its own module so that it is easier to find, and makes the modules contain more immediately relevant information.

core/src/contracts/stablex_contract.rs
nlordell: So, from what I understand from this, you want to provide a `get_last_block_for_batch` implementation for any type that implements `BatchIdGetter`. You have two options for this:
```rust
trait BatchIdGetter {
    fn batch_id_from_block<'a>(&'a self, block_number: BlockNumber) -> BoxFuture<'a, Result<u32>>;
    fn current_batch_id_and_block_number<'a>(&'a self) -> BoxFuture<'a, Result<(u32, u64)>>;

    fn get_last_block_for_batch<'a>(&'a self, batch_id: u32) -> BoxFuture<'a, Result<u64>> {
        // NOTE: Default trait method implementation, types that implement `BatchIdGetter` get this
        // implementation automatically.
        // ...
    }
}
```

Alternatively, if you want to separate it into a separate trait, you could:
```rust
trait IncrementalBinarySearch: BatchIdGetter {
    fn get_last_block_for_batch<'a>(&'a self, batch_id: u32) -> BoxFuture<'a, Result<u64>> {
        // NOTE: Default trait method implementation, types that implement `BatchIdGetter` get this
        // implementation automatically.
        // ...
    }
}
```

Or if you really want to keep it a separate trait, you can do it as @e00E suggested.

core/src/contracts/stablex_contract.rs
nlordell: I like the suggestion on changing the name as well, especially since pretty much all of our traits in this codebase conform to this naming, making them easy to identify by name. But maybe `BatchIdRetrieving` or `BatchIdSearching` as `BatchIdGetting` sounds a little funny to my ears.

core/src/contracts/stablex_contract.rs
nlordell, line 522: Nice! One pro-tip is we have a `mockall` crate, which can automatically generate these mock implementations for you.

core/src/contracts/stablex_contract/search_batches.rs
```diff
@@ -0,0 +1,202 @@
...
+        upper: u64,
+    }
+    impl Bounds {
+        fn diff(&self) -> u64 {
+            self.upper - self.lower
+        }
+        fn mid(&self) -> u64 {
+            (self.upper + self.lower) / 2
+        }
+    }
```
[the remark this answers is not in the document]
fedgiac: Nope! I put it inside the scope at first because it was unrelated to everything else in the code, and I didn't think of moving it out after moving the code to a new file. I also like it more outside of the scope as it is now.

core/src/contracts/stablex_contract/search_batches.rs
```diff
@@ -0,0 +1,202 @@
...
+        .compat()
+        .await?
+        .ok_or_else(|| anyhow!("block {:?} is missing", block_number))
+}
+
+#[cfg_attr(test, automock)]
+pub trait BatchIdRetrieving {
+    fn batch_id_from_block<'a>(&'a self, block_number: BlockNumber) -> BoxFuture<'a, Result<u32>>;
+
+    fn current_batch_id_and_block_number<'a>(&'a self) -> BoxFuture<'a, Result<(u32, u64)>>;
```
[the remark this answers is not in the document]
fedgiac: I also don't like much the second method, I considered making a struct just for it but then I decided it would have been overkill.
I'm inclined towards keeping it as it is however. The name of the function reflects the order of the output values, and the two types are different, so `u32` is naturally mapped to batch numbers while `u64` means block numbers, I think this should be enough to avoid confusion.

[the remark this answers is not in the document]
nlordell: Sorry, what I meant is include a doc comment to clarify this a bit more (a `///` comment so it shows up in the tooltip in your IDE for example). I also don’t think the method should change.

bh2smith:
Lets get this thing merged please.

PR #975, opened 2020-06-10
Fix not running most tests in CI

e00E:
I remembered that Ben said in our last meeting that coverage was very
low. The reason is that we only run tests for default cargo workspace
members but when I added the core and now the price estimator crates I
did not add them there. I never noticed this because I always run tests
only for the crate I am working on like `cargo test -p price-estimator`.
Alternatively we could run all tests in CI like `cargo test --workspace`
but currently the e2e tests are rust tests too which would run and
fail because the needed docker containers haven't been started.

### Test Plan
CI, see that all tests are run, see that coverage goes up.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,9 @@ members = [
     "pricegraph/wasm",
 ]
 default-members = [
+    "core",
     "driver",
+    "price-estimator",
     "pricegraph",
 ]
 


fleupold (approved):
Maybe we should add notifications in our coverage decreases significantly on a single PR: https://docs.coveralls.io/coveralls-notifications to be notified of issues in the future?

bh2smith:
> Maybe we should add notifications in our coverage decreases significantly on a single PR: https://docs.coveralls.io/coveralls-notifications to be notified of issues in the future?

Not only this, but we should not allow travis to pass if coverage goes down. We have it configured this way for dex-contracts.

e00E:
I disagree with that. But we shouldn't discuss it here. Maybe next meeting?

PR #978, opened 2020-06-10
Refactor filter building

e00E:
I looked through a larger warp example and found a nicer way to
build the filter.
Also added a check that the http method is `get` and added parsing of
the query part of the url.

`atoms` is not yet being checked, that will be another PR.

### Test Plan
CI


nlordell (approved):
Looks great!

fleupold (approved):
Nice! Out of curiosity, which example did you use?

Issue #842, opened 2020-05-04
Implement PriceEstimation Source based on price graph


e00E:
I understand why we do this (it helps the solver) but conceptually it doesn't make sense. The prices should be determined by the orderbook and not by some other exchange's orderbook.

fleupold:
@koeppelmann also would like to see our price estimation from external sources disappear. 

I think as an interim step we can try to estimate prices using our own orderbook and the pricegraph approach that @nlordell implemented.

nlordell:
Only thing to watch out for is whether or not prices can be manipulated this way. (Not saying the my can, just that we should consider this).

bh2smith:
A nice suggestion made was to use a small amount of OWL when estimating prices (the minimum tradable amount). This would help with extreme slippage in estimations? Although, we may want to investigate if such an estimation might be misleading to the solver. 

Price estimation first reduces the orderbook by removing overlaps, and the resulting, non-overlapping, orderbook yields the price estimation.

fleupold:
We need to implement a [PriceSource](https://github.com/gnosis/dex-services/blob/f1a0b703acb7512b767644415e12a2c6d1597b8e/core/src/price_estimation/price_source.rs#L61) that uses the internal pricegraph crate.

For this we can probably reuse the [`Orderbook` struct](https://github.com/gnosis/dex-services/blob/f1a0b703acb7512b767644415e12a2c6d1597b8e/price-estimator/src/orderbook.rs#L11) that @e00E implemented. I'd move this from the `price-estimator` into core.

The other part is to adjust the current price estimation to be performed on all tokens and not just a whitelisted subset of tokens (cf `tokens` passed around in `get_prices`). For this we would need to get the alias and decimals for each token. We could use [our Graph endpoint](https://thegraph.com/explorer/subgraph/gnosis/dfusion) for this, or if we don't want to depend on another service read it from the chain directly (requires more work and batch requests).

Alternatively we could make the TokenInfo part of source dependencies and not pass it in each invocation of `get_prices`. This would allow to keep the TokenInfo hardcoded and scoped to tokens we estimate via Kraken/Dex.AG. However, alias and decimals are also serialized into the [instance file](https://gnosis-dev-dfusion.s3.amazonaws.com/data/mainnet/best-ring-solver/instances/2020-06-02/instance_5303580_2020-06-02T05%3A06%3A28.016082323%2B00%3A00.json) and I'm not sure what would be the implications if we provide an estimate but no alias or decimals.

[...]

PR #979, opened 2020-06-10
Remove tokens from price source trait

fleupold:
Part of #842. For this issue we need to implement a PriceSource, that is using the price graph to estimate prices. The way the PriceGraph trait is currently designed is that it passes a list of tokens that should be estimated. That list of tokens contains detailed information such as symbol name and decimals. 

This is required for our existing external data sources. However for the internal, pricegraph based estimation all we need to know about a token is its ID. Moreover, given that arbitrary tokens can be added to the exchange we are not guaranteed to have full detail information for all tokens.

This PR therefore changes the signature of the PriceSource trait to only pass the token Id. The more detailed information (required for kraken and dex.ag) is now passed as instance state via the constructor.

In a next step I want to get rid of the `shouldEstimatePrice` flag and make it so that we attempt to estimate all tokens for which there are orders.

Eventually I would like to get rid of the hardcoded TokenData altogether and fetch the information either from the graph or from the chain (the only thing we need to hardcode is a fallback estimate for our core tokens).

### Test Plan

No logic changes. Therefore:

- cargo test
- cargo test price_estimation  -- --ignored

### DIFF
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -306,30 +306,4 @@ mod tests {
             }
         );
     }
-
-    #[test]
-    fn token_get_price() {
-        for (token, usd_price, expected) in &[
-            (Token::new(4, "USDC", 6), 0.99, 0.99 * 10f64.powi(30)),
-            (Token::new(7, "DAI", 18), 1.01, 1.01 * 10f64.powi(18)),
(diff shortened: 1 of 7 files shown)


e00E (approved):

nlordell (approved):
Really nice refactor!

bh2smith (commented):
Really cool!

PR #983, opened 2020-06-11
Refactor TransitiveOrder to include price and capacity

e00E:
The transitive orderbook as returned from price estimator is supposed to
return price and volume. Pricegrpah has this internally but converts it
into buy and sell amount instead.
By expressing this in terms of methods we can pick the internal
representation as we like while making use of the struct more
convenient.

@nlordell What do you think of this?

### Test Plan
CI

### DIFF
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -10,6 +10,9 @@ mod data;
 pub use encoding::*;
 pub use orderbook::Orderbook;
 
+/// The fee factor that is applied to each order's buy price.
+pub const FEE_FACTOR: f64 = 1.0 / 0.999;
+
 /// A struct representing a transitive orderbook for a base and quote token.
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct TransitiveOrderbook {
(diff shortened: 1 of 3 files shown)


pricegraph/src/lib.rs
```diff
@@ -36,8 +39,34 @@ pub struct TransitiveOrderbook {
 /// Additionally, a transitive order over a single order is equal to that order.
 #[derive(Clone, Debug, PartialEq)]
 pub struct TransitiveOrder {
+    price: f64,
+    capacity: f64,
+}
+
+impl TransitiveOrder {
+    /// ?
```
e00E, line 47: What is a good doc comment here?

pricegraph/src/lib.rs
```diff
@@ -36,8 +39,34 @@ pub struct TransitiveOrderbook {
...
+    pub fn price(&self) -> f64 {
+        self.price
+    }
+
+    /// ?
```
e00E, line 52: What is a good doc comment here?

nlordell:
> The transitive orderbook as returned from price estimator is supposed to return price and volume. 

Note that the price is probably in a different unit that the internal pricegraph representation. The frontend expects exchange prices (buy amount / sell amount) while the pricegraph internally represents them as effective prices (buy amount / sell amount / fee factor).

I suggest adding a `price` method that returns `self.buy / self.sell` as, AFIAK, this is what the front-end expects.

e00E:
Just checked, I think you're right. In that case I don't see a reason for this PR anymore.

PR #984, opened 2020-06-11
Refactor BatchID in its own module

fleupold:
Part of #842 

Pricegraph requires to be initialized with a BatchID. We already have an abstraction for this type in the scheduler. This PR moves this abstraction in the model module for reuse.

### Test Plan

CI

### DIFF
--- a/core/src/driver/scheduler.rs
+++ b/core/src/driver/scheduler.rs
@@ -9,9 +9,6 @@ use anyhow::{anyhow, Error, Result};
 use std::str::FromStr;
 use std::time::Duration;
 
-/// The total time in a batch.
-const BATCH_DURATION: Duration = Duration::from_secs(300);
-
 /// The time in a batch where a solution may be submitted.
 const SOLVING_WINDOW: Duration = Duration::from_secs(240);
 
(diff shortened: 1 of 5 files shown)


core/src/models/order.rs
```diff
@@ -40,6 +41,29 @@ impl Order {
             self.remaining_sell_amount,
         )
     }
+
+    pub fn to_element(&self, balance: primitive_types::U256) -> Element {
+        // Some conversions are needed because the primitive types crate is on different versions in
```
nlordell, line 46: I would say we should just pin the price graph version to be the same as core. Then we don’t need to do these conversions. It’s not really used anyway.

nlordell (approved):
Nice! My only comment is that we should make the `pricegraph` and `core` crates use the same `primitive-types` version.

Also, I’m a bit surprised the price graph needs a batch ID, but maybe I’m forgetting something.

fleupold:
Sorry I included some unrelated code earlier. Updated the PR now

fleupold:
> Also, I’m a bit surprised the price graph needs a batch ID, but maybe I’m forgetting something.

You are right, it's the OrderbookReading that needs it.

nlordell (approved):
Looks awesome now!

core/src/models/order.rs
fleupold: This code is actually meant to be in the next PR. Cleaned up the mess and repushed the change.

e00E (approved):

PR #985, opened 2020-06-11
Adding a pricegraph based price source

fleupold:
Part of #842 

Introducing the price graph based price source. Uses the orderbook reader to get the right state of the orderbook and then estimates the amount of each token we'd be getting for selling one OWL. OWL itself is priced at 1 (10^18).

### Test Plan
For now added some unit test. Will wire it into our system in a separate PR and then e2e test it.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -18,6 +18,8 @@ isahc = { version = "0.9.4", features = ["json"] }
 lazy_static = "1.4.0"
 log = "0.4.8"
 num = { version = "0.2", features = ["serde"] }
+pricegraph = { path = "../pricegraph" }
+primitive-types = "0.7"
 prometheus = "0.9.0"
 rouille = "3.0.0"
 rustc-hex = "2.1.0"
--- a/core/src/price_estimation/orderbook_based.rs
+++ b/core/src/price_estimation/orderbook_based.rs
@@ -16,21 +16,22 @@ pub struct PricegraphEstimator {
 const ONE_OWL: f64 = 1_000_000_000_000_000_000.0;
 
 impl PriceSource for PricegraphEstimator {
-    fn get_prices<'b>(
-        &'b self,
-        tokens: &'b [TokenId],
-    ) -> BoxFuture<'b, Result<HashMap<TokenId, u128>>> {
+    fn get_prices<'a>(
+        &'a self,
+        tokens: &'a [TokenId],
+    ) -> BoxFuture<'a, Result<HashMap<TokenId, u128>>> {
         async move {
             let batch = BatchId::currently_being_solved(SystemTime::now())?;
             let (account_state, orders) =
                 self.orderbook_reader.get_auction_data(batch.into()).await?;
-            let orderbook = Orderbook::from_elements(orders.iter().map(|order| {
+            let mut orderbook = Orderbook::from_elements(orders.iter().map(|order| {
                 order.to_element(U256(
                     account_state
                         .read_balance(order.sell_token, order.account_id)
                         .0,
                 ))
             }));
+            orderbook.reduce_overlapping_orders();
             Ok(self.estimate_prices(tokens, orderbook))
         }
         .boxed()
--- a/core/src/price_estimation/orderbook_based.rs
+++ b/core/src/price_estimation/orderbook_based.rs
@@ -40,12 +40,13 @@ impl<'a> PriceSource for PricegraphEstimator<'a> {
 // Private trait to facilitate testing this module
 #[cfg_attr(test, mockall::automock)]
 trait Pricegraph {
-    fn estimate_price(&mut self, pair: TokenPair, volume: f64) -> Option<f64>;
+    fn estimate_price(&self, pair: TokenPair, volume: f64) -> Option<f64>;
 }
 
 impl Pricegraph for Orderbook {
-    fn estimate_price(&mut self, pair: TokenPair, volume: f64) -> Option<f64> {
-        self.fill_market_order(pair, volume)
+    fn estimate_price(&self, pair: TokenPair, volume: f64) -> Option<f64> {
+        // Clone the pricegraph so that each estimation is independent of one another.
+        self.clone().fill_market_order(pair, volume)
     }
 }
 
@@ -58,7 +59,7 @@ impl<'a> PricegraphEstimator<'a> {
     fn estimate_prices(
         &self,
         tokens: &[TokenId],
-        mut pricegraph: impl Pricegraph,
+        pricegraph: impl Pricegraph,
     ) -> HashMap<TokenId, u128> {
         tokens
             .iter()
@@ -66,6 +67,7 @@ impl<'a> PricegraphEstimator<'a> {
                 let price = if token == &TokenId::reference() {
                     1.0
                 } else {
+                    // Estimate price by selling 1 unit of the reference token for each token to estimate.
                     let pair = TokenPair {
                         buy: token.0,
                         sell: TokenId::reference().0,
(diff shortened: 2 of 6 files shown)


core/src/price_estimation/orderbook_based.rs
```diff
@@ -0,0 +1,144 @@
...
+            .iter()
+            .flat_map(|token| {
+                let price = if token == &TokenId::reference() {
+                    1.0
+                } else {
+                    let pair = TokenPair {
+                        buy: token.0,
+                        sell: TokenId::reference().0,
+                    };
+                    pricegraph.estimate_price(pair, ONE_OWL)?
```
fleupold, line 73: @nlordell do you think we should be cloning the price graph here so that we don't "eat" up the liquidity in between estimates?

nlordell: Yes. I think the pricegraph needs a new API that uses the orderbook graph internally.

e00E (approved):
Looks good assuming the orderbook cloning happens.

core/src/price_estimation/orderbook_based.rs
```diff
@@ -0,0 +1,149 @@
...
+impl PriceSource for PricegraphEstimator {
+    fn get_prices<'b>(
+        &'b self,
+        tokens: &'b [TokenId],
+    ) -> BoxFuture<'b, Result<HashMap<TokenId, u128>>> {
+        async move {
+            let batch = BatchId::currently_being_solved(SystemTime::now())?;
+            let (account_state, orders) =
+                self.orderbook_reader.get_auction_data(batch.into()).await?;
+            let orderbook = Orderbook::from_elements(orders.iter().map(|order| {
```
fleupold, line 27: Note, that I changed this to no longer reduce the orderbook beforehand, as price estimates should be based on the overlapping orderbook.

core/src/price_estimation/orderbook_based.rs
```diff
@@ -0,0 +1,149 @@
...
+use anyhow::Result;
+use futures::future::{BoxFuture, FutureExt as _};
+use pricegraph::{Orderbook, TokenPair};
+use primitive_types::U256;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::SystemTime;
+
+pub struct PricegraphEstimator {
+    orderbook_reader: Arc<dyn StableXOrderBookReading>,
```
fleupold, line 13: When wiring this into the threaded price source I wasn't able to get this to work with lifetimes (PriceOracle itself doesn't have any references so the only lifetime I could have used was static which the OrderbookReading isn't)

core/src/price_estimation/orderbook_based.rs
```diff
@@ -0,0 +1,149 @@
...
+                    // Estimate price by selling 1 unit of the reference token for each token.
+                    // We sell rather than buy the reference token because volume is denominated
+                    // in the sell token, for which we know the number of decimals.
+                    let pair = TokenPair {
+                        buy: token.0,
+                        sell: TokenId::reference().0,
+                    };
+                    pricegraph.estimate_price(pair, ONE_OWL)?
+                };
+                let price_in_reference = 1.0 / price_in_token;
```
fleupold, line 79: When testing, I noticed that this was backwards before (price directions are hard)

core/src/price_estimation/orderbook_based.rs
nlordell: I think we should be using `Arc` in most places instead of `&` anyway, it just makes things easier and the cost of arc is an atomic increment and decrement on `clone` and `drop`.

core/src/price_estimation/orderbook_based.rs
nlordell: The first step to estimating prices is to reduce the orderbook. I think from a performance standpoint it makes sense to reduce it first.

core/src/price_estimation/orderbook_based.rs
```diff
@@ -0,0 +1,149 @@
...
+            Ok(self.estimate_prices(tokens, orderbook))
+        }
+        .boxed()
+    }
+}
+
+// Private trait to facilitate testing this module
+#[cfg_attr(test, mockall::automock)]
+trait Pricegraph {
+    fn estimate_price(&self, pair: TokenPair, volume: f64) -> Option<f64>;
```
nlordell, line 43: Nice! I had saw it last night with the `&mut self` and wanted to suggest changing it to `&self` because it needs to be cloned anyway.

core/src/price_estimation/orderbook_based.rs
```diff
@@ -0,0 +1,149 @@
...
+impl Pricegraph for Orderbook {
+    fn estimate_price(&self, pair: TokenPair, volume: f64) -> Option<f64> {
+        // Clone the pricegraph so that each estimation is independent of one another.
+        self.clone().fill_market_order(pair, volume)
+    }
+}
+
+impl PricegraphEstimator {
+    #[allow(dead_code)]
+    pub fn new(orderbook_reader: Arc<dyn StableXOrderBookReading>) -> Self {
```
nlordell, line 55: Just a tidbit that I learned from @e00E and I really liked and wanted to share, for methods that are unused, you can name them with a prefixed `_`. This automatically makes it not subject to the `dead_code` lint, and has the added benefit that when you do start using it, you don't forget to remove the `#[allow]` attribute, as you will see the ugly `_function` name and remember to rename it!

core/src/price_estimation/orderbook_based.rs
nlordell: I agree :disappointed:

nlordell (approved):
Looks good!

core/Cargo.toml
```diff
@@ -18,6 +18,8 @@ isahc = { version = "0.9.3", features = ["json"] }
 lazy_static = "1.4.0"
 log = "0.4.8"
 num = { version = "0.2", features = ["serde"] }
+pricegraph = { path = "../pricegraph" }
+primitive-types = "0.7"
```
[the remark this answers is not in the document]
fleupold: Yes, will do that but in a separate PR as it will touch other parts of the code as well.

core/src/price_estimation/orderbook_based.rs
fleupold: Can we not estimate prices without reducing the orderbook first? I don't think it makes a huge difference (prices won't swing more than a few % before/after reduction), but technically the solver needs to know the estimated prices to clear the orderbook as is. This should include prices of overlapping trade.

E.g. the price of an esoteric non-liquid token might be infinity if all its supply has been covered by a matching order. However the price that the solver needs to know should be in the overlapping range.

On the other hand, this might anchor the prices for market orders very low (in the extreme case estimating the price at 0 before reducing and at infinity after). This could be bad for the fairness of such orders.

e00E: I think what Nick means is that the orderbook is reduced internally anyway when you estimate a price with `fill_market_order`. Otherwise we wouldn't be able to calculate a price because the negative cycle makes the cost infinitely negative.
Since the reductions happens anyway it is more performant to do it once before estimating any prices.

nlordell: > Can we not estimate prices without reducing the orderbook first?

The price estimate in its current form, is really about finding an exchange rate such that there is an overlapping order with the current orderbook. When there are negative cycles (i.e. existing overlaps in the orderbook), its unclear which order this overlap will match with, so it gets reduced first.

I agree, however, that for price source estimation, it would be nice to include this as well so that the estimate is closer to the middle of the spread instead of being on one side of it.

PR #989, opened 2020-06-12
Add transititive orderbook request scaffolding

e00E:
This includes everything except the actual functionality of computing
the transitive orderbook.

### Test Plan
Some unit tests. Can really test when fully implemented by comparing to old price estimator.

### DIFF
--- a/price-estimator/src/models.rs
+++ b/price-estimator/src/models.rs
@@ -1,6 +1,6 @@
 use serde::{Deserialize, Serialize};
 use serde_with::rust::display_fromstr;
-use std::num::ParseIntError;
+use std::{convert::From, num::ParseIntError, str::FromStr};
 
 #[derive(Debug, Copy, Clone)]
 pub struct TokenPair {
@@ -25,7 +25,7 @@ pub enum ParseTokenPairError {
     ParseIntError(#[from] ParseIntError),
 }
 
-impl std::str::FromStr for TokenPair {
+impl FromStr for TokenPair {
     type Err = ParseTokenPairError;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let mut split = s.split('-');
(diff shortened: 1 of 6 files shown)


price-estimator/src/models.rs
```diff
@@ -67,13 +67,45 @@ pub struct EstimatedBuyAmountResult {
...
+pub struct TransitiveOrder {
+    pub price: f64,
+    pub volume: f64,
+}
+
+impl From<&pricegraph::TransitiveOrder> for TransitiveOrder {
+    fn from(order: &pricegraph::TransitiveOrder) -> Self {
+        Self {
+            price: order.buy / order.sell,
+            volume: order.sell,
```
e00E, line 80: I think this is correct based on https://github.com/gnosis/dex-price-estimator/blob/a7b1f5793e95501174865df140f783338213a107/src/orderbook_fetcher.ts#L109 but to be sure we'll need to wait until this is fully implemented and compare the output to the nodejs price estimator.

fleupold: This depends on which way the transitive order is expressed. We need the price in the quote token (2nd token in the pair) and the amount in base token (1st token in the pair). I think the way the [transitive orderbook](https://github.com/gnosis/dex-services/blob/6c707afc5810175514e3e152e075446cc6d47262/pricegraph/src/lib.rs#L15) is structured right now, it would have to differ between bids and asks.

I always find it confusing to now "what" exactly a price is expressing. @nlordell was mentioning that he was thinking about a refactoring to annotate prices with the pair (or at least the denominated token). This would have also helped me in https://github.com/gnosis/dex-services/pull/985#discussion_r439098477

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F7WSsTFgNsfubS%2Fgiphy.gif&token=awesome_sauce" data-gitmeme-token="awesome_sauce"></a>

bh2smith (approved):

nlordell (approved):

price-estimator/src/models.rs
nlordell: I see now, asks and bids need prices to be in the quote token, so for asks, its `order.buy / order.sell` and for bids its `order.sell / order.buy`.

e00E: I have turned this into `unimplemented` for now and will add it back in the next PR.

PR #995, opened 2020-06-12
Implement a top-level pricegraph API

nlordell:
This PR implements a top level Pricegraph API so that all the confusing price format and quantity conversions happen in one place.

### Test Plan

Moved unit test to top level.

### DIFF
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -72,7 +72,7 @@ impl Pricegraph {
     ///
     /// The auction elements are in the standard exchange format.
     pub fn new(elements: impl IntoIterator<Item = Element>) -> Self {
-        let orderbook = Orderbook::new(elements);
+        let orderbook = Orderbook::from_elements(elements);
         Pricegraph::from_orderbook(orderbook)
     }
 
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -10,6 +10,9 @@ mod data;
 pub use encoding::*;
 pub use orderbook::Orderbook;
 
+/// The fee factor that is applied to each order's buy price.
+const FEE_FACTOR: f64 = 1.0 / 0.999;
+
 /// A struct representing a transitive orderbook for a base and quote token.
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct TransitiveOrderbook {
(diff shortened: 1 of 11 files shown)


pricegraph/src/lib.rs
```diff
@@ -41,3 +44,123 @@ pub struct TransitiveOrder {
...
+    /// be no overlapping transitive orders.
+    ///
+    /// This method returns a clone of the reduced orderbook because orderbook
+    /// operations are destructive (as they require filling orders).
+    pub fn reduced_orderbook(&self) -> Orderbook {
+        self.reduced_orderbook.clone()
+    }
+
+    /// Estimates an exchange rate for the specified token pair and sell volume.
+    /// Returns `None` if the volume cannot be fully filled because there are
```
e00E, line 112:
```suggestion
    /// Returns `None` if the volume cannot be fully filled because there is
```

pricegraph/src/lib.rs
```diff
@@ -41,3 +44,123 @@ pub struct TransitiveOrder {
...
+    /// not enough liquidity in the current batch.
+    ///
+    /// Note that this price is in exchange format, that is, it is expressed as
+    /// the ratio between buy and sell amounts, with implicit fees.
```
e00E, line 116: We might want to repeat some documentation from  `fill_market_order` to explain what None and Some(value) mean.

    /// Fill a market order in the current orderbook graph returning the maximum
    /// price the order can have while overlapping with existing orders. Returns
    /// `None` if the order cannot be filled because the token pair is not
    /// connected, or the connection cannot support enough capacity to fully
    /// fill the specified volume.

Or we could mention the documentation of that method.

pricegraph/src/lib.rs
```diff
@@ -41,3 +44,123 @@ pub struct TransitiveOrder {
...
+    pub fn estimate_exchange_rate(&self, pair: TokenPair, sell_amount: f64) -> Option<f64> {
+        self.reduced_orderbook()
+            .fill_market_order(pair, sell_amount as _)
+    }
+
+    /// Returns a transitive order with a buy amount calculated such that there
+    /// exists overlapping transitive orders to completely fill the speicified
```
e00E, line 123:
```suggestion
    /// exists overlapping transitive orders to completely fill the specified
```

pricegraph/src/lib.rs
```diff
@@ -41,3 +44,123 @@ pub struct TransitiveOrder {
...
+    /// `sell_amount`. As such, this is an estimated order that is *likely* to
+    /// be matched given the **current** state of the batch.
```
e00E, line 125: This is nicely clear.

e00E (approved):

pricegraph/src/lib.rs
nlordell: Hmm, I'm not sure how to word this better. I was hoping to keep the "graph" concepts out of the top level API as much as possible and document it from and exchange perspective a little more (so the documentation reads a bit more intuitively). Any suggestions?

pricegraph/src/lib.rs
```diff
@@ -41,3 +44,123 @@ pub struct TransitiveOrder {
...
+impl TransitiveOrder {
+    /// Retrieves the exchange price for this order.
+    pub fn price(&self) -> f64 {
+        self.buy / self.sell
+    }
+
+    /// Retrieves the effective price for this order after fees are condidered.
+    ///
+    /// Note that `effective_price > price`.
+    pub fn effective_price(&self) -> f64 {
```
fleupold, line 57: Thinking about the naming (I'm fine with the way it is). Is this the "matchable" price, e.g. the price at which someone could take this order? If so maybe we can call it matcheable or takeable price?

pricegraph/src/lib.rs
```diff
@@ -41,3 +44,123 @@ pub struct TransitiveOrder {
...
+            full_orderbook,
+            reduced_orderbook,
+        }
+    }
+
+    /// Gets a copy of the full orderbook for operations that need to contain
+    /// the existing overlapping transitive orders for accuracy.
+    ///
+    /// This method returns a clone of the reduced orderbook because orderbook
+    /// operations are destructive (as they require filling orders).
```
fleupold, line 97: Some copy-pasta-gate here.

Also nanonit, the two sentences could probably be combined.

pricegraph/src/lib.rs
fleupold: To simplify the interface, could we return Infinity in case there is not enough liquidity? Or would this just hide a possible error case people need to think about.

fleupold (approved):

pricegraph/src/lib.rs
nlordell: I think it would return `0` in that case, since its buy / sell, and if there isn't enough liquidity then your buy amount needs to be infinitely small.

pricegraph/src/lib.rs
nlordell: :see_no_evil:

pricegraph/src/lib.rs
nlordell: No, this is the effective limit price, that is, the minimum exchange rate that can be used in a solution because of fees. If the exchange rate were lower, than the solution would not be accepted because the executed_buy_amount / executed_sell_amount would be lower than the order limit price.

nlordell: I think for the orderbook visualization, we should probably use this value, or else we might see a small amount of overlap that is not actually there once fees are factored in.

nlordell: For the order to be "matchable", you need to place an order whose effective exchange rate overlaps with this order's effective exchange rate.

nlordell: speaking of which, I renamed price to exchange rate :tada:. Lets try and keep this top level API using the language proposed by @marcovc.

pricegraph/src/lib.rs
nlordell: I'm OK with returning a `0` here if you guys thing it makes more sense.

e00E: I much prefer Optional over infinity. I slightly prefer Optional over 0 because when talking about exchange rates and prices it doesn't make sense for it to be zero. Not being able to trade is distinct from that which is what Option symbolizes.

nlordell: Any tips on the docs? I was hoping to keep the graph lingo out as much as possible.

nlordell: I *ever so slightly* prefer `Option` as well, as it indicates "there is no exchange rate for what you asked" instead of "here is an invalid exchange rate because it isn't possible".

e00E: To explain the optional in docs we can say that it will be None if a trade is not possible because no combination of trades is able to trade this amount of the sell token into the buy token.

e00E: I don't mind *Fill a market order in the current orderbook graph returning the maximum price the order can have while overlapping with existing orders.*. It doesn't even contain graph lingo if you delete the word "graph".

nlordell: `Fill a market order in the current orderbook graph` is not super accurate anymore, because the `orderbook` is always cloned now.

nlordell: I updated it to include your suggestion! I hope it sounds better now :)

PR #996, opened 2020-06-12
Fix overlapping transitive orderbook direction

nlordell:
This PR fixes the direction of the ask and bid orders, as they were inverted.

### Test Plan

Fixed old unit test.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -180,10 +180,6 @@ impl Orderbook {
             });
         }
 
-        for orders in &mut [&mut overlap.asks, &mut overlap.bids] {
-            orders.sort_unstable_by(|a, b| num::compare(a.price(), b.price()));
-        }
-
         overlap
     }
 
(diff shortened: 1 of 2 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -192,6 +192,10 @@ impl Orderbook {
             });
         }
 
+        for orders in &mut [&mut overlap.asks, &mut overlap.bids] {
+            orders.sort_unstable_by(|a, b| num::compare(a.price(), b.price()));
+        }
```
e00E, line 197: If we do this we should probably mention it in the api documentation.

e00E (approved):

pricegraph/src/orderbook.rs
nlordell: I removed it in the next PR. It turns out ordering first was a bad idea. I will just remove it here as well.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2Fft05jBFkMCUuinFTdu%2Fgiphy.gif&token=direction" data-gitmeme-token="direction"></a>

PR #997, opened 2020-06-12
Add API method for computing transitive orderbook

nlordell:
This PR adds a new API method for computing a transitive orderbook.

### Test Plan

```
$ cargo test -p pricegraph real_orderbooks -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.10s
     Running target/debug/deps/pricegraph-397047f497bad3b3

running 1 test
#5298183: estimated order for buying 194.87633796952346 DAI for 1 WETH
#5298183: DAI-WETH market contains 91 ask orders and 100 bid orders within a 5% spread:
 - Ask orders
    ...
 - Buy orders
    ...
```

### DIFF
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -138,7 +138,7 @@ impl Pricegraph {
     /// Computes a transitive orderbook for the given market specified by the
     /// base and quote tokens.
     ///
-    /// This method optionally accepts a spread is a decimal fraction that
+    /// This method optionally accepts a spread that is a decimal fraction that
     /// defines the maximume transitive order price with the equation:
     /// `first_transitive_price + first_transitive_price * spread`. This means
     /// that given a spread of 0.5 (or 50%), and if the cheapest transitive
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -72,7 +72,7 @@ impl Pricegraph {
     ///
     /// The auction elements are in the standard exchange format.
     pub fn new(elements: impl IntoIterator<Item = Element>) -> Self {
-        let orderbook = Orderbook::new(elements);
+        let orderbook = Orderbook::from_elements(elements);
         Pricegraph::from_orderbook(orderbook)
     }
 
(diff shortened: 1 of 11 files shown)


pricegraph/src/lib.rs
```diff
@@ -134,6 +134,59 @@ impl Pricegraph {
             sell: sell_amount,
         })
     }
+
+    /// Computes a transitive orderbook for the given market specified by the
+    /// base and quote tokens.
+    ///
+    /// This method optionally accepts a spread is a decimal fraction that
```
e00E, line 141:
```suggestion
    /// This method optionally accepts a spread that is a decimal fraction that
```

pricegraph/src/lib.rs
```diff
@@ -134,6 +134,59 @@ impl Pricegraph {
...
+                    sell: base,
+                },
+                spread,
+            ));
+
+        for orders in &mut [
+            &mut transitive_orderbook.asks,
+            &mut transitive_orderbook.bids,
+        ] {
+            orders.sort_unstable_by(|a, b| num::compare(a.price(), b.price()));
```
fleupold, line 185: If I understand correctly, the prices for bids and asks are going to be represented as inverse prices (thus the value will jump drastically between best bid/best ask). 
I believe normally bids and asks in an orderbook have the same price unit (e.g. https://poloniex.com/public?command=returnOrderBook&currencyPair=BTC_ETH&depth=50)

fleupold (approved):
Looks good. Just one thing to consider if the price unit should not unified for bids & asks.

pricegraph/src/lib.rs
nlordell: One thing to note is that we return ask and bid orders, not prices. That is buy and sell amounts (as they are represented in the exchange).

I guess, one thing we can add is `inverse` method for inverting the orders and calculating the inverse price.

The sorting here is just so the ask orders are sorted, and the bid orders are sorted, and they are sorted interdependently.

nlordell: Maybe we can change the `TransitiveOrderbook` representation to align with expectations a bit more? We can also add iterator methods to produce `volume/price` pairs for the price estimator service?

nlordell: Created #1002  to capture this work.

e00E (approved):

PR #1001, opened 2020-06-12
Fix orders appearing multiple times in the transitive orderbook

### LINKED ISSUE #999: Transitive orders that are overlapping can get considered more than once.
There is an issue with the current algorithm where transitive orders may be considered more than once if they are part of more than one negative cycle. This leads to incorrect reporting of transitive orders and inflates the actual available volume to the system.

nlordell:
Fixes #999 

This PR modifies the algorithm for computing overlapping transitive orderbook by completely reducing each transitive order instead of reducing the cycle. This changes the result so that not all initially overlapping transitive orders are returned, but each transitive order gets considered exactly once. This is OK since these initially overlapping orders will be included in the `fill_transitive_orders` result.

### Test Plan

Add new test case to illustrate problem.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -871,15 +871,16 @@ mod tests {
         assert_approx_eq!(overlap.bids[0].sell, 1_000_000.0);
 
         // NOTE: This is counter-intuitive, but there should only be one
-        // transitive order from `0 -> 1`. This is because whenever a negative
-        // cycle is found, it gets split into two transitive orders, one from
-        // `base -> quote` (ask) and the other from `quote -> base` (bid). These
-        // transitive orders are **completely** reduced, so even if there are
-        // more orders that could overlap with the remaining amount of a
-        // transitive order if only the negative cycle we reduced, they might
-        // not get found by `reduce_overlapping_transitive_orderbook`. Rest
-        // assured, they will get found by `fill_transitive_orderbook` and
-        // ultimately included in the final transitive orderbook.
+        // transitive order from `1 -> 0` even if there are two orders that
+        // overlap with the large `0 -> 1` order. This is because whenever a
+        // negative cycle is found, it gets split into two transitive orders,
+        // one from `base -> quote` (ask) and the other from `quote -> base`
+        // (bid). These transitive orders are **completely** reduced, so even if
+        // there are other orders that overlap with the remaining amount of one
+        // of these transitive orders, they might not get found by
+        // `reduce_overlapping_transitive_orderbook`. Rest assured, they will
+        // get found by `fill_transitive_orderbook` and ultimately included in
+        // the final transitive orderbook.
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


e00E (approved):

pricegraph/src/orderbook.rs
```diff
@@ -818,14 +823,65 @@ mod tests {
...
+        assert_approx_eq!(overlap.bids[0].sell, 1_000_000.0);
+
+        // NOTE: This is counter-intuitive, but there should only be one
+        // transitive order from `0 -> 1`. This is because whenever a negative
+        // cycle is found, it gets split into two transitive orders, one from
+        // `base -> quote` (ask) and the other from `quote -> base` (bid). These
+        // transitive orders are **completely** reduced, so even if there are
+        // more orders that could overlap with the remaining amount of a
+        // transitive order if only the negative cycle we reduced, they might
+        // not get found by `reduce_overlapping_transitive_orderbook`. Rest
```
fleupold, line 880: Something with this sentence is weird (it's also a bit too long for my brain stack)

pricegraph/src/orderbook.rs
```diff
@@ -173,14 +179,23 @@ impl Orderbook {
...
-                    overlap.asks.push(
-                        self.find_transitive_order(ask)
-                            .expect("ask transitive path not found after being detected"),
-                    );
-                    overlap.bids.push(
-                        self.find_transitive_order(bid)
-                            .expect("bid transitive path not found after being detected"),
-                    );
+                    let (capacity, transitive_price) = self
+                        .fill_path(ask)
```
fleupold, line 183: I really don't see the difference between filling the two halfs of the path individually here and `continue` vs. filling the full cycle in one go outside of this if statement (as it was before). I read all your comments but I don't get how e.g. the new unit test was not passing before.

fleupold: Ok, I think I see it now (after reading the associated issue). We are filling the "entirety" of the order and not just the potential fraction that is part of the ring.

pricegraph/src/orderbook.rs
```diff
@@ -818,14 +823,65 @@ mod tests {
...
+        assert_eq!(overlap.asks.len(), 1);
+        assert_approx_eq!(overlap.asks[0].buy, 50_000_000.0);
+        assert_approx_eq!(overlap.asks[0].sell, 100_000_000.0);
+
+        // Transitive order `1 -> 0` buying 1 selling 0
+        assert_eq!(overlap.bids.len(), 1);
+        assert_approx_eq!(overlap.bids[0].buy, 1_000_000.0);
+        assert_approx_eq!(overlap.bids[0].sell, 1_000_000.0);
+
+        // NOTE: This is counter-intuitive, but there should only be one
```
fleupold, line 873: Maybe explain shortly why this is counter-intuitive.

fleupold (approved):

PR #1003, opened 2020-06-12
[Easy] Downgrade primitive_types lib

fleupold:
Getting rid of the conversions between the different Address and U256 types

### Test Plan
CI

### DIFF
--- a/core/src/price_estimation/orderbook_based.rs
+++ b/core/src/price_estimation/orderbook_based.rs
@@ -4,7 +4,6 @@ use crate::orderbook::StableXOrderBookReading;
 use anyhow::Result;
 use futures::future::{BoxFuture, FutureExt as _};
 use pricegraph::{Orderbook, TokenPair};
-use primitive_types::U256;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::SystemTime;
@@ -25,11 +24,7 @@ impl PriceSource for PricegraphEstimator {
             let (account_state, orders) =
                 self.orderbook_reader.get_auction_data(batch.into()).await?;
             let mut orderbook = Orderbook::from_elements(orders.iter().map(|order| {
-                order.to_element(U256(
-                    account_state
-                        .read_balance(order.sell_token, order.account_id)
-                        .0,
-                ))
+                order.to_element(account_state.read_balance(order.sell_token, order.account_id))
             }));
             orderbook.reduce_overlapping_orders();
             Ok(self.estimate_prices(tokens, orderbook))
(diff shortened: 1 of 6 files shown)


nlordell (approved):
Great! Thanks for the cleanup :)

core/src/price_estimation/orderbook_based.rs
```diff
@@ -25,11 +24,7 @@ impl PriceSource for PricegraphEstimator {
             let (account_state, orders) =
                 self.orderbook_reader.get_auction_data(batch.into()).await?;
             let mut orderbook = Orderbook::from_elements(orders.iter().map(|order| {
-                order.to_element(U256(
-                    account_state
-                        .read_balance(order.sell_token, order.account_id)
-                        .0,
-                ))
+                order.to_element(account_state.read_balance(order.sell_token, order.account_id))
```
bh2smith, line 27: So what type is it now then?

price-estimator/src/orderbook.rs
```diff
@@ -64,12 +63,8 @@ fn order_to_element(order: &Order, account_state: &AccountState) -> Element {
     // Some conversions are needed because the primitive types crate is on different versions in
     // core and pricegraph.
     Element {
-        user: UserId::from_slice(order.account_id.as_fixed_bytes()),
-        balance: U256(
-            account_state
-                .read_balance(order.sell_token, order.account_id)
-                .0,
-        ),
+        user: order.account_id,
```
bh2smith, line 66: I guess these are now just regular old `u16`?

pricegraph/Cargo.toml
```diff
@@ -11,7 +11,7 @@ harness = false
 [dependencies]
 arbitrary = { version = "0.4", optional = true, features = ["derive"] }
 petgraph = "0.5"
-primitive-types = "0.7"
+primitive-types = "0.6.2"
```
bh2smith, line 14: I find this weird because it looks like you removing the imports of this. I  guess then we are still using it somewhere.

bh2smith (approved):
Looks much better. When did you spot this?

e00E (approved):

core/src/price_estimation/orderbook_based.rs
fleupold: It's still a U256 but now also defined in the same library and the same version as the one that the `Element` struct is expecting.

price-estimator/src/orderbook.rs
fleupold: Address

pricegraph/Cargo.toml
fleupold: It's transitively imported by our web3 dependency. There we import an older version that the current latest. By making our explicit imports use the same version as our indirect dependency inclusion, we don't have to explicitly convert types that are basically the same (but not exactly since different version)

PR #1006, opened 2020-06-15
use `timeLimit` instead of `solverTimeLimit`

nlordell:
I noticed that a new `timeLimit` parameter to constrain the total time (processing + solving) of the solver. I imagine that this should be used instead of the `solverTimeLimit` parameter, since the latter does not take processing time into account.

Maybe @josojo and @marcovc can confirm that this time limit is indeed what we should be using in this case.

### Test Plan

CI, solvers should accept new `timeLimit` parameter.

### DIFF
--- a/core/src/price_finding/price_finder_interface.rs
+++ b/core/src/price_finding/price_finder_interface.rs
@@ -138,10 +138,10 @@ pub fn execute_private_solver(
         .arg(format!("{}{}", "/app/", input_file.to_owned()))
         .arg(format!("--outputDir={}{}", "/app/", result_folder))
         .arg("--logging=WARNING")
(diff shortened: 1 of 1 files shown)


marcovc:
That is correct! That PR was supposed to wait on a subsequent one that makes sure the solver only takes --timeLimit in all cases, which is not true currently. But I think Alex tagged a new version with just this part, which is fine, but I should have warned you. Sorry about that!

nlordell:
> but I should have warned you.

@josojo did :smile:, This PR just implements the change.

bh2smith (approved):

nlordell:
This PR is no longer needed with #1013

PR #1008, opened 2020-06-15
Create new type for computing historic exchange informatino

nlordell:
This PR introduces a new type (that is currently just a giant list of events) to query historic exchange data.

The hopes is for this eventually be used to produces metrics on how "good" the current pricegraph estimates are using historic data.

### Test Plan

New type in `core`, added a unit test but not really being used anywhere.


e00E (approved):
Nothing wrong with the code, just some ideas:

Should this be in core or in a standalone binary?
Should this be reduced to only reading events from the on disk backup file? That way it doesn't use the network making it more pure. Then we could have another tool that creates the file. Not sure if that is better or worse in the end.

nlordell:
> Should this be in core or in a standalone binary?

I was going to make an `e2e/src/bin/historic_prices.rs` file. I just wanted to put the historic data querying logic in the `core` crate (as it "feels" like something that should be shared).

> Should this be reduced to only reading events from the on disk backup file?

I think that is a great idea. In that case we should have an easy `bin` target (in `e2e` crate? WDYT) that can update the disk backup file easily.

e00E:
Sure, could be in e2e or could be another crate. In my mind e2e is only for the test but I don't think it matters that much where the binary is.

core/src/history.rs
```diff
@@ -0,0 +1,195 @@
+//! This module contains an implementation for quering historic echange data by
```
fleupold, line 1: nanonit: querying

core/src/history.rs
```diff
@@ -0,0 +1,195 @@
...
+use std::{collections::HashMap, convert::TryFrom, path::Path};
+
+/// Historic exchange data.
+pub struct ExchangeHistory {
+    _events: Vec<(batch_exchange::Event, BatchId)>,
+}
+
+impl ExchangeHistory {
+    /// Initializes a new indexed orderbook from a web3 provider.
+    pub async fn initialize(
```
fleupold, line 20: Is there a reason we cannot reuse the existing logic from the streamed orderbook for this kind of thing? Is it because the existing code always fetches to the latest block and we might only want to be efficient and fetch less data for the historic queries?
If so, could we adjust the existing code to take an optional block number until which to update?

nlordell: The main reason is that this is not just going to retrieve the orderbook, but also needs additional querying methods such as:
- historic prices (to calculate how far off price estimates are)
- list of orders placed in a single batch
- trades that happened in a single batch.

Note that we should be able to use all the event-based orderbook `State` code (i.e. reconstructing the orderbook from a list of events).

I also really like @e00E's suggestion of making this only use events from the filestore (making this more "pure" and reproducible) but ultimately I strongly believe they should be different types. Once we start adding things (like generating the metrics for each batch while running in the driver) I think that the updating orderbook and the exchange history can definitely share logic for keeping the list of events up to date.

Additionally, to me, the `ExchangeHistory` is more analogous to the `core::streamed::orderbook::Orderbook` type (which keeps a collection of events with their matching batch IDs) and not the `core::streamed::updating_orderbook::UpdatingOrderbook` which has the logic for keeping the former up to date.

To sum up, I definitely think they can share code, but I don't know if its worth the hassle quite yet since they do have very different uses.

core/src/history.rs
nlordell: :see_no_evil:

core/src/history.rs
fleupold: I see your point about needing additional methods, but would think that ExchangeHistory can be composed of a (shared) component to read the orderbook as one part of the data it needs.

It does feel that the vast majority of code in this method already exists somewhere else in the codebase. If you think that `UpdatingOrderbook` does too much (fetching events once **and** doing it periodically afterwards), maybe it can be broken up. I do feel that having a component that can generate an OrderbookReading for a given batchID will have multiple uses.

I'm not sure if I see a functional point of having this function only read from file (after all, the file has to be generated somehow and given historic chain-data is pretty much immutable unless you hack a big DAO that Vitalik himself is invested in it shouldn't change). I see it as a huge performance benefit, but think it should be as optional as for the orderbook fetching.

I don't want to push too hard for sharing logic if you think it's premature.

nlordell: > If you think that UpdatingOrderbook does too much

I don't, and I definitely agree and see shared behaviour between the two, I'm just not sure exactly how this type will end up looking in the end, that is why I'm a bit hesitant to factor it out right now.

> composed of a (shared) component to read the orderbook as one part of the data it needs.

I think that is the purpose of `State`. We can definitely make a utility function that creates `State` from an `Iterator<Item = Element>` which I intend to do.

> It does feel that the vast majority of code in this method already exists somewhere else in the codebase.

You may be right! Maybe I missed something - but to me `UpdatingOrderbook` is very much about having a thread-safe updating logic for querying latest events, while the `initialize` method just queries past events once. This is a pretty significant difference, especially WRT ownership semantics, where the `initialize` method just borrows the `StableXContract` and `BatchedBatchIdReader` implementations for the future while the `UpdatingOrderbook` needs to own them. Additionally it contains paths to deal with reorgs which should be considered errors for the `ExchangeHistory` initialization.

I do agree that the code looks similar and feels like it should use some common logic somewhere. I also think that as this type grows and adds new features that we can end up with duplicated logic here and in the `UpdatingOrderbook` type, which wouldn't be ideal.

> I'm not sure if I see a functional point of having this function only read from file

I've had some intermittent issues with nodes being down or a few weeks ago having Infura not returning all events for blocks when scanning. This is **super** rare though.

Nevertheless, I think I will make the `ExchangeHistory` only read from file for now, my main motivation being that I want to focus on producing historic price estimation metrics at the moment and then I can come back and refactor `UpdatingOrderbook`, `Orderbook` and `ExchangeHistory` to use common event retrieval and updating logic, so that we can keep a running metric on price estimates.

> I don't want to push too hard for sharing logic if you think it's premature.

The thing is I still haven't figured out exactly how this will look in the end. The part that is confusing me is that we really have different use cases:
- Run through historic data and determine is price estimates would have actually matched orders
- Check on every batch that price estimates for placed orders would have been matched.

While ultimately, I think they should share 95% of their code, I haven't quote figured out exactly what the 5% will be.

fleupold: >  I think I will make the ExchangeHistory only read from file for now

I'd assume most of the shared logic potential (create orderbook from past events) goes away if we only read from file since `Orderbook` alread implements from<File>.

Are you planning to do this in this PR, or are you looking at merging this version first?

nlordell: I'm going to do it this PR. I'll request review once its ready.

e00E (approved):

core/src/orderbook/streamed/orderbook.rs
```diff
@@ -111,15 +129,16 @@ impl TryFrom<File> for Orderbook {
     type Error = anyhow::Error;
 
     fn try_from(mut file: File) -> Result<Self> {
-        let mut contents = Vec::new();
-        let bytes_read = file
-            .read_to_end(&mut contents)
+        let orderbook = Orderbook::read(&mut file)
```
fleupold, line 132: Can we not impl `TryFrom<Read>` here?

fleupold (approved):

core/src/orderbook/streamed/orderbook.rs
nlordell: No because Rust doesn't support trait specializations:
```
struct Foo;
impl<T: std::io::Read> std::convert::TryFrom<T> for Foo {
    type Error = ();
    fn try_from(path: T) -> Result<Self, Self::Error> {
        Ok(Foo)
    }
}
````

Gives the error:
```
error[E0119]: conflicting implementations of trait `std::convert::TryFrom<_>` for type `Foo`:
 --> src/lib.rs:2:1
  |
2 | impl<T: std::io::Read> std::convert::TryFrom<T> for Foo {
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: conflicting implementation in crate `core`:
          - impl<T, U> std::convert::TryFrom<U> for T
            where U: std::convert::Into<T>;
```

Specifically, this is because there is a potential conflicting implementation for all U that implement both `Read` and `Into<Foo>`.

Also, from the `TryFrom` docs:
> Simple and safe type conversions that may fail in a controlled way under some circumstances.

So to me, the `TryFrom` is more for modeling conversion behaviour and not decoding, explaining why there is a separate `FromStr` trait for string parsing and not just `TryFrom<&str>` (lifetime variance differences aside).

PR #1009, opened 2020-06-15
Update price estimator to new pricegraph

e00E:
The pricegraph now has a helper method that does exactly what we want.
The "estimate_buy_amount" module became so small that I've removed it.

### Test Plan
Tested with curl to see that price is still correct.


pricegraph/src/lib.rs
```diff
@@ -62,6 +62,7 @@ impl TransitiveOrder {
 
 /// API entry point for computing price estimates and transitive orderbooks for
 /// a give auction.
+#[derive(Clone, Debug)]
```
nlordell, line 65: Oops! Forgot about those :stuck_out_tongue:

nlordell (approved):
Looks good!

fleupold (approved):

PR #1010, opened 2020-06-15
add utility method for computing ask and bid prices

### LINKED ISSUE #1002: Improve `TransitiveOrderbook` API
Currently the `TransitiveOrderbook` API is a bit disjointed from how its actually going to be used. Specifically, orderbooks for a market expect prices to be in the `quote` tokens for both ask and bid orders.

This issue captures the work to improve the `TransitiveOrderbook` struct and extend it with some new methods to:
- Calcualte the inverse price and effective price for a `TransitiveOrder`.
- Enumerate price/volume pairs for ask and bid orders, this is what the front-end displays.

This should be done in the `pricegraph` module so that the intricacies of prices (and fees) are handled internally, so consumers of the library don't really need to think about it.

nlordell:
Fixes #1002 

This PR implements new methods for retrieving transitive orderbook prices expressed in the quote token. This aligns the `TransitiveOrderbook` with the data the `price-estimator` service is intended to produce.

### Test Plan

New unit test :tada:


e00E (approved):

fleupold (approved):
Very nice! TIL about DoubleEndedIterator

nlordell:
> TIL about DoubleEndedIterator

The reason I added this explicitly, is that generally `TransitiveOrderbook`s `asks` and `bids` are ordered, and this allows consumers of the `ask_prices` and `bid_prices` to reverse the order.

PR #1013, opened 2020-06-15
solver bump: hotfix for faster best ring solver

josojo:
Solver bump: See https://github.com/gnosis/dex-solver/pull/277

Additionally, in a previous PR of the solver, the time constraint was changed. Now we have available:
```
 ArgHolder(
            '--timeLimit',
            type=ArgparseHelpers.nonnegative(int),
            default=180,
            help="Time limit (preprocessing + solving + postprocessing) in [s]."
        ),
        ArgHolder(
            '--solverTimeLimit',
            type=ArgparseHelpers.nonnegative(int),
            default=180,
            help="Time limit (just for solving) in [s]."
        ),
```
This PR changes the code so that we always use `timeLimit`, as the driver intended to cap the time of all procedures in the solver.

### Test Plan
e2e only

### DIFF
--- a/core/src/price_finding/price_finder_interface.rs
+++ b/core/src/price_finding/price_finder_interface.rs
@@ -138,7 +138,7 @@ pub fn execute_private_solver(
         .arg(format!("{}{}", "/app/", input_file.to_owned()))
         .arg(format!("--outputDir={}{}", "/app/", result_folder))
         .arg("--logging=WARNING")
-        .args(&["--solverTimeLimit", &time_limit])
+        .args(&["--timeLimit", &time_limit])
         .arg(format!("--minAvgFeePerOrder={}", min_avg_fee_per_order))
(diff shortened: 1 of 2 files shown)


core/src/price_finding/price_finder_interface.rs
```diff
@@ -138,7 +138,7 @@ pub fn execute_private_solver(
         .arg(format!("{}{}", "/app/", input_file.to_owned()))
         .arg(format!("--outputDir={}{}", "/app/", result_folder))
         .arg("--logging=WARNING")
-        .args(&["--solverTimeLimit", &time_limit])
+        .args(&["--timeLimit", &time_limit])
```
nlordell, line 141: uber-nano-nit: for consistency, we should probably change this to be formatted like the other arguments
```suggestion
        .arg(format!("--timeLimit={}", time_limit))
```

nlordell (approved):
:tada: Thanks for looking into this!

PR #1015, opened 2020-06-16
Deploy price estimator

e00E:
And a couple of small changes needed for our staging environment:
* fix urls to start with /api/v1
* allow choosing bind address

Corresponding staging PR: https://gitlab.gnosisdev.com/devops/gnosis-staging/merge_requests/288

## Test Plan
I have tested that this is compatible with the frontend. This is possible with https://pr1019--dexreact.review.gnosisdev.com/ which points to our staging price estimator which I have updated with this docker image.
Alternatively the real frontend https://mesa.eth.link/ can be adapted to use a localhost price estimator by running `dexPriceEstimatorApi.urlsByNetwork[1] = "http://localhost:8080/api/v1/"` in the browser console. For this to work "cors" has to be disabled which I only managed to do with `chromium --disable-web-security --user-data-dir=temp`.
We can see that prices are the same and that the orderbook visualization loads much faster. The visuals are not the same as the old price estimator especially at the edges of the graph but they are similar.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ jobs:
         - ./grcov --branch --ignore-not-existing --llvm --excl-start "mod test" --excl-line "#\[cfg_attr\(test, mockall::automock\)\]|#\[derive" --ignore "/*" --ignore "target/debug/build/**" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-pro --service-job-id $TRAVIS_JOB_ID
         - curl --form json_file=@coveralls.json https://coveralls.io/api/v1/jobs
 
-    - name: Deploy
+    - name: "Deploy Driver"
       if: (type != pull_request) AND (tag is present OR branch = master)
       rust: 1.43.0
       before_install:
@@ -99,5 +99,22 @@ jobs:
             branch: master
         - provider: script
           script: ./docker/deploy.sh $TRAVIS_TAG
+          on:
+            tags: true
+    - name: "Deploy Price Estimator"
+      if: (type != pull_request) AND (tag is present OR branch = master)
+      rust: 1.43.0
+      before_script:
+        - ./scripts/setup_contracts.sh
+      script:
+        - true
+      deploy:
+        - script:
+        - provider: script
+          script: ./price-estimator/docker/deploy.sh $TRAVIS_BRANCH
+          on:
+            branch: master
+        - provider: script
+          script: ./price-estimator/docker/deploy.sh $TRAVIS_TAG
           on:
             tags: true
--- a/price-estimator/docker/Dockerfile
+++ b/price-estimator/docker/Dockerfile
@@ -1,0 +1,21 @@
+FROM rust:1.43-stretch as builder
+WORKDIR /usr/src/app
+COPY Cargo.toml .
+COPY Cargo.lock .
+COPY core core
+COPY driver driver
+COPY dex-contracts dex-contracts
+COPY e2e e2e
+COPY pricegraph pricegraph
+COPY price-estimator price-estimator
+RUN ls -l && cargo build --release -p price-estimator
+
+FROM debian:stretch-slim
+# Needed so that price-estimator works with https node urls.
+RUN apt-get update && apt-get --yes install openssl ca-certificates
+COPY --from=builder /usr/src/app/target/release/price-estimator /bin/
+ADD docker/rust/tini /tini
+RUN chmod +x /tini
+ENTRYPOINT ["/tini", "--"]
+CMD ["price-estimator"]
+
--- a/price-estimator/docker/deploy.sh
+++ b/price-estimator/docker/deploy.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+image_tag=$1
+
+echo "Docker login"
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin;
+
+echo "Building price estimator docker";
+docker build --tag gnosispm/dex-price-estimator-rust:$image_tag -f price-estimator/docker/Dockerfile .
+
+echo "Pushing price estimator docker"
+docker push gnosispm/dex-price-estimator-rust:$image_tag;
+
+echo "done"
+
+if [ "$image_tag" == "master" ] && [ -n "$AUTODEPLOY_URL_PRICE_ESTIMATOR" ]; then
+    # Notifying webhook
+    curl -s  \
+      --output /dev/null \
+      --write-out "%{http_code}" \
+      -H "Content-Type: application/json" \
+      -X POST \
+      -d '{"push_data": {"tag": "'$AUTODEPLOY_TAG_PRICE_ESTIMATOR'" }}' \
+      $AUTODEPLOY_URL_PRICE_ESTIMATOR
+fi
+
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -4,7 +4,7 @@ The service exposes the following endpoints:
 
 ### Markets
 
-`GET /markets/:market`
+`GET /api/v1/markets/:market`
 
 * `market` is of the form `<base_token_id>-<quote_token_id>`. The token ids the same as in the smart contract.
 
@@ -12,7 +12,7 @@ Url Query:
 * `atoms`: If set to `true` (for now this is the only implemented method) all amounts will be denominated in the smallest available unit (base quantity) of the token.
 * `hops`: TODO: document this once it has been implemented.
 
-Example Request: `/markets/1-7/?atoms=true`
+Example Request: `/api/v1/markets/1-7/?atoms=true`
 
 Example Response:
 
@@ -31,14 +31,14 @@ Returns the transitive orderbook (containing bids and asks) for the given base a
 
 ### Estimated Buy Amount
 
-`GET /markets/:market/estimated-buy-amount/:sell-amount-in-quote-token`
+`GET /api/v1/markets/:market/estimated-buy-amount/:sell-amount-in-quote-token`
 
 * `market` is as above
 * `sell_amount_in_quote_token` is a positive integer.
 
 Url query is as above.
 
-Example Request: `/markets/1-7/estimated-buy-amount/20000000000000000000?atoms=true`
+Example Request: `/api/v1/markets/1-7/estimated-buy-amount/20000000000000000000?atoms=true`
 
 Example Response:
 
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -13,7 +13,7 @@ pub fn all<T: Send + Sync + 'static>(
 }
 
 /// Validate a request of the form
-/// `/markets/<baseTokenId>-<quoteTokenId>/estimated-buy-amount/<sellAmountInQuoteToken>`
+/// `/api/v1/markets/<baseTokenId>-<quoteTokenId>/estimated-buy-amount/<sellAmountInQuoteToken>`
 /// and answer it.
 pub fn estimated_buy_amount<T: Send + Sync + 'static>(
     orderbook: Arc<Orderbook<T>>,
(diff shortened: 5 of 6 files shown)


price-estimator/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+WORKDIR /usr/src/app
+# Copy sources
+COPY Cargo.toml .
+COPY Cargo.lock .
+COPY core core
+COPY driver driver
+COPY dex-contracts dex-contracts
+COPY e2e e2e
+COPY pricegraph pricegraph
+COPY price-estimator price-estimator
```
fleupold, line 13: Instead of copying all files 1 by 1, would it make sense to create a Dockerignore file for the folders we don't care about and then just copy `.` 

This would have the advantage that we don't have to add files here (and potentially  in other Dockerfiles) if we do a rename or add a new subproject.

e00E: We already have a dockerignore file but I'm not sure if the same file is going to be correct for the driver and the price estimator. We should look into this but not sure I'll do it in this PR.

price-estimator/docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM debian:stretch-slim
```
nlordell, line 1: If we really want to have an optimized (and tiny) image, maybe we should use `alpine` linux and build `price-estimator` for `musl`?

Just neat a suggestion, not actually something we need to do ATM.

.travis.yml
```diff
@@ -101,3 +101,24 @@ jobs:
...
+      deploy:
+        - script:
+        #- provider: script
+        #  script: ./price-estimator/docker/deploy.sh $TRAVIS_BRANCH
+        #  on:
+        #    branch: master
+        #- provider: script
+        #  script: ./price-estimator/docker/deploy.sh $TRAVIS_TAG
+        #  on:
+        #    tags: true
```
nlordell, line 124: nit: missing newline at the end of file.

price-estimator/docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+# Needed so that price-estimator works with https node urls.
+RUN apt-get update && apt-get --yes install openssl ca-certificates
+COPY target/release/price-estimator /bin/
+ADD docker/rust/tini /tini
+RUN chmod +x /tini
+ENTRYPOINT ["/tini", "--"]
+CMD price-estimator
```
nlordell, line 8: Should this be:
```suggestion
CMD ["price-estimator"]
```

From the docker reference: https://docs.docker.com/engine/reference/builder/#cmd

> The `CMD` instruction has three forms:
> - `CMD ["executable","param1","param2"]` (exec form, this is the preferred form)
> - `CMD ["param1","param2"]` (as default parameters to ENTRYPOINT)
> - `CMD command param1 param2` (shell form)
>
> [..]
>
> If `CMD` is used to provide default arguments for the `ENTRYPOINT` instruction, both the `CMD` and `ENTRYPOINT` instructions should be specified with the JSON array format.

price-estimator/docker/deploy.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
nlordell, line 1: I like this in its own sub-directory! Maybe (probably in a separate PR) we should do this for the driver as well?

price-estimator/docker/Dockerfile
e00E: Agree that we could do it and agree that we don't need this right now. Can keep in mind for the future.

price-estimator/docker/deploy.sh
e00E: I agree that we should do this.

.travis.yml
```diff
@@ -101,3 +101,20 @@ jobs:
           script: ./docker/deploy.sh $TRAVIS_TAG
           on:
             tags: true
+    - name: "Deploy Price Estimator"
```
e00E, line 104: This shows one disadvantage of the monorepo: We have to deploy the price estimator every time a new commit lands in master. It wouldn't be enough to check to the last modification date of the price-estimator folder because a change in pricegraph or core could affect the priceestimator. If the price estimator was in it's own repo (for example using a git submodule for core and pricegraph or a cargo git dependency) then this wouldnt' be the case. Of course in that case we would have to manually update the dependency when needed.
I'm not saying we should do this just want to point it out.

price-estimator/docker/deploy.sh
```diff
@@ -0,0 +1,27 @@
...
+image_name=$1
+
+echo "Docker login"
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin;
+
+echo "Building price estimator docker";
+docker build --tag gnosispm/dex-price-estimator-rust:$image_name -f price-estimator/docker/Dockerfile .
+
+echo "Pushing price estimator docker"
+docker push gnosispm/dex-price-estimator-rust:$image_name;
```
nlordell, line 14: nit: I would just call it `dex-price-estimator`, I feel like it being Rust is an implementation detail that can be encoded in the tag name.

price-estimator/docker/deploy.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+image_name=$1
```
nlordell, line 5: nit: In Docker lingo, I believe this is called a "tag". I find calling the variable "image_name" a bit confusing (to me the image name is "dex-price-esimator-rust"):

```suggestion
image_tag=$1
```

price-estimator/docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN ls -l && cargo build --release -p price-estimator
+
+FROM debian:stretch-slim
+# Needed so that price-estimator works with https node urls.
+RUN apt-get update && apt-get --yes install openssl ca-certificates
+COPY --from=builder /usr/src/app/target/release/price-estimator /bin/
+ADD docker/rust/tini /tini
+RUN chmod +x /tini
+ENTRYPOINT ["/tini", "--"]
+CMD price-estimator
```
nlordell, line 20: DId `CMD ["price-estimator"]` not work?

nit: missing newline.

nlordell (commented):
Looks good overall, just some minor nits

price-estimator/docker/Dockerfile
nlordell: Reading a bit more from the Docker reference, it looks like:
> If you use the shell form of the CMD, then the <command> will execute in /bin/sh -c:

So this is running `price-estimator` without `tini`. You can probably verify with `docker top`.

price-estimator/docker/deploy.sh
e00E: The reasoning for this name is that the nodejs price estimator image is called `dex-price-estimator`. I did not want to replace that name so I appended `rust`.

price-estimator/docker/Dockerfile
e00E: I think I accidentally reverted that. Will put it back.

price-estimator/docker/deploy.sh
e00E: Will change it to that. Reason for this name is that it's the same as in the driver's deploy script.

price-estimator/docker/deploy.sh
nlordell: I would argue that we are replacing the nodejs price estimator with the Rust one, so we should re-appropriate the image name. I think having different tags is enough to differentiate the different versions.

We could even create a `:legacy` tag just for the latest nodejs version so its archived and easily retrievable.

.travis.yml
nlordell: I think this is also a "pro" since we have CD and integration errors with `price-estimator` and `driver` with `core` and `pricegraph` are caught early and tested often. It just adds CI overhead, but Travis is a :robot: and doesn't mind :smile:.

One point though (that you've made before) that I find more of an issue is regarding deploying tagged versions. We currently have no way of individually deploying the price estimator and driver which can be annoying, especially if we need to do a hotfix.

price-estimator/docker/deploy.sh
e00E: Just to be safe I don't want to every accidentally override one image with the other (like we bump the old price estimator for some reason and it overrrides the master tag). So I'd rather have separate names.

nlordell: What about `image_tag="$1-rust"`? IMO its better to have more tags and a single image than a bunch of images.

For example, the NodeJS image does this to select between [different operating systems](https://hub.docker.com/_/node/). Especially because these are intended to be the same service, it makes more sense for them to have the same name.

.travis.yml
fleupold: If we want to be fancy we can look at the project dependency graph and skip a build if no file in the dependency graph was changed. This is how large scale monorepos do it if I'm not mistaken.

price-estimator/docker/deploy.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo "Building price estimator docker";
+docker build --tag gnosispm/dex-price-estimator-rust:$image_name -f price-estimator/docker/Dockerfile .
+
+echo "Pushing price estimator docker"
+docker push gnosispm/dex-price-estimator-rust:$image_name;
+
+echo "done"
+
+if [ "$image_name" == "master" ] && [ -n "$AUTODEPLOY_URL_PRICE_ESTIMATOR" ] && [ -n "$AUTODEPLOY_TOKEN_PRICE_ESTIMATOR" ]; then
```
fleupold, line 18: I don't think the token is needed anymore. @davidalbela can you confirm?

price-estimator/docker/deploy.sh
```diff
@@ -0,0 +1,27 @@
...
+    # Notifying webhook
+    curl -s  \
+      --output /dev/null \
+      --write-out "%{http_code}" \
+      -H "Content-Type: application/json" \
+      -X POST \
+      -d '{"push_data": {"tag": "'$AUTODEPLOY_TAG_PRICE_ESTIMATOR'" }}' \
+      $AUTODEPLOY_URL_PRICE_ESTIMATOR
+fi
```
fleupold, line 27: It's a shame that we blatantly copy this deploy script across repos but now even inside the same repo. I think there should be a way to share this logic in a function and just invoke it with the few arguments that are different. Maybe it could even be imported from a shared source across repos.

fleupold (approved):

price-estimator/docker/deploy.sh
e00E: But then it's an exception in our images that this one doesn't use "master" or the git tag as the image tag. Seems like more disadvantages just to avoid an imperfect name.

price-estimator/docker/deploy.sh
e00E: It's set to "notrequired" in CI so probably not but we are using it in the script in the nodejs price estimator even though it has this value.

price-estimator/docker/deploy.sh
e00E: The driver script does more things than that one, I don't see a good way to unify them without making it overly complicated. Unless you just mean the curl but in that case it's still only one command so I don't see anything wrong with duplicating it.

.travis.yml
nlordell: I think, as @e00E suggested, most of the code is shared anyway:

```
$ find $(ls -d */ | grep -v dex-contracts | grep -v target) -iname "*.rs" -exec cat {} \; | wc -l     
16529
$ find driver price-estimator -iname "*.rs" -exec cat {} \; | wc -l 
900
```

So a change in roughly 95% of the codebase would require building both driver and the price-estimator (note I did not include `e2e/` in the second line count as I imagine we might want to add price estimator e2e tests eventually).

price-estimator/docker/deploy.sh
nlordell: I disagree, but I also don't think this super important. I'm fine with `dex-price-estimator-rust` but suggest revisiting this if/when we deprecate and archive the gnosis/dex-price-estimator so this wouldn't be a concern any more.

price-estimator/docker/deploy.sh
nlordell: I think the biggest issue here is if we change the deployment process from a DevOps standpoint, then each of these scripts in all of our repos need to be updated manually. I agree with @fleupold that perhaps investing some time in creating a CI repos with common scripts would be beneficial to streamline some of these processes and simplify the update process in case of changes to our deployment process.

That being said, I also think it is outside the scope of this PR.

nlordell (approved):

price-estimator/docker/deploy.sh
fleupold: > The driver script does more things than that one

You are right, it's pushing two images in the same script, but in theory the logic for the two public images is exactly the same.

PR #1018, opened 2020-06-16
Fix pricegraph edge cases

e00E:
`reduce_overlapping_transitive_orderbook` was missing a check for the
token pair validity which would cause a panic in bellman_ford::search
when the quote token didn't exist.

`fill_transitive_orders` had an infinite loop with an empty orderbook
and the token pair buying and selling the fee token. `path::find_path`
would in this case keep returning `Some`.

### Test Plan
Added unit test and noticed originally in #923  .

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -166,6 +166,13 @@ impl Orderbook {
         base: TokenId,
         quote: TokenId,
     ) -> TransitiveOrderbook {
+        if !self.is_token_pair_valid(TokenPair {
+            buy: base,
+            sell: quote,
+        }) {
+            return Default::default();
+        }
+
         let (base, quote) = (node_index(base), node_index(quote));
 
         let mut overlap = TransitiveOrderbook::default();
@@ -578,11 +585,15 @@ impl Orderbook {
         Some((order, user))
     }
 
-    /// Returns `true` if the specified token pair is valid, and `false`
-    /// otherwise.
+    /// Returns whether the specified token pair is valid.
+    ///
+    /// A token pair is considered valid if both the buy and sell token
+    /// exist in the current orderbook and are unequal.
     fn is_token_pair_valid(&self, pair: TokenPair) -> bool {
-        let max_token = (self.projection.node_count() - 1) as u16;
-        pair.buy <= max_token && pair.sell <= max_token
+        let node_count = self.projection.node_count();
+        pair.buy != pair.sell
+            && (pair.buy as usize) < node_count
+            && (pair.sell as usize) < node_count
     }
 }
 
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -217,6 +217,14 @@ mod tests {
     use super::*;
 
     #[test]
+    fn transitive_orderbook_empty_same_token() {
+        let pricegraph = Pricegraph::new(std::iter::empty());
+        let orderbook = pricegraph.transitive_orderbook(0, 0, None);
+        assert!(orderbook.asks.is_empty());
+        assert!(orderbook.bids.is_empty());
+    }
+
+    #[test]
     fn transitive_orderbook_prices() {
         let transitive_orderbook = TransitiveOrderbook {
             asks: vec![


pricegraph/src/orderbook.rs
```diff
@@ -166,6 +166,14 @@ impl Orderbook {
         base: TokenId,
         quote: TokenId,
     ) -> TransitiveOrderbook {
+        if !self.is_token_pair_valid(TokenPair {
+            buy: base,
+            sell: quote,
```
e00E, line 171: This is the missing bounds check.

pricegraph/src/orderbook.rs
```diff
@@ -233,7 +241,7 @@ impl Orderbook {
             assert!(spread > 0.0, "invalid spread");
         }
 
-        if !self.is_token_pair_valid(pair) {
+        if !self.is_token_pair_valid(pair) || pair.buy == pair.sell {
```
e00E, line 244: This is the infinite loop.

pricegraph/src/orderbook.rs
```diff
@@ -578,11 +586,9 @@ impl Orderbook {
         Some((order, user))
     }
 
-    /// Returns `true` if the specified token pair is valid, and `false`
-    /// otherwise.
     fn is_token_pair_valid(&self, pair: TokenPair) -> bool {
-        let max_token = (self.projection.node_count() - 1) as u16;
-        pair.buy <= max_token && pair.sell <= max_token
+        let node_count = self.projection.node_count();
+        (pair.buy as usize) < node_count && (pair.sell as usize) < node_count
```
e00E, line 591: This change isn't needed but at first I thought the `- 1` was a mistake until I realized that there is always a fee token so node_count is never 0. I find this way clearer.
Removed the doc comment because it doesn't add information to the signature.

I don't mind reverting this if we disagree, just included it because I came across it.

pricegraph/src/orderbook.rs
```diff
@@ -166,6 +166,14 @@ impl Orderbook {
         base: TokenId,
         quote: TokenId,
     ) -> TransitiveOrderbook {
+        if !self.is_token_pair_valid(TokenPair {
+            buy: base,
+            sell: quote,
+        }) || base == quote
```
e00E, line 172: Without this check the new test passes but the fuzzer finds an instance where a transitive order ends up as `NaN`. I haven't saved the instance but if it's interesting it should be able to find it again.

Conceptually it doesn't make sense to request the transitive orderbook for the same token, does it? If it does then we might have to fix this in a different way.

pricegraph/src/orderbook.rs
nlordell: Good catch!

nlordell: nit: I would have this be a `is_market_valid` method that is `is_token_pair_valid && buy != sell`.

pricegraph/src/orderbook.rs
nlordell: I'm fine with this change, I also agree it is a little more readable this way.

pricegraph/src/orderbook.rs
nlordell: > Conceptually it doesn't make sense to request the transitive orderbook for the same token, does it?

Hmm, I can't think of a particular use case for it. I think its fine for now.

pricegraph/src/orderbook.rs
```diff
@@ -578,11 +586,9 @@ impl Orderbook {
         Some((order, user))
     }
 
-    /// Returns `true` if the specified token pair is valid, and `false`
-    /// otherwise.
```
nlordell, line 582: nit: I would prefer to leave the doc comment in, maybe we can modify it to be more informative:
```rust
    /// Returns `true` if the specified token pair is valid, and `false`
    /// otherwise.
    ///
    /// A token pair is considered valid if both the buy and sell token
    /// exist in the current orderbook.
```

pricegraph/src/orderbook.rs
```diff
@@ -578,11 +586,9 @@ impl Orderbook {
...
     fn is_token_pair_valid(&self, pair: TokenPair) -> bool {
```
nlordell, line 589: Now that I think about it, maybe even for price estimates we should ignore `buy == sell` token pairs. My main argument here is that (AFAIK) the solver currently ignores these orders anyway, so you can't really specify an overlapping price for them.

fleupold (approved):
I'd argue that the `base == quote` check should be part of the `is_token_pair_valid` method since for the purpose of a price graph a pair needs to have different buy and sell tokens. It's also easy to forget this check so it would be nice if it could be included in the general validity check.

nlordell (approved):
Nice!

PR #1021, opened 2020-06-17
Finish implementing transitive orderbook

e00E:
Thanks to Nick's helper method this is very easy now.

### Test Plan
Run locally and compare to nodejs price estimator. There are a lot of differences so it's hard to tell which side is more correct but I do see rough correspondences. It's easier to tell for token pairs that have less transitive orders like `1-10`.

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -78,15 +78,17 @@ async fn estimate_buy_amount<T>(
 }
 
 async fn get_markets<T>(
-    _token_pair: TokenPair,
+    token_pair: TokenPair,
     _query: QueryParameters,
-    _orderbook: Arc<Orderbook<T>>,
+    orderbook: Arc<Orderbook<T>>,
 ) -> Result<impl warp::Reply, Infallible> {
-    // TODO: implement
-    Ok(warp::reply::json(&MarketsResult {
-        asks: Vec::new(),
-        bids: Vec::new(),
-    }))
+    let transitive_orderbook = orderbook.get_pricegraph().await.transitive_orderbook(
+        token_pair.sell_token_id,
+        token_pair.buy_token_id,
+        None,
+    );
+    let result = MarketsResult::from(&transitive_orderbook);
+    Ok(warp::reply::json(&result))
 }
 
 #[cfg(test)]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2192,7 +2192,6 @@ name = "price-estimator"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "assert_approx_eq",
  "core",
  "env_logger",
  "ethcontract",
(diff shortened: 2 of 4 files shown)


price-estimator/src/filter.rs
```diff
@@ -78,15 +78,17 @@ async fn estimate_buy_amount<T>(
...
 ) -> Result<impl warp::Reply, Infallible> {
-    // TODO: implement
-    Ok(warp::reply::json(&MarketsResult {
-        asks: Vec::new(),
-        bids: Vec::new(),
-    }))
+    let transitive_orderbook = orderbook.get_pricegraph().await.transitive_orderbook(
+        token_pair.sell_token_id,
+        token_pair.buy_token_id,
+        None,
```
e00E, line 88: Should we use a spread?

    /// This method optionally accepts a spread that is a decimal fraction that
    /// defines the maximume transitive order price with the equation:
    /// `first_transitive_price + first_transitive_price * spread`. This means
    /// that given a spread of 0.5 (or 50%), and if the cheapest transitive
    /// order has a price of 1.2, then the maximum price will be `1.8`.

price-estimator/src/filter.rs
```diff
@@ -78,15 +78,17 @@ async fn estimate_buy_amount<T>(
...
+    orderbook: Arc<Orderbook<T>>,
 ) -> Result<impl warp::Reply, Infallible> {
-    // TODO: implement
-    Ok(warp::reply::json(&MarketsResult {
-        asks: Vec::new(),
-        bids: Vec::new(),
-    }))
+    let transitive_orderbook = orderbook.get_pricegraph().await.transitive_orderbook(
+        token_pair.sell_token_id,
+        token_pair.buy_token_id,
```
e00E, line 87: To get the same results as the nodejs price estimator I had to put sell_token as the first argument here. Not sure if this is expected. This is clear when trying a token pair that only either has asks or bids but not the other like `1-10`.

price-estimator/src/filter.rs
e00E: Since the nodejs price estimator didn't have this I haven't set it here either.

price-estimator/src/filter.rs
nlordell: Hmm, this may be an issue with the direction of ask and bid orders... They *might* be inverted in *pricegraph*. @fleupold , do you know if "ask" orders are "buying base token and selling quote token" or "buying quote token and selling base token"? I'm pretty sure "pricegraph" uses the first definition (at least, that is the intention).

nlordell (approved):
Looks great! My only concern is the same one you brought up about the ask and bid order directions.

price-estimator/src/filter.rs
nlordell: Originally, I thought that the runtime for calculating all transitive orders for a token pair would be high, so the spread was supposed to limit this. It turned out not to be the case, so I think we can leave it out.

price-estimator/src/filter.rs
e00E: If you have `/markets/x-y/estimated-buy-amount/1` return something (not `None`), that means you can sell y for x, so should `/markets/x-y/` be guaranteed to have at least one bid or at least one ask?

nlordell: > at least one bid or at least one ask?

That is the question!

nlordell: According to wikipedia:
> Bids (buyers) on the left, asks (sellers)

And since the price is in quote token, to me this implies bids are buying base token and selling quote token, and asks are buying quote token and selling base token. So the `pricegraph` has it backwards.

I will wait on @fleupold to confirm before making `pricegraph` changes.

Eitherway, I think the base token should be the "buy" token, and the quote token should be the "sell" token in the call to `Pricegraph::transitive_orderbook` (as that was the expectation), and that the direction should be fixed internally in the `pricegraph` crate.

fleupold: If you are looking at an ETH/USD orderbook you can buy ETH from the asks for USD. Thus the asks buying USD (quote token) and selling ETH (base token). So @nlordell, your statement above is correct. This is another thing that is super ambiguous and I always have to look at existing orderbooks for this.

fleupold (approved):

price-estimator/src/filter.rs
e00E: I am going to merge this then as it is correct right now. Then in separate PR we will switch buy/sell here and in the pricegraph so that is stays correct but has the expected parameter naming.

PR #1025, opened 2020-06-17
Implement querying historic orderbooks

nlordell:
This PR implements querying finalized orderbooks at historic batches.

### Test Plan

Added unit test!

### DIFF
--- a/core/src/history.rs
+++ b/core/src/history.rs
@@ -2,14 +2,17 @@
 //! inspecting indexed events.
 
 use crate::{
-    contracts::stablex_contract::batch_exchange, models::BatchId, orderbook::streamed::orderbook,
+    contracts::stablex_contract::batch_exchange,
+    models::BatchId,
+    models::{AccountState, Order},
+    orderbook::streamed::{orderbook, State},
 };
 use anyhow::Result;
 use std::{fs::File, io::Read, path::Path};
 
 /// Historic exchange data.
 pub struct ExchangeHistory {
-    _events: Vec<(batch_exchange::Event, BatchId)>,
+    events: Vec<(batch_exchange::Event, BatchId)>,
 }
 
 impl ExchangeHistory {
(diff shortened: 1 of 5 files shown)


core/src/history.rs
```diff
@@ -22,18 +25,62 @@ impl ExchangeHistory {
...
             .collect();
 
-        Ok(ExchangeHistory { _events: events })
+        Ok(ExchangeHistory::from_events(events))
+    }
+
+    fn from_events(events: Vec<(batch_exchange::Event, BatchId)>) -> Self {
+        // NOTE: Since history methods depend on events being sorted, assert
+        // this invariant is true!
+        debug_assert!(events.windows(2).all(|event_pair| {
```
e00E, line 34: In case someone is wondering about this. There is an `is_sorted` function in std but it doesn't allow specifying a comparison function. That function is called `is_sorted_by` but it is not yet stable.

nlordell: The other option is to use `itertools` which adds this. I wouldn't mind adding the dependency as I can see it being useful in other situations.

core/src/history.rs
```diff
@@ -22,18 +25,62 @@ impl ExchangeHistory {
...
+        self.events
+            .iter()
+            .take_while(move |(_, batch_id)| *batch_id <= end_batch_id)
+            .map(|(event, batch_id)| (event, *batch_id))
+    }
+
+    /// Returns the finalized orderbook at a historic batch.
+    ///
+    /// The finalized orderbook includes all account state changes and orders
+    /// that are consisdered by the smart contract for solving the subsequent
```
e00E, line 64:
```suggestion
    /// that are considered by the smart contract for solving the subsequent
```

core/src/history.rs
```diff
@@ -22,18 +25,62 @@ impl ExchangeHistory {
...
+        &self,
+        batch_id: impl Into<BatchId>,
+    ) -> Result<(AccountState, Vec<Order>)> {
+        let batch_id = batch_id.into();
+
+        State::from_events(
+            self.events_until_end_of_batch(batch_id)
+                .map(|(event, batch_id)| (event, batch_id.into())),
+        )?
+        .normalized_auction_state_at_beginning_of_batch(batch_id.next().into())
```
e00E, line 76: Probably fine but for some historical analysis we might not want to normalize as it removes some information.

e00E (approved):

core/src/history.rs
nlordell: I'm OK either way. My main reasoning for doing this is:
- AFAIK, In the SC there is no real difference between an untouched balance (either a user that never used the exchange or never used a specific token) and a 0 balance (deposit + withdraw request + withdraw)
- This is not true for orders, but we already lose information about cancelled and deleted orders, so I thought it would be easier to use some normalized view for orders to remove any ambiguity.

fleupold: What information do we lose? Is this about orders that don't have sufficient balance or are not valid? I agree that for now we can normalize it away.

Technically though, this could be an issue since we can receive proceeds in a trade that can be used in the same settlement for selling a token (e.g. if I am willing to sell and buy DAI but have no DAI at the moment I could be getting it in a first trade from token A and sell it in the same batch for token B). However, the price estimator does not cope with this today so this information is not needed.

core/src/history.rs
```diff
@@ -22,18 +25,62 @@ impl ExchangeHistory {
...
+            .take_while(move |(_, batch_id)| *batch_id <= end_batch_id)
+            .map(|(event, batch_id)| (event, *batch_id))
+    }
+
+    /// Returns the finalized orderbook at a historic batch.
+    ///
+    /// The finalized orderbook includes all account state changes and orders
+    /// that are considered by the smart contract for solving the specified
+    /// batch.
+    pub fn auction_data_for_batch(
```
fleupold, line 66: Couldn't this be expressed via the OrderbookReading trait? I'm still not entirely sure why we cannot make the existing OrderbookReading handle historic batches.

fleupold (approved):
For the "bad UX" metrics we probably want to go even more fine-grained and allow to get the orderbook at a certain block (to simulate the state of the orderbook/price estimation service) close to when the user placed their order.

core/src/history.rs
nlordell: > Technically though, this could be an issue since we can receive proceeds in a trade that can be used in the same settlement for selling a token

Thats a really good point! I I would be curious to know if the solver considers these orders ATM (@marcovc or @twalth3r can you confirm confirm?).

nlordell:
> For the "bad UX" metrics we probably want to go even more fine-grained and allow to get the orderbook at a certain block

That's a good point, although some of this is outside our control - for example another user may place an order after that offers more utility and/or less disregarded utility and will be preferred over the original order. Nevertheless it is still useful to know how often this happens, so might as well collect this data.

core/src/history.rs
nlordell: :see_no_evil: I totally brainfarted. Yeah that's a super good point, I will refactor this.

core/src/history.rs
marcovc: The current situation is to forbid it - orders max sell amount are capped by the account balances. That means that a person cannot sell tokens if their account balance is zero. However, there is an ongoing discussion to lift this constraint.

nlordell: I just realized,
```
        .filter(|order| order.remaining_sell_amount > 0)
```

We only filter out orders with 0 remaining sell amount, but orders with remaining sell amount and 0 balance are still left in. So this shouldn't be a concern :smile:

bh2smith: I had to dig into the code to figure out what it meant to "normalize_auction_data" and found that this is more like a reduce or filter auction data that simply removed irrelevant details. More precisely:

```
/// Removes empty orders and token balances for which there is
/// not at least one sell order by that user
```

I would like to propose that maybe we use a term like `reduce_auction_data` since normalization usually refers to factoring out a scalar from the whole. Of course, this would be for a separate PR, but I this confusing term took me down a rabbit hole.

fleupold: I also find the term normalize a bit vague here (I believe it's more based on https://en.wikipedia.org/wiki/Database_normalization than scalar extraction). I'd be very happy about a rename.

nlordell: I think this should be `canonicalize`, since we are creating a canonical representation of an orderbook.

bh2smith: Yes. I’d be much happier with this term here.

nlordell:
Closing this PR as it has been superceeded by other PRs. Notably #1069  and #1066

core/src/history.rs
nlordell: This was implemented as part of https://github.com/gnosis/dex-services/pull/1063 :smile:

PR #1029, opened 2020-06-17
fix transitive orderbook ask/bid direction

nlordell:
Fixes the direction of ask/bid orders (again :rofl:).

:warning: Please review with caution! :warning: 

### Test Plan

Updated unit tests by flipping ask and bid orders without changing values. They seem to pass!

### DIFF
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -16,11 +16,11 @@ const FEE_FACTOR: f64 = 1.0 / 0.999;
 /// A struct representing a transitive orderbook for a base and quote token.
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct TransitiveOrderbook {
-    /// Transitive "ask" orders, i.e. transitive orders buying the base token
-    /// and selling the quote token.
-    pub asks: Vec<TransitiveOrder>,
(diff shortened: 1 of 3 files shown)


e00E (approved):
Tested with price estimator without and with this PR. Looks correct.

fleupold (approved):
I'm sorry I didn't notice this before 😢

nlordell:
> I'm sorry I didn't notice this before cry

No worries! These logic bugs are super subtle and hard to find.

pricegraph/src/lib.rs
```diff
@@ -32,7 +32,7 @@ impl TransitiveOrderbook {
     pub fn ask_prices(&self) -> impl DoubleEndedIterator<Item = (f64, f64)> + '_ {
         self.asks
             .iter()
-            .map(|order| ((order.buy / order.sell) * FEE_FACTOR, order.sell))
+            .map(|order| ((order.sell / order.buy) / FEE_FACTOR, order.buy))
```
fleupold, line 35: Trying to find the issue (cf #1037). Is this change correct? Asks sell quote and buy base. E.g. in ETH-USD they sell USD for ETH. The price (in quote) should be around $200, but if I sell 1ETH for $200 this evaluates to 0.005 (which is backwards).

Same for bids.

PR #1034, opened 2020-06-19
Build price estimator image based on alpine

### LINKED ISSUE #1027: Optimize the price estimator docker image
We could try using a very small linux distro like alpine with musl to decrease the container size.

This isn't important, just nice to have.

nlordell:
Fixes #1027 

This PR adds the necessary changes in order to build an alpine based image for the price estimator.

Note that the image is still fairly large (price estimator is a 20MB?! binary).
```
$ docker images
REPOSITORY                                      TAG             IMAGE ID       CREATED          SIZE
pe                                              latest          6a21ba0a7202   35 minutes ago   26.4 MB
```

This is currenty implemented by pulling a `crux/muslrust` image that is setup for cross compiling x86_64 musl targets and includes musl libraries for linking (such as openssl). 

For comparison, here is the current image, so its about 1/4 of the size (also the Rust image is 1.23GB).

```
$ docker images
REPOSITORY                                      TAG             IMAGE ID       CREATED              SIZE
pe                                              latest          6ea8d7a1e1d1   About a minute ago   101 MB
```

### Test Plan

```
$ docker build -t pe -f price-estimator/docker/Dockerfile .
...
$ docker run -it --rm -e RUST_LOG=info -e ETHEREUM_NODE_URL pe
[2020-06-19T10:25:02Z INFO  price_estimator] Starting price estimator with runtime options: Options {
```

### DIFF
--- a/price-estimator/docker/Dockerfile
+++ b/price-estimator/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.43-stretch as builder
+FROM clux/muslrust:1.43.1-stable as builder
 WORKDIR /usr/src/app
 COPY Cargo.toml .
 COPY Cargo.lock .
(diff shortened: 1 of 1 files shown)


e00E:
If it's just a change in the Dockerfile then I don't mind using alpine. I was afraid it would be more work to get the binary working. How would cross compiling openssl look? Is that also contained to the docker image.
On the other hand I still see the argument that this makes the build process more complicated in the sense that the docker image used is more arcane than the standard debian rust based one so if something goes wrong it might be harder to fix. But if that happens we can still switch back to the debian one.
In practice whether the image is 30 MB or 100 MB doesn't matter but I do like the idea of being more minimal.

nlordell:
> How would cross compiling openssl look?

I imagine it would involve `apt install`-ing things to the Travis image namely a `musl` toolchain, and enabling `vendor` feature of the `openssl` crate. I didn't get it to work with clang, but it should would with `musl-gcc`, as it works in the `crux/muslrust` image which uses the `musl-gcc` cross compiling toolchain.

> But if that happens we can still switch back to the debian one.

My thoughts exactly.

nlordell:
Building with Travis is also not difficult, it requires:
- musl C toolchain `apt install musl-tools`
- enabling the `vendored` feature for the `openssl-sys` crate which requires a Cargo.toml change

The upside of using a Docker build is that there are no `Cargo.toml` changes and the price estimator musl binary and Docker image can easily be built without installing additional tools.

e00E (approved):

PR #1037, opened 2020-06-22
Revert "fix transitive orderbook ask/bid direction (#1029)"

e00E:
This reverts commit f9abf3ea0b4c31724874d63a04de5a85c0ebeed8.

The reason for this is that when I tested the rust price estimator
against the node price estimator again just now I noticed that the asks
and bids are reverted: visualizing the orderbook for pair A-B in the
node price estimator is the same as pair B-A in the rust price
estimator.
The reason we made the PR I am reverting here is given in https://github.com/gnosis/dex-services/pull/1021#discussion_r441311440 .

I am not yet sure where the bug is but until we find it we should
revert.

### Test Plan
Visually comparing both price estimators through the front end.

### DIFF
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -16,11 +16,11 @@ const FEE_FACTOR: f64 = 1.0 / 0.999;
 /// A struct representing a transitive orderbook for a base and quote token.
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct TransitiveOrderbook {
-    /// Transitive "ask" orders, i.e. transitive orders buying the quote token
-    /// and selling the base token.
-    pub asks: Vec<TransitiveOrder>,
-    /// Transitive "bid" orders, i.e. transitive orders buying the base token
+    /// Transitive "ask" orders, i.e. transitive orders buying the base token
     /// and selling the quote token.
+    pub asks: Vec<TransitiveOrder>,
+    /// Transitive "bid" orders, i.e. transitive orders buying the quote token
+    /// and selling the base token.
     pub bids: Vec<TransitiveOrder>,
 }
 
@@ -32,7 +32,7 @@ impl TransitiveOrderbook {
     pub fn ask_prices(&self) -> impl DoubleEndedIterator<Item = (f64, f64)> + '_ {
         self.asks
             .iter()
-            .map(|order| ((order.sell / order.buy) / FEE_FACTOR, order.buy))
+            .map(|order| ((order.buy / order.sell) * FEE_FACTOR, order.sell))
     }
 
     /// Returns an iterator with bid prices (expressed in the quote token) and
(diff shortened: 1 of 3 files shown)


fleupold (approved):
I trust your testplan and am surprised this fixes it as I would have thought the PR we are reverting did not changed the logic (it only inversed the order in two places). But you are right, we should rethink this part with care and revert for now.

nlordell (commented):
I'm assuming that reverting this makes it match the nodejs price estimator? Its so strange because the PR being reverted is just changes the transitive order direction in the pricegraph instead of the route handler.

Personally I would prefer to get a little more intuition on what is causing this inconsistency before reverting (as the description of ask and bid orders seem to have been corrected by the PR), especially since this isn't live right now and will need to figure out what is going on before releasing to production anyway.

e00E:
I'm adding a test to this PR right now to make it more clear what's going on.

fleupold:
> Personally I would prefer to get a little more intuition on what is causing this inconsistency before reverting 

@nlordell could https://github.com/gnosis/dex-services/pull/1029/files#r443476929 be the issue

nlordell:
@fleupold Yes, I think it is. It also makes sense why in the FE the red and green sides are flipped.

fleupold:
If that's the case, we could also fix forward (instead of reverting), up to you.

e00E:
I agree that we should fix forward. Could you look at the test I added to see if this it is correct?
If so then I will close this PR and make a one with just a fix.

nlordell:
Here are the two graphs that @e00E shared on Slack:
image

The top image, (the Rust version - what we have here) seems to have the prices inverted since 1 OWL is definitely not 200 WETH. Maybe, the issue is just in the ask/bid price calculation as @fleupold pointed out. In which case, I think a more targeted PR to fix that issue would be better.

pricegraph/src/lib.rs
```diff
@@ -224,6 +225,42 @@ mod tests {
...
+        }]);
+        let orderbook = pricegraph.transitive_orderbook(0, 1, None);
+        assert_eq!(
+            orderbook.asks,
+            vec![TransitiveOrder {
+                buy: 2.0 * base as f64,
+                sell: base as f64,
+            }]
+        );
+        assert_eq!(orderbook.bids, vec![]);
```
nlordell, line 252: I think this test is backwards (at this point, I'm not really sure though)... from what I understand ask orders are buying quote token and selling base token, and the transitive orderbook was created with `base = 0` and `quote =1`, here the test asserts that the order buying base for quote is an ask.

nlordell: Maybe @fleupold can confirm though, because I've gotten this backwards so many times I don't trust myself anymore :stuck_out_tongue:

e00E:
New PR https://github.com/gnosis/dex-services/pull/1038

PR #1038, opened 2020-06-22
Fix transitive orderbook prices being inverted

e00E:
The prices were being inverted as seen when visually comparing the transitive orderbook from the node price estimator and the rust price estimator.

### Test Plan
Added new test.

### DIFF
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -302,7 +302,7 @@ mod tests {
         let bid_prices = transitive_orderbook.bid_prices().collect::<Vec<_>>();
         assert_approx_eq!(bid_prices[0].0, 2.0 / FEE_FACTOR);
         assert_approx_eq!(bid_prices[0].1, 1_000_000.0);
-        assert_approx_eq!(bid_prices[1].0, 9.0 / 5.0 / FEE_FACTOR);
+        assert_approx_eq!(bid_prices[1].0, (9.0 / 5.0) / FEE_FACTOR);
         assert_approx_eq!(bid_prices[1].1, 500_000.0);
     }
 
--- a/pricegraph/src/lib.rs
+++ b/pricegraph/src/lib.rs
@@ -32,7 +32,7 @@ impl TransitiveOrderbook {
     pub fn ask_prices(&self) -> impl DoubleEndedIterator<Item = (f64, f64)> + '_ {
         self.asks
             .iter()
-            .map(|order| ((order.sell / order.buy) / FEE_FACTOR, order.buy))
+            .map(|order| ((order.buy / order.sell) * FEE_FACTOR, order.sell))
     }
 
     /// Returns an iterator with bid prices (expressed in the quote token) and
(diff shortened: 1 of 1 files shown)


pricegraph/src/lib.rs
```diff
@@ -224,6 +226,49 @@ mod tests {
         assert!(orderbook.bids.is_empty());
     }
 
+    #[test]
+    fn transitive_orderbook_simple() {
```
e00E, line 230: Let's discuss based on this test, if this is the correct result (before I changed the other test too).

I think the price (0.5 and 2.0) is correct but is the multiplication/division with the fee factor correct or inversed?

nlordell: The fees should push the two prices further apart, so they *appear* to be correct... since `0.5 / FEE_FACTOR < 0.5` and `2.0 * FEE_FACTOR > 2.0`.

I think the issue with my reasoning on the prices is that we want the price to be expressed in `quote` token and in my head, that meant that it was `base_amount / quote_amount` which is inverted, and should be `quote_amount / base_amount`, so that if we multiply some quantity of `base`, we will get the total price in `quote`. Things appear to be correct now.

nlordell (approved):
Looks good! But don't trust my review :stuck_out_tongue:

pricegraph/src/lib.rs
```diff
@@ -249,20 +293,17 @@ mod tests {
...
-            transitive_orderbook.bid_prices().collect::<Vec<_>>(),
-            vec![
-                (0.5 * FEE_FACTOR, 2_000_000.0),
-                ((1.0 / 1.8) * FEE_FACTOR, 900_000.0)
-            ]
-        );
+        let ask_prices = transitive_orderbook.ask_prices().collect::<Vec<_>>();
+        assert_approx_eq!(ask_prices[0].0, 2.0 * FEE_FACTOR);
+        assert_approx_eq!(ask_prices[0].1, 10_000_000.0);
+        assert_approx_eq!(ask_prices[1].0, 1.5 / 0.9 * FEE_FACTOR);
```
nlordell, line 299: nit: I would prefer if the order of operations were clarified to remove some ambiguity when reading the code.
```suggestion
        assert_approx_eq!(ask_prices[1].0, (1.5 / 0.9) * FEE_FACTOR);
```

pricegraph/src/lib.rs
```diff
@@ -249,20 +293,17 @@ mod tests {
...
+        assert_approx_eq!(ask_prices[1].1, 900_000.0);
+
+        let bid_prices = transitive_orderbook.bid_prices().collect::<Vec<_>>();
+        assert_approx_eq!(bid_prices[0].0, 2.0 / FEE_FACTOR);
+        assert_approx_eq!(bid_prices[0].1, 1_000_000.0);
+        assert_approx_eq!(bid_prices[1].0, 9.0 / 5.0 / FEE_FACTOR);
```
nlordell, line 305: Here as well.
```suggestion
        assert_approx_eq!(bid_prices[1].0, (9.0 / 5.0) / FEE_FACTOR);
```

nlordell (approved):

fleupold (approved):
Yes, looks correct to me. Good catch on the volume as well! Will schedule merge when green.

PR #1039, opened 2020-06-23
Turn threaded price source background thread into async task

e00E:
using async-std.
Simplified the threaded price source in several places:
* get rid of code only compiled for tests
* use weak ptr to observe dropping of parent struct

Now that we have async_std I turned futures::Mutex into async_std::Mutex
and futures_timer::sleep into async_std::sleep. If this gets merged we
can do the same changes in other places and remove the futures-timer
dependency.

### Test Plan
CI

### DIFF
--- a/core/src/price_estimation/threaded_price_source.rs
+++ b/core/src/price_estimation/threaded_price_source.rs
@@ -1,90 +1,48 @@
 use super::price_source::PriceSource;
-use crate::{models::TokenId, util::FutureWaitExt as _};
+use crate::models::TokenId;
 use anyhow::Result;
-use futures::{
-    future::{BoxFuture, FutureExt as _},
-    lock::Mutex,
+use async_std::{
+    sync::Mutex,
+    task::{self, JoinHandle},
 };
+use futures::future::{BoxFuture, FutureExt as _};
 use std::collections::HashMap;
-#[cfg(test)]
-use std::sync::mpsc::Receiver;
-use std::sync::{
-    mpsc::{self, RecvTimeoutError, Sender},
-    Arc,
-};
-use std::thread::{self, JoinHandle};
+use std::sync::Arc;
 use std::time::Duration;
 
-/// Implements `PriceSource` in a non blocking way by updating prices in a
-/// thread and reusing previous results.
+/// Implements a `PriceSource` that is always immediately ready by updating itself in a background
+/// task and using the most recent update.
 pub struct ThreadedPriceSource {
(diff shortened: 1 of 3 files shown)


core/src/price_estimation/threaded_price_source.rs
```diff
@@ -1,90 +1,48 @@
...
-                        // Make sure we don't hold the mutex while this blocking call happens.
-                        match price_source.get_prices(&tokens).wait() {
-                            Ok(prices) => price_map.lock().wait().extend(prices),
-                            Err(err) => log::warn!("price_source::get_prices failed: {}", err),
-                        }
+        let join_handle = task::spawn({
+            let price_map = Arc::downgrade(&price_map);
+            async move {
+                while let Some(price_map) = price_map.upgrade() {
+                    match price_source.get_prices(&tokens).wait() {
```
fleupold, line 37: Should this not be an await?

core/src/price_estimation/threaded_price_source.rs
```diff
@@ -1,90 +1,48 @@
...
-            move || loop {
-                match receiver.recv_timeout(update_interval) {
-                    Ok(()) | Err(RecvTimeoutError::Timeout) => {
-                        // Make sure we don't hold the mutex while this blocking call happens.
-                        match price_source.get_prices(&tokens).wait() {
-                            Ok(prices) => price_map.lock().wait().extend(prices),
-                            Err(err) => log::warn!("price_source::get_prices failed: {}", err),
-                        }
+        let join_handle = task::spawn({
+            let price_map = Arc::downgrade(&price_map);
```
fleupold, line 34: 👍Elegant!

core/src/price_estimation/threaded_price_source.rs
```diff
@@ -1,90 +1,48 @@
...
+            async move {
+                while let Some(price_map) = price_map.upgrade() {
+                    match price_source.get_prices(&tokens).wait() {
+                        Ok(prices) => price_map.lock().wait().extend(prices),
```
fleupold, line 38: Does MutexGuard automatically convert to a Promis or what 🧙is happening here? Can this also be an await?

fleupold (approved):

core/src/price_estimation/threaded_price_source.rs
e00E: Good thing you saw that. All `wait()`s should be await. This was leftover from the threaded code.

nlordell (approved):
I really like this change.

PR #1040, opened 2020-06-23
Use generic client for dex.ag price fetching

fedgiac:
Makes the current dex.ag client reusable by changing the structure of the code.

No functionalities are changed. 

### Test Plan
Test coverage should not change.

Try out ignored tests related to this PR:
```
cargo test online_dexag_client -- --ignored --nocapture
cargo test online_dexag_api -- --ignored --nocapture
cargo test online_kraken_prices -- --ignored --nocapture
cargo test online_kraken_api -- --ignored --nocapture
```


core/src/price_estimation/clients/apiclient.rs
```diff
@@ -1,79 +1,88 @@
...
 use anyhow::{anyhow, Context, Result};
-use api::{DexagApi, DexagHttpApi};
 use futures::{
     future::{self, BoxFuture, FutureExt as _},
     lock::Mutex,
 };
 use std::collections::HashMap;
 
-struct ApiTokens {
+pub trait Symbolic {
```
fleupold, line 11: Since Token is only data I wouldn't mock it but instead use the plain old data model and create dummy objects in the tests. We only really need mocks for components that contain logic.

core/src/price_estimation/clients/apiclient.rs
```diff
@@ -1,79 +1,88 @@
...
+pub trait Api {
+    type Token: Symbolic + Sync + Send;
+
+    fn get_token_list<'a>(&'a self) -> BoxFuture<'a, Result<Vec<Self::Token>>>;
+    /// Returns the price of one unit of `from` expressed in `to`.
+    /// For example `get_price("ETH", "DAI")` is ~220.
+    fn get_price<'a>(&'a self, from: &Self::Token, to: &Self::Token) -> BoxFuture<'a, Result<f64>>;
+}
+
+pub trait HttpConnecting: Sized {
```
fleupold, line 24: Can this be part of the Api trait?

core/src/price_estimation/clients/apiclient.rs
```diff
@@ -1,79 +1,88 @@
...
         const STABLE_COIN: &str = "DAI";
-        let stable_coin = tokens.remove(STABLE_COIN).ok_or_else(|| {
-            anyhow!(
-                "dexag exchange does not track our stable coin {}",
-                STABLE_COIN
-            )
-        })?;
+        let stable_coin = tokens
+            .remove(STABLE_COIN)
+            .ok_or_else(|| anyhow!("1inch exchange does not track {}", STABLE_COIN))?;
```
fleupold, line 74: "1inch" seems incorrect here.

core/src/price_estimation/clients/apiclient.rs
```diff
@@ -1,79 +1,88 @@
...
+        // OWL tracks USD so we use another stable coin as an approximate USD price.
         const STABLE_COIN: &str = "DAI";
-        let stable_coin = tokens.remove(STABLE_COIN).ok_or_else(|| {
-            anyhow!(
-                "dexag exchange does not track our stable coin {}",
-                STABLE_COIN
-            )
-        })?;
+        let stable_coin = tokens
+            .remove(STABLE_COIN)
```
fleupold, line 73: Stable coin should probably be a constructor argument so that we can use DAI for Dex.Ag and OWL for 1inch

core/src/price_estimation/clients/apiclient.rs
```diff
@@ -134,31 +143,71 @@ where
...
+        pub HttpApi {}
+        pub trait Api {
+            type Token = MockSymbolic;
+            fn get_token_list<'a>(&'a self) -> BoxFuture<'a, Result<Vec<MockSymbolic>>>;
+            fn get_price<'a>(&'a self, from: &MockSymbolic, to: &MockSymbolic) -> BoxFuture<'a, Result<f64>>;
+        }
+        pub trait HttpConnecting: Sized {
+            fn bind(http_factory: &HttpFactory) -> Result<Self>;
+        }
+    }
```
fleupold, line 189: I think we don't need all this extra setup if we use the original token data type and instantiate our dummy test objects as such structs. This should make this change less intrusive.

core/src/price_estimation/clients/apiclient.rs
```diff
@@ -1,79 +1,88 @@
...
     // This is cached in the struct because we don't expect it to change often.
-    tokens: HashMap<String, api::Token>,
-    stable_coin: api::Token,
+    tokens: HashMap<String, T::Token>,
+    stable_coin: T::Token,
 }
 
-pub struct DexagClient<Api> {
-    api: Api,
+pub struct Client<T: Api> {
```
fleupold, line 36: nit: maybe the module can be called `generic_client` (as it works with multiple APIs, whereas  kraken requires a specific client)? No strong feelings (and I'm struggling to find a good reason beyond finding the word apiclient a bit weird).

fleupold (commented):
Looks good. I would try to simplify the traits a bit (don't mock data objects) and pass in the stablecoin token as a dependecny (so that 1inch can use OWL).

bh2smith (commented):
This is really starting to look like some very nice, generic, clean and impressive rust code! Nice work here. Will come back for a proper review once some of the existing suggestions have been included!

core/src/price_estimation/clients/apiclient.rs
fedgiac: I'm not sure I understand the suggestion, maybe I don't understand the word "data model". Is it different from my manually implemented `MockSymbolic`?

core/src/price_estimation/clients/apiclient.rs
fedgiac: Agreed that `STABLE_COIN` should not be constant. It's now returned by a method of the `Api` trait, in this way it's easily customizable when creating a new Api implementation.

core/src/price_estimation/clients/apiclient.rs
fedgiac: I don't like much the idea of having the test for a generic client implementation use a token defined for a specific API implementation. On the other hand, merging `Api` and `HttpConnecting` means that there is no need for the `mock!` macro anymore, so maybe the new changes are already enough to address this comment?

core/src/price_estimation/clients/apiclient.rs
fleupold: So if a struct has an Impl with lots of logic, I think it's a very good idea to mock it. However if a struct is just about the data it carries and the unit under test doesn't call any functions on it other than accessing its members, we don't need to mock and can just use an instantiation of this "plain old data object" itself (https://en.wikipedia.org/wiki/Passive_data_structure)

core/src/price_estimation/clients/apiclient.rs
fleupold: Ah I didn't see before, but the Token struct will be different for the 1Inch API? That's interesting but I still think you don't necessarily need a trait for that and it would be much simpler for readability if we used a concrete token type in the tests instead (see comment above about passive data objects).

fedgiac: Yeah, the token struct will be different for every API implementation. That's a key change I need to make to include 1inch, and likely other APIs. The reason for `Symbolic` is that I still need a way to convert the custom token structure to a more "standardized" token structure that is the same for different API. The conversion is done by `.symbol()`, because the only thing we really need of a token to implement `PriceSource` is its symbol. (That's also nice because it allows you to handle weird conversions when implementing `Symbolic`. For example, some APIs use "ETH" to mean "WETH", so this conversion would be done cleanly in `.symbol()`.)
So, whithin these limitations the extra `Symbolic` trait seemed a good option.
For the tests, I could rename `MockSymbolic` to `MockToken` to clarify the purpose of it (i could even rename `Symbolic` to `Tokenized`, `Tokenizing` or something like that), but I don't see how I could simplify the implementation of `MockSymbolic` in any other way.

fleupold: My main point is that when you just need something that carries data, instead of going through all the code required for the trait, we could just make a struct that contains the data and pass this around being. 
In this very example you even only have one field (a string), so why not just pass symbol as a string directly?

fedgiac:
New changes based on a helpful discussion between me and Felix.
Thanks to Nick for the help with the race condition making the output of `cargo test` not deterministic.

core/src/price_estimation/clients/dexag/api.rs
```diff
@@ -89,6 +88,10 @@ impl DexagApi for DexagHttpApi {
         }
         .boxed()
     }
+
+    fn stable_coin_symbol() -> String {
```
fleupold, line 92: ubertinynanonit (feel free to ignore if you otherwise wouldn't touch this): could return &str and caller can decide to own it.

core/src/price_estimation/clients/generic_client.rs
```diff
@@ -1,79 +1,111 @@
...
+    fn get_token_list<'a>(&'a self) -> BoxFuture<'a, Result<Vec<Self::Token>>>;
+
+    /// Returns the price of one unit of `from` expressed in `to`.
+    /// For example `get_price("ETH", "DAI")` is ~220.
+    fn get_price<'a>(&'a self, from: &Self::Token, to: &Self::Token) -> BoxFuture<'a, Result<f64>>;
+
+    /// Returns a string representing the reference coin in the stead of OWL for this API
+    /// Could be different from "OWL", e.g., when the API does not offer prices with
+    /// respect to OWL
+    fn stable_coin_symbol() -> String;
```
fleupold, line 54: Nit: any reason this is not returning `Self::Token`?

core/src/price_estimation/clients/generic_client.rs
```diff
@@ -1,79 +1,111 @@
...
+    }
+}
+#[cfg(test)]
+impl From<&str> for MockGenericToken {
+    fn from(name: &str) -> Self {
+        MockGenericToken(name.to_string())
+    }
+}
+
+#[cfg_attr(test, mockall::automock(type Token=MockGenericToken;))]
```
fleupold, line 38: You should be able to mock MockGenericToken in the tests module below and reference it here with `tests::MockGenericToken`

core/src/price_estimation/clients/generic_client.rs
```diff
@@ -134,31 +166,51 @@ where
...
+    // new contexts for each test would cause race conditions on execution.
+    fn initialize_mockapi_context() {
+        static SETUP_ONCE: Once = Once::new();
+
+        SETUP_ONCE.call_once(|| {
+            let ctx = MockApi::stable_coin_symbol_context();
+            ctx.expect().returning(|| "DAI".to_string());
+            std::mem::forget(ctx);
+        });
+    }
```
fleupold, line 189: 😭 So much complication to mock a non instance method? That's a shame. I briefly looked into what would happen if we made STABLECOIN_SYMBOL an associated constant but that doesn't seem supported by mockall.

Passing the stablecoun symbol into the constructor of the GenericClient would work but separate the logic of what the stablecoin is away from the API. 

So leaving this is fine, i guess.

fleupold (approved):
Thanks for the fruitful discussion. Looks good!

core/src/price_estimation/clients/generic_client.rs
fedgiac: To make the function return `Self::Token` we would have to either hardcode all components of the token struct as the output of the function or perform external network calls, since the token cache is currently external to the API.
The second solution is inefficient, the first means that we have to add extra checks to confirm that the reference coin is indeed tracked by the exchange when calling `.stable_coin_symbol()`.
So, overall I don't think this change is worth the extra complexity. It would be a different thing however if the cache was internal to the API.

core/src/price_estimation/clients/generic_client.rs
```diff
@@ -1,79 +1,91 @@
...
+    fn get_token_list<'a>(&'a self) -> BoxFuture<'a, Result<Vec<Self::Token>>>;
+
+    /// Returns the price of one unit of `from` expressed in `to`.
+    /// For example `get_price("ETH", "DAI")` is ~220.
+    fn get_price<'a>(&'a self, from: &Self::Token, to: &Self::Token) -> BoxFuture<'a, Result<f64>>;
+
+    /// Returns a string representing the reference coin in the stead of OWL for this API
+    /// Could be different from "OWL", e.g., when the API does not offer prices with
+    /// respect to OWL
+    fn stable_coin_symbol() -> &'static str;
```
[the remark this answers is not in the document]
nlordell: One think I considered would be to change the `stable_coin_symbol` to take `&self`. So the reasons I can think of for this would be:
- Makes it so the reference token can be configurable (you could have a Dexag client with DAI as the reference and one with USDC) - this could eventually allow us to configure via command line or environment variable what we want to use for the reference token.
- Makes the mocking and testing a little easier.

Although, its 100% fine how it is.

nlordell (approved):
Sorry it took me so long to review this! It looks very nice, and very Rust-y!

I remember mentioning something about not needing a `GenericToken` trait and just using a simple data model, but now that I took some time to read though the code, I like this abstraction. It allows implementations to have API-specific cached token data, that may differ from one API to the other.

Awesome stuff!

core/src/price_estimation/clients/generic_client.rs
```diff
@@ -364,42 +366,4 @@ mod tests {
             }
         );
     }
-
-    // Run with `cargo test online_dexag_client -- --ignored --nocapture`.
-    #[test]
-    #[ignore]
-    fn online_dexag_client() {
```
bh2smith, line 371: Why was this removed? Is it no longer needed?

core/src/price_estimation/clients/generic_client.rs
```diff
@@ -265,26 +291,16 @@ mod tests {
...
-                    symbol: "DAI".to_string(),
-                    address: None,
-                },
-                super::api::Token {
-                    name: String::new(),
-                    symbol: "ETH".to_string(),
-                    address: None,
-                },
-            ];
+            static ref API_TOKENS: [MockGenericToken; 2] = ["DAI".into(), "ETH".into(),];
```
bh2smith, line 303: trailing comma?

core/src/price_estimation/clients/dexag.rs
```diff
@@ -0,0 +1,56 @@
...
+    use crate::http::HttpFactory;
+    use crate::models::TokenId;
+    use crate::price_estimation::data::TokenBaseInfo;
+    use crate::price_estimation::price_source::PriceSource;
+    use crate::util::FutureWaitExt as _;
+
+    // Run with `cargo test online_dexag_client -- --ignored --nocapture`.
+    #[test]
+    #[ignore]
+    fn online_dexag_client() {
```
bh2smith, line 20: I see, it was moved.

bh2smith (approved):
Looks really nice. It appears you have already received very well detailed comments. I have nothing important to add.

core/src/price_estimation/clients/generic_client.rs
```diff
@@ -1,79 +1,91 @@
...
-    api_tokens: Mutex<Option<ApiTokens>>,
+    api_tokens: Mutex<Option<Tokens<T>>>,
     tokens: TokenData,
 }
 
-impl DexagClient<DexagHttpApi> {
-    /// Create a DexagClient using DexagHttpApi as the api implementation.
+impl<T: Api> GenericClient<T> {
+    /// Create a GenericClient using the api implementation from HttpConnecting.
     pub fn new(http_factory: &HttpFactory, tokens: TokenData) -> Result<Self> {
```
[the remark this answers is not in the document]
bh2smith: This might be one of the best code review comments I have ever seen. I am now just a lurker/spectator tagged as a reviewer but now learning from a rust guru!

core/src/price_estimation/clients/generic_client.rs
fedgiac: I thought rustfmt would take care of it, but this is not the case!
https://github.com/rust-lang/rustfmt/issues/1034

core/src/price_estimation/clients/generic_client.rs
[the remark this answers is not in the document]
fedgiac: An issue I see with this suggestion is that it's going to move some complexity down to `price_estimation.rs`.

Currently the API is already part of the client, and we only need to import:
```
use self::clients::{DexagClient, KrakenClient};
```

With the proposed changes we'd need to import all "genericized" APIs in `price_estimation.rs` and use each of them to "specificize" a generic client (which must then receive a different treatment compared to an ad hoc clients). With the current state of the PR, this complexity is hidden in the type creation in `dexag.rs` and there is no difference between ad hoc clients and clients built from `GenericClient` in `price_estimation.rs`.

Also, the only user of the clients discussed in this PR is `PriceOracle`, which is intrinsically bound to HTTP since its constructor requires an HTTP factory. So, if we need to make this PR more general it might be enough to replace every occurrence of `HttpFactory` with `impl Connector` for some custom `Connector` trait used to generalize `HttpFactory` that is specified somewhere else in the code. In general, I think if we'll need to use something that is not HTTP in the future, then the code surrounding this PR might end up looking quite different anyway. 

In conclusion, I'd not change the current code until we actually need to use an API that is not served through HTTP.

core/src/price_estimation/clients/generic_client.rs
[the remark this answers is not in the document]
fedgiac: Taking `&self` is reasonable, but the changes needed to implement this proposal would make a similar issue to the one explained in [my other comment](https://github.com/gnosis/dex-services/pull/1040/files#r448462317) arise: The only place where something that is not an abstract type is created is in `price_estimation.rs`, so the call to the constructor would have to look like `DexagClient::new(http_factory, tokens.clone(), &"DAI")`. I don't quite like that because it's not immediate what the role of this "DAI" is.
I see however that if the changes from the linked comment were somehow introduced then this change would definitely be easy to do.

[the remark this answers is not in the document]
nlordell: We can implement this if it is needed in the future.

PR #1043, opened 2020-06-24
Implement estimated sell amount route

nlordell:
This PR implements an estimated sell amount route for computing the and order using maximum sell amount from a given exchange rate.

This will be useful for various GP services to know about how much of a token they can sell in the next batch at an exchange rate.

If you guys think its worth it, I can split the pricegraph and price-estimator changes into two separate PRs.

### Test Plan

Added unit test to verify route parameters are correctly parsed, also, check run the price estimator and check the route:
```
$ curl -s http://localhost:8080/api/v1/markets/1-7/estimated-sell-amount/245.5\?atoms\=true | jq
{
  "baseTokenId": "1",
  "quoteTokenId": "7",
  "buyAmountInBase": "64016789977586752",
  "sellAmountInQuote": "15731853793290840064"
}
```

### DIFF
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -53,3 +53,31 @@ Example Response:
 
 * `buyAmountInBase` estimates the buy amount (in base tokens) a user can set as a limit order while still expecting to be completely matched when selling the given amount of quote token.
 * The other fields repeat the parameters in the url back.
+
+### Estimated Sell Amount
+
+`GET /api/v1/markets/:market/estimated-sell-amount/:price-in-quote`
+
+* `market` is as above
+* `price_in_quote` is a decimal number.
+
+Url query is as above.
+
+Example Request: `/api/v1/markets/1-7/estimated-sell-amount/245.5?atoms=true`
+
+Example Response:
+
+```json
+{
+    "baseTokenId": "1",
+    "quoteTokenId": "7",
+    "buyAmountInBase": "1103276584712580736",
+    "sellAmountInQuote": "271125527074012594176"
+}
+```
+
+The following result indicates that if we wanted to buy ETH (token 1) for DAI (token 7) and pay 245.5 DAI per unit of ETH, we would be able to sell an estimated maximum 271.13 DAI.
+
+* `sellAmountInBase` estimates the sell amount (in quote tokens) a user can completely fill in the following batch at the specified `price_in_quote`.
+* `buyAmountInBase` is the computed buy amount (in base tokens) for the order from the specified price and estimated sell amount.
+* The other fields repeat the parameters in the url back.
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -56,14 +56,14 @@ Example Response:
 
 ### Estimated Sell Amount
 
-`GET /api/v1/markets/:market/estimated-sell-amount/:price-in-quote`
+`GET /api/v1/markets/:market/estimated-amounts-at-price/:price-in-quote`
 
 * `market` is as above
 * `price_in_quote` is a decimal number.
 
 Url query is as above.
 
-Example Request: `/api/v1/markets/1-7/estimated-sell-amount/245.5?atoms=true`
+Example Request: `/api/v1/markets/1-7/estimated-amounts-at-price/245.5?atoms=true`
 
 Example Response:
 
@@ -71,8 +71,8 @@ Example Response:
 {
     "baseTokenId": "1",
     "quoteTokenId": "7",
-    "buyAmountInBase": "1103276584712580736",
-    "sellAmountInQuote": "271125527074012594176"
+    "buyAmountInBase": "6098377078823660544",
+    "sellAmountInQuote": "1497151572851208749056"
 }
 ```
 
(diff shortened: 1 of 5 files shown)


pricegraph/src/lib.rs
```diff
@@ -158,6 +158,33 @@ impl Pricegraph {
...
     }
 
+    /// Estimates a sell amount for the specified token pair and limit exchange
+    /// rate.
+    pub fn estimate_sell_amount(&self, pair: TokenPair, exchange_rate: f64) -> f64 {
+        self.reduced_orderbook()
+            .fill_order_at_price(pair, exchange_rate)
+    }
+
+    /// Returns a transitive order with a buy and sell amount computed such that
```
e00E, line 168: probably: "with the **largest** buy and sell amount"

e00E (approved):
I don't mind the single PR. Still need to update readme but we can do this is another PR.

nlordell:
> I don't mind the single PR. Still need to update readme but we can do this is another PR.

Already did :see_no_evil:

price-estimator/README.md
```diff
@@ -53,3 +53,31 @@ Example Response:
...
+### Estimated Sell Amount
+
+`GET /api/v1/markets/:market/estimated-sell-amount/:price-in-quote`
+
+* `market` is as above
+* `price_in_quote` is a decimal number.
+
+Url query is as above.
+
+Example Request: `/api/v1/markets/1-7/estimated-sell-amount/245.5?atoms=true`
```
fleupold, line 66: Given you are returning both buy and sell amount maybe we should call this route amount-at-price?

price-estimator/src/filter.rs
```diff
@@ -52,6 +71,13 @@ fn markets_filter() -> impl Filter<Extract = (TokenPair, QueryParameters), Error
         .and(warp::query::<QueryParameters>())
 }
 
+fn estimated_sell_amount_filter(
+) -> impl Filter<Extract = (TokenPair, f64, QueryParameters), Error = Rejection> + Copy {
+    warp::path!("api" / "v1" / "markets" / TokenPair / "estimated-sell-amount" / f64)
```
fleupold, line 76: nit: Can we use [path prefixes](https://docs.rs/warp/0.2.3/warp/macro.path.html#path-prefixes) for the common parts?

price-estimator/src/filter.rs
```diff
@@ -91,6 +115,41 @@ async fn get_markets<T>(
...
+    let exchange_rate = 1.0 / price_in_quote;
+    let transitive_order = orderbook
+        .get_pricegraph()
+        .await
+        .order_for_exchange_rate(token_pair.into(), exchange_rate);
+    let (buy_amount_in_base, sell_amount_in_quote) = transitive_order
+        .map(|order| {
+            (
+                apply_rounding_buffer(order.buy, price_rounding_buffer),
+                apply_rounding_buffer(order.sell, price_rounding_buffer),
```
fleupold, line 136: I don't fully understand this logic. We are applying the rounding buffer to both buy and sell amounts (why not just one?). Also, I think the way this is implemented now it will make both the buy and sell amounts smaller, which wouldn't really create a buffer, would it?

nlordell: So, my understanding was that the price rounding buffer was to make the exchange rate slightly more pessimistic to match the solver. Initially I had (incorrectly) thought that we just wanted to decrease the sell amount so that it is more pessimistic as well (and also decrease the buy amount by the same ratio so that the resulting order has exactly the quote price specified by the call).

Now that I think about it, maybe this is incorrect and we should be applying it to the input price so that the exchange rate is more pessimistic and restrictive (making it match potentially fewer orders). Then we can also apply this to the buy amount (as you suggested) which would make the input price exactly match the order exchange rate.

price-estimator/src/filter.rs
nlordell: Addressed in #1045

price-estimator/src/filter.rs
fleupold: That argument makes sense. Do I understand correctly that the ratio of buy/sell amount still doesn't always match the requested price? E.g. if I want to know how much DAI I can sell at a price of $1 and there is only one order in the book buying 100DAI at 2$, I would get sellAmount:100DAI and buyAmount:200$, right?

However, by rounding the price before computing the sell amount we ensure that the ratio is always strictly better than the requested price. If we only rounded the buy amount down, the response ratio could be slightly worse than the requested one.

fleupold (approved):

price-estimator/src/filter.rs
nlordell: > Do I understand correctly that the ratio of buy/sell amount still doesn't always match the requested price?

This is true for the estimate, however, the route is implemented to return 100:100 though. This is something I wasn't sure about how we should implement it. Basically, from an exchange rate we can estimate two orders:

1. Best possible order where the buy and sell amounts are as large as possible such that `buy_amount / sell_amount >= xrate`.
2. Order at exchange rate where the sell amount is as large as possible, and `buy_amount / sell_amount == xrate`.

I opted for 2, since it more accurately expresses the willingness to pay, but this can be easily changed.

> If we only rounded the buy amount down, the response ratio could be slightly worse than the requested one.

Correct.

PR #1049, opened 2020-06-25
Merge transitive orders of same price

### LINKED ISSUE #1046: Orderbook route response should only have one element per price point
The orderbook should aggregate offers at the same price level. At the moment we can e.g. query https://dex-price-estimator.gnosis.io/api/v1/markets/1-0?atoms=true& and see the following two elements in the bids:

> {
      "price": 199.79999999999998,
      "volume": 50000000000000000000
    },
    {
      "price": 199.79999999999998,
      "volume": 166022205859424830
    },

Since the have the exact same string representation for price they should be aggregated into the sum of their volumes. This is causing a weird effect in the UI:

orderbook-WETH-OWL

e00E:
Having several orders at the same price caused ui issues in the
frontend. To fix this we merge orders of the same price by adding their
volumes.
Conceptually it is not clear whether merged or unmerged is more
correct as merging loses the information that the same price was reached
through different paths through the order graph (i think).
For this reason I implemented this in price-estimator and not
pricegraph.

Fixes #1046 

### Test Plan
CI and manually compared the query listed in the issue.

### DIFF
--- a/price-estimator/src/models.rs
+++ b/price-estimator/src/models.rs
@@ -1,6 +1,6 @@
 use serde::{Deserialize, Serialize};
 use serde_with::rust::display_fromstr;
-use std::{convert::From, num::ParseIntError, str::FromStr};
+use std::{cmp::Ordering, convert::From, num::ParseIntError, str::FromStr};
 
 #[derive(Debug, Copy, Clone)]
 pub struct TokenPair {
@@ -67,7 +67,7 @@ pub struct EstimatedBuyAmountResult {
     pub sell_amount_in_quote: u128,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, PartialEq, Serialize)]
 pub struct TransitiveOrder {
     pub price: f64,
     pub volume: f64,
@@ -79,11 +79,25 @@ pub struct MarketsResult {
     pub bids: Vec<TransitiveOrder>,
 }
 
+/// Sort orders by their price and merge orders with the same price into one.
+fn sanitize_orders(mut orders: Vec<TransitiveOrder>) -> Vec<TransitiveOrder> {
+    orders.sort_unstable_by(|a, b| a.price.partial_cmp(&b.price).unwrap_or(Ordering::Less));
+    let mut result = Vec::<TransitiveOrder>::with_capacity(orders.len());
+    for order in orders.into_iter().filter(|order| !order.price.is_nan()) {
+        match result.last_mut() {
+            #[allow(clippy::float_cmp)]
+            Some(last) if last.price == order.price => last.volume += order.volume,
+            _ => result.push(order),
+        }
+    }
+    result
+}
+
 impl From<&pricegraph::TransitiveOrderbook> for MarketsResult {
     fn from(orderbook: &pricegraph::TransitiveOrderbook) -> Self {
         let to_order = |(price, volume)| TransitiveOrder { price, volume };
-        let asks = orderbook.ask_prices().map(to_order).collect();
-        let bids = orderbook.bid_prices().map(to_order).collect();
+        let asks = sanitize_orders(orderbook.ask_prices().map(to_order).collect());
+        let bids = sanitize_orders(orderbook.bid_prices().map(to_order).collect());
         Self { asks, bids }
     }
 }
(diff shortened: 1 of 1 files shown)


price-estimator/src/models.rs
```diff
@@ -132,4 +146,53 @@ mod tests {
         });
         assert_eq!(json, expected);
     }
+
+    #[test]
+    fn sanitize_orders_works() {
```
fleupold, line 151: nit: I find testing that it "works" a bit lax (requires implicit knowledge of how this method is supposed to work or jumping to the doc string of the method which can be outdated). Can we call it `sanitize_orders_sums_volume_at_same_price`

price-estimator/src/models.rs
```diff
@@ -79,11 +79,25 @@ pub struct MarketsResult {
...
 }
 
+/// Sort orders by their price and merge orders with the same price into one.
+fn sanitize_orders(mut orders: Vec<TransitiveOrder>) -> Vec<TransitiveOrder> {
+    orders.sort_unstable_by(|a, b| a.price.partial_cmp(&b.price).expect("price isn nan"));
+    let mut result = Vec::<TransitiveOrder>::with_capacity(orders.len());
+    for order in orders {
+        match result.last_mut() {
+            #[allow(clippy::float_cmp)]
+            Some(last) if last.price == order.price => last.volume += order.volume,
```
fleupold, line 89: I couldn't find any explicit documentation, but I assume `display_fromstr` is precise enough to give every distinct float a distinct string (as in there are not two floats evaluating to 199.79999999999998). Otherwise we would need to compare the string representations here.

fleupold (approved):
Thanks for addressing this so quickly!

> Conceptually it is not clear whether merged or unmerged is more correct as merging loses the information that the same price was reached through different paths through the order graph (i think).

This is right. In regular orderbooks you usually only see one volume per price point (even though there could be multiple orders at that price point). I see a use case for getting the different paths when estimating a price (to see which orders the estimate relied on and if they were all still in the system at solving time), but for the orderbook it's really just about how much volume is there at what price.

I therefore think it could well be implemented in pricegraph directly (it would also not require sorting again).

price-estimator/src/models.rs
```diff
@@ -79,11 +79,25 @@ pub struct MarketsResult {
     pub bids: Vec<TransitiveOrder>,
 }
 
+/// Sort orders by their price and merge orders with the same price into one.
+fn sanitize_orders(mut orders: Vec<TransitiveOrder>) -> Vec<TransitiveOrder> {
```
nlordell, line 83: nit: I think this should be part of the pricegraph API, specifically `ask_prices` and `bid_prices`.

nlordell (approved):

price-estimator/src/models.rs
nlordell: `serde` uses https://docs.rs/ryu/1.0.3/ryu/ for float to string conversion, just FYI.

price-estimator/src/models.rs
nlordell, line 83: nit: I think this should be part of the pricegraph API, specifically ask_prices and bid_prices.

price-estimator/src/models.rs
e00E: I thought about this too but I came to the conclusion that this is what you would expect from number formatting. I didn't check then but through ryu I find:

> Ryu generates the shortest decimal representation of a floating point number that maintains round-trip safety. That is, a correct parser can recover the exact original number.

price-estimator/src/models.rs
```diff
@@ -79,11 +79,25 @@ pub struct MarketsResult {
...
+    orders.sort_unstable_by(|a, b| a.price.partial_cmp(&b.price).expect("price isn nan"));
```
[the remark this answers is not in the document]
e00E: I do think it's reasonable that pricegraph cannot output NaN but it's true that in this case we don't need that assumption.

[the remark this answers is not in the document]
nlordell: If pricegraph returns a NaN price, this is a bug. We don't necessarily need to have that assumption here though (as you both suggested) but a log message would be helpful if a NaN os encountered so we can debug the issue.

Maybe pricegraph should use NonNaNF64 type or something...

price-estimator/src/models.rs
e00E: > I therefore think it could well be implemented in pricegraph directly (it would also not require sorting again).

pricegraph can be used by different projects. We shouldn't sort in pricegraph at all because different consumers might want different orderings. For example currently pricegraph orders both asks and bids by exchange by exchange_rate which is not the same thing as bid_prices. Someone else might want to order by volume etc.

nlordell: > We shouldn't sort in pricegraph at all because different consumers might want different orderings

I agree... I think pricegraph should just return the ask_prices and bid_prices in exchange rate order (which is how they are stored in the vec). My point was that I think this should be implemented as part of the ask_prices and bid_prices methods.

e00E: The way i imagine this working out when implemented directly in `ask_prices` is more complicated if we don't want to collect into a vec internally. This is because we can't easily implement this with the iterator combinators so we we would have to create a custom struct implementing `DoubleEndedIterator`.
The need for merging the asks and bids is specific enough to the price estimator that I prefer to keep it there.

nlordell: > The need for merging the asks and bids is specific enough to the price estimator

I agree this is true for the orders, but not for `ask_prices` and `bid_prices`.

nlordell (approved):
I think we can merge this as-is, especially since it would be nice to deploy a fix to prod on Monday for the UI glitch that we are currently seeing.

e00E:
I agree. Will merge and we can move the code to pricegraph later.

Issue #970, opened 2020-06-09
[Price Estimation] Make price estimation rounding tighter


fleupold:
Original issue: https://github.com/gnosis/dex-price-estimator/issues/52

Looking into other stablecoin swaps I noticed that spreads are extremely competitive. E.g., at the moment our price estimator offers 0.99690 USDC/DAI while [mStable](https://app.mstable.org/swap) offers marginally better 0.997.

However when actually doing a trade we end up achieving a better price of 0.9979.
While this is a great "surprise" for the user it would be nice if we could make our price estimation "tighter", as people might not even attempt to trade otherwise. At the moment we just add another 0.1% rounding buffer on top.

fleupold:
Summary of the discussion with @marcovc and @twalth3r to explain the current solver rounding:

- Current solver rounding code can be found here: https://github.com/gnosis/dex-solver/blob/master/src/opt/process/Rounding.py#L67-L90
- It is using the price estimate given to the solver (average prices from dex.ag/kraken) and multiplying this with a constant.
- This makes it so that tokens with less decimals (e.g. GUSD or USDC) receive a higher rounding buffer (price of to GUSD in OWL is in the order of 10**34, whereas the price of DAI in OWL is in the order of 10**18)
- The price is not depending on the amount sold (thus rounding buffer for selling 10 DAI is as big as for 1000 DAI, whereas in our current approach we use a % of the amount).

For now, the goal would be to imitate this rounding behaviour for the price estimate.

We can use the price estimated by the price graph to approximate the rounding buffer in a similar way, however there are two issues with that:
1. Prices for esoteric tokens may have a significantly different price on GP compared to Kraken/1Inch (our current price sources)
2. The price at the time of estimation might not be the same as the price at the time of solver invocation.

Point 2.) can be addressed by multiplying the price itself with a reasonable buffer (expected maximal swing in the next n minutes)
Point 1.) could be addresses by using the same averaged price source (querying kraken and dex.ag) in the price estimator when computing the rounding buffer.

fleupold:
Pulling in the external price sources has the added disadvantage that it makes replaying price quotes in the past harder (price estimation is no longer a pure function given the orderbook) - cf. #967.
However this is likely going to become more challenging anyways as we start adding "virtual orders" (thus any solution applied there could also be applied to the external price estimations)

twalth3r:
[...]

PR #1051, opened 2020-06-29
Calculate rounding buffer like solver

e00E:
Currently we are applying a rounding buffer as a fixed factor but this
is not how the solver works and thus ends up usually too conservative. By
calculating the rounding buffer like the solver does we can get a more
realistic result.
This commit adds the rounding buffer calculation but does not yet apply
it or actually retrieve prices.

Related to https://github.com/gnosis/dex-services/issues/970 .

### Test Plan
This change is not complete so it can't be tested fully yet. I have added a unit test that I checked with the real solver.

### DIFF
--- a/price-estimator/src/solver_rounding_buffer.rs
+++ b/price-estimator/src/solver_rounding_buffer.rs
@@ -1,0 +1,167 @@
+use core::models::{AccountState, Order, TokenId, TokenInfo};
+use ethcontract::{Address, U256};
+use std::collections::{BTreeMap, HashMap};
+
+// This code is closely related to dex-solver/src/opt/process/Rounding.py .
+// Discussion of motivation happened in https://github.com/gnosis/dex-services/issues/970 .
+
+type Tokens = BTreeMap<TokenId, TokenInfo>;
+
+const MAX_ROUNDING_VOLUME: f64 = 100_000_000_000.0;
+const PRICE_ESTIMATION_ERROR: f64 = 10.0;
+
+fn max_rounding_amount(token_price: f64, fee_token_price: f64) -> f64 {
+    let estimated_price_in_fee_token = token_price / fee_token_price;
+    let max_rounding_amount = MAX_ROUNDING_VOLUME / estimated_price_in_fee_token;
+    max_rounding_amount.max(1.0)
+}
+
+/// Perform the same rounding buffer calculation as our solvers in order to increase the correctness
+/// of our estimates.
+/// All token prices must be nonzero.
+#[allow(dead_code)]
+pub fn apply_rounding_buffer(
+    tokens: &Tokens,
+    orders: &mut Vec<Order>,
+    account_state: &mut AccountState,
+    pessimistic_factor: f64,
+) {
+    let fee_token_price = tokens
+        .get(&TokenId(0))
+        .expect("fee token does not have price")
+        .external_price as f64;
+    assert!(fee_token_price > 0.0, "fee token price is 0");
+
+    // The maximum rounding buffer over all orders from this address selling this token.
+    let mut account_balance_buffers = HashMap::<(Address, TokenId), u128>::new();
+    // Apply rounding buffer to account balances and order sell amounts.
+    for order in orders.iter_mut() {
+        // Compute rounding buffer.
+        // If a token doesn't have a price it means it isn't connected to the fee. In this case we
+        // use a price of 1 to get some reasonable default rounding buffer.
+        let (sell_token, buy_token) = (TokenId(order.sell_token), TokenId(order.buy_token));
+        let estimated_buy_token_price = match tokens.get(&buy_token) {
+            Some(TokenInfo { external_price, .. }) if *external_price > 0 => *external_price as f64,
+            _ => 1.0,
+        };
+        let estimated_sell_token_price = match tokens.get(&sell_token) {
+            Some(TokenInfo { external_price, .. }) if *external_price > 0 => *external_price as f64,
+            _ => 1.0,
+        };
+        let estimated_xrate = estimated_buy_token_price / estimated_sell_token_price;
+        let rounding_buffer = max_rounding_amount(estimated_buy_token_price, fee_token_price)
+            * estimated_xrate
+            * PRICE_ESTIMATION_ERROR.powi(2);
+        // Multiply by an extra factor which the solver doesn't do, as added safety in case the
+        // prices move.
+        let rounding_buffer = (pessimistic_factor * rounding_buffer) as u128;
+
+        // Update rounding buffer for account balances;
+        let entry = account_balance_buffers
+            .entry((order.account_id, sell_token))
+            .or_default();
+        *entry = (*entry).max(rounding_buffer);
+
+        // Reduce order sell amount.
+        order.denominator = order.denominator.saturating_sub(rounding_buffer);
+        order.remaining_sell_amount = order.remaining_sell_amount.saturating_sub(rounding_buffer);
+    }
+
+    // Reduce account balances.
+    for ((address, token_id), rounding_buffer) in account_balance_buffers {
+        if let Some(balance) = account_state.0.get_mut(&(address, token_id.0)) {
+            *balance = balance.saturating_sub(U256::from(rounding_buffer));
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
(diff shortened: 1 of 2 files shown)


price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    // The maximum rounding buffer over all orders from this address selling this token.
+    let mut account_balance_buffers = HashMap::<(Address, TokenId), u128>::new();
+    // Apply rounding buffer to account balances and order sell amounts.
+    for o in orders.iter_mut() {
+        // Compute rounding buffer.
+        let (t_s, t_b) = (TokenId(o.sell_token), TokenId(o.buy_token));
+        let estimated_buy_token_price = match tokens.get(&t_b) {
+            Some(TokenInfo { external_price, .. }) if *external_price > 0 => *external_price as f64,
+            _ => continue,
```
nlordell, line 42: Does the solver just not add a rounding buffer for tokens without estimates? I imagine it still needs some rounding buffer for these tokens.

nlordell (approved):
Very nice!

price-estimator/src/solver_rounding_buffer.rs
e00E: Ah I meant to have a comment here about how we should handle missing prices. The solver calculates the prices in a more complicated way through an optimization model that the estimated prices are part of so it's not clear to me what we should do because we can't implement that optimization model.
I'll ask in the issue about it.

price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -0,0 +1,162 @@
...
+        .expect("fee token does not have price")
+        .external_price as f64;
+    assert!(fee_token_price > 0.0, "fee token price is 0");
+
+    // The maximum rounding buffer over all orders from this address selling this token.
+    let mut account_balance_buffers = HashMap::<(Address, TokenId), u128>::new();
+    // Apply rounding buffer to account balances and order sell amounts.
+    for o in orders.iter_mut() {
+        // Compute rounding buffer.
+        let (t_s, t_b) = (TokenId(o.sell_token), TokenId(o.buy_token));
```
fleupold, line 39: ubernit: can we please spell out the variable names? I know the python codebase uses abbreviations to make it more look like a mathematical equation/proof but I think it's less cognitive effort to understand this code if I don't have to interpret a one character variable name.

price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    #[test]
+    fn apply_rounding_buffer_ok() {
+        let mut tokens = Tokens::new();
+        tokens.insert(TokenId(0), token_info(1));
+        tokens.insert(TokenId(1), token_info(2));
+        tokens.insert(TokenId(2), token_info(10));
+
+        let accounts = vec![
+            account(0, 0, 100_000_000_000_000_000),
```
fleupold, line 123: I assume you don't think the test macros are worth the effort here? They would allow reading the test setup in a non IDE without jumping back and forth to the account/order function to remember in which order the 3/6 numbers are passed into the named struct.

price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -0,0 +1,162 @@
...
+            order(0, 0, 1, 0, 600_000_000_000_000, 490_000_000_000_000),
+            order(1, 0, 0, 1, 600_000_000_000_000, 495_000_000_000_000),
+            order(2, 0, 2, 0, 600_000_000_000_000, 490_000_000_000_000),
+            order(3, 0, 0, 2, 600_000_000_000_000, 499_000_000_000_000),
+            order(4, 0, 2, 1, 600_000_000_000_000, 495_000_000_000_000),
+            order(5, 0, 1, 2, 600_000_000_000_000, 499_000_000_000_000),
+        ];
+        for (order, expected) in orders.iter().zip(expected_orders.iter()) {
+            assert_eq!(order, expected);
+        }
```
fleupold, line 150: Does equality on the vec not hold?

price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -0,0 +1,162 @@
...
+            numerator,
+            denominator,
+            remaining_sell_amount: denominator,
+            valid_from: 0,
+            valid_until: 0,
+        }
+    }
+
+    #[test]
+    fn apply_rounding_buffer_ok() {
```
fleupold, line 116: Is this test taken from the python codebase?

While it is nice, and reassures that the logic is correct, it doesn't really document the expected behaviour (e.g. that we take the max rounding per token to compute the rounding on the balance, that rounding depends on prices but not on amounts, etc.). Maybe we could add a few more documentary tests given that this is a quite custom and 1-off algortithm.

price-estimator/src/solver_rounding_buffer.rs
fleupold: We should have price estimates for all tokens that are already connected to the fee token. 
I'm not entirely sure what happens if we have an order that is not yet connected to the fee token, but the price estimator is using it to suggest an order that will make it connected. But then this is such an exotic case that we can probably ignore it and not apply any rounding in that case.

fleupold (approved):
I do share your concerns to maintain this logic in two different places. Maybe we can later export this code with a python binding or even move the pre/post-processing stages from the solver to the driver?

price-estimator/src/solver_rounding_buffer.rs
e00E: The only reason for these names is to keep them the same as in the python code but I don't mind either way. Definitely wouldn't have used them if this wasn't moved over from there. I'll change it.

price-estimator/src/solver_rounding_buffer.rs
e00E: Equality is the same but this gives better output in case it doesn't match because the output is reduced to the one failing element. It was useful when I wrote the test so I kept it.

price-estimator/src/solver_rounding_buffer.rs
e00E: I don't think it makes sense to test details of the algorithm. After all it has simply been copied from python and we will always keep it up to date with the python version. It's not the job of this code to verify that the python algorithm is sensible. Unless you think it would help catch bugs introduced by converting to rust.

I wrote this test by running the python function manually with the same inputs to get the expected result.

price-estimator/src/solver_rounding_buffer.rs
e00E: For readability I want these account and order definitions to fit in one line (or maybe 2). With the full structs it becomes too large. In this case since the functions are right above the test and there are only two functions I think it's fine and when reading you'll remember the parameters after looking them up once.
If it was more functions and more tests maybe macro would be worth it but not at this point.

e00E:
> Maybe we can later export this code with a python binding or even move the pre/post-processing stages from the solver to the driver?

We could calculate and pass the rounding buffers directly. I worry that this takes too much control away from the solver and that it couples them too tightly. This is very specific to this kind of solver so I don't want to do the calculation for it.

On the other hand writing the logic once and sharing it between the solver and price-estimator (not driver) could be worth it. Still it would make it harder for the solver team to tweak this because they probably prefer writing Python and not Rust (long term idea: rewrite solver in Rust :) ). If we can use a python module from Rust then we would still have to refactor that python module but it probably wouldn't put more strain on them. Still I don't think that this is worth it for this one function.

bh2smith (approved):
I bet you had a great time digging through the solver code! Could we please use `order` instead of `o` here?

fleupold:
@e00E is there something blocking this merge? I would suggest we take 2x the exact rounding buffer to begin with (so that we have a little bit of extra buffer in case price estimates move slightly). @nlordell do we already have a way of measuring the impact of that change on our top level accuracy?

e00E:
It's not blocked but further steps depend on #1055 (whether we want to use external prices) but sure I'll merge this since this isn't used by the rest of the code yet anyway. That will be next PR.

nlordell:
> do we already have a way of measuring the impact of that change on our top level accuracy?

No, the accuracy is measured directly with pricegraph estimates without any buffers applied. The current e2e script would have to be augmented to also include the price buffer.

PR #1052, opened 2020-06-29
Handle node going back in time more gracefully

### LINKED ISSUE #993: Event based orderbook block interval weirdness
Today we saw errors like

> 2020-06-12T08:37:59.718Z ERRO [core::driver::scheduler::system] Batch 5306502 failed with retryable error: web3 error: RPC error: Error { code: ServerError(-32000), message: "One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found", data: Some(String("0x9c65e6")) }
 
coupled with this line right before

> [core::orderbook::streamed::updating_orderbook] Updating event based orderbook with from block 10249702 to block 10249684.

0x9c65e6 is the from block 10249702 in hex.

This is weird because in this case the from block was larger than the to block which is likely what caused the error. The relevant code is `UpdatingOrderb::update_with_events`.
We calculate `from_block` as `last_handled_block - 25`  and use the current block as reported by the node as `to_block`. Finally, we set `last_handled_block` to `to_block`.
This means that what the node reported as the current block went backwards by more than 25 blocks.
Looking at the logs from at least (didn't scroll back more) 2020-06-12T08:31:06.887Z to 2020-06-12T09:02:59.818Z the node claimed the current block was 10249684. With a short time in between the current block went up again to 10249874 for 3 batches and then back down to 10249684.
This is erratic and wrong behavior by the node. Still, we could at least detect this in updating orderbook and handle it somehow. For example, we could decrease the from block even more even if this goes earlier than the hardcoded 25 limit that currently exists.

e00E:
by printing an error instead of calling past_events with a nonsensical
range.

Fixes #993 

### Test Plan
Run driver and see that event based orderbook still updates.


nlordell (approved):
Nice!

bh2smith (approved):

Issue #992, opened 2020-06-12
Deploy in release mode


e00E:
We have been building the driver binary in debug mode for the deployed docker images. This was never a problem because there was no cpu intensive work.

Recently we got the event based orderbook, which in it's current version re-creates the whole orderbook from scratch every batch based on all past events. We're also going to start using the pricegraph to estimate prices in #990 .
Both of these operations are cpu intensive and while I don't think the time they take is a problem yet we might want do be future proof by building in release mode which should make them significantly faster.

advantages:
* everything is faster and more efficient in terms of cpu usage

disadvantages:
* deploy step takes longer because compiling in release mode takes longer

Another difference is that we lose implicit integer overflow checks that are only enabled in debug mode. This hasn't ever happened and we shouldn't be relying on implicit checks like this in the first place (and as far as I'm aware, we don't).

The change should be very simple: add `--release` to `cargo build --locked -p driver` in travis but I'm looking for other opinions on this issue.

nlordell:
I think you can configure release mode to build with overflow checks.

fleupold:
Release build would give us less debug information in cases of crashes, right? I don't think we are crashing a lot, so this might not be an issue. We could also consider building our tags with release and have latest in debug mode.

nlordell:
> Release build would give us less debug information in cases of crashes, right?

Just checked and it looks like we get full back traces in release mode:
```
   Compiling playground v0.0.1 (/playground)
    Finished release [optimized] target(s) in 1.13s
     Running `target/release/playground`
thread 'main' panicked at 'test', src/main.rs:2:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
...
```

From the reference https://doc.rust-lang.org/cargo/reference/profiles.html it looks like release turns on more optimizations, turns off incremental building and sets opt-level to 3. These make the compilation take longer but in theory help produce faster code (:+1:)

[...]

PR #1054, opened 2020-06-29
Deploy in release mode

e00E:
See discussion in #992 .

The PR itself is trivial so let's keep the discussion in the issue to not split the conversation.

### Test Plan
I have tested that the driver still runs locally when built in release mode. The real test will be actually merging this PR and looking at staging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,9 +89,9 @@ jobs:
         - sudo apt-get update && sudo apt-get install -y python3-pip python3-setuptools && pip3 install --upgrade --user awscli
         - $(aws ecr get-login --no-include-email --region $AWS_REGION)
       script:
-        - cargo build --locked -p driver
-        - docker build --tag stablex-binary-public --build-arg SOLVER_BASE=gnosispm/dex-open-solver:$OPEN_SOLVER_VERSION --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
-        - docker build --tag stablex-binary-private --build-arg SOLVER_BASE=163030813197.dkr.ecr.eu-central-1.amazonaws.com/dex-solver:$PRIVATE_SOLVER_VERSION --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
+        - cargo build --locked -p driver --release
+        - docker build --tag stablex-binary-public --build-arg SOLVER_BASE=gnosispm/dex-open-solver:$OPEN_SOLVER_VERSION --build-arg RUST_BASE=rust-binary --build-arg BINARY_PATH=target/release/driver -f docker/rust/Dockerfile .
+        - docker build --tag stablex-binary-private --build-arg SOLVER_BASE=163030813197.dkr.ecr.eu-central-1.amazonaws.com/dex-solver:$PRIVATE_SOLVER_VERSION --build-arg RUST_BASE=rust-binary --build-arg BINARY_PATH=target/release/driver -f docker/rust/Dockerfile .
       deploy:
         - provider: script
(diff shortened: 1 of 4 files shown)


nlordell (approved):
:tada: Love it!

Note that we still have `opt-level = "s"` in the workspace Cargo.toml file. I think its fine (and might even produce smaller binaries, which is nice for the Docker image). On the other hand, one argument I see against leaving `opt-level = "s"` is that it is a bit premature to tweak compiler optimization flags without benchmarking the impact.

Either way, I don't think it matters much for the driver, just wanted to point it out and start a discussion.

e00E:
Removed the opt-level line as again this is the most sensible default to me. If we find a good argument for adding it back we can do that later but I don't think we should use it.

Cargo.toml
```diff
@@ -15,11 +15,7 @@ default-members = [
...
 ]
 
-[profile.release]
-opt-level = "s"
-
 [patch.crates-io]
 # Needed to make coverage work on MacOS (https://github.com/servo/core-foundation-rs/pull/357)
 core-foundation = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
 core-foundation-sys = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
-
```
e00E, line 25: There is still an empty line here btw (there were two).

nlordell: Hahaha, don't worry, there is a prominent :no_entry_sign: icon when there is a missing newline at the end of file. There is none here :smile:

nlordell:
> f we find a good argument for adding it back we can do that later but I don't think we should use it.

I think that's sensible. I don't think just blindly doing it without comparing resulting binary sizes is worth it.

docker/rust/Dockerfile
```diff
@@ -34,7 +36,7 @@ ONBUILD RUN apt-get update \
   && apt-get install -y --no-install-recommends libpq-dev libssl1.0.0 libssl-dev ca-certificates \
   && rm -rf /var/lib/apt/lists/*
 
-ONBUILD COPY target/debug/driver stablex
+ONBUILD COPY ${RUST_BASE} stablex
```
fleupold, line 39: should this be BINARY_PATH?

.travis.yml
```diff
@@ -89,9 +89,9 @@ jobs:
         - sudo apt-get update && sudo apt-get install -y python3-pip python3-setuptools && pip3 install --upgrade --user awscli
         - $(aws ecr get-login --no-include-email --region $AWS_REGION)
       script:
-        - cargo build --locked -p driver
-        - docker build --tag stablex-binary-public --build-arg SOLVER_BASE=gnosispm/dex-open-solver:$OPEN_SOLVER_VERSION --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
-        - docker build --tag stablex-binary-private --build-arg SOLVER_BASE=163030813197.dkr.ecr.eu-central-1.amazonaws.com/dex-solver:$PRIVATE_SOLVER_VERSION --build-arg RUST_BASE=rust-binary -f docker/rust/Dockerfile .
+        - cargo build --locked -p driver --release
```
fleupold, line 92: Judging from your comment on the issue (that it would be sensible to keep running debug builds in staging), would it be easy enough to only build release on tagged builds and set the flag conditionally based on whether we are on a tag or master like this: https://travis-ci.community/t/how-do-i-conditionally-define-environment-variables/3376/2?

fleupold (approved):

.travis.yml
e00E: I'd like to do this PR as is first and then we can change master to be release mode but with debug assertions and overflow checks.

Issue #1007, opened 2020-06-15
Query TokenData for all traded tokens


fleupold:
We currently have a hardcoded list with Symbol and Decimal information per token. For the visualization und understanding of solution results this can be helpful (e.g. to show user readable prices). 

Moreover, we are likely going to need this information in the price estimation service when we want to give price estimations in base units (rather than atoms)

fleupold:
Waiting on https://github.com/gnosis/dex-contracts/pull/860

PR #1057, opened 2020-06-29
[TokenInfo 1] Extract TokenData into TokenInfoFetching trait

fleupold:
Part of #1007

In this PR I abstract the hardcoded TokenData behavior we have at the moment into a trait that can implemented in different ways (e.g. fetching from chain, fetching from a cache that is backed by another fetcher, etc).

I'm implementing this in a separate core module, so that it can be reused by the price estimator when implementing routes without the `atoms=true` flag.

In the next PR I will make the trait async so that it can be implemented using long-running operations. I will then implement an on-chain fetcher that uses the new BatchExchangeViewer contract. Afterwards I will put an in memory cache in between the on chain fetcher to speed up look ups after they have been once fetched. Ultimately we will have an implementation that looks up information in the primary (hardcoded) list or fetches the data from (cached) chain.

### Test Plan
CI and making sure instance files still look like today when running the driver locally (using the gnosis-staging hardcoded TOKEN_DATA).


core/src/token_info/hardcoded.rs
```diff
@@ -0,0 +1,115 @@
+//! This module contains fallback token data that should be used by the price
```
fleupold, line 1: This file has been moved 1:1 from `core/src/price_estimation/data.rs` (TokenBaseInfo has been moved up into `core/src/token_info.rs`).

e00E:
I think you forgot to `add core/src/token_info.rs` to the commit.

e00E (approved):

core/src/token_info.rs
```diff
@@ -0,0 +1,77 @@
...
+        SYMBOL_OVERRIDES.get(&self.alias).unwrap_or(&self.alias)
+    }
+
+    /// Converts the prices from USD into the unit expected by the contract.
+    /// This price is relative to the OWL token which is considered pegged at
+    /// exactly 1 USD with 18 decimals.
+    pub fn get_owl_price(&self, usd_price: f64) -> u128 {
+        let pow = 36 - (self.decimals as i32);
+        (usd_price * 10f64.powi(pow)) as _
+    }
```
bh2smith, line 66: Nice!

core/src/token_info/hardcoded.rs
```diff
@@ -0,0 +1,115 @@
...
+            (
+                TokenBaseInfo::new("DAI", 18, 0),
+                1.01,
+                1.01 * 10f64.powi(18),
+            ),
+            (TokenBaseInfo::new("FAKE", 32, 0), 1.0, 10f64.powi(4)),
+            (
+                TokenBaseInfo::new("SCAM", 42, 0),
+                10f64.powi(10),
+                10f64.powi(4),
```
bh2smith, line 95: `36 - 42 + 10 = 4` check.


What happens when this results in a negative exponent?

bh2smith (approved):
Looks good, might include another test of what happens when the owl price results in negative exponents.

core/src/token_info/hardcoded.rs
fleupold: These tests were just moved from the original location. Ammending them is not really in the scope of this PR. A negative exponent would be a number <1, right? This is checking amounts, which are expressed as ints as far as I know.

PR #1062, opened 2020-06-30
Change ordering of transitive orders in price estimator response

e00E:
The frontend requires the specific ordering implemented in this PR. Noticed this yesterday evening when one half of the orderbook visualization in the frontend was mirrored.

### Test Plan
Adapted unit test. Tested visually with local frontend that it looks correct.

### DIFF
--- a/price-estimator/src/models.rs
+++ b/price-estimator/src/models.rs
@@ -67,7 +67,7 @@ pub struct EstimatedOrderResult {
     pub sell_amount_in_quote: u128,
 }
 
-#[derive(Debug, PartialEq, Serialize)]
+#[derive(Clone, Debug, PartialEq, Serialize)]
 pub struct TransitiveOrder {
     pub price: f64,
     pub volume: f64,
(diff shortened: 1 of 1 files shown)


nlordell (approved):

price-estimator/src/models.rs
```diff
@@ -80,8 +80,13 @@ pub struct MarketsResult {
 }
 
 /// Sort orders by their price and merge orders with the same price into one.
-fn sanitize_orders(mut orders: Vec<TransitiveOrder>) -> Vec<TransitiveOrder> {
-    orders.sort_unstable_by(|a, b| a.price.partial_cmp(&b.price).unwrap_or(Ordering::Less));
+/// If reverse is false sorts by price ascending, if true descending.
+fn sanitize_orders(mut orders: Vec<TransitiveOrder>, reverse: bool) -> Vec<TransitiveOrder> {
```
fleupold, line 84: Nit: seeing that there is a bunch going on in the comments of this method. Maybe the flag could be called `descending` in which case we don't have to explain what we mean by reverse. Also `sanitize` seem to be a very general term meaning

> make clean and hygienic; disinfect.

How about naming this method `sort_and_aggregate_orders_by_price`?

fleupold (approved):

PR #1064, opened 2020-06-30
Event iterators for batch ranges

nlordell:
Adds `EventRegstry` utility methods for getting iterators to events for batch ranges.

### Test Plan

Added unit test.

### DIFF
--- a/core/src/history/events.rs
+++ b/core/src/history/events.rs
@@ -110,10 +110,20 @@ impl EventRegistry {
         )
     }
 
+    /// Returns an iterator over all owned events and their corresponding batch
+    /// IDs.
     pub fn into_events(self) -> impl Iterator<Item = (batch_exchange::Event, BatchId)> {
         self.events
             .into_iter()
             .map(|(_, Value { event, batch_id })| (event, batch_id))
+    }
+
+    /// Returns an iterator of all events and their corresponding batch IDs.
+    /// This is a borrowed version of the `into_events` method.
+    pub fn events(&self) -> impl Iterator<Item = (&'_ batch_exchange::Event, BatchId)> + '_ {
+        self.events
+            .values()
+            .map(|Value { event, batch_id }| (event, *batch_id))
     }
 
     /// Returns an iterator containing all events up to and including the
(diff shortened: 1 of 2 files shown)


core/src/history/events.rs
```diff
@@ -118,6 +115,40 @@ impl EventRegistry {
...
+    fn events_for_batch_range(
+        &self,
+        start_batch_id: BatchId,
+        end_batch_id: BatchId,
+    ) -> impl Iterator<Item = (&'_ batch_exchange::Event, BatchId)> + '_ {
+        self.events
+            .values()
+            .skip_while(move |entry| entry.batch_id < start_batch_id)
+            .take_while(move |entry| entry.batch_id <= end_batch_id)
+            .map(|entry| (&entry.event, entry.batch_id))
```
e00E, line 150: This is possible in a more efficient way with `BTreeMap::range`. Then we might as well accept a range as the parameter instead of start and end batch ids.
I'm also wondering if it wouldn't be better to give direct access to `self.events` instead of going through a layer between. Could even provide the `map(|entry| (&entry.event, entry.batch_id))` as a free function if this is what most consumers want. This approach has less encapsulation which is a negative if we want to refactor the internals in the future but a positive in terms of amount of code needed just for the encapsulation right now. I'm fine with either approach.

e00E (approved):

core/src/history/events.rs
nlordell: >  is possible in a more efficient way with `BTreeMap::range`.

Unfortunately, this does not work for values, only for keys :disappointed:. I also couldn't find any way to do a binary search on values (despite them being ordered in our binary tree).

 > I'm also wondering if it wouldn't be better to give direct access to self.events instead [...] This approach has less encapsulation

Personally, I prefer more encapsulation. Specifically, I imagine their may be some invariants that are upheld by `handle_event_data` that would be possible to break otherwise. Additionally, it would require making the key and value structs public as well.

> Could even provide the map(|entry| (&entry.event, entry.batch_id)) as a free function

I think having an `events` iterator like this makes a lot of sense... I will update.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FyyZRSvISN1vvW%2Fgiphy.gif&token=nice" data-gitmeme-token="nice"></a>

PR #1065, opened 2020-06-30
[TokenInfo 2] Making TokenInfoFetching async

fleupold:
Part of #1007 

In order to fetch data from chain we need an async way of fetching token info. This PR refactors the existing trait and callsites to work with async assumptions.

### Test Plan
CI

### DIFF
--- a/core/src/macros.rs
+++ b/core/src/macros.rs
@@ -28,3 +28,9 @@ macro_rules! std_map {
         map
     }}
 }
+
+macro_rules! immediate {
+    ($expression:expr) => {
+        async { $expression }.boxed()
+    };
+}
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -21,7 +21,7 @@ use anyhow::Result;
 use average_price_source::AveragePriceSource;
 use futures::future::{BoxFuture, FutureExt as _};
 use log::warn;
-use price_source::{NoopPriceSource, PriceSource};
+use price_source::PriceSource;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::iter;
 use std::iter::FromIterator;
@@ -55,26 +55,21 @@ impl PriceOracle {
         token_info_fetcher: Arc<dyn TokenInfoFetching>,
         update_interval: Duration,
     ) -> Result<Self> {
-        let source: Box<dyn PriceSource + Send + Sync> = if token_info_fetcher.all_ids()?.is_empty()
-        {
-            Box::new(NoopPriceSource)
-        } else {
-            let (kraken_source, _) = ThreadedPriceSource::new(
-                token_info_fetcher.all_ids()?,
-                KrakenClient::new(http_factory, token_info_fetcher.clone())?,
-                update_interval,
-            );
-            let (dexag_source, _) = ThreadedPriceSource::new(
-                token_info_fetcher.all_ids()?,
-                DexagClient::new(http_factory, token_info_fetcher.clone())?,
-                update_interval,
-            );
-            Box::new(AveragePriceSource::new(vec![
-                Box::new(kraken_source),
-                Box::new(dexag_source),
-                Box::new(PricegraphEstimator::new(orderbook_reader)),
-            ]))
-        };
+        let (kraken_source, _) = ThreadedPriceSource::new(
+            token_info_fetcher.clone(),
+            KrakenClient::new(http_factory, token_info_fetcher.clone())?,
+            update_interval,
+        );
+        let (dexag_source, _) = ThreadedPriceSource::new(
+            token_info_fetcher.clone(),
+            DexagClient::new(http_factory, token_info_fetcher.clone())?,
+            update_interval,
+        );
+        let source = Box::new(AveragePriceSource::new(vec![
+            Box::new(kraken_source),
+            Box::new(dexag_source),
+            Box::new(PricegraphEstimator::new(orderbook_reader)),
+        ]));
 
         Ok(PriceOracle {
             token_info_fetcher,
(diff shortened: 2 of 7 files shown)


core/src/price_estimation/kraken.rs
```diff
@@ -49,6 +50,16 @@ where
     /// that are used when computing the price map.
     #[allow(clippy::needless_lifetimes)]
     async fn get_token_asset_pairs(&self, tokens: &[TokenId]) -> Result<HashMap<String, TokenId>> {
+        let token_infos: Vec<_> = stream::iter(tokens)
```
nlordell, line 53: I feel like this snippet it repeated across `PriceSource` implementations. Does it make sense for just `PriceOracle` to fetch all the on-chain token data and then just pass it as a parameter to each `PriceSource`?

core/src/macros.rs
```diff
@@ -28,3 +28,9 @@ macro_rules! std_map {
         map
     }}
 }
+
+macro_rules! immediate {
```
nlordell, line 32: Love this!

nlordell (approved):
Looks good, my only comment would that it would be nice to retrieve on-chain token information only in one place and pass it down to each of the price sources. Not sure it that makes sense though, I still have to review #1057 more closely.

fleupold:
> my only comment would that it would be nice to retrieve on-chain token information only in one place and pass it down to each of the price sources.

While this is possible my reason for not doing so is that we want to do the lookup at runtime, not at instantiation time (if e.g. in the case of prod the service runs for a week it would not get any new token Info otherwise). We could pass tokenData in the getPrices method on every lookup (this would significantly affect the the plan I outlined in #1057 though)

core/src/price_estimation/kraken.rs
fleupold: It depends on which tokens are listed on the exchange (e.g. if we have 100 💩 coins we might only fetch token data for a the small subset that is listed on kraken & dex.AG)

nlordell: Makes sense. I think if we add a caching layer as you suggested it will be fine.

core/src/macros.rs
bh2smith: Moi aussie!

core/src/price_estimation.rs
```diff
@@ -21,7 +21,7 @@ use anyhow::Result;
 use average_price_source::AveragePriceSource;
 use futures::future::{BoxFuture, FutureExt as _};
 use log::warn;
-use price_source::{NoopPriceSource, PriceSource};
```
bh2smith, line 24: What was `Noop`supposed to mean anyway?

core/src/token_info.rs
```diff
@@ -1,17 +1,21 @@
...
 
+#[cfg_attr(test, automock)]
 pub trait TokenInfoFetching: Send + Sync {
     /// Retrieves some token information from a token ID.
-    fn get_token_info(&self, id: TokenId) -> Result<TokenBaseInfo>;
+    fn get_token_info<'a>(&self, id: TokenId) -> BoxFuture<'a, Result<TokenBaseInfo>>;
 
     /// Returns a vector with all the token IDs available
-    fn all_ids(&self) -> Result<Vec<TokenId>>;
+    fn all_ids<'a>(&self) -> BoxFuture<'a, Result<Vec<TokenId>>>;
```
bh2smith, line 18: Could you tell me a bit about the reason for introducing this... lifetime parameter? `<`a>`

bh2smith (approved):

core/src/token_info.rs
fleupold: Great question!

So a BoxFuture always needs a lifetime (https://docs.rs/futures/0.3.5/futures/future/type.BoxFuture.html) which cannot be inferred from the [elision rules](https://doc.rust-lang.org/reference/lifetime-elision.html) therefore it has to be explicitly stated. 
Now, just using `'a` here might be pointless since no other reference on the method definition has the same lifetime (so we could use `'static` as well). If I understand correctly this lifetime basically tells the compiler what references it can borrow inside the future (since the future might live/be active beyond this function call).

In other cases we have shared the `'a` lifetime with `&'a self` which means that the future can borrow variables from self (we are basically guaranteeing the future will be consumed before self goes out of scope).

I didn't need to borrow `self` in the code so far, but maybe we should still allow it in the abstract trait? What do our rust experts think?

core/src/price_estimation.rs
fleupold: https://en.wikipedia.org/wiki/NOP_(code)

core/src/token_info.rs
nlordell: > What do our rust experts think?

I think it was a great explanation!

> I didn't need to borrow self in the code so far, but maybe we should still allow it in the abstract trait? 

I think so, or else you will get "does not live long enough" errors if you borrow self for the future. When implementing the caching layer as you had suggested in the previous PR, this might become a problem as you may need to update cached values which would require `self` to at least live for part of the future.

e00E:
```rust
fn all_ids<'a>(&self) -> BoxFuture<'a, Result<Vec<TokenId>>>;
```

I know it's already changed but want to point out that I think this doesn't make sense. This signature says that the caller of the function can pick the lifetime of the return value freely because no other parameter uses this lifetime. This can't be right because the caller's choice doesn't actually influence body of this function. In practice it would cause all implementation to have to return `'static` because that is the only way you can guarantee to outlive `'a` because `'a` might also be `'static` in which case we should just write `'static` in the signature.

core/src/price_estimation/threaded_price_source.rs
```diff
@@ -46,6 +47,14 @@ impl ThreadedPriceSource {
     }
 }
 
+async fn update_prices<T: PriceSource>(
+    price_source: &T,
+    token_info_fetching: &Arc<dyn TokenInfoFetching>,
```
e00E, line 52: In general `&Arc<dyn Trait>` should be `&dyn Trait` because either way it's just a reference but the second signature is more flexible. It would be called by turning `&token_info_fetcher` into `token_info_fetcher.as_ref()`.
Don't need to change it here because this is really only meant to be called from that one place.

core/src/price_estimation/kraken.rs
```diff
@@ -91,19 +102,22 @@ where
...
-                        .get_owl_price(info.p.last_24h());
-
-                    Some((*token, price))
+            let prices = stream::iter(ticker_infos)
+                .filter_map(|(pair, info)| {
+                    let token_info_fetcher = self.token_info_fetcher.clone();
+                    let token = token_asset_pairs.get(&pair);
+                    async move {
+                        let price = token_info_fetcher
+                            .get_token_info(*(token?))
```
e00E, line 111: It's weird that we do `get_token_info` on the same tokens that we already have fetched previously through `get_token_asset_pairs`. Doesn't really matter if it's cached though.

e00E (approved):

core/src/price_estimation/threaded_price_source.rs
fleupold: Good point! Will change it to make it a habit of not using &Arc

nlordell: Good catch! One "valid" use of `&Arc` is if you intend to clone it. Although, in that case its probably better to pass in an owned value `Arc` instead, and have the caller clone it if it needs to.

core/src/price_estimation/kraken.rs
fleupold: Good point. I was able to get rid of it by putting the info into the asset pair hash map we collect already.

PR #1066, opened 2020-06-30
compute historic auction elements and settlement

nlordell:
Add utility methods for computing historic auction elements and settlement data.

### Test Plan

Unit test :tada:


core/src/history.rs
```diff
@@ -99,4 +155,76 @@ mod tests {
...
+            ExchangeHistory { events }
+        };
+
+        assert_eq!(
+            history.settlement_for_batch(batch).unwrap(),
+            Settlement { trades, solution },
+        );
+    }
+
+    #[test]
```
fleupold, line 223: Nice, we could test that trade reversion works as expected but I can sympathize if this is a lot of extra boilerplate code to add.

fleupold (approved):
🍠 

Looks great!

e00E (approved):

core/src/history.rs
nlordell: I already test trade revision :smile: 
https://github.com/gnosis/dex-services/blob/bddbdfae9d2325709e9409b34e64dcdb70f97213/core/src/history.rs#L195-L199

I'll split it into two tests though, so its clearer.

PR #1067, opened 2020-07-01
Added estimaged exchange rate API

nlordell:
This PR adds a new route for estimating the price for the best ask in a market.

Note that there still exists the possibility that dust orders affect the estimate even if they aren't actually tradable. This, however, requires a larger fix independent of the addition of this route.

### Test Plan

Start the price estimator and:
```
$ curl -s 'http://localhost:8080/api/v1/markets/1-7/estimated-best-ask-price?atoms=true'
215.68243731260944
```


anxolin:
Thanks a lot for this PR! this is going to be super useful 🔥🔥

unrelated, but, git support for Rust feels a bit odd. For example it was not super smart comparing the code around `get_market` 

For the name, and signature I like it. I would double check how other exchanges call this endpoint.

Also, unrelated to this PR. Is there any HTTP error we should control for this or other endpoints? what happens if there's no orders? no token, or an internal exception? what would be the return data signature in case of an error?

e00E:
> git support for Rust feels a bit odd. For example it was not super smart comparing the code around get_market

Afaik git has no special language support except for trying to show you which function or module you are in if the diff itself is only a small part of it. Locally you can try different diff algorithms but I don't think the Github ui does this.

> Also, unrelated to this PR. Is there any HTTP error we should control for this or other endpoints? what happens if there's no orders? no token, or an internal exception? what would be the return data signature in case of an error?

From the perspective of users of the api I would say anything except a http success status code is an error. If we return success then the resulting json always matches the expected schema. We should probably still document which status codes are used in case of error but it shouldn't be part of the api guarantees made.

Currently if the request path and parameters are valid we never return an error, so for example if there are no orders and you try to get markets then asks and bids are empty and if you try to estimate the price and there are no orders you will get a price 0.

price-estimator/src/filter.rs
```diff
@@ -157,6 +183,24 @@ async fn estimate_amounts_at_price<T>(
...
+    let price = orderbook
+        .get_pricegraph()
+        .await
+        .estimate_exchange_rate(token_pair.into(), 0.0)
+        .map(|xrate| {
+            // NOTE: Exchange rate is the inverse of price for an ask order.
+            1.0 / apply_rounding_buffer(xrate, price_rounding_buffer)
+        });
+
+    Ok(warp::reply::json(&price))
```
e00E, line 201: We should probably use a type from `models` here so it's clear what the schema is. Currently price is an `Option<f64>`.

nlordell:
> Is there any HTTP error we should control for this or other endpoints?

I think we only have a couple errors in 400s and 404s when either endpoints don't exist or some part of the filter fails to parse.

> what happens if there's no orders? no token?

Usually empty results, (like no bid/ask orders, or an `null` price or amounts).

> or an internal exception?

In theory there should be no exceptions (`panic`s in Rust). If there are any, I believe the TCP connection is closed, so you would have an HTTP protocol error.

price-estimator/src/filter.rs
nlordell: Let me know if this is what you meant.

e00E: That's fine. We could still think about whether we always want this to be a json object but most importantly I wanted to see the type.

nlordell: A number is valid JSON though. We can modify the schema if needed.

price-estimator/src/models.rs
```diff
@@ -102,6 +102,11 @@ impl From<&pricegraph::TransitiveOrderbook> for MarketsResult {
     }
 }
 
+/// A type representing a price estimate result.
+#[derive(Debug, Serialize)]
+#[serde(transparent)]
+pub struct PriceEstimateResult(pub Option<f64>);
```
fleupold, line 108: I assume the fact that the price is in quote is evident?

fleupold (approved):
Looks good! I'm wondering if the volume at that price is relevant and should be part of the response or not?

On a side note unrelated to this PR, maybe we can get rid of the `estimated` prefix in all the routes at some point. Given the context of this service and the URL it is deployed under it's pretty clear that all values are be estimates.

e00E (approved):

nlordell:
> I'm wondering if the volume at that price is relevant and should be part of the response or not?

We can add it, but it would require pricegraph changes, so if we decide to do it, I think it should be in a follow-up PR as it would be a bit more involved.

> On a side note unrelated to this PR, maybe we can get rid of the estimated prefix in all the routes at some point

I would like that.

price-estimator/src/models.rs
nlordell: I'll add a note in the comments somewhere.

PR #1068, opened 2020-07-01
PriorityPriceSource

fleupold:
Partially related to #1007. While trying to read `TokenBaseInformation` from chain I noticed that this struct also contains the external price which is not part of the data we fetch from chain (we only fetch symbol and decimals). 

I therefore started looking at what it takes to separate the price information from the token base info (in line with @e00E's suggestion to make the token base info a separate piece of metadata that the solver does not need to care about).

My suggestion for separating the two is as follows:

1. Have the hardcoded TokenData contain a struct that can have any of the three pieces of information: decimals, symbol, external price (this will allow us to override any and all aspects of a token). This struct will be read from the config like today.
2. Implement `PriceSource` on the hardcoded `TokenData` (on top of `TokenInfoFetching`)
3. Use the abstract `PriceSource` trait to fetch the final price (instead of having custom override logic in `PriceOracle`)
4. Make the TokenInfoFetching return a struct that only contains `symbol` and `decimals`. This can then be implemented by the hardcoded TokenData as well as the Onchain TokenInfoFetching.
5. Use a similar override chain for `TokenInfo` (first looking at TokenData then in the cache, then onchain) as for the different price sources

This PR specifically is part of step 3 and will allow the PriceOracle to get rid of custom override logic and instead rely on its price source hierarchy to chose the right price. Hardcoded TokenData will become the highest priority source (overriding all other sources) followed by the current AveragePriceSource.

This will allow us to get rid of some logic in PriceOracle

### Test Plan
Unit tests

### DIFF
--- a/core/src/price_estimation/priority_price_source.rs
+++ b/core/src/price_estimation/priority_price_source.rs
@@ -30,10 +30,7 @@ impl PriceSource for PriorityPriceSource {
             for source in &self.sources {
                 match source.get_prices(&remaining_tokens).await {
                     Ok(partial_result) => {
-                        remaining_tokens = remaining_tokens
-                            .into_iter()
-                            .filter(|token| !partial_result.contains_key(token))
-                            .collect();
+                        remaining_tokens.retain(|token| !partial_result.contains_key(token));
                         result.extend(partial_result);
                     }
                     Err(err) => log::warn!("Price Source failed: {:?}", err),
--- a/core/src/price_estimation/priority_price_source.rs
+++ b/core/src/price_estimation/priority_price_source.rs
@@ -2,7 +2,6 @@ use super::PriceSource;
 use crate::models::TokenId;
 use anyhow::Result;
 use futures::future::{BoxFuture, FutureExt as _};
-use std::collections::hash_map::RandomState;
 use std::collections::{HashMap, HashSet};
 use std::iter::FromIterator;
 
@@ -27,8 +26,7 @@ impl PriceSource for PriorityPriceSource {
         tokens: &'a [TokenId],
     ) -> BoxFuture<'a, Result<HashMap<TokenId, u128>>> {
         async move {
-            let mut remaining_tokens: HashSet<TokenId, RandomState> =
-                HashSet::from_iter(tokens.iter().cloned());
+            let mut remaining_tokens: HashSet<_> = HashSet::from_iter(tokens.iter().cloned());
             let mut result = HashMap::new();
             for source in &self.sources {
                 let remaining_token_vec = Vec::from_iter(remaining_tokens.iter().cloned());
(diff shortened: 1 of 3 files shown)


core/src/price_estimation/priority_price_source.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+impl PriceSource for PriorityPriceSource {
+    fn get_prices<'a>(
+        &'a self,
+        tokens: &'a [TokenId],
+    ) -> BoxFuture<'a, Result<HashMap<TokenId, u128>>> {
+        async move {
+            let mut remaining_tokens: HashSet<TokenId, RandomState> =
```
nlordell, line 30: nit: `RandomState` is the default hash state for the a `HashSet` so it is superfluous here.
```suggestion
            let mut remaining_tokens: HashSet<TokenId> =
```

Also, the type for the `HashSet` item can also be elided in this case:
```suggestion
            let mut remaining_tokens: HashSet<_> =
```

nlordell (approved):
Looks great! A couple of very small nits.

core/src/price_estimation/priority_price_source.rs
```diff
@@ -0,0 +1,141 @@
...
+            let mut remaining_tokens: HashSet<_> = HashSet::from_iter(tokens.iter().cloned());
+            let mut result = HashMap::new();
+            for source in &self.sources {
+                let remaining_token_vec = Vec::from_iter(remaining_tokens.iter().cloned());
+                match source.get_prices(&remaining_token_vec).await {
+                    Ok(partial_result) => {
+                        remaining_tokens = remaining_tokens
+                            .into_iter()
+                            .filter(|token| !partial_result.contains_key(token))
+                            .collect();
```
e00E, line 38: Can probably be replaced with `HashSet::retain`. Or even make remaining tokens be `let mut remaining_tokens = tokens.to_vec()` then use `retain` on that and pass `&remaining_tokens` to `get_prices` since we aren't actually making use of this being a set instead of a vec.

e00E (approved):
I like the outline you described for future changes.

bh2smith (approved):
Really nice to review. Just great!

core/src/price_estimation/priority_price_source.rs
fleupold: Right, O(1) lookup is only used for partial result so hash set is not needed. Will use a vec instead.

e00E: instead of filter and collect we can still use `retain` on the vector here

fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FkVVEP6SmfEXvy%2Fgiphy.gif&token=syntactic_sugar" data-gitmeme-token="syntactic_sugar"></a>

PR #1069, opened 2020-07-01
Historic prices with pricegraph spread.

nlordell:
This PR implements a simple program that outputs historic prices with pricegraph price estimate spreads. This program records for each solved batch the settled price and the estimated bid-ask spread. The spread is computed with `pricegraph` by reducing all overlapping ring trades and then finding the best remaining ask and bid transitive orders for each traded market. Note that this implies that `best_bid < price < best_ask` should hold. In cases where it doesn't it means there is a likelihood of producing bad price estimates for that market.

### Test Plan

```
$ cargo run --release -p e2e -- --orderbook-file target/orderbook-file-mainnet
    Finished release [optimized] target(s) in 0.16s
9402 / 45998 [=============>-----------------------------------------------------] 20.44 % 13666.57/s 3s
```

We can plot the results (here for WETH-DAI market) and see that the price estimates are usually exactly between the bid/ask spread. There are currently a few outliers (~6%) that I still need to figure out what happened. But I suspect that they are related to the solver and pricegraph choosing different orders when reducing overlap (i.e. matching orders with overlapping exchange rates).

image


e2e/src/bin/historic_prices.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    let mut spreads = File::create(&options.output_dir.join("spreads.csv"))?;
+    writeln!(&mut spreads, "market,best_bid,price,best_ask")?;
+
+    let first_batch = history
+        .first_batch()
+        .ok_or_else(|| anyhow!("exchange has no events"))?;
+    let (count, batches) = batches_until_now_from(first_batch);
+
+    let mut progress = ProgressBar::on(io::stderr(), count);
```
fleupold, line 43: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F134iGJcqa8A9lC%2Fgiphy.gif&token=fancy" data-gitmeme-token="fancy"></a>

fleupold (approved):
Looks really cool! So we want `best_bid < price < best_ask` because in our estimation we have already reduced any overlaps (intuitively I thought we need overlapping orderbooks and thus a negative spread).

How do you know which batchID is the datapoint from in your graph? I don't see that. info in the CSV.

e00E (approved):

nlordell:
> So we want best_bid < price < best_ask because in our estimation we have already reduced any overlaps (intuitively I thought we need overlapping orderbooks and thus a negative spread).

Yes, and this would be a good indication of bad estimates. In fact, we are seeing this is off around 6% of the time so it in line with what you previously calculated with how often price estimates were off.

> How do you know which batchID is the datapoint from in your graph? I don't see that. info in the CSV.

:see_no_evil: forgot to push the commit

nlordell:
@fleupold I updated the PR comment to be a bit more descriptive.

core/src/history.rs
```diff
@@ -32,6 +32,11 @@ impl ExchangeHistory {
         ExchangeHistory::read(File::open(filestore)?)
     }
 
+    /// Returns the very first batch for the exchange.
+    pub fn first_batch(&self) -> Option<BatchId> {
+        self.events.events().next().map(|(_, batch)| batch)
```
bh2smith, line 37: Does this work on the assumption that the first event emitted contains a batch information element? If not, how does this get the first batch?

PR #1070, opened 2020-07-01
Implementing PriceSource for TokenData

fleupold:
Depends on #1068

This PR implements PriceSource on TokenData and uses it in combination with the priority price source to get rid of the custom override logic in the PriceOracle. We can now rely on the PriceSource to give us a correct price estimate and can orthogonally and optionally (later we might want to do this somewhere else), fetch token information via the TokenInfoFetching trait.

### Test Plan

Unit tests as well as making sure that our instance file still has the correct metadata when running locally on mainnet.

### DIFF
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -5,14 +5,14 @@ mod average_price_source;
 mod dexag;
 mod kraken;
 mod orderbook_based;
-mod price_source;
+pub mod price_source;
 mod priority_price_source;
 mod threaded_price_source;
 
 use self::dexag::DexagClient;
 use self::kraken::KrakenClient;
 use self::orderbook_based::PricegraphEstimator;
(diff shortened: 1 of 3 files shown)


nlordell (approved):
This is awesome! Really putting those abstractions to work!

e00E (approved):

core/src/price_estimation.rs
```diff
@@ -123,22 +129,20 @@ impl PriceEstimating for PriceOracle {
...
-                            ..base_info.into()
-                        })
-                    } else if let Some(price) = price {
-                        Some(TokenInfo {
-                            alias: None,
-                            decimals: None,
-                            external_price: price,
-                        })
-                    } else {
-                        None
```
fleupold, line 140: This PR actually changes this logic in an undesired way. We only use the external price of price is not available (I though we always use the external price if there is custom token info).

PR #1071, opened 2020-07-01
Use two-stage model for standard solver

twalth3r:
This PR:
* passes the `--optModel` parameter to the private solvers (standard and best-ring)
* uses the two-stage model for the standard solver (see https://github.com/gnosis/dex-solver/pull/296)
* updates the solver version to v0.7.7

### Test Plan
e2e tests

### DIFF
--- a/core/src/price_finding/price_finder_interface.rs
+++ b/core/src/price_finding/price_finder_interface.rs
@@ -50,6 +50,33 @@ impl InternalOptimizer {
     }
 }
 
+/// The optimization algorithm used by standard and best-ring solvers.
+#[derive(Clone, Copy, Debug)]
+pub enum OptModel {
+    Mip,
+    TwoStage,
+}
+
+impl FromStr for OptModel {
+    type Err = Error;
+    fn from_str(string: &str) -> Result<Self> {
+        match string {
+            "mip" => Ok(Self::Mip),
+            "two_stage" => Ok(Self::TwoStage),
+            _ => Err(anyhow!("optimization model does not exit")),
+        }
+    }
+}
+
+impl OptModel {
+    fn to_argument(self) -> &'static str {
+        match self {
+            OptModel::Mip => "mip",
+            OptModel::TwoStage => "two_stage",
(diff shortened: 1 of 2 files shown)


core/src/price_finding/price_finder_interface.rs
```diff
@@ -50,6 +50,33 @@ impl InternalOptimizer {
...
+impl FromStr for OptModel {
+    type Err = Error;
+    fn from_str(string: &str) -> Result<Self> {
+        match string {
+            "mip" => Ok(Self::Mip),
+            "two_stage" => Ok(Self::TwoStage),
+            _ => Err(anyhow!("optimization model does not exit")),
+        }
+    }
+}
```
nlordell, line 69: nit: This shouldn't be needed here since we aren't reading an optimization model from a string, but instead it is being determined by its solver type. Maybe this should be a configurable option? If so, we can leave this in and add the command line parsing stuff in a follow-up PR.

twalth3r: Removed in 18450de

core/src/price_finding/price_finder_interface.rs
```diff
@@ -50,6 +50,22 @@ impl InternalOptimizer {
     }
 }
 
+/// The optimization algorithm used by standard and best-ring solvers.
+#[derive(Clone, Copy, Debug)]
+pub enum OptModel {
+    Mip,
```
fleupold, line 56: ubernit: given that we are not super familiar with the optimization models and their abbreviations, maybe we can write MixedIntegerProgramming out in the enum definition?

twalth3r: That's an awesome suggestion, having to write MIP as Mip gave me a sleepless night already! I went with `OptModel::MixedInteger` which is much more pleasing to my eyes now. Thanks!

twalth3r:
> How would you rate you experience with Rust on a scale from 1 to 5?

I would definitely recommend it to my friends!

PR #1073, opened 2020-07-02
On-Chain TokenInfoFetching

fleupold:
Part of #1007 - depends on #1075

This PR implements TokenInfoFetching via the BatchExchange and BatchExchangeWrapper smart contract. Requires an update of the submodule to get the latest contract deployments and exposing the new methods so that they can be used to implement the trait.

### Test Plan
Integration Test

### DIFF
--- a/core/src/token_info/onchain.rs
+++ b/core/src/token_info/onchain.rs
@@ -7,11 +7,8 @@ use crate::contracts::stablex_contract::StableXContractImpl;
 impl TokenInfoFetching for StableXContractImpl {
     fn get_token_info<'a>(&'a self, id: TokenId) -> BoxFuture<'a, Result<TokenBaseInfo>> {
         async move {
-            let info = self.get_token_info(id.into()).await?;
-            Ok(TokenBaseInfo {
-                alias: info.1,
-                decimals: info.2,
-            })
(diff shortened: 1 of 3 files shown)


nlordell (approved):
Nice! All coming together :smile:. Should the dex-contracts upgrade be a separate PR? I'm OK either way.

core/src/token_info/onchain.rs
```diff
@@ -0,0 +1,27 @@
+use anyhow::Result;
+use futures::future::{BoxFuture, FutureExt};
+
+use super::{TokenBaseInfo, TokenId, TokenInfoFetching};
+use crate::contracts::stablex_contract::StableXContractImpl;
+
+impl TokenInfoFetching for StableXContractImpl {
+    fn get_token_info<'a>(&'a self, id: TokenId) -> BoxFuture<'a, Result<TokenBaseInfo>> {
+        async move {
+            let info = self.get_token_info(id.into()).await?;
```
nlordell, line 10: ubernit: I would prefer destructuring here instead of indexing into the tuple:
```suggestion
            let (alias, decimals) = self.get_token_info(id.into()).await?;
```

nlordell (approved):

e00E (approved):

PR #1074, opened 2020-07-02
Cached TokenInfoFetching

fleupold:
Part of #1007 

TokenInfo is unlikely to change over time (contracts would have to do something really 🤡 like changing decimals based on block number or the like).
Hitting the chain every time we want to fetch data is costly, and can easily be cached. This PR implements an in-memory cache for fetching token Info. It still hits the chain every-time we want to fetch all token Ids, since this can theoretically happen on every block (depending on load, we might consider caching this as well with a time to live).

I had to use a thread safe, interior mutability datatype and chose a RWLock as on cache hit we only read the cache and this can happen in parallel (a Mutex does not allow parallel reads). Not that this performance difference is critical, but I didn't see a reason to use a Mutex (couldn't find clear evidence on overhead and [this post](https://www.reddit.com/r/rust/comments/5bx34b/mutex_vs_rwlock/?utm_source=share&utm_medium=web2x) suggests the main difference are trait bounds on the inner type).

### Test Plan
Unit tests

### DIFF
--- a/core/src/token_info/cached.rs
+++ b/core/src/token_info/cached.rs
@@ -25,7 +25,7 @@ impl TokenInfoCache {
     }
 
     #[allow(dead_code)]
-    pub fn new_with_existing_cache(
+    pub fn new_with_cache(
         inner: Arc<dyn TokenInfoFetching>,
         cache: HashMap<TokenId, TokenBaseInfo>,
     ) -> Self {
@@ -138,7 +138,7 @@ mod tests {
             alias: "Foo".to_owned(),
             decimals: 42,
         };
-        let cache = TokenInfoCache::new_with_existing_cache(
+        let cache = TokenInfoCache::new_with_cache(
             Arc::new(inner),
             hash_map! {
                 TokenId::from(1) => hardcoded.clone()
--- a/core/src/token_info/cached.rs
+++ b/core/src/token_info/cached.rs
@@ -22,6 +22,17 @@ impl TokenInfoCache {
             inner,
         }
     }
+
+    #[allow(dead_code)]
+    pub fn new_with_existing_cache(
+        inner: Arc<dyn TokenInfoFetching>,
+        cache: HashMap<TokenId, TokenBaseInfo>,
+    ) -> Self {
+        Self {
+            inner,
+            cache: RwLock::new(cache),
+        }
+    }
 }
 
 impl TokenInfoFetching for TokenInfoCache {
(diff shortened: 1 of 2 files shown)


core/src/token_info/cached.rs
```diff
@@ -0,0 +1,159 @@
+use super::{TokenBaseInfo, TokenId, TokenInfoFetching};
+
+use anyhow::Result;
+use futures::future::{BoxFuture, FutureExt};
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
```
e00E, line 6: Since this is async code we have to use an async mutex: https://docs.rs/async-std/1.6.2/async_std/sync/struct.RwLock.html .

core/src/token_info/cached.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Implementation of TokenInfoFetching that stores previously fetched information in an in-memory cache for fast retrieval.
+ * TokenIds will always be fetched from the inner layer, as new tokens could be added at any time.
+ */
+struct TokenInfoCache {
+    cache: RwLock<HashMap<TokenId, TokenBaseInfo>>,
```
nlordell, line 13: As @e00E mentioned, this should definitely use `async-std` synchronization primitives. This is especially true because of the potential for deadlocks here (seeing as it is possible for futures to run all on a single thread, which may be blocked waiting on a lock, preventing the future holding the lock to progress because the thread is not available for polling it).

core/src/token_info/cached.rs
```diff
@@ -0,0 +1,159 @@
...
+    #[allow(dead_code)]
+    pub fn new(inner: Arc<dyn TokenInfoFetching>) -> Self {
+        Self {
+            cache: RwLock::new(HashMap::new()),
+            inner,
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn new_with_existing_cache(
```
nlordell, line 27: ubernit: Typically, Rust functions that create a type with options, is just named `with_*` (like `Vec::new()` and `Vec::with_capacity()`). While its not a big deal at all, I think it would be better to follow the convention here.

nlordell (approved):
Nice! Just one issue with using standard synchronization primitives in an async context - which can lead to deadlock.

core/src/token_info/cached.rs
fleupold: Makes perfect sense and will change. Just to understand the concept of async a bit better would it really be possible to run into a deadlock here? We are not yielding while we hold the lock, so I don't really see how we can deadlock. Again this is a detail we should not rely on but for educational purposes, how could we deadlock in the current code?

nlordell: Looking at the code closely, I think you are right and I don't see how the current code would deadlock. For some reason I thought that the write handle was held for the whole `async { ... }.boxed()` scope, but it is only held for the `if` statement scope, and as you mentioned we never yield the future while it is held.

core/src/token_info/cached.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[allow(dead_code)]
+    pub fn new(inner: Arc<dyn TokenInfoFetching>) -> Self {
+        Self {
+            cache: RwLock::new(HashMap::new()),
+            inner,
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn new_with_cache(
```
nlordell, line 28: Hehe, I was suggesting more:
```suggestion
    pub fn with_existing_cache(
```

or 
```suggestion
    pub fn with_cache(
```

fleupold: 🤦 of course!

e00E (approved):

PR #1075, opened 2020-07-02
Remove price from TokenBaseInfo

fleupold:
Step 4 of plan outlined in #1068

This PR introduces a new type `TokenOverrideInfo` that is  read from the config `TOKEN_DATA` for overriding token information, such as symbol or price. `TokenBaseInfo` can then only contain `symbol` and `decimals` as prices are now fetched exclusively via the `PriceSource` trait.

Originally I planned to make all fields optional, however this will create some weirdness when we override symbol but still need to go on-chain for fetching decimals (the current abstraction doesn't work as nicely and would involve introducing a lot more optionals and `if let Some(x)` logic). It seems reasonable to override symbol and decimals in any case when doing a price override (they never change and symbol will serve as documentation which token we are overriding).

### Test Plan

CI

### DIFF
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -152,15 +152,15 @@ impl PriceEstimating for PriceOracle {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::token_info::{hardcoded::TokenData, TokenBaseInfo};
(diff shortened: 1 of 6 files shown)


nlordell (approved):
Easy `s/TokenBaseInfo/TokenInfoOverride/g`

e00E (approved):

PR #1077, opened 2020-07-02
Add 1inch API

### LINKED ISSUE #720: New price api: 1inch
It would be valuable to include another price API:
https://1inch.exchange/#/ETH/aDAI

1. 1inch delivers prices for tokens that dex.ag does not support: e.g. aDAI.
2. In the case of dex.ag being down, we still get a price feed for the majority of the tokens.
3. With 3 price-feeds we can throw alerts if we don't get any price for one token without getting too much noise

fedgiac:
Fixes #720. 

Adds a new price source ([1inch](https://1inch.exchange/#/)) to the price estimator.

### Test Plan

Make real requests to 1inch and see that the output prices are reasonable:
```
cd core
cargo test online_oneinch_api -- --ignored --nocapture
cargo test online_oneinch_client -- --ignored --nocapture
```

Compare line by line the prices of 1inch to the prices of dex.ag:
```
cargo build --tests && paste -d "#" <(cargo test online_oneinch_client -- --ignored --nocapture) <(cargo test online_dexag_client -- --ignored --nocapture) | sed "s/#/ <- 1inch\n/"
```

Sample output of the last command (currently 1 OWL ≈ 1 DAI):
```
Took 10.306066243 seconds to get prices. <- 1inch
Took 3.076605871 seconds to get prices.
Token ETH has OWL price of 230216987158628073472. <- 1inch
Token ETH has OWL price of 271568087876241457152.
Token USDT price could not be determined. <- 1inch
Token USDT has OWL price of 988542174834931369985118830592.
Token TUSD price could not be determined. <- 1inch
Token TUSD has OWL price of 1016945049680955520.
Token USDC has OWL price of 1084226059718864440266741252096. <- 1inch
Token USDC has OWL price of 989262897579293506779113062400.
Token PAX price could not be determined. <- 1inch
Token PAX has OWL price of 984341620006141056.
Token GUSD has OWL price of 7329162705030018616441094040190976. <- 1inch
Token GUSD price could not be determined.
Token DAI has OWL price of 1093977005319718912. <- 1inch
Token DAI has OWL price of 1000000000000000000.
Token sETH has OWL price of 230548517962684497920. <- 1inch
Token sETH has OWL price of 227093142966047801344.
Token sUSD has OWL price of 1102349188370144896. <- 1inch
Token sUSD has OWL price of 1069104617560581120.
Token SNX price could not be determined. <- 1inch
Token SNX has OWL price of 2323483974903152640.
test price_estimation::clients::oneinch::tests::online_oneinch_client ... ok <- 1inch
test price_estimation::clients::dexag::tests::online_dexag_client ... ok
```

Things to note:
1. dex.ag currently returns a wrong price for ETH/DAI. Compare to some ETH price from another source.
2. The price of USDC is quite different because the 1inch API tries to trade only 1 USDC. If it tried to use 10USDC the two prices would be similar. No plans to mitigate this issue currently.
3. 1inch is slow. I believe all missing prices are timeouts (>10s to retrieve a price).

### DIFF
--- a/core/src/price_estimation/clients/oneinch.rs
+++ b/core/src/price_estimation/clients/oneinch.rs
@@ -1,0 +1,61 @@
+mod api;
+
+use super::generic_client::GenericClient;
+use api::OneinchHttpApi;
+
+pub type OneinchClient = GenericClient<OneinchHttpApi>;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::http::HttpFactory;
+    use crate::models::TokenId;
+    use crate::price_estimation::price_source::PriceSource;
+    use crate::token_info::{hardcoded::TokenData, TokenBaseInfo};
+    use crate::util::FutureWaitExt as _;
+    use std::sync::Arc;
+
+    // Run with `cargo test online_oneinch_client -- --ignored --nocapture`.
+    #[test]
+    #[ignore]
+    fn online_oneinch_client() {
+        use std::time::Instant;
+
+        let tokens = hash_map! {
+            TokenId(1) => TokenBaseInfo::new("WETH", 18, 0),
+            TokenId(2) => TokenBaseInfo::new("USDT", 6, 0),
+            TokenId(3) => TokenBaseInfo::new("TUSD", 18, 0),
+            TokenId(4) => TokenBaseInfo::new("USDC", 6, 0),
+            TokenId(5) => TokenBaseInfo::new("PAX", 18, 0),
+            TokenId(6) => TokenBaseInfo::new("GUSD", 2, 0),
+            TokenId(7) => TokenBaseInfo::new("DAI", 18, 0),
+            TokenId(8) => TokenBaseInfo::new("sETH", 18, 0),
+            TokenId(9) => TokenBaseInfo::new("sUSD", 18, 0),
+            TokenId(15) => TokenBaseInfo::new("SNX", 18, 0)
+        };
+        let mut ids: Vec<TokenId> = tokens.keys().copied().collect();
+
+        let client = OneinchClient::new(
+            &HttpFactory::default(),
+            Arc::new(TokenData::from(tokens.clone())),
+        )
+        .unwrap();
+        let before = Instant::now();
+        let prices = client.get_prices(&ids).wait().unwrap();
+        let after = Instant::now();
+        println!(
+            "Took {} seconds to get prices.",
+            (after - before).as_secs_f64()
+        );
+
+        ids.sort();
+        for id in ids {
+            let symbol = tokens.get(&id).unwrap().symbol();
+            if let Some(price) = prices.get(&id) {
+                println!("Token {} has OWL price of {}.", symbol, price);
+            } else {
+                println!("Token {} price could not be determined.", symbol);
+            }
+        }
+    }
+}
--- a/core/src/price_estimation/clients/oneinch.rs
+++ b/core/src/price_estimation/clients/oneinch.rs
@@ -11,7 +11,7 @@ mod tests {
     use crate::http::HttpFactory;
     use crate::models::TokenId;
     use crate::price_estimation::price_source::PriceSource;
-    use crate::token_info::{hardcoded::TokenData, TokenBaseInfo};
+    use crate::token_info::hardcoded::{TokenData, TokenInfoOverride};
     use crate::util::FutureWaitExt as _;
     use std::sync::Arc;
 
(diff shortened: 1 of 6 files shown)


core/src/price_estimation/clients/oneinch/api.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+        .boxed()
+    }
+
+    fn get_price<'a>(&'a self, from: &Token, to: &Token) -> BoxFuture<'a, Result<f64>> {
+        let mut url = self.api_url.join("quote").unwrap();
+        // 1inch requires the user to specify the amount traded in atoms.
+        // We compute the price when selling one full token to avoid unavoidable rounding
+        // artifacts when selling exactly one token atom.
+        let mut one_token_from = String::from("1");
```
fleupold, line 79: nit: `let amount = 10u128.pow(from.decimals)` would be slightly nicer than pushing n "0"s

core/src/price_estimation/clients/oneinch/api.rs
```diff
@@ -0,0 +1,179 @@
...
+            Ok(token_mapping
+                .into_iter()
+                .map(|(_token_symbol, token_data)| token_data)
+                .collect())
+        }
+        .boxed()
+    }
+
+    fn get_price<'a>(&'a self, from: &Token, to: &Token) -> BoxFuture<'a, Result<f64>> {
+        let mut url = self.api_url.join("quote").unwrap();
```
fleupold, line 75: nit: I assume we also unwrap in dex.ag, but technically we could `?` here to avoid panics.

fleupold (approved):
Awesome stuff! Lots of integration tests.

core/src/price_estimation/clients/oneinch.rs
```diff
@@ -0,0 +1,61 @@
+mod api;
+
+use super::generic_client::GenericClient;
+use api::OneinchHttpApi;
+
+pub type OneinchClient = GenericClient<OneinchHttpApi>;
```
nlordell, line 6: The fruits of #1040

core/src/price_estimation/clients/oneinch/api.rs
```diff
@@ -0,0 +1,179 @@
...
+    to_token_amount: String,
+    from_token_amount: String,
+}
+
+#[derive(Debug)]
+pub struct OneinchHttpApi {
+    api_url: Url,
+    client: HttpClient,
+}
+impl OneinchHttpApi {
```
nlordell, line 35: nit: 
```suggestion

impl OneinchHttpApi {
```

core/src/price_estimation/clients/oneinch/api.rs
```diff
@@ -0,0 +1,179 @@
...
+        // We compute the price when selling one full token to avoid unavoidable rounding
+        // artifacts when selling exactly one token atom.
+        let mut one_token_from = String::from("1");
+        one_token_from.push_str(&String::from("0").repeat(from.decimals as usize));
+        {
+            // This is in its own block because we are supposed to drop `q`.
+            let mut q = url.query_pairs_mut();
+            q.append_pair("fromTokenSymbol", &from.symbol);
+            q.append_pair("toTokenSymbol", &to.symbol);
+            q.append_pair("amount", &one_token_from);
```
nlordell, line 86: nit: `query_pairs_mut` returns a `Serializer<UrlQuery>` which supports the builder pattern, so you may even be able to:
```suggestion
            url.query_pairs_mut()
                .append_pair("fromTokenSymbol", &from.symbol)
                .append_pair("toTokenSymbol", &to.symbol)
                .append_pair("amount", &one_token_from);
```

This might make it so you don't need the block since you never create a binding to the intermediate binding.

core/src/price_estimation/clients/oneinch/api.rs
```diff
@@ -0,0 +1,179 @@
...
+impl GenericToken for Token {
+    fn symbol(&self) -> &str {
+        &self.symbol
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct TradedAmounts {
+    to_token_amount: String,
```
nlordell, line 26: nit: I think that the traded amounts should be parsed at deserialization time, since `{"toTokenAmoun":"blabla"}` should fail to deserialize.
```suggestion
    #[serde(with = "display_fromstr")]
    to_token_amount: u128,
```

This also requires:
```rust
use serde_with::rust::display_fromstr;
```

nlordell (approved):
Nice! Just a couple of minor comments.

core/src/price_estimation/clients/oneinch/api.rs
fedgiac: Very true! I forgot about `u128`.

core/src/price_estimation/clients/oneinch/api.rs
fedgiac: Actually URL management is a bit different in 1inch, so we don't need to unwrap there.

I can't just use `?` because I need to return a `BoxFuture`, and I can't wrap everything inside `async move` because of lifetime issues with `from` and `to`. I could clone `from.symbol` and `to.symbol` and then move that, but it's more code and I think it's actually better if the program fails when failing to create an url when joining two compile-time constants.

core/src/price_estimation/clients/oneinch/api.rs
fedgiac: Nice! I also changed it in dex.ag.

core/src/price_estimation/clients/oneinch/api.rs
fedgiac: Done! I parsed it as `f64` instead of `u128` since there is no `.into()` or any easy conversion I could find from `u128` to `f64`.

core/src/price_estimation/clients/oneinch/api.rs
nlordell: I would suggest moving the URL computation into the `async { ... }` block so that you can use `?`.

core/src/price_estimation/clients/oneinch/api.rs
nlordell: You should be able to do a numeric cast with `as`: `100u128 as f64`.

nlordell: I think it would be *slightly* more correct to parse into `u128` since the values are atoms, which would avoid rounding errors for large numbers (although to be fair, there would be rounding errors when computing the exchange rate, so ultimately it makes no difference).

core/src/price_estimation/clients/oneinch/api.rs
nlordell: Hmm, maybe you could do:
```
let mut url = self.api_url.join("quote").and_then(move |url| {
    // use `to` and `from` to compute the URL here...
});
```

And use `url?` inside the `async { ... }` block?

The reason I suggest this is that the API `with_url` can be created with a caller specified URL which might error when `join`-ing `"quote"`.

fedgiac: It makes sense, and I like this solution!

core/src/price_estimation/clients/oneinch/api.rs
```diff
@@ -0,0 +1,183 @@
...
+impl Api for OneinchHttpApi {
+    type Token = Token;
+
+    fn bind(http_factory: &HttpFactory) -> Result<Self> {
+        Self::with_url(http_factory, DEFAULT_API_URL)
+    }
+
+    fn get_token_list<'a>(&'a self) -> BoxFuture<'a, Result<Vec<Token>>> {
+        async move {
+            let url = self.api_url.join("tokens").unwrap();
```
nlordell, line 64:
```suggestion
            let url = self.api_url.join("tokens")?;
```

e00E (approved):

PR #1078, opened 2020-07-03
Remove driver docker container special case for driver compilation

e00E:
I don't think anyone was using this because it has been broken since the
price estimator has been introduced. The "quick compilation" option
through volume mounts has been broken for longer.
I don't see why this is useful anyway when you can use `cargo run` instead.

While making this I also noticed that the readme needs updating in a few places but I won't make that part of this PR.

### Test Plan
Check that CI still works, check that running e2e tests works locally


nlordell (approved):
Trimmed down! My only concern is that running the driver with the solver on a Mac is a little hard now.

fleupold (approved):
I share you concern @nlordell, but I think in that case it might make sense to create a new Dockerfile for that instead of making this one as complicated as it was before.

nlordell:
>  I think in that case it might make sense to create a new Dockerfile for that instead of making this one as complicated as it was before.

I think that is a great idea. 

Just to clarify I gave this PR the big old :+1: because I also believe that the docker file was more complicated than it should be. I like this change, and just wanted to note that it may be a bit more complicated to build and run a driver with a solver locally on macs right now. That being said, I really like your point, that this should probably be done with a separate dockerfile instead of making a really complicated one like we have today.

bh2smith (approved):
I've got my fingers crossed it wont be too difficult to run on my precious mac in the future. If the test plan directly contained some of the updated README content, I would try this myself.

PR #1079, opened 2020-07-03
Move driver related docker files into driver folder

e00E:
Now that we have several top level projects it makes sense to keep them
in their own folders like the price estimator already does.

### Test Plan
CI and tested locally that e2e still works.


.dockerignore
```diff
@@ -1,13 +1,8 @@
...
-README.md
-LICENSE
-.vscode
 !target/debug/driver
 !target/release/driver
+dex-contracts
+!dex-contracts/build
+**/node_modules
+.git
+.vscode
```
e00E, line 8: I cleaned this up because the point of the dockerignore file is to make docker have to transfer less data to the docker daemon that builds container. For example if we didn't exclude target it would have to send the whole target folder to the docker daemon which can be quite large.
For small files/folders this doesn't matter unless we were copying the whole root folder into the container which are not doing anywhere.

driver/docker-compose.yml
```diff
@@ -11,17 +11,16 @@ services:
   truffle:
```
nlordell, line 11: I imagine it makes sense for the docker compose file to stay in the root of the repository. Its imaginable that we want the docker compose file to start up the driver and the price estimator together for example.

driver/docker-compose.yml
```diff
@@ -11,17 +11,16 @@ services:
...
     command: /app/run.sh
     build:
-      context: .
-      dockerfile: docker/truffle/Dockerfile
+      context: docker/truffle
```
nlordell, line 14: nit: Personally I would prefer that the context just be the root of the project. My main argument is that then all paths are relative to the root of the project and no path math is needed.

nlordell (commented):
To me it makes sense for the `docker-compose.yaml` to stay in the root of the project, while the driver-specific Docker files move to the driver directory. My main argument there is that the Docker compose file is to orchestrate multiple services and I can imagine a scenario where we want to start the driver and the price estimator with `docker-compose up`.

driver/docker-compose.yml
e00E: Yeah I'm unsure about this. I don't mind either way. Since the truffle docker doesn't need any file from anywhere else I did this which has the advantage of  staying correct even when /driver/docker/truffle moves. As I said I know what's better so let's see what the second reviewer thinks.

driver/docker-compose.yml
e00E: That's reasonable if/when we want to do that but until then it makes more sense in the driver folder to me.

driver/docker-compose.yml
nlordell: > which has the advantage of staying correct even when /driver/docker/truffle moves

Thats a good point

> Yeah I'm unsure about this

> As I said I know what's better

:rofl:

e00E: Added `don't` :)

nlordell: Hehe, I figured that's what you meant. I just thought it was funny :stuck_out_tongue_closed_eyes:

bh2smith (approved):
Looks good. Can we ensure there are no changes needed in the README! That is, specifically that `docker-compose up` still works from the project root.

driver/docker-compose.yml
nlordell: The other argument is that `docker-compose up stablex` no longer works from the root (akin to `cargo run -p driver` for example). To me the Docker compose file is really for orchestrating multiple docker images throughout the repository (kind of how a Cargo workspace aggregates multiple crates).

e00E: I agree with this now. Moving this file back to the root.

fleupold (approved):
Looks good. I think `tini` shouldn't be driver specific and we should probably use it in the price estmiator as well.

nlordell:
> and we should probably use it in the price estmiator as well.

We do, its installed via the Alpine Linux package manager :smile:

README.md
```diff
@@ -274,5 +274,5 @@ docker-compose build
 In order to start BatchExchange for the Rinkeby network, make sure that the env variables in common-rinkeby.env are up to date and then start the specific docker:
 
 ```
-docker-compose -f docker-compose.yml -f docker-compose.rinkeby.yml up stablex
+docker-compose -f driver/docker-compose.rinkeby.yml up stablex
```
nlordell, line 277: I think this needs to be
```suggestion
docker-compose -f docker-compose.yml -f driver/docker-compose.rinkeby.yml up stablex
```

nlordell (approved):
Cool! And I like the docker compose overrides in the `driver` folder.

PR #1080, opened 2020-07-03
[Easy] Fix PriceSource priority order

fleupold:
https://github.com/gnosis/dex-services/pull/1070 made the price estimation use the PriorityPrice source with TokenData being the primary source and the average source being the secondary.

However, this was an undesired change in logic. I misread in particular this part of the code https://github.com/gnosis/dex-services/pull/1070/files#r449536863

This PR reverts the change, making the average source primary and the overrides secondary.

### Test Plan
Run driver locally and see price for WETH is no longer hardcoded at $250 but instead around $247.


josojo (approved):
Thanks! for the quick fix

e00E (approved):

twalth3r (approved):
Thanks a lot @fleupold for this quick fix!

Issue #825, opened 2020-05-03
[GasEstimation] Don't double gas price if system time is past cutoff time for accepted solutions


fleupold:
We currently double the gas price up to a maximum of 60 GWEI every two blocks regardless of what the current time is. 
In case we already passed the four minute mark up to which solutions are accepted we should probably not double the gas price but instead try to cancel the tx by sending a no-op tx with a slighty higher gas price to clear the tx queue.

https://github.com/gnosis/dex-services/blob/6f794300512e6332e1a80b4ad3ad6700f89fccf1/driver/src/solution_submission/mod.rs#L178-L190

bh2smith:
Cool idea! I like the cancellation proposal!

fleupold:
As described in https://github.com/gnosis/dex-services/issues/853 we should also not send the tx again if we didn't increase the gas price (due to the cap) as this leads to errors:

> 'submitSolution(uint32,uint256,address[],uint16[],uint128[],uint128[],uint16[]):(uint256)' failure: web3 error: RPC error: Error { code: ServerError(-32010), message: "Transaction gas price supplied is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.", data: None }

nlordell:
Regarding #853, maybe there was something else going on as well, since if the last gas price is the gas cap, then we do not resubmit the solution.

I checked the logs and it looks like the solution was submitted twice, once with a gas price of 57500000001 and the second (and final) submission was with a gas price of 60000000000 (which is the gas cap).

fleupold:
You are right @nlordell that this should not be an issue in our code. I missed the break condition. Let's discuss the actual root cause on the original issue then.

fleupold:
Related task that I will close in favor of this one:

> This is regarding the following error:
https://gnosisinc.slack.com/archives/CQZBU9T5J/p1589080051000100
If the gas_cap has been reached, then we don't increase the nonce or the gas price anymore and we will get this error message.
Hence, the code should check whether we had already increased the gas price to gas_cap already beforehand, and if this happened, then we should not try to resubmit

In fact the code should also check that the increase >12.5% as otherwise parity doesn't accept the a transaction with the same nonce.

PR #1085, opened 2020-07-06
Make sure gas price is increased by at least minimum factor

e00E:
Openethereum requires that a resubmitted transaction increases it's gas
price by at least 12.5%.

Part of #825 .

### Test Plan
New unit test in CI


fleupold (approved):
👍

core/src/solution_submission.rs
```diff
@@ -185,6 +185,10 @@ async fn retry_with_gas_price_increase(
     const INCREASE_FACTOR: u32 = 2;
     const BLOCK_TIMEOUT: usize = 2;
     const DEFAULT_GAS_PRICE: u64 = 15_000_000_000;
+    // openethereum requires that the gas price of the resubmitted transaction has increased by at
+    // least 12.5%. Rounded to 13% here in case of floating point error or in case the bound becomes
+    // exclusive.
+    const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.13;
```
nlordell, line 191: Why is the rounding buffer so big? Why not just:
```suggestion
    const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125000001;
```

Or even:
```suggestion
    const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
```

core/src/solution_submission.rs
```diff
@@ -332,6 +338,50 @@ mod tests {
         .unwrap();
     }
 
+    #[test]
+    fn test_retry_with_gas_price_respects_minimum_increase() {
```
nlordell, line 342: Nice unit test :tada:

nlordell (approved):
Looks good. Just one comment about the factor constant. I think you don't need such a big rounding buffer.

core/src/solution_submission.rs
e00E: Why is it so big? The exact value here doesn't really matter (1.125 vs 1.3) and I found this more obvious.
On the other hand I don't see anything wrong with the epsilon and I realize now there should probably also be a `ceil` in there so I'll change it.

e00E: Changed it to calculate the effective bound once at the start instead of at every iteration and to use epsilon.

nlordell: > The exact value here doesn't really matter (1.125 vs 1.3)

I guess its not super important, it might just cause 1 fewer submission to happen because of the gas price cap (if the next gas price is `previous_gas_price * 1.125 < next_gas_price < previous_gas_price * 1.13`.

core/src/solution_submission.rs
```diff
@@ -219,8 +225,9 @@ async fn retry_with_gas_price_increase(
             )
             .await;
 
+        let gas_price_can_still_increase = gas_price < effective_gas_cap;
```
nlordell, line 228: very cleaver!

nlordell (approved):
Like it!

core/src/solution_submission.rs
bh2smith: Clever, but what does it bring to the table apart from readability? I see it is only used in one place...

e00E: I like this because it gives the condition a name. It makes it clear what `gas_price < effective_gas_cap` means on a higher level than just what it does.

core/src/solution_submission.rs
```diff
@@ -186,6 +186,12 @@ async fn retry_with_gas_price_increase(
     const BLOCK_TIMEOUT: usize = 2;
     const DEFAULT_GAS_PRICE: u64 = 15_000_000_000;
 
+    // openethereum requires that the gas price of the resubmitted transaction has increased by at
+    // least 12.5%.
+    let min_gas_price_increase_factor: f64 = 1.125 * (1.0 + f64::EPSILON);
```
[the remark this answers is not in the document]
e00E: Epsilon is always > 0 afaik since it is related to the error floating point operations have.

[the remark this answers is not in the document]
nlordell: `EPSILON` is a Rust standard library constant that is:
> This is the difference between 1.0 and the next larger representable number.

So basically, because of how double precision floating point numbers work, there are no values that can be represented between `1.125` and `1.125 * (1.0 + f64::EPSILON) == 1.125_000_000_000_000_2`. If you try to write express any number between those two, it will be either rounded up or down. Specifically:
```
println!("{}", 1.125_000_000_000_000_05 == 1.125);                          // true
println!("{}", 1.125_000_000_000_000_1 == 1.125);                           // true
println!("{}", 1.125_000_000_000_000_15 == (1.125 * (1.0 + f64::EPSILON))); // true
```

PR #1089, opened 2020-07-06
Cancel transaction when solution submission time expires

### LINKED ISSUE #821: Solution from previous batch clog solutions for next batch
> ERROR (pod: dev-dfusion-standard-solver-mainnet-master-68bcc79797-2g5r8):
Apr 30 10:05:56.824 ERRO [driver::driver::scheduler::system] Batch 5294136 failed with unretryable error: method 'submitSolution(uint32,uint256,address[],uint16[],uint128[],uint128[],uint16[]):(uint256)' failure: web3 error: RPC error: Error { code: ServerError(-32010), message: "Transaction gas price supplied is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.", data: None }

I am not (yet) sure what this means. I suspect it could be related to the gas price increases we recently implemented.

### LINKED ISSUE #825: [GasEstimation] Don't double gas price if system time is past cutoff time for accepted solutions

e00E:
There is no reason to continue to try submitting the solution with
higher gas prices when the deadline is already over.
With this commit we cancel the transaction and exit when this happens.

Fixes #825 
Fixes #821 

### Test Plan
Added unit test for this. Wasn't able to test the full behavior with ganache but I did test that sending the noop transaction works in ganache.

### DIFF
--- a/core/src/contracts/stablex_contract.rs
+++ b/core/src/contracts/stablex_contract.rs
@@ -7,11 +7,11 @@ use crate::{
     contracts,
     models::{ExecutedOrder, Solution},
 };
-use anyhow::{Error, Result};
+use anyhow::{anyhow, Context as _, Error, Result};
 use ethcontract::{
     contract::Event,
     errors::{ExecutionError, MethodError},
-    transaction::{confirm::ConfirmParams, GasPrice, ResolveCondition},
+    transaction::{confirm::ConfirmParams, Account, GasPrice, ResolveCondition, TransactionResult},
     Address, BlockNumber, PrivateKey, U256,
 };
 use futures::{
@@ -50,13 +50,8 @@ impl StableXContractImpl {
         Ok(StableXContractImpl { instance, viewer })
     }
 
-    pub fn account(&self) -> Address {
-        self.instance
-            .defaults()
-            .from
-            .as_ref()
-            .map(|from| from.address())
-            .unwrap_or_default()
+    pub fn account(&self) -> Option<Account> {
+        self.instance.defaults().from.clone()
     }
 
     pub fn address(&self) -> Address {
@@ -148,6 +143,13 @@ pub trait StableXContract {
     fn stream_events<'a>(
         &'a self,
     ) -> BoxStream<'a, Result<Event<batch_exchange::Event>, ExecutionError>>;
+
+    /// Create a noop transaction. Useful to cancel a previous transaction that is stuck due to
+    /// low gas price.
+    fn send_noop_transaction<'a>(
+        &'a self,
+        gas_price: U256,
+    ) -> BoxFuture<'a, Result<TransactionResult>>;
 }
 
 impl StableXContract for StableXContractImpl {
@@ -323,6 +325,30 @@ impl StableXContract for StableXContractImpl {
             .from_block(BlockNumber::Latest)
             .stream()
             .boxed()
+    }
+
+    fn send_noop_transaction<'a>(
+        &'a self,
+        gas_price: U256,
+    ) -> BoxFuture<'a, Result<TransactionResult>> {
+        async move {
+            let web3 = self.instance.raw_instance().web3();
+            let account = self.account().ok_or_else(|| anyhow!("no account"))?;
+            let address = account.address();
+            let transaction = ethcontract::transaction::TransactionBuilder::new(web3)
+                .from(account)
+                .to(address)
+                .gas(U256::from(21000))
+                .gas_price(GasPrice::Value(gas_price))
+                .value(U256::zero())
+                .resolve(ResolveCondition::Confirmed(ConfirmParams::mined()));
+            let transaction_result = transaction
+                .send()
+                .await
+                .context("failed to send noop transaction")?;
+            Ok(transaction_result)
+        }
+        .boxed()
     }
 }
 
--- a/core/src/driver/scheduler.rs
+++ b/core/src/driver/scheduler.rs
@@ -5,12 +5,10 @@ use self::evm::EvmScheduler;
 use self::system::SystemScheduler;
 use crate::contracts::stablex_contract::StableXContract;
 use crate::driver::stablex_driver::StableXDriver;
+use crate::models::batch_id::SOLVING_WINDOW;
 use anyhow::{anyhow, Error, Result};
 use std::str::FromStr;
 use std::time::Duration;
-
-/// The time in a batch where a solution may be submitted.
-const SOLVING_WINDOW: Duration = Duration::from_secs(240);
 
 /// A scheduler that can be started in order to run the driver for each batch.
 pub trait Scheduler {
--- a/core/src/models/batch_id.rs
+++ b/core/src/models/batch_id.rs
@@ -4,6 +4,8 @@ use std::time::{Duration, SystemTime, SystemTimeError};
 
 /// The total time in a batch.
 pub const BATCH_DURATION: Duration = Duration::from_secs(300);
+/// The time in a batch where a solution may be submitted.
+pub const SOLVING_WINDOW: Duration = Duration::from_secs(240);
 
 /// Wraps a batch id as in the smart contract to add functionality related to
 /// the current time.
@@ -44,6 +46,10 @@ impl BatchId {
 
     pub fn solve_start_time(self) -> SystemTime {
         self.order_collection_start_time() + BATCH_DURATION
+    }
+
+    pub fn solve_end_time(self) -> SystemTime {
+        self.solve_start_time() + SOLVING_WINDOW
     }
 
     pub fn next(self) -> BatchId {
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -2,9 +2,9 @@
 
 use crate::contracts::stablex_contract::StableXContract;
 use crate::gas_station::GasPriceEstimating;
-use crate::models::Solution;
+use crate::models::{BatchId, Solution};
 
-use anyhow::{Error, Result};
+use anyhow::{anyhow, Error, Result};
 use ethcontract::errors::{ExecutionError, MethodError};
 use ethcontract::web3::types::TransactionReceipt;
 use ethcontract::U256;
@@ -12,7 +12,7 @@ use futures::future::{BoxFuture, FutureExt as _};
 use log::info;
 #[cfg(test)]
 use mockall::automock;
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 use thiserror::Error;
 
 /// The amount of time the solution submitter should wait between polling the
@@ -22,6 +22,10 @@ use thiserror::Error;
 const POLL_TIMEOUT: Duration = Duration::from_secs(5);
 #[cfg(test)]
 const POLL_TIMEOUT: Duration = Duration::from_secs(0);
+
+// openethereum requires that the gas price of the resubmitted transaction has increased by at
+// least 12.5%.
+const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
 
 #[cfg_attr(test, automock)]
 pub trait StableXSolutionSubmitting {
@@ -155,19 +159,48 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
         solution: Solution,
         claimed_objective_value: U256,
     ) -> BoxFuture<Result<(), SolutionSubmissionError>> {
+        const GAS_PRICE_CAP: u64 = 60_000_000_000;
         async move {
-            match retry_with_gas_price_increase(
+            let submit_future = retry_with_gas_price_increase(
                 self.contract,
                 batch_index,
                 solution.clone(),
                 claimed_objective_value,
                 self.gas_price_estimating,
-                60_000_000_000u64.into(),
-            )
-            .await
-            {
-                Ok(()) => Ok(()),
-                Err(err) => Err(self.make_error(batch_index, solution, err).await),
+                GAS_PRICE_CAP.into(),
+            );
+            // Add some extra time in case of desync between real time and ethereum node current block time.
+            let deadline = BatchId::from(batch_index).solve_end_time() + Duration::from_secs(30);
+            let remaining = deadline
+                .duration_since(SystemTime::now())
+                .unwrap_or(Duration::from_secs(0));
+            if let Ok(submit_result) = async_std::future::timeout(remaining, submit_future).await {
+                match submit_result {
+                    Ok(()) => Ok(()),
+                    Err(err) => Err(self.make_error(batch_index, solution, err).await),
+                }
+            } else {
+                let gas_price = U256::from(
+                    (GAS_PRICE_CAP as f64 * MIN_GAS_PRICE_INCREASE_FACTOR).ceil() as u128,
+                );
+                log::info!(
+                    "cancelling transaction because it took too long, using gas price {}",
+                    gas_price
+                );
+                match self.contract.send_noop_transaction(gas_price).await {
+                    Ok(_) => log::info!(
+                        "cancelled solution submission of batch {} because of deadline",
+                        batch_index
+                    ),
+                    Err(err) => log::error!(
+                        "failed to cancel solution submission of batch {} after deadline: {:?}",
+                        batch_index,
+                        err
+                    ),
+                }
+                Err(SolutionSubmissionError::Unexpected(anyhow!(
+                    "solution submission transaction not confirmed in time"
+                )))
             }
         }
         .boxed()
@@ -232,9 +265,6 @@ async fn retry_with_gas_price_increase(
 ) -> Result<(), MethodError> {
     const BLOCK_TIMEOUT: usize = 2;
     const DEFAULT_GAS_PRICE: u64 = 15_000_000_000;
-    // openethereum requires that the gas price of the resubmitted transaction has increased by at
-    // least 12.5%.
-    const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
 
     let effective_gas_price_cap = U256::from(
         (gas_price_cap.as_u128() as f64 / MIN_GAS_PRICE_INCREASE_FACTOR).floor() as u128,
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -1,10 +1,13 @@
 #![allow(clippy::ptr_arg)] // required for automock
 
-use crate::contracts::stablex_contract::StableXContract;
-use crate::gas_station::GasPriceEstimating;
-use crate::models::{BatchId, Solution};
+use crate::{
+    contracts::stablex_contract::StableXContract,
+    gas_station::GasPriceEstimating,
+    models::{BatchId, Solution},
+};
 
 use anyhow::{anyhow, Error, Result};
+use async_std::future::TimeoutError;
 use ethcontract::errors::{ExecutionError, MethodError};
 use ethcontract::web3::types::TransactionReceipt;
 use ethcontract::U256;
@@ -22,6 +25,8 @@ use thiserror::Error;
 const POLL_TIMEOUT: Duration = Duration::from_secs(5);
 #[cfg(test)]
 const POLL_TIMEOUT: Duration = Duration::from_secs(0);
+
+const GAS_PRICE_CAP: u64 = 60_000_000_000;
 
 // openethereum requires that the gas price of the resubmitted transaction has increased by at
 // least 12.5%.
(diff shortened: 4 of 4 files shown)


core/src/contracts/stablex_contract.rs
```diff
@@ -324,6 +323,36 @@ impl StableXContract for StableXContractImpl {
...
+                .compat()
+                .await
+                .context("failed to get nonce")?;
+            let transaction_result = ethcontract::contract::MethodBuilder::fallback(
+                web3.clone(),
+                address,
+                Default::default(),
+            )
+            .from(account)
+            .gas(U256::from(21000))
```
e00E, line 345: Is this the right gas amount?

core/src/solution_submission.rs
```diff
@@ -9,10 +9,9 @@ use ethcontract::errors::{ExecutionError, MethodError};
 use ethcontract::web3::types::TransactionReceipt;
 use ethcontract::U256;
 use futures::future::{BoxFuture, FutureExt as _};
-use log::info;
```
e00E, line 12: This file had mixed uses of logging macros `log::info!` vs `info!`. I prefer having the module name to see where this is from especially for `log::error!`.

core/src/contracts/stablex_contract.rs
```diff
@@ -324,6 +323,33 @@ impl StableXContract for StableXContractImpl {
...
+            let nonce = web3
+                .eth()
+                .transaction_count(address, None)
+                .compat()
+                .await
+                .context("failed to get nonce")?;
+            let transaction_result =
+                ethcontract::contract::MethodBuilder::fallback(web3, address, Default::default())
+                    .from(account)
+                    .gas(U256::from(21000))
```
e00E, line 342: Is this the right gas amount?

core/src/contracts/stablex_contract.rs
nlordell: AFAIK that is the base transaction gas amount.

core/src/solution_submission.rs
nlordell: I also prefer this for the log macros. I think a lot of the `use log::{info, error}` come from the migration from edition 2015 to 2018, where the `#[macro_use] extern crate log;` was removed.

core/src/contracts/stablex_contract.rs
nlordell: AFAIK that is the base transaction gas amount.

core/src/solution_submission.rs
```diff
@@ -215,29 +222,60 @@ async fn retry_with_gas_price_increase(
...
+                        gas_price_cap
+                    } else {
+                        U256::from(
+                            (gas_price.as_u128() as f64 * MIN_GAS_PRICE_INCREASE_FACTOR) as u128,
+                        )
+                    };
+                    log::info!(
+                        "trying to cancel solution submission transaction for batch {}",
+                        batch_index
+                    );
```
e00E, line 255: More logging than necessary maybe but I'd like to have this until we see this working correctly at least once.

core/src/contracts/stablex_contract.rs
```diff
@@ -324,6 +323,35 @@ impl StableXContract for StableXContractImpl {
...
             .boxed()
     }
+
+    fn noop_transaction<'a>(&'a self, gas_price: U256) -> BoxFuture<'a, Result<TransactionResult>> {
+        use futures::compat::Future01CompatExt as _;
+        async move {
+            let web3 = self.instance.raw_instance().web3();
+            let account = self.account().ok_or_else(|| anyhow!("no account"))?;
+            let address = account.address();
+            let nonce = web3
```
nlordell, line 333: This shouldn't be needed, this is how `ethcontract` determines the nonce.

core/src/solution_submission.rs
```diff
@@ -215,29 +222,60 @@ async fn retry_with_gas_price_increase(
...
+                    // `contract.noop_transaction` gets the current nonce from the node which is
+                    // technically a race condition. It would be better to retrieve the nonce from
+                    // the `submit_solution` transaction but I couldn't find a way to do that.
```
nlordell, line 258: This can be done a couple of ways:
1. Retrieve the transaction data with the transaction hash: https://docs.rs/web3/0.11.0/web3/api/struct.Eth.html#method.transaction. Unfortunately, the transaction hash is not super easy to determine right now and requires first `TransactionBuilder::build`-ing the transaction and using `web3` to send it. I created gnosis/ethcontract-rs#345 to address the issue.
2. Manually setting the nonce for the submit solution transaction by as you do in the `noop_transaction`, and then returning along side the future.

core/src/solution_submission.rs
```diff
@@ -215,29 +222,60 @@ async fn retry_with_gas_price_increase(
...
         // Breaking condition for our loop
-        gas_price_can_still_increase
-            && matches!(
-                result,
-                Err(MethodError {
-                    inner: ExecutionError::ConfirmTimeout,
-                    ..
-                })
-            )
+        match result {
```
nlordell, line 237: I find this match statement a little hard to follow and it feels a little deeply nested. Maybe it can be refactored a bit? (PS. I have no good ideas on how to do this).

core/src/solution_submission.rs
e00E: Let's keep it this way until the ethcontract PR is available.

core/src/solution_submission.rs
nlordell: I think this logging is fine. Its not like this path should happen often, and when it does its nice to have lots of logs in order to determine the exact sequence that was executed.

If this were a common path that would be hit multiple times, then maybe the extra log messages could pollute the logs a bit, but I don't think its the case here.

core/src/solution_submission.rs
e00E: Agree. Tried this a bit. I think it ends up changing the whole function. In that case I'd like keep this PR as is and make another one to refactor the function.

nlordell: Now that I think about it, maybe the function can be refactored in this PR already? Seeing as it is already changing significantly to include the `noop_transaction` to cancel the solution submission?

e00E: For the reviewers and when looking at individual commits later it is cleaner to have this separate since the refactoring will change more things than just adding this functionality.

fleupold: I cannot follow the code in this match statement (I wasn't even able to properly understand a much simpler flavor [here](https://github.com/gnosis/dex-services/pull/1070#discussion_r449536863)). 

Would it be possible to refactor before adding logic? Maybe we can have a method that returns an `enum SolutionSubmissionAction` and change the loop into something more straight-forward such as:

```rust
loop {
  match next_action(previous_gas_price, gas_price_estimating) {
     SolutionSubmissionAction::Resubmit(gas_price) => { ... }
     SolutionSubmissionAction::LeaveInMemPool => break
  }
}
```

and then in this PR change `LeaveInMemPool` to `CancelTx`. `next_action` could have some unit tests describing what action we expect based on the previous gas price, current estimate and time passed.

core/src/solution_submission.rs
```diff
@@ -358,6 +397,10 @@ mod tests {
             });
         // There should not be a second call to submit_solution because 90 to 100 is not a large
         // enough gas price increase.
+        contract
```
fleupold, line 400: This test doesn't really match with the comment above. I'm probably missing something, but don't we have to adjust the time somehow to trigger the noop transaction?

core/src/solution_submission.rs
fleupold: The same race condition applies for increasing the gas price for solution submission.

core/src/solution_submission.rs
```diff
@@ -217,28 +225,59 @@ async fn retry_with_gas_price_increase(
...
+                } else {
+                    let gas_price = if gas_price_can_still_increase {
+                        gas_price_cap
+                    } else {
+                        U256::from(
+                            (gas_price.as_u128() as f64 * min_gas_price_increase_factor).ceil()
+                                as u128,
+                        )
+                    };
+                    log::info!(
```
fleupold, line 254: Are we cancelling the tx if the gas price can no longer increase? I think in that case we should hope that the current gas price is high enough to complete before the deadline and only cancel after the deadling.

fleupold (changes requested):
My main concern is that the deadline seems to be the one from the solver (3:30) and not the one from the smart contract (4minutes) and that I think we cancel orders even before the deadline arises in case gas becomes too expensive (instead we should leave it in the mempool until the deadline in that case).

It would also be great if we can refactor the do/match statement beforehand to make reasoning about the changes in this PR easier.

core/src/solution_submission.rs
nlordell: In general though, I think it would be better to ensure the same nonce that was used by the `submit_solution` transaction is used for the cancellation transaction.

core/src/solution_submission.rs
nlordell: I would also be in favour of a refactor before the change, as this loop has become a bit unruly and would be nice to split up a bit. I like @fleupold's suggestion of a method that tries to submit a solution and returns an enum that can be used to determine what to do next.

e00E: I'll do the refactor first in a separate PR then rebase this PR.

core/src/contracts/stablex_contract.rs
```diff
@@ -324,6 +323,27 @@ impl StableXContract for StableXContractImpl {
             .stream()
             .boxed()
     }
+
+    fn noop_transaction<'a>(&'a self, gas_price: U256) -> BoxFuture<'a, Result<TransactionResult>> {
```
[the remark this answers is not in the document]
nlordell: I'm not a huge fan of `cancel_pending_transaction`, especially since there are no parameter relating to a transaction. I think once we include gnosis/ethcontract-rs#345 then `fn cancel_pending_transaction(self, tx: TransactionResult)` makes a lot more sense, since we would be able to recover the tx nonce and gas price to send a proper transaction cancellation.

Until then `send_noop_transaction` feels better for me. Although, I like using an action verb like `send` to indicate its actually affecting the EVM.

core/src/driver/scheduler.rs
```diff
@@ -5,13 +5,11 @@ use self::evm::EvmScheduler;
 use self::system::SystemScheduler;
 use crate::contracts::stablex_contract::StableXContract;
 use crate::driver::stablex_driver::StableXDriver;
+use crate::models::batch_id::SOLVING_WINDOW;
 use anyhow::{anyhow, Error, Result};
 use std::str::FromStr;
 use std::time::Duration;
 
-/// The time in a batch where a solution may be submitted.
-const SOLVING_WINDOW: Duration = Duration::from_secs(240);
```
fleupold, line 13: nice find!

core/src/solution_submission.rs
```diff
@@ -229,6 +232,24 @@ async fn retry_with_gas_price_increase(
         )
     };
 
+    let wait_for_deadline = || async {
```
fleupold, line 235: Very clever!

core/src/solution_submission.rs
```diff
@@ -238,15 +259,40 @@ async fn retry_with_gas_price_increase(
...
+                    gas_price_factor,
+                );
+            }
+            Either::Right(_) => {
+                let gas_price = U256::from(
+                    (gas_price.as_u128() as f64 * MIN_GAS_PRICE_INCREASE_FACTOR).ceil() as u128,
+                );
+                match contract.send_noop_transaction(gas_price).await {
+                    Ok(_) => log::info!(
+                        "cancel solution submission of batch {} because of deadline",
```
fleupold, line 282: ubernit: cancelled

core/src/solution_submission.rs
```diff
@@ -229,6 +232,24 @@ async fn retry_with_gas_price_increase(
...
+        {
+            // Add some extra time because of desync between real current time and ethereum current
+            // block time.
+            Ok(remaining) => async_std::task::sleep(remaining + Duration::from_secs(30)).await,
+            Err(err) => {
+                log::warn!(
+                    "could not calculate remaining time to solve batch: {:?}",
+                    err
+                );
+                future::pending().await
```
fleupold, line 248: Should this rather resolve immediately? Could this happen if we increased the gas limit with 15s remaining and so the next time we try to compute the remaining time it is negative?

core/src/solution_submission.rs
```diff
@@ -238,15 +259,40 @@ async fn retry_with_gas_price_increase(
...
+                        batch_index
+                    ),
+                    Err(err) => log::error!(
+                        "failed to cancel solution submission of batch {} after deadline: {:?}",
+                        batch_index,
+                        err
+                    ),
+                }
+                // TODO: What to return? Should we make up a method error? Or we could make this
+                // function return Result<(), Either<MethodError, DeadlineReached>> .
```
fleupold, line 292: Equivalent to your suggestion, I think the cleanest would be our own SolutionSubmissionError enum which can itself be either MethodError or DeadlineReached

core/src/solution_submission.rs
```diff
@@ -238,15 +259,40 @@ async fn retry_with_gas_price_increase(
         } else {
             Some(BLOCK_TIMEOUT)
         };
-        let result = submit_solution(gas_price, block_timeout).await;
-        if !is_confirm_timeout(&result) || is_last_iteration {
-            return result;
+        let solution_future = submit_solution(gas_price, block_timeout);
+        let deadline_future = wait_for_deadline();
+        futures::pin_mut!(deadline_future);
```
fleupold, line 264: Why is this needed?

core/src/solution_submission.rs
```diff
@@ -238,15 +259,40 @@ async fn retry_with_gas_price_increase(
...
+        match future::select(solution_future, deadline_future).await {
+            Either::Left((result, _)) => {
+                if !is_confirm_timeout(&result) || is_last_iteration {
+                    return result;
```
fleupold, line 268: Before, we were cancelling as soon as the gas price reached the max. If I read the code correctly, we are now not cancelling at all if the gas price became too expensive. I think we should still cancel in that case, but only after the deadline.

fleupold (commented):
Very clever use of `future::select`!

Could you add a few unit test for the new behaviour (specifically around how gas price becoming too expensive and running over time is handled)? 
I can see how this code can be hard to test but given it has a lot of complex logic I think it's really important.

I would think that having a function that computes the next action could make it slightly simpler to test. If need be we can only check for cancellation after each run of submit solution (since we only wait 2 blocks for confirmation) but I do like your parallel solution.

core/src/solution_submission.rs
e00E: `select` requires the future to be pinned (or unpin). Afaik this is the normal way to accomplish this. It's not needed for `solution_future` because that future happens to already be unpin.

core/src/solution_submission.rs
e00E: My reasoning is that in the last iteration the solution_future should never return a confirm timeout because for the last iteration we set the block timeout to `None`. So in the last iteration we will either hit the deadline or not get a timeout on the solution.
This isn't clear in the code though so I should add a comment.

core/src/solution_submission.rs
e00E: Good point.

core/src/solution_submission.rs
```diff
@@ -108,20 +107,31 @@ impl<'a> StableXSolutionSubmitter<'a> {
...
+                            .await
+                        {
+                            return SolutionSubmissionError::from(err);
+                        }
+                    }
                 }
+                SolutionSubmissionError::Unexpected(err.into())
+            }
+            RetryError::TransactionNotConfirmedInTime => {
+                SolutionSubmissionError::Benign("transaction not confirmed in time".to_string())
```
e00E, line 132: Should this be a benign error?

core/src/solution_submission.rs
```diff
@@ -244,36 +272,86 @@ async fn retry_with_gas_price_increase(
...
+    IncreaseGasPrice,
+    Cancel,
+    Done(std::result::Result<(), RetryError>),
+}
+
+// Either::Left is the solution submission future, Either::Right the deadline future.
+fn determine_retry_action<T0, T1>(
+    either: Either<(Result<(), MethodError>, T0), ((), T1)>,
+    is_last_iteration: bool,
+) -> RetryAction {
```
e00E, line 342: I know you suggested a test for this but with how simple the function is a test would just repeat the definition so I don't think it's needed.

core/src/solution_submission.rs
fleupold: I would say this is a bad error, since in this case we are not guaranteed that some solver provided a solution. Thus we should log this as a failure and work towards getting those cases to 0 (we can consider not alerting on them if they are frequent, but they should be measured somewhere).

core/src/solution_submission.rs
fleupold: I tried to figure out why exactly the deadline future is not unpin, but wasn't able to (might have something to do with `pin_project!` wrapper on the Timeout struct). Do you have an explanation?

core/src/solution_submission.rs
```diff
@@ -244,36 +272,86 @@ async fn retry_with_gas_price_increase(
...
+    match either {
+        Either::Left((result, _)) => {
+            if !is_confirm_timeout(&result) || is_last_iteration {
```
fleupold, line 345: Do we even need the `|| is_last_iteration` if in the last iteration the confirm will never timeout?

core/src/solution_submission.rs
fleupold: Agree that just the `determine_retry_action` is not worth testing. I would however like the overall block given how complex it is unless we can factor the timeout in a more orthogonal way (see PR level comment).

We could expose the deadline future part of a trait "wait_for_solution_submission_end" which we can inject and control in unit tests. This would have the added benefit that we could implement it based on the current block timestamp instead of adding 30s to the system time (although I do think that approach is pragmatic and fine too)

fleupold (commented):
While investigating why the deadline_future is not unpin I stumbled across this: https://docs.rs/async-std/1.6.2/async_std/future/fn.timeout.html which could be a simpler way to build an "orthogonal" abstraction around this logic. 

We could basically keep the existing `retry_with_gas_price_increase` as the inner future (no logic change needed) and await the Timeout future in a `retry_with_gas_price_increase_and_timout` function.

core/src/solution_submission.rs
e00E: With medium confidence I believe that that async functions simply do not propagate Unpin. The return type of `wait_for_deadline` is `impl Future` so even if we did nothing in its body it would still be treated as Unpin. The reason that `solution_future` is unpin is because it isn't an `impl Future` but a `BoxFuture`.

core/src/solution_submission.rs
e00E: Updated this if condition

core/src/solution_submission.rs
nlordell: > With medium confidence I believe that that async functions simply do not propagate Unpin

Correct, `async` functions are marked as `!Unpin` regardless of their body. For example even the following `async` function is `!Unpin`:
```rust
async fn answer() -> i32 { 42 }
```
In fact, an `async` functions get wrapped by the `const fn std::future::from_generator<T>(gen: T) -> impl Future<Output = T::Return>`, and their underlying type is explicitly `!Unpin`: 
https://github.com/rust-lang/rust/blob/8aa18cbdc5d4bc33bd61e2d9a4b643d87f5d21de/src/libcore/future/mod.rs#L64

> is unpin is because it isn't an impl Future but a BoxFuture

You could also `.boxed()` instead of `futures::pin_mut!`. You might also be able to `.fuse()` the future as well which I thought was needed for `select!`.

core/src/solution_submission.rs
```diff
@@ -174,77 +184,175 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
     }
```
nlordell, line 184: I think the easier solution would be to:
```rust
let mut submission = retry_with_gas_price_increase(...);
let mut deadline = wait_for_deadline(...);

select! {
    result = submission => result.map_err(self.make_error(...)),
    _ = b => {
        self.contract.send_noop_transaction(...);
        Err(...)
    },
};
```

nlordell (changes requested):
I don't think we need to select on the deadline inside the retry method but instead around it. I think this is what @fleupold mentioned.

core/src/solution_submission.rs
e00E: I thought about this. The disadvantage is that we don't know what gas price was used so we have to be maximally pessimistic. Since this hopefully happens rarely it is probably worth it for the simpler code. Will try and see how it looks.

e00E:
This is ready. I implemented the suggestion for handling the deadline one level higher and it turns out much cleaner. We still can't easily test that level but imo it is sufficiently simple that it shouldn't be a problem.

core/src/solution_submission.rs
```diff
@@ -155,19 +159,48 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
         solution: Solution,
         claimed_objective_value: U256,
     ) -> BoxFuture<Result<(), SolutionSubmissionError>> {
+        const GAS_PRICE_CAP: u64 = 60_000_000_000;
         async move {
-            match retry_with_gas_price_increase(
+            let submit_future = retry_with_gas_price_increase(
```
nlordell, line 164: I think `submit_solution` method should take a `nonce` parameter, and we should use `web3` here to read it. That way the `send_noop_transaction` is guaranteed to have the same nonce as the solution submission transaction.

I would be OK with adding that in a follow up PR though.

nlordell (approved):
Looks good. Just one suggestions (but I would be OK in including it as a follow up PR).

nlordell:
>  We still can't easily test that level but imo it is sufficiently simple that it shouldn't be a problem.

Maybe we should mock a `Deadline` trait that creates a deadline future to unit test the error condition?

fleupold (approved):
Looks very nice

> Maybe we should mock a Deadline trait that creates a deadline future to unit test the error condition?

I agree that testing this would be nice and should be possible by not depending directly on async_std::future::timeout in the code. Given the rest of the logic is fairly well covered in the tests I think we should add coverage for this part as well.

core/src/models/batch_id.rs
```diff
@@ -46,6 +48,10 @@ impl BatchId {
         self.order_collection_start_time() + BATCH_DURATION
     }
 
+    pub fn solve_end_time(self) -> SystemTime {
+        self.solve_start_time() + SOLVING_WINDOW
+    }
```
bh2smith, line 53: What happens if `self.solve_start_time() + SOLVING_WINDOW` is greater than the batch length? Do we have anything checking or imposing such a constraint?

bh2smith (approved):
Looks good, just one concern about the solve stop time.

core/src/models/batch_id.rs
e00E: The constants for batch duration and solving window are defined right next to each other with one clearly being larger than the other so I don't think there is a need to assert this more.

nlordell: Rust needs `static_assert!`s :stuck_out_tongue:

PR #1091, opened 2020-07-07
Simplify retry_with_gas_price_increase

e00E:
* create a couple of helper functions
* turn the while / do-while into a loop

Makes #1089 easier.

I recommend looking at the full new function (or side by side), not just the diff.

### Test Plan
Existing unit tests cover this.

### DIFF
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -174,77 +174,108 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
     }
 }
 
+fn is_confirm_timeout(result: &Result<(), MethodError>) -> bool {
+    matches!(
+        result,
+        &Err(MethodError {
+            inner: ExecutionError::ConfirmTimeout,
+            ..
+        })
+    )
+}
+
+struct InfallibleGasPriceEstimator<'a> {
+    gas_price_estimating: &'a (dyn GasPriceEstimating + Sync),
+    previous_gas_price: U256,
+}
+
+impl<'a> InfallibleGasPriceEstimator<'a> {
+    fn new(
+        gas_price_estimating: &'a (dyn GasPriceEstimating + Sync),
+        default_gas_price: U256,
+    ) -> Self {
+        Self {
+            gas_price_estimating,
+            previous_gas_price: default_gas_price,
+        }
+    }
+
+    /// Get a fresh price estimate or if that fails return the most recent previous result.
+    async fn estimate(&mut self) -> U256 {
+        match self.gas_price_estimating.estimate_gas_price().await {
+            Ok(gas_estimate) => {
+                self.previous_gas_price = gas_estimate.fast;
+            }
+            Err(ref err) => {
+                log::warn!(
+                    "failed to get gas price from gnosis safe gas station: {}",
+                    err
+                );
+            }
+        };
+        self.previous_gas_price
+    }
+}
+
+fn gas_price(estimated_price: U256, price_increase_count: u32, cap: U256) -> U256 {
+    let factor = 2u32.pow(price_increase_count);
+    cap.min(estimated_price * factor)
+}
+
 async fn retry_with_gas_price_increase(
     contract: &(dyn StableXContract + Sync),
     batch_index: u32,
     solution: Solution,
     claimed_objective_value: U256,
     gas_price_estimating: &(dyn GasPriceEstimating + Sync),
-    gas_cap: U256,
+    gas_price_cap: U256,
 ) -> Result<(), MethodError> {
-    const INCREASE_FACTOR: u32 = 2;
     const BLOCK_TIMEOUT: usize = 2;
     const DEFAULT_GAS_PRICE: u64 = 15_000_000_000;
-
     // openethereum requires that the gas price of the resubmitted transaction has increased by at
     // least 12.5%.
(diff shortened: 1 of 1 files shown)


fleupold (approved):
Looks good, I'm a bit concerned that adding the cancellation logic will blow up the loop logic again but let's see...

core/src/solution_submission.rs
```diff
@@ -174,77 +174,76 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+            None
+        } else {
+            Some(BLOCK_TIMEOUT)
+        };
+        let result = submit_solution(gas_price, block_timeout).await;
+        if !is_confirm_timeout(&result) || is_last_iteration {
+            return result;
+        }
         gas_price_factor *= INCREASE_FACTOR;
-        info!(
```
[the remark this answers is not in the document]
e00E: Added it back. I had moved it initially then forgot to add it back.

e00E:
I refactored this more to remove even more independently testable logic from the function.

core/src/solution_submission.rs
```diff
@@ -388,42 +465,8 @@ mod tests {
...
-                ))
-                }
-                .boxed()
-            });
-        contract
-            .expect_submit_solution()
-            .with(always(), always(), always(), eq(U256::from(15)), eq(None))
-            .return_once(|_, _, _, _, _| {
+            .times(3)
+            .returning(|_, _, _, _, _| {
```
e00E, line 469: Since the gas price calculation is independently tested we don't need this test to test the exact values.

core/src/solution_submission.rs
```diff
@@ -174,77 +174,105 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+    matches!(
+        result,
+        &Err(MethodError {
+            inner: ExecutionError::ConfirmTimeout,
+            ..
+        })
+    )
+}
+
+struct InfallibleGasPriceEstimator<'a> {
```
nlordell, line 187: I like this!

core/src/solution_submission.rs
```diff
@@ -174,77 +174,105 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+            "solution submission try {} with gas price {}",
+            gas_price_increase_count, gas_price
         );
+        let is_last_iteration = gas_price >= effective_gas_price_cap;
+        let block_timeout = if is_last_iteration {
+            None
+        } else {
+            Some(BLOCK_TIMEOUT)
+        };
+        let result = submit_solution(gas_price, block_timeout).await;
```
nlordell, line 270: ubernit: I don't think the closure has so much value anymore seeing as this is the only place the `submit_solution` is called.

nlordell (approved):
Nice refactor! The loop became so much easier to follow now!

core/src/solution_submission.rs
fleupold: Me too! Cool abstraction...

core/src/solution_submission.rs
```diff
@@ -174,77 +174,104 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
-                })
-            )
-    } {
-        // Increase gas
-        gas_price_factor *= INCREASE_FACTOR;
-        info!(
-            "retrying solution submission with increased gas price factor of {}",
-            gas_price_factor,
-        );
+        if !is_confirm_timeout(&result) || is_last_iteration {
```
fleupold, line 270: Given we don't timeout in the last iteration the or condition here is not needed, right?

fleupold (approved):
Nice, I really like the Infallible gas estimation wrapper.

core/src/solution_submission.rs
e00E: That's true. Since we always set the block timeout to None in the last iteration, getting a confirm timeout means that we cannot be in the last iteration.
There is still some merit to having this in case we make the change that we do still have a block timeout even in the last iteration or in case there unexpectedly is a ConfirmTimeout anyway (ethcontract doesn't say *there will never be a confirm timeout if block timeout is set to None*) so I'd like to keep it like this with this here explanation added as a comment.
This is also the same way the previous code worked where we could make the same assumptions to remove the `gas_price_can_still_increase` check.

core/src/solution_submission.rs
```diff
@@ -174,77 +174,108 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
+        let estimated_price = gas_price_estimator.estimate().await;
+        let gas_price = gas_price(estimated_price, gas_price_increase_count, gas_price_cap);
+        assert!(gas_price <= gas_price_cap);
+        info!(
+            "solution submission try {} with gas price {}",
+            gas_price_increase_count, gas_price
+        );
+        let is_last_iteration = gas_price >= effective_gas_price_cap;
+        let block_timeout = if is_last_iteration {
+            None
```
nlordell, line 257: Now that I think of it, it may be a little dangerous to have `None` as a block timeout for confirmations. The reason being it can be that there is a gas price spike and it takes minutes (or even hours) for the transaction to be mined. Maybe we should use a more reasonable block timeout, something like 5 to 10 (75 - 150 seconds on average). In that case `is_last_iteration` would be needed.

On the other hand, we will be selecting this future with a timeout future, so it probably doesn't really matter.

PR #1093, opened 2020-07-07
Introduce `Market` struct

nlordell:
This PR introduces a new `Market` structure for specifying base and quote tokens so there is no ambiguity in the order in which they are specified.

### Test Plan

No significant changes, unit tests should catch any logic errors that may have been introduced (that may accidentally reverse the order of the ask and bid orders).


pricegraph/src/lib.rs
```diff
@@ -84,6 +84,37 @@ impl TransitiveOrder {
     }
 }
 
+/// A struct representing a market.
+///
+/// This is used for computing transitive orderbooks.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub struct Market {
+    /// The base token being purchased in a market.
```
fleupold, line 93: Technically the base token can also be sold in a market, against the bids.

pricegraph/src/lib.rs
```diff
@@ -84,6 +84,37 @@ impl TransitiveOrder {
...
+    pub base: TokenId,
+    /// The quote token being sold in a market. Prices in a market are always
+    /// expressed in the quote token.
+    pub quote: TokenId,
+}
+
+impl Market {
+    /// Returns the token pair for ask orders.
+    pub fn ask_pair(self) -> TokenPair {
```
fleupold, line 102: It's a bit counter-intuitive to me that you can turn a TokenPair into one market but one market into two token pairs. Maybe we should clarify the `Into` more explicitly into something like from_buy_order?

fleupold (approved):
Small note about <Into> being unique but <From> not, otherwise looks good.

pricegraph/src/lib.rs
nlordell: I really wasn't sure how to describe base and quote tokens... I'll look up a Wikipedia definition or something.

pricegraph/src/lib.rs
nlordell: Oh! You mean the `price_estimator::models::TokenPair`. I'll rename that, I think that type should be a "market" anyway.

e00E (approved):

pricegraph/src/lib.rs
nlordell: I'm going to use language from: https://en.wikipedia.org/wiki/Currency_pair

pricegraph/src/lib.rs
nlordell: @fleupold  Hey, I refactored the `price_estimator::models::TokenPair` to not use `Into<pricegraph::TokenPair>` and `Into<pricegraph::Market>` which was admittedly a bit confusing. Its now `price_estimator::models::Market` and implements `Deref<Target = Market>` and we have a `Filter` prefix that gets the bid token pair for a market. Let me know if this is a little clearer.

Issue #1082, opened 2020-07-06
Handle `atoms=false` (should be default)


fleupold:
Since #1007 we have decimal information for all valid tokens available in rust. We should use this information to expose all price-estimation APIs in base units (where 1ETH == 1 instead of 10**18).

fleupold:
We should also expose baseunits for the market route. EFrontier is trying to use it for their market making strategy and really struggling to do the conversion.

e00E:
This is done.

PR #1095, opened 2020-07-08
Implement atoms=false for estimated_buy_amount

e00E:
and add error code handling of rejections

First step for #1082 . So far this commit only implements this for one route but not the others. If this way of implementation makes sense I will add it to the other routes (should it apply to all routes, even `/marktets`?) and update the readme. Then we probably also want the token info overrides like the driver has for tokens that don't implement ERC20 properly.

### Test Plan
Unit test for error codes and manual test:
```
$ curl "localhost:8080/api/v1/markets/4-7/estimated-buy-amount/1000000000000000000?atoms=true"
{"baseTokenId":"4","quoteTokenId":"7","buyAmountInBase":"1006194.8387101036","sellAmountInQuote":"1000000000000000000"}

$ curl "localhost:8080/api/v1/markets/4-7/estimated-buy-amount/1?atoms=false"
{"baseTokenId":"4","quoteTokenId":"7","buyAmountInBase":"1.0061948387101036","sellAmountInQuote":"1"}
```
Tested that changing the type of token amounts from u128 to f64 doesn't cause problems in frontend.

### DIFF
--- a/core/src/token_info.rs
+++ b/core/src/token_info.rs
@@ -51,6 +51,11 @@ impl TokenBaseInfo {
         SYMBOL_OVERRIDES.get(&self.alias).unwrap_or(&self.alias)
     }
 
+    /// One unit of the token taking decimals into account, given in number of atoms.
+    pub fn base_unit_in_atoms(&self) -> f64 {
+        10f64.powi(self.decimals as i32)
+    }
+
     /// Converts the prices from USD into the unit expected by the contract.
     /// This price is relative to the OWL token which is considered pegged at
     /// exactly 1 USD with 18 decimals.
@@ -92,5 +97,13 @@ mod tests {
     #[test]
     fn weth_token_symbol_is_eth() {
         assert_eq!(TokenBaseInfo::new("WETH", 18).symbol(), "ETH");
+    }
+
+    #[test]
+    #[allow(clippy::float_cmp)]
+    fn base_unit_in_atoms() {
+        assert_eq!(TokenBaseInfo::new("", 0).base_unit_in_atoms(), 1.0);
+        assert_eq!(TokenBaseInfo::new("", 1).base_unit_in_atoms(), 10.0);
+        assert_eq!(TokenBaseInfo::new("", 2).base_unit_in_atoms(), 100.0);
     }
 }
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -1,18 +1,25 @@
 ## API
 
+All endpoints use the query part of the url with these key-values:
+
+* `atoms`: Required. If set to `true` all amounts are denominated in the smallest available unit (base quantity) of the token. If `false` all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. TODO: `false` is currently only implemented for estimated-buy-amount .
+* `hops`: Optional. TODO: document this once it has been implemented.
+
+Example: `<path>?atoms=true`
+
+The endpoint documentation references these types:
+
+* A *token id* is a natural number >= 0 in decimal notation. It refers to the token ids in the smart contract. Example: `0`
+* A *token amount* is given as a floating point number formatted according to https://doc.rust-lang.org/std/primitive.f64.html#impl-FromStr which resembles numbers in json closely. Examples: `0`, `1.1`.
+* A *market* is of the form `<base token id>-<quote token id>`. Example: `0-1`
+
 The service exposes the following endpoints:
 
 ### Markets
 
 `GET /api/v1/markets/:market`
 
-* `market` is of the form `<base_token_id>-<quote_token_id>`. The token ids the same as in the smart contract.
-
-Url Query:
-* `atoms`: If set to `true` (for now this is the only implemented method) all amounts will be denominated in the smallest available unit (base quantity) of the token.
-* `hops`: TODO: document this once it has been implemented.
-
-Example Request: `/api/v1/markets/1-7/?atoms=true`
+Example Request: `/api/v1/markets/1-7?atoms=true`
 
 Example Response:
 
@@ -33,11 +40,6 @@ Returns the transitive orderbook (containing bids and asks) for the given base a
 
 `GET /api/v1/markets/:market/estimated-buy-amount/:sell-amount-in-quote-token`
 
-* `market` is as above
-* `sell_amount_in_quote_token` is a positive integer.
-
-Url query is as above.
-
 Example Request: `/api/v1/markets/1-7/estimated-buy-amount/20000000000000000000?atoms=true`
 
 Example Response:
@@ -57,11 +59,6 @@ Example Response:
 ### Estimated Amounts At Price
 
 `GET /api/v1/markets/:market/estimated-amounts-at-price/:price-in-quote`
-
-* `market` is as above
-* `price_in_quote` is a decimal number.
-
-Url query is as above.
 
 Example Request: `/api/v1/markets/1-7/estimated-amounts-at-price/245.5?atoms=true`
 
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -1,27 +1,66 @@
 use crate::models::*;
 use crate::orderbook::Orderbook;
+use core::{
+    models::TokenId,
+    token_info::{TokenBaseInfo, TokenInfoFetching},
+};
 use pricegraph::TokenPair;
 use std::convert::Infallible;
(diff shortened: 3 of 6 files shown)


price-estimator/src/filter.rs
```diff
@@ -127,20 +168,43 @@ async fn get_markets<T>(
...
     orderbook: Arc<Orderbook<T>>,
-) -> Result<impl warp::Reply, Infallible> {
+    token_infos: Arc<dyn TokenInfoFetching>,
+) -> Result<impl Reply, Rejection> {
+    let sell_amount_in_quote_atoms = if query.atoms {
+        sell_amount_in_quote
+    } else {
+        let token_info = get_token_info(token_pair.sell_token_id, token_infos.as_ref()).await?;
+        sell_amount_in_quote * token_info.unit_in_atoms()
+    };
```
fleupold, line 194: I was wondering if this logic should rather live in the estimate_buy_amount_filter (so that the actual handler can only deal with atom based logic) but that doesn't nicely work with the way that results are returned at the moment since they also have to be changed if `atoms = false`. 
Given that the handler itself just delegates a call into the pricegraph, maybe it's fine to keep this logic here.

price-estimator/src/models.rs
```diff
@@ -62,9 +62,9 @@ pub struct EstimatedOrderResult {
     #[serde(with = "display_fromstr")]
     pub quote_token_id: u16,
     #[serde(with = "display_fromstr")]
-    pub buy_amount_in_base: u128,
+    pub buy_amount_in_base: f64,
     #[serde(with = "display_fromstr")]
-    pub sell_amount_in_quote: u128,
+    pub sell_amount_in_quote: f64,
```
fleupold, line 67: Does this work with the frontend in the case that atoms=true?

price-estimator/src/filter.rs
```diff
@@ -209,9 +273,78 @@ fn apply_rounding_buffer(amount: f64, price_rounding_buffer: f64) -> f64 {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures::future::FutureExt as _;
+    use anyhow::{anyhow, Result};
+    use core::orderbook::StableXOrderBookReading;
+    use futures::future::{BoxFuture, FutureExt as _};
+
+    fn empty_orderbook() -> impl StableXOrderBookReading {
```
fleupold, line 280: Is there a reason you are not using the Mocked versions here?

core/src/token_info.rs
```diff
@@ -51,6 +51,11 @@ impl TokenBaseInfo {
         SYMBOL_OVERRIDES.get(&self.alias).unwrap_or(&self.alias)
     }
 
+    /// One unit of the token taking decimals into account, given in number of atoms.
+    pub fn unit_in_atoms(&self) -> f64 {
```
fleupold, line 55: ubernit: maybe base_unit instead of just unit (since technically an atom is also a unit)

price-estimator/src/models.rs
e00E: Either way all json numbers can be fractions and something that previously wasn't a fraction as u128 still won't be a fraction as f64 so I didn't expect this to be a problem but I'll test it.

price-estimator/src/filter.rs
e00E: Can't use mocks in other crates because they are only compiled for tests.

price-estimator/src/models.rs
```diff
@@ -130,6 +130,11 @@ impl From<&pricegraph::TransitiveOrderbook> for MarketsResult {
 #[serde(transparent)]
 pub struct PriceEstimateResult(pub Option<f64>);
 
+#[derive(Debug, Serialize)]
+pub struct ErrorMessage {
+    pub message: &'static str,
+}
```
nlordell, line 136: nit: I think the struct shouldn't be an `ErrorMessage` but more of an `Error`. Also I think the `message` field should be used for a human readable messages that may include some extra context and not status constants (which should probably be an enum instead of `&'static str`).
```suggestion
#[derive(Debug, Serialize)]
pub struct Error {
    pub status: ErrorStatus,
}

#[derive(Debug, Serialize)]
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
pub enum ErrorState {
    NotFound,
    NoTokenInfo,
    // ...
}
```

price-estimator/src/filter.rs
nlordell: Maybe it can be implemented in the `filter` by using a `map` function?

price-estimator/src/main.rs
```diff
@@ -127,3 +133,18 @@ fn setup_driver_metrics() -> HttpMetrics {
     });
     HttpMetrics::new(&prometheus_registry).unwrap()
 }
+
+async fn initialize_token_info_cache(cache: &TokenInfoCache) -> anyhow::Result<()> {
+    for token_id in cache.all_ids().await.context("failed to get all ids")? {
+        // Individual tokens might not conform to erc20 in which case we are unable to retrieve
+        // their info.
+        if let Err(err) = cache.get_token_info(token_id).await {
```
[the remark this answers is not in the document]
nlordell: :+1: I think it should be a `all_token_infos` method on the `TokenInfoFetching` trait that provides a default implementation.

price-estimator/src/models.rs
e00E: I disagree. This struct represents the body of the response we send when we do not respond with success. The `handle_rejection` function decides what goes in there so I see no reason to introduce an enum with more complicated serialization.
I agree with the human readable part so i'll change the messages a bit to contain more informatin.

price-estimator/src/main.rs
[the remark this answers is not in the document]
e00E: Will do this in another PR.

price-estimator/src/models.rs
e00E: Tested it without problems.

price-estimator/src/filter.rs
e00E: I don't think this will turn out better by doing at a higher level so I'll keep it this way. I'm not opposed to that in the future either just don't think it's worth trying for this PR.

e00E:
Ready for review now. I've included a readme update as well.

price-estimator/src/models.rs
nlordell: My main argument was that if the error message was going to be `NOT_FOUND` style string constants, then it should be an enum (and a `status` or `kind` as opposed to a `message`). This could be very valuable for clients to recover an exact error type from the error data (kind of like the `ErrorKind` pattern in Rust).

`messages` are intended to be human readable error messages, kind of like `Error::description` or `Error::to_string` in Rust.

price-estimator/src/models.rs
nlordell: I think, **in a future PR**, we should change these to serialize as numbers instead of strings.

e00E: Agree, I believe we have an issue for this.

price-estimator/src/models.rs
e00E: I'm not sure if we should make any guarantees about the contents of the error. Human readable has practical benefits but I don't think any client can recover in a meaningful way so I don't want to make more work for us to implement this. This is something I'm not super confident about so I could be convinced but I would also like to have that discussion in another issue/pr and keep this simple way of handling errors for now.

nlordell: > I would also like to have that discussion in another issue/pr

:+1: 

I just have a nit with the naming, maybe
```suggestion
#[derive(Debug, Serialize)]
pub struct ErrorResult {
    pub message: &'static str,
}
```

price-estimator/README.md
```diff
@@ -1,18 +1,25 @@
 ## API
 
+All endpoints use the query part of the url with these key-values:
+
+* `atoms`: Required. If set to `true` all amounts are denominated in the smallest available unit (base quantity) of the token. If `false` all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. TODO: `false` is currently only implemented for estimated-buy-amount .
```
fleupold, line 5: I think it would make sense to default this to false (can be a separate PR given the size of this one).

price-estimator/src/models.rs
fleupold: One reason error kinds are nice is that they frontend can easily localize them to other languages (which a backend API usually doesn't). But I agree that it is not needed in our current state.

Issue #1084, opened 2020-07-06
The Great `pricegraph::Orderbook` Refactor


nlordell:
The `pricegraph::Orderbook` module (`pricegraph/src/orderbook.rs` is currently well over 1000 lines and becoming hard to navigate and add new features. This issue captures the work to break it down into smaller pieces as well as move away from the `Orderbook` name which is a bit overused and makes searching for code harder. Currently on the list of things to do:
1. Rename `Orderbook` -> `Auction` (?)
2. Move primitive operations (path finding and reducing path) into its own module.
3. Implement high level operations (reducing overlapping order, exchange rate estimates) into separate modules and implement them directly on the `Pricegraph` root type.

PR #1097, opened 2020-07-08
Added errors for operations that require reduced orderbooks

nlordell:
This PR adds an error for operations that require non-overlapping orderbooks instead of silently reducing them. This is a step towards #1084 where we want to expose some primitive operations for the orderbook graph. Specifically:
- Fill ring trade including token pair (for computing overlapping transitive orders, used for computing transitive orderbook)
- Fill transitive order (for computing transitive orders on a reduced graph, used for computing price estimates and transitive orderbook)
- Reduce orderbook.

So, by adding a new error for operations that expect a reduced orderbook, we can work towards these "slimmer" primitives.

### Test Plan

Unit tests still pass :tada:, unit test changes were only to reflect new error return types. Also added new unit test to verify overlapping orderbooks trigger errors.


e00E (approved):

pricegraph/src/orderbook.rs
```diff
@@ -93,43 +93,44 @@ impl Orderbook {
...
                     Ok((_, predecessor)) => ControlFlow::Continue(predecessor),
-                    Err(NegativeCycle(predecessor, _)) => {
-                        if predecessor[0].is_some() {
-                            // The negative cycle is connected to the fee token.
-                            ControlFlow::Break(true)
-                        } else {
-                            ControlFlow::Continue(predecessor)
-                        }
-                    }
+                    Err(NegativeCycle(..)) => ControlFlow::Break(true),
```
fleupold, line 110: Why did this logic change?

fleupold (approved):
Nice refactor! Took me a while to find where we are actually returning the Overlap error because rust is so nice to convert the NegativeCycle for you automatically thanks to the Into implementation.

pricegraph/src/orderbook.rs
nlordell: So, the reason I changed it here is that with the old implementation it was possible to have:
```
assert!(!orderbook.is_overlapping());
orderbook.fill_optimal_transitive_order(...).unwrap(); // panic
```

The original implementation had tried to ensure that the ring was connected to the fee token, but had a pretty critical flaw:
```
0 <---- 1 <---- 2 <---- 3 <---- 4
                         `------^
```

This would be identified as "solvable", but it in fact is not, since the fee token is not directly connected to the ring.

Additionally to determine a "solvable" ring trade, there are other constraints that were being ignored, so I decided to just simplify the `is_overlapping` method so the language is more coherent with the rest of the methods (specifically to do with `OverlapError`).

nlordell: Also we're not *really* using this method anywhere.

PR #1098, opened 2020-07-09
Use `Flow` Instead of Capacity/Xrate Tuple

nlordell:
This PR refactors all inner operations to use `Flow` instead of tuples to avoid confusion. Also allows a method to be removed as it is no longer used.

### Test Plan

Simple refactor, unit tests continue to pass :tada: .


e00E (approved):

fleupold (approved):
😎

PR #1102, opened 2020-07-09
Allow hardcoded token info

e00E:
Used as an override to the ERC20 information we fetch from the block chain
in case that information is wrong or unavailable which can happen for
example when tokens do not implement the standard properly.

### Test Plan
Manually tested with `'{69:{"alias":"aaa","decimals":1,"externalPrice":2,}}'` and checked that the info is overriden.

### DIFF
--- a/price-estimator/src/main.rs
+++ b/price-estimator/src/main.rs
@@ -8,7 +8,7 @@ use core::{
     http::HttpFactory,
     metrics::{HttpMetrics, MetricsServer},
     orderbook::EventBasedOrderbook,
-    token_info::{cached::TokenInfoCache, TokenInfoFetching as _},
+    token_info::{cached::TokenInfoCache, hardcoded::TokenData, TokenInfoFetching as _},
     util::FutureWaitExt as _,
 };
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Easy

fleupold (approved):
Nice! I don't think we need it right now but will be good to have if we need it later. Just a shame that this list will likely have to be duplicated in the config. I'm wondering if we should instead make it part of the code repo?

e00E:
We should merge this anyway in case we need it. Kubernetes might allow us to share parts of the config in a way that makes this possible. It feels wrong to have hard coded constants that only match one specific deployment of the smart contract.

nlordell:
>  I'm wondering if we should instead make it part of the code repo?

One of the main advantages of having it as a command line argument is that we can quickly modify the value and restart the pod in case of any issues.

What we could do, is implement a `TokenData::default_for_network(network_id: impl AsRef<str>)` so we have some hard coded default values in the future when the need arises. For now I also prefer having this in our pocket in case we have some issues with a specific token.

PR #1103, opened 2020-07-09
Use types to represent prices and exchange rates

nlordell:
This PR introduces two new scalar types:
- `Price` to represent exchange order limit prices (with implicit fees). That is, an orders buy amount (or price numerator) divided by an order's sell amount (or price denominator).
- `ExchangeRate` to represent effective order exchange rates that include fees.

These types help us convert between prices and exchange rates in a type-safe way, preventing using a `Price` where an `ExchangeRate` is expected.

### Test Plan

Unit tests still pass with same values! :tada:

### DIFF
--- a/pricegraph/src/orderbook/scalar.rs
+++ b/pricegraph/src/orderbook/scalar.rs
@@ -1,0 +1,153 @@
+//! This module contains definitions for measurement scalars used by the
+//! orderbook graph representation.
+
+use crate::encoding;
+use crate::FEE_FACTOR;
+use std::cmp;
+use std::ops;
+
+/// An exchange price. Limit prices on the exchange are represented by a
+/// fraction of two `u128`s representing a buy and sell amount. These limit
+/// prices implicitly include fees, that is they must be respected **after**
+/// fees are applied. As such, the actual exchange rate for a trade (the ratio
+/// of executed buy amount over executed sell amount) is strictly greater than
+/// and **never equal to** the limit price of an order.
+///
+/// Prices are guaranteed to be a strictly positive finite real numbers.
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
+pub struct Price(f64);
+
+impl Price {
+    /// Creates a new price from a `f64` value. Returns `None` if the price
+    /// value is not valid. Specifically, it must be in the range `(0, +∞)`.
+    pub fn new(value: f64) -> Option<Self> {
+        if is_strictly_positive_and_finite(value) {
+            Some(Price(value))
+        } else {
+            None
+        }
+    }
+
+    /// Creates a new price from an exchange price fraction.
+    pub fn from_fraction(price: &encoding::Price) -> Option<Self> {
+        if price.numerator != 0 && price.denominator != 0 {
+            Some(Price(assert_strictly_positive_and_finite(
+                price.numerator as f64 / price.denominator as f64,
+            )))
+        } else {
+            None
+        }
+    }
+
+    /// Converts a price into an effective exchange rate with explicit fees.
+    pub fn exchange_rate(self) -> ExchangeRate {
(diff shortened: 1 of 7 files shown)


pricegraph/src/orderbook/scalar.rs
```diff
@@ -0,0 +1,159 @@
...
+/// and **never equal to** the limit price of an order.
+///
+/// Prices are guaranteed to be a strictly positive finite real numbers.
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
+pub struct Price(f64);
+
+impl Price {
+    /// Creates a new price from a `f64` value. Returns `None` if the price
+    /// value is not valid. Specifically, it must be in the range `(0, +∞)`.
+    pub fn new(value: f64) -> Option<Self> {
```
fleupold, line 23: Any reason to not use try_from and into traits here?

pricegraph/src/orderbook/scalar.rs
```diff
@@ -0,0 +1,159 @@
...
+/// prices implicitly include fees, that is they must be respected **after**
+/// fees are applied. As such, the actual exchange rate for a trade (the ratio
+/// of executed buy amount over executed sell amount) is strictly greater than
+/// and **never equal to** the limit price of an order.
+///
+/// Prices are guaranteed to be a strictly positive finite real numbers.
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
+pub struct Price(f64);
+
+impl Price {
```
fleupold, line 20: maybe calling it LimitPrice is even more clear?

pricegraph/src/orderbook/scalar.rs
```diff
@@ -0,0 +1,159 @@
...
+// to not be `NaN`.
+impl Eq for ExchangeRate {}
+
+impl Ord for ExchangeRate {
+    fn cmp(&self, rhs: &Self) -> cmp::Ordering {
+        self.partial_cmp(rhs).expect("exchange rate cannot be NaN")
+    }
+}
+
+impl PartialEq<f64> for ExchangeRate {
```
fleupold, line 110: cool, could even be implemented for Limit Price if needed.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FEwAmOj2aLfs1q%2Fgiphy.gif&token=price" data-gitmeme-token="price"></a>

pricegraph/src/orderbook/scalar.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// projection graph.
+    ///
+    /// This is the base-2 logarithm of the exchange rate. This eanbles path
+    /// weights to be computed using addition instead of multiplication.
+    pub fn weight(self) -> f64 {
+        self.0.log2()
+    }
+}
+
+// NOTE: We can implement `Eq` for `ExchangeRate` since its value is garanteed
```
e00E, line 100:
```suggestion
// NOTE: We can implement `Eq` for `ExchangeRate` since its value is guaranteed
```

e00E (approved):

pricegraph/src/orderbook/scalar.rs
nlordell: Not really, just was lazy to implement an error type for it. Rust standard has some wrapper number types for non-zero numbers that also follow this `new(inner) -> Option<Self>` pattern so I just went with it (https://doc.rust-lang.org/std/num/struct.NonZeroU64.html#method.new). If you think there is value in adding an error type for invalid prices and implementing `TryFrom`, we can definitely do that (and I think it would be very correct).

pricegraph/src/orderbook/scalar.rs
nlordell: For sure!

PR #1105, opened 2020-07-10
Unit test solution submission noop transaction

e00E:
In #1089 we talked about testing the new functionality of sending the noop transaction. This PR adds a test by moving the logic that handles the result of deadline-solution-submission into its own function. This sidesteps having to mock `async_std::future::timeout`. Mocking this is not worth the extra code for what little benefit we get. The mocked version looks something like this

```rust
use std::{error::Error, fmt, future::Future, time::Duration};

/// A mockable version of async_std::future::timeout .
pub trait Timeouting {
    // must use static because mockall limitations
    fn timeout<F, T>(dur: Duration, f: F) -> BoxFuture<'static, Result<T, TimeoutError>>
    where
        F: Future<Output = T> + Send + 'static,
        T: 'static,
    {
        async move {
            async_std::future::timeout(dur, f)
                .await
                .map_err(|_| TimeoutError {})
        }
        .boxed()
    }
}

/// An error returned when a future times out.
#[derive(Clone, Copy, Debug)]
pub struct TimeoutError {}

impl Error for TimeoutError {}

impl fmt::Display for TimeoutError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        "future has timed out".fmt(f)
    }
}
```
and in addition to that we have to store a `Timeouter` in the solution submission struct. 

It can still be argued that there is some untested code left in `submit_solution` but this function has no control flow and only trivial arithmetic that doesn't need to be tested.

### DIFF
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -1,10 +1,13 @@
 #![allow(clippy::ptr_arg)] // required for automock
 
-use crate::contracts::stablex_contract::StableXContract;
-use crate::gas_station::GasPriceEstimating;
-use crate::models::{BatchId, Solution};
+use crate::{
+    contracts::stablex_contract::StableXContract,
+    gas_station::GasPriceEstimating,
+    models::{BatchId, Solution},
+};
 
 use anyhow::{anyhow, Error, Result};
+use async_std::future::TimeoutError;
 use ethcontract::errors::{ExecutionError, MethodError};
 use ethcontract::web3::types::TransactionReceipt;
 use ethcontract::U256;
@@ -23,6 +26,7 @@ const POLL_TIMEOUT: Duration = Duration::from_secs(5);
 #[cfg(test)]
 const POLL_TIMEOUT: Duration = Duration::from_secs(0);
 
+const GAS_PRICE_CAP: u64 = 60_000_000_000;
 // openethereum requires that the gas price of the resubmitted transaction has increased by at
 // least 12.5%.
 const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
@@ -127,6 +131,41 @@ impl<'a> StableXSolutionSubmitter<'a> {
         }
         SolutionSubmissionError::Unexpected(err.into())
     }
+
+    async fn handle_submit_solution_result(
+        &self,
+        batch_index: u32,
+        solution: Solution,
+        result: std::result::Result<Result<(), MethodError>, TimeoutError>,
+    ) -> Result<(), SolutionSubmissionError> {
+        if let Ok(submit_result) = result {
+            match submit_result {
+                Ok(()) => Ok(()),
+                Err(err) => Err(self.make_error(batch_index, solution, err).await),
+            }
+        } else {
+            let gas_price =
+                U256::from((GAS_PRICE_CAP as f64 * MIN_GAS_PRICE_INCREASE_FACTOR).ceil() as u128);
+            log::info!(
+                "cancelling transaction because it took too long, using gas price {}",
+                gas_price
+            );
+            match self.contract.send_noop_transaction(gas_price).await {
+                Ok(_) => log::info!(
+                    "cancelled solution submission of batch {} because of deadline",
+                    batch_index
+                ),
+                Err(err) => log::error!(
+                    "failed to cancel solution submission of batch {} after deadline: {:?}",
+                    batch_index,
+                    err
+                ),
+            }
+            Err(SolutionSubmissionError::Unexpected(anyhow!(
+                "solution submission transaction not confirmed in time"
+            )))
+        }
+    }
 }
 
 impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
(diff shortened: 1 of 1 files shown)


core/src/solution_submission.rs
```diff
@@ -639,4 +652,41 @@ mod tests {
             }
         };
     }
+
+    // Silly way to create a TimeoutError because the type can't be constructed directly.
+    fn timeout_error() -> TimeoutError {
```
nlordell, line 657: :rofl:, I've definitely done this with some std library error types that have the same problem. Not super important but you can also make it a `lazy_static!` constant - but that's really nit-picky.

core/src/solution_submission.rs
```diff
@@ -23,6 +26,7 @@ const POLL_TIMEOUT: Duration = Duration::from_secs(5);
 #[cfg(test)]
 const POLL_TIMEOUT: Duration = Duration::from_secs(0);
 
+const GAS_PRICE_CAP: u64 = 60_000_000_000;
```
nlordell, line 29:
```suggestion
const GAS_PRICE_CAP: u64 = 60_000_000_000;

```

nlordell (approved):
Nice, we get to unit test the error handling which was my biggest concern.

fleupold (approved):
Looks fine. 

I don't feel strongly, but generally I like to avoid directly testing functions that are not part of the public contract (`handle_submit_solution_result` in this case) since I consider them an implementation detail that we should be able to change without affecting the passing of the tests for the entire unit.

[This article](https://anthonysciamanna.com/2016/02/14/should-private-methods-be-tested.html) takes a somewhat arrogant standpoint but the overall reasoning resonates with me. It might apply slightly less to Rust as it is not as strictly object oriented as e.g. Java.

Instead of having to write a full fledged mock, could we pass the timeout function as a somewhat generic dependency into the constructor and have two test implementation to resolve of fail immediately?

e00E:
Only testing through the public interface is a good guideline but not something that should be taken as an absolute.
Looking at the article I find it silly that we can follow its advice by doing the "ExtractClass" refactoring, as in we could turn `handle_submit_solution_result` into a freestanding public function and now we are allowed to test it. The function hasn't actually changed and it's still used in only one place.
Taking this to the extreme we realize that most of our code goes into the driver binary, so really the true public interface is this binary and nearly all tests should be integration tests. But again it is more practical to test in smaller steps.
Especially unit tests still make sense for private functionality. On the other hand I wouldn't assert implementation details in integration tests.

Related is the separation of pure computation and "io". It is good to separate the pure and impure parts of the code because this makes testing easier and the code easier to understand. We can do this by splitting out functions like in this PR or introducing traits. I find the splitting out into separately testable function cleaner in this case. The function is already simple and to make it pure we would have to move the system time usage and the timeouting into a trait. The amount of extra code is not worth it for this function. Maybe if those traits would be useful in more places they would be worth having.

nlordell:
> I consider them an implementation detail that we should be able to change without affecting the passing of the tests for the entire unit.

In general I agree with this, as if modifying an internal implementation requires modifying unit tests, it is possible that you have less confidence in the test result as a new bug may have been introduced by the unit test modification.

> We can do this by splitting out functions

Here, the public API is a thin wrapper around `call_failing_io() -> call_pure_handler()` so I'm also OK with testing a private method `call_pure_handler()` method.

That being said using trait "feels" slightly more correct to me.

fleupold:
As said before I'm fine with this, but would still like to continue the theoretical discussion as I think it's an interesting fundamental of software engineering philosophy.

I think one advantage of the `ExtractClass` refactoring is that it keeps the scope of each test uniform. Right now a simplified high level structure of our test subject looks like this:

```rust
pub fn foo(input) {
  let result = some_computation(input);
  let result = some_impure_fn(result);
  return handle_result(result)
}
```

Before we were testing `foo` by giving different inputs and asserting on different results. Now, we still have these tests but we added some that directly test `handle_result(input)` to skip one aspect of the impure fn that is hard to mock (this is testing the same subject but with less scope). 

The original suggestion was to factor the impure computation into a trait that can be passed in from above, but you are right that we could also extract our `handle_result` and `some_computation` into public functions and have `foo` only contain of "wiring" and no real logic.
In both refactorings, the unit tests regain a uniform scope (each one tests a subject entirely) and we wouldn't have some tests shortcutting the logic. 

The code-smell mentioned in the article is that this is a sign of a single unit doing too many things (here deciding what to do _and_ how to wait on Futures). As such I'd argue that our previous tests have now become integration tests and we added a single unit test for one sub-component - the response handler.

nlordell:
@fleupold I agree with this! While now this is a fairly minimal code smell, it may be nice to avoid these to not accumulate tech debt. Testing is hard.

e00E:
My main gripe is when testing makes the non test code harder. 

> here deciding what to do and how to wait on Futures

How to wait on futures (i think we're talking about timeout) is not a decision in the real code. There is only one way. The only reason we need a decision is that having a real timeout (impurities) makes testing harder. So we're changing the code by introducing traits or new functions because of testing. In a perfect world we would be able to test what we want without having to change the code. It is true that having to do this indicates problems with the real code but this is only an indiciation, its not always the case. In my experience introducing a new function is usually good but using the `enum Action` pattern or Io-Traits can make the real code harder to understand. Depending on the situation the increased testability can still be worth it.
In the end there is no objective metric on what is correct to do. My feeling here is that Io-Traits could be good but they take some getting used to and we have a problem with mockall not being powerful enough to represent them well.

I'll merge this now but definitely feel free to continue to discuss or we can try changing this test in future PRs to compare.

nlordell:
> My feeling here is that Io-Traits could be good but they take some getting used to and we have a problem with mockall not being powerful enough to represent them well.

I think this is a very valid point. I especially agree that `mockall` has some limitations that we've unfortunately run into on a number of occasions.

PR #1108, opened 2020-07-10
Cache all token infos in parallel

e00E:
and move this into TokenInfoCache. Imo this fits better into the cache
than the trait.

### Test Plan
Unit test and starting price estimator shows the output about failed tokens faster.

### DIFF
--- a/core/src/token_info/cached.rs
+++ b/core/src/token_info/cached.rs
@@ -1,8 +1,11 @@
 use super::{TokenBaseInfo, TokenId, TokenInfoFetching};
 
-use anyhow::Result;
+use anyhow::{Context as _, Result};
 use async_std::sync::RwLock;
-use futures::future::{BoxFuture, FutureExt};
+use futures::{
+    future::{BoxFuture, FutureExt},
+    stream::{self, StreamExt as _},
+};
 use std::collections::HashMap;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


core/src/token_info/cached.rs
```diff
@@ -33,6 +36,25 @@ impl TokenInfoCache {
...
+
+    /// Attempt to retrieve and cache all token info that is not already cached.
+    /// Fails if `all_ids` fails. Does not fail if individual token infos fail.
+    pub async fn cache_all(&self, number_of_parallel_requests: usize) -> Result<()> {
+        stream::iter(self.all_ids().await.context("failed to get all ids")?)
+            .for_each_concurrent(number_of_parallel_requests, |token_id| async move {
+                // Individual tokens might not conform to erc20 in which case we are unable to retrieve
+                // their info.
+                if let Err(err) = self.get_token_info(token_id).await {
+                    log::info!(
```
nlordell, line 48: Maybe `log::warn` so it gets a bit more attention? I imagine if we run into tokens that have this issue we will need to configure the token data with the correct decimals and such for it.

nlordell (approved):

core/src/token_info/cached.rs
```diff
@@ -151,4 +173,38 @@ mod tests {
...
+        fn token_ids() -> Vec<TokenId> {
+            [0, 1, 2].iter().cloned().map(TokenId).collect()
+        }
+
+        let mut inner = MockTokenInfoFetching::new();
+        inner
+            .expect_all_ids()
+            .times(1)
+            .returning(|| immediate!(Ok(token_ids())));
+        inner
```
fleupold, line 188: Would be nice to test the documented behavior that it doesn't of one of the subqueries fails.

fleupold (approved):
Looks good.

As a longterm improvement I think we should implement `get_token_infos` on the top level trait and use the BatchTransport similar to how we use it [here](https://github.com/gnosis/dex-services/blob/c9cea46f7b561c0f0b2830cbb5f5e4ae62f758e9/core/src/orderbook/streamed/block_timestamp_reading.rs#L62-L65).

We have seen in the past that while parallelising the roundtrip over network has a good improvement, just hitting the node only once has an even stronger effect (it doesn't seem super efficient in handling concurrent requests).

PR #1112, opened 2020-07-13
Fix `panic` because of partially reduced orderbook graph when computing transitive orderbook

nlordell:
This PR fixes a panic in the `pricegraph` when calculating the transitive orderbook for separate subgraphs, where the market's base token's subgraph contained negative cycles. This would happen because only one of the subgraphs (the quote token's subgraph) would end up completely reduced after computing the overlapping transitive orderbook, while the code assumed that both were completely reduced.

This issue was discovered by the pricegraph fuzzing.

This PR also moves the `orderbook!` macro to its own module so that it can be reused from the `pricegraph/lib.rs` module for unit tests.

### Test Plan

- `git checkout 2c0dad400608e1049dedafcdb88ad16cefb7271f` to checkout the commit with the failing test and no fix.
- `cargo test -p pricegraph` and see that it fails
- `git checkout fix-fuzz-crash && cargo test -p pricegraph` and see that it was fixed

### DIFF
--- a/pricegraph/src/encoding.rs
+++ b/pricegraph/src/encoding.rs
@@ -1,7 +1,7 @@
 //! This module implements decoding for the standard `BatchExchange` contract
 //! encoded orders.
 
-use primitive_types::{H160, U256};
+pub use primitive_types::{H160, U256};
 use thiserror::Error;
 
 /// The stride of an orderbook element in bytes.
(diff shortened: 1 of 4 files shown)


e00E (approved):

fleupold (approved):
Nice! Fuzzing for the win...

nlordell:
> Nice! Fuzzing for the win...

Yeah... especially since it would have been AGES before we actually ran into this, seeing how dense our graph is.

PR #1113, opened 2020-07-13
Serialize price estimator token ids as numbers istead of strings

### LINKED ISSUE #969: Change token ids in result from string to number
Currently the token ids are returned as strings even though they are numbers. This is inline with the nodejs price estimator but it would be cleaner to have numbers.

We need to ask frontend if this would break their code.

e00E:
and add documentation on how to test this with the production frontend.

Fixes #969 

### Test Plan
Unit test changed and tested that production frontend still works.

### DIFF
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -46,8 +46,8 @@ Example Response:
 
 ```json
 {
-    "baseTokenId": "1",
-    "quoteTokenId": "7",
+    "baseTokenId": 1,
+    "quoteTokenId": 7,
     "buyAmountInBase": "79480982311034354",
     "sellAmountInQuote": "20000000000000000000",
 }
@@ -66,8 +66,8 @@ Example Response:
 
 ```json
 {
-    "baseTokenId": "1",
-    "quoteTokenId": "7",
+    "baseTokenId": 1,
+    "quoteTokenId": 7,
     "buyAmountInBase": "6098377078823660544",
     "sellAmountInQuote": "1497151572851208749056"
 }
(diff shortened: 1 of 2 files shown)


price-estimator/README.md
```diff
@@ -78,3 +78,19 @@ The following result indicates that if we wanted to buy ETH (token 2) for DAI (t
...
+2. Open the frontend.
+3. Open the developer tools with `F12`.
+4. In the browser console enter `dexPriceEstimatorApi.urlsByNetwork[1] = "http://localhost:8080/api/v1/"`.
+5. Induce a request by changing the sell amount and check that price estimator prints that it handled the request.
+
+It is useful to start the price estimator with logging enabled, using the gnosis staging node url and using a permanent orderbook file:
+
+```
+env RUST_LOG=warn,price_estimator=info,core=info cargo run -p price-estimator -- --node-url https://staging-openethereum.mainnet.gnosisdev.com --orderbook-file ../orderbook-file-mainnet
+```
```
nlordell, line 96: nit: missing newline

nlordell (approved):
🎉

fleupold (approved):
Just to understand the JSON spec correctly, this change means that clients now have to be able to parse responses where the tokenId is e.g. `7.0`?

e00E:
According to json yes. (Not that they would have to parse it seeing as it is already part of of the request and just echoed in the response.)

nlordell:
> Just to understand the JSON spec correctly, this change means that clients now have to be able to parse responses where the tokenId is e.g. 7.0?

No. While the JSON spec accepts both `7` and `7.0` as a number, the serde serialization writes integers without this, so the JSON will always be a natural number. The point I was trying to make before was this:
```rust
extern crate serde_json; // 1.0.56
fn main() {
    dbg!(serde_json::to_string(&10.0_f64.powi(18)).unwrap()); // "1e18"
    dbg!(serde_json::to_string(&10u128.pow(18)).unwrap()); // "1000000000000000000"
}
```

The string representation of a floating point number is ambiguous, so how `serde` chooses to write it out (as a JSON number or otherwise) is not really in our control (even if it were rounded).

e00E:
The question was whether clients need to be able to handle `7.0` to which the answer is yes because we are not making guarantee that we will always encode natural numbers without a decimal point or exponent notation. You are correct that our current implementation does do this but I don't want to guarantee this.

nlordell:
> The question was whether clients need to be able to handle 7.0 to which the answer is yes because we are not making guarantee that we will always encode natural numbers without a decimal point or exponent notation.

No, I don't believe they do. With `serde_json` we are **guaranteeing** that `u16`s will be serialized as natural numbers (i.e. without `.0` and without exponential notation):

> ```rust
> extern crate serde_json; // 1.0.56
> fn main() {
>     dbg!(serde_json::to_string(&10.0_f64.powi(18)).unwrap()); // "1e18"
>     dbg!(serde_json::to_string(&10u128.pow(18)).unwrap()); // "1000000000000000000"
> }
> ```

fleupold:
I think what @e00E is saying is that we don't want to guarantee we are using `serde_json` for serialisation as this is an implementation detail (e.g. we could decide to rewrite the service in an even cooler language than rust - maybe [lolcode](https://en.wikipedia.org/wiki/LOLCODE) - tomorrow).

nlordell:
My point is that we should guarantee this, even if we decide to rewrite the service in lolcode. I think it is fairly common practice for JSON serializers to distinguish between integers and floats and I don't think it's reasonable for clients to have to parse these token IDs as floats instead of integers.

PR #1115, opened 2020-07-14
[Mergify] auto merge dependabot PRs

fleupold:
This PR creates a rule to automatically merge passing dependabot updates (instead of having to do so manually every morning). 

Looking forward to hearing security concerns if there are any. The regex exact matches for dependabot and depenabot-preview bots (but usernames already taken by the real bot). Therefore I think in order to automerge using this rule, it would require to capture the original dependabot account in which case it would be simpler to hide malicious code in a benign looking update.

The rule has been taken from the [mergify docs](https://doc.mergify.io/examples.html#dependabot) and adjusted to the squash based merge we normally do in this repo.

### Test Plan

See dependabot PRs being auto-merged in the future.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -10,3 +10,14 @@ pull_request_rules:
         method: squash
         strict: smart+fasttrack
         commit_message: title+body
+  - name: automatic merge for Dependabot pull requests
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
+      - status-success=Travis CI - Pull Request
+      - base=master
+    actions:
+      merge:
+        method: squash
+        strict: smart+fasttrack
+        commit_message: title+body
+    


.mergify.yml
```diff
@@ -9,4 +9,15 @@ pull_request_rules:
       merge:
         method: squash
         strict: smart+fasttrack
-        commit_message: title+body
\ No newline at end of file
+        commit_message: title+body
+  - name: automatic merge for Dependabot pull requests
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
```
e00E, line 15: If I read this correctly this allows these accounts: `dependabot[bot]`, `dependabot-preview[bot]`. Is the `[bot]` really part of the author field?

e00E (approved):
No problem with this conceptually.
Updating dependencies without reviewing them is a security risk but we are already not doing that so this doesn't change. And it would be a lot of work to review every dependency we have every time we update.

nlordell (approved):
:tada: :rocket: :metal:

.mergify.yml
nlordell: It looks like the author field is `dependabot-preview[bot]`:
```
$ git log 9b9189ff755ebe3bca0878caff9d1183e4eed790 -1  | grep Author
Author: dependabot-preview[bot] <27856297+dependabot-preview[bot]@users.noreply.github.com>
```

e00E: But if this is the git author field then anyone can set their name to this right? I thought this was referring to the github account, otherwise it doesn't protect against malicious PRs.

nlordell: > But if this is the git author field then anyone can set their name to this right?

Hmm, good point. Maybe we can check the commit GPG signature, since `dependabot` uses signed commits.

fleupold: I created https://github.com/gnosis/dex-services/pull/1125/commits and simulated it on `mergify.io` with the given config:

Screen Shot 2020-07-14 at 16 41 36

While the author of a commit can be forged easily the author of a pull request seems to have to be an authenticated github user. As such the dependabot-bot username needs to be unique.

> Maybe we can check the commit GPG signature, since dependabot uses signed commits.

I don't think we can make mergify rules based on signatures (couldn't find it here: https://doc.mergify.io/conditions.html)

bh2smith (approved):

.mergify.yml
nlordell: > author of a pull request

Oh yea! that makes a lot of sense...

thanks for looking into this.

fleupold:
Turns out mergify never automerges PRs that touch the mergify config 🤦.

Make sense

PR #1116, opened 2020-07-14
Explicitly set nonce in solution submission transaction

e00E:
To make it more clear that this is always the same nonce and that we are
cancelling the same nonce in case of deadline too.

### Test Plan
CI. (Btw, in staging we saw yesterday that cancelling worked)

### DIFF
--- a/core/src/contracts/stablex_contract.rs
+++ b/core/src/contracts/stablex_contract.rs
@@ -132,6 +132,7 @@ pub trait StableXContract {
         claimed_objective_value: U256,
         gas_price: U256,
         block_timeout: Option<usize>,
+        nonce: U256,
     ) -> BoxFuture<'a, Result<(), MethodError>>;
 
     fn past_events<'a>(
@@ -149,7 +150,11 @@ pub trait StableXContract {
     fn send_noop_transaction<'a>(
         &'a self,
         gas_price: U256,
+        nonce: U256,
     ) -> BoxFuture<'a, Result<TransactionResult>>;
+
+    /// The current nonce aka transaction_count.
+    fn get_transaction_count<'a>(&'a self) -> BoxFuture<'a, Result<U256>>;
 }
 
 impl StableXContract for StableXContractImpl {
@@ -273,6 +278,7 @@ impl StableXContract for StableXContractImpl {
         claimed_objective_value: U256,
         gas_price: U256,
         block_timeout: Option<usize>,
+        nonce: U256,
     ) -> BoxFuture<'a, Result<(), MethodError>> {
         async move {
             let (prices, token_ids_for_price) = encode_prices_for_contract(&solution.prices);
@@ -293,7 +299,8 @@ impl StableXContract for StableXContractImpl {
                 // NOTE: Gas estimate might be off, as we race with other solution
                 //   submissions and thus might have to revert trades which costs
                 //   more gas than expected.
-                .gas(5_500_000.into());
+                .gas(5_500_000.into())
+                .nonce(nonce);
 
             method.tx.resolve = Some(ResolveCondition::Confirmed(ConfirmParams {
                 block_timeout,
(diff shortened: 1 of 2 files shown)


nlordell (approved):
:+1: Awesome change! I think this could have led to some subtle issues where different nonces could be used in some racy situations, so I'm happy this if fixed.

core/src/contracts/stablex_contract.rs
```diff
@@ -330,6 +337,7 @@ impl StableXContract for StableXContractImpl {
     fn send_noop_transaction<'a>(
         &'a self,
         gas_price: U256,
+        nonce: U256,
```
fleupold, line 340: Can this method now be called `cancel_transaction`?

core/src/contracts/stablex_contract.rs
```diff
@@ -149,7 +150,11 @@ pub trait StableXContract {
...
     ) -> BoxFuture<'a, Result<TransactionResult>>;
+
+    /// The current nonce aka transaction_count.
+    fn get_nonce<'a>(&'a self) -> BoxFuture<'a, Result<U256>>;
```
fleupold, line 157: Can the name reflect which nonce we are fetching (something like `get_nonce_for_next_transaction` or even `get_transaction_count`)?

core/src/contracts/stablex_contract.rs
```diff
@@ -350,6 +359,21 @@ impl StableXContract for StableXContractImpl {
...
     }
+
+    fn get_nonce<'a>(&'a self) -> BoxFuture<'a, Result<U256>> {
+        use futures::compat::Future01CompatExt as _;
+        async move {
+            let web3 = self.instance.raw_instance().web3();
+            let account = self.account().ok_or_else(|| anyhow!("no account"))?;
+            let address = account.address();
+            web3.eth()
+                .transaction_count(address, None)
```
fleupold, line 370: I think ultimately we might even want to share this state across run-loops so that the next batch uses the current batch's nonce +1 even if the cancellation tx is still pending (right now it would still use the same nonce since transaction_count is based on only mined txs).

fleupold (approved):
The new functions are not really GP contract related. Maybe we can split them out into a separate stablex contract independent trait?

e00E:
> The new functions are not really GP contract related. Maybe we can split them out into a separate stablex contract independent trait?

I have considered this. The most appropriate place is probably something related to our web3 type but since this functionality is currently only used for solution submission I found it more practical and easier to understand to keep it next to that. I'm not opposed to changing this in the future though.

core/src/contracts/stablex_contract.rs
e00E: I'm fine either way but this came up already in https://github.com/gnosis/dex-services/pull/1089#discussion_r450740145 .
Imo send_noop_transaction says exactly what this does since it still works even when there is no transaction to cancel.

core/src/contracts/stablex_contract.rs
e00E: Made an issue for this.

core/src/contracts/stablex_contract.rs
nlordell: I'm on the fence about this one. If the signature was `fn(tx: ethcontract::web3::types::Transaction)` then I would definitely prefer the `cancel_transaction` name, but as is it isn't *really* cancelling a transaction, but sending a no-op transaction with the specified nonce and gas price. That being said, this method is exclusively used for cancelling transactions, so I also think it might make the name better describe what is its intended use (instead of what it does).

I'm pretty indifferent and am happy with both.

bh2smith (approved):

bh2smith:
It looks like coverage decreased all the way to zero here. I am guessing this is the result of a failed build.

core/src/contracts/stablex_contract.rs
fleupold: As we now have changed the other method also to be called exactly what it does and not what it's intended for, I'm fine to leaving this here too. 
Just felt that stating the intention might help to understand what we are trying to do when we use it in another part of the codebase.

PR #1117, opened 2020-07-14
Introduce `Ring` flow type and refactor transitive orderbook computation

nlordell:
This PR introduces a new method to compute an overlapping ring trade flow over a market. This method is used for computing transitive orderbooks (specifically the part of the transitive orderbook with overlapping orders).

### Test Plan

All unit tests still pass :tada:

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -200,11 +200,14 @@ impl Orderbook {
             let path = path::find_cycle(&predecessors, node, Some(quote))
                 .expect("negative cycle not found after being detected");
 
-            let base_index = match if path.first() == Some(&quote) {
+            let base_index_search = if path.first() == Some(&quote) {
                 path.iter().position(|node| *node == base)
             } else {
+                // NOTE: If the path doesn't start with the quote token, don't
+                // even need to search for the base token.
                 None
-            } {
+            };
+            let base_index = match base_index_search {
                 Some(value) => value,
                 None => {
                     // NOTE: Skip negative cycles that are not along the
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -189,7 +189,7 @@ impl Orderbook {
     /// negative cycles.
     pub fn fill_market_ring_trade(&mut self, market: Market) -> Option<Ring> {
         if !self.is_token_pair_valid(market.bid_pair()) {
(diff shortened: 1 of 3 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -146,53 +146,93 @@ impl Orderbook {
...
+    ///
+    /// Note that if this method returns `None`, then the orderbook is
+    /// **partially** reduced. Specifically, the subgraph containing the market
+    /// `quote` token is fully reduced, however, other not connected subgraphs,
+    /// specifically the market `base`'s subgraph in the case where the `quote`
+    /// and `base` token are not part of the same subgraph, may still contain
+    /// negative cycles.
+    pub fn fill_market_ring_trade(&mut self, market: Market) -> Option<Ring> {
         if !self.is_token_pair_valid(market.bid_pair()) {
             return Default::default();
```
e00E, line 192: feels more natural with the changed signature
```suggestion
            return None;
```

e00E (approved):

pricegraph/src/orderbook.rs
```diff
@@ -146,53 +146,93 @@ impl Orderbook {
...
-                    let bid = self
-                        .fill_path(bid)
-                        .expect("bid transitive path not found after being detected");
-                    overlap.bids.push(bid.as_transitive_order());
+            let base_index = match if path.first() == Some(&quote) {
+                path.iter().position(|node| *node == base)
+            } else {
+                None
+            } {
+                Some(value) => value,
```
fleupold, line 208: ubernit: I found the two consecutive match statements slightly hard to read. Maybe splitting them into two is a bit easier on the eye?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FB5Nvq20uYuSBy%2Fgiphy.gif&token=ring" data-gitmeme-token="ring"></a>

pricegraph/src/orderbook.rs
nlordell: Too bad cause I thought this pattern looked kind of neat :smile: and I read it like "match, if the first node is the quote, then the position of the base token, else match nothing".

That being said, I completely agree it's a little hard to parse at a quick glance and not a pattern we are used to seeing.

PR #1118, opened 2020-07-14
Separate API operations into modules

nlordell:
This PR just splits the `Pricegraph` `impl` blocks into multiple submodules. The ultimate plan is to lift `Orderbook` methods into the `Pricegraph` type (for computing transitive orderbooks and price estimates) so that the `Orderbook` type only contains primitives for manipulating the orderbook graph while the `Pricegraph` type actually computes things by combining these primitives.

This is in an effort to reduce the size of `orderbook.rs` which is currently >1000L long :disappointed:. The motivation for splitting operations into their own modules is to allow the `Pricegraph` API surface to grow without creating another run-away module with way too much code that becomes hard to navigate.

In the following PRs, I plan on:
- move `Orderbook::{reduce_overlapping_transitive_orderbook, fill_transitive_orders}` to `src/api/transitive_orderbook.rs`, implemented with the new `Orderbook::{fill_optimal_transitive_order_if, fill_market_ring_trade}` primitive operations
- move `Orderbook::{fill_market_order, fill_order_at_price}` to `src/api/price_estimation.rs` module implemented with the `Orderbook::fill_optimal_transitive_order_if` primitive.

This has the added benefit that each module will contain unit tests specific to the operation, so hopefully the code will be easier to test and develop for.

One thing I don't 100% like about the structure is that `Pricegraph` methods are split into multiple files, which might be a bit counter-intuitive and make methods harder to find. Another option here would be to implement "helper" functions for each operations in these modules and have `Pricegraph` methods wrap them in `src/lib.rs`. However, this feels like extra work (double method signature for each operation) so I'm not sure if its worth it. Yet another option would be to implement operations as traits. This would be nice since the traits and implementations could live in the same file, and make `Pricegraph` more easily mockable, but with the downside that traits need to be `use`-ed in order to call the methods. I'm open to suggestions.

### Test Plan

Just moved code around, no substantial changes.

### DIFF
--- a/pricegraph/src/api.rs
+++ b/pricegraph/src/api.rs
@@ -1,0 +1,75 @@
+//! Module containing the high-level `Pricegraph` API operation implementations.
+
+mod price_estimation;
+mod transitive_orderbook;
+
+pub use self::transitive_orderbook::TransitiveOrderbook;
+use crate::encoding::{TokenId, TokenPair};
+use crate::FEE_FACTOR;
+
+/// A struct representing a transitive order for trading between two tokens.
+///
+/// A transitive order is defined as the transitive combination of multiple
+/// orders into a single equivalent order. For example consider the following
+/// two orders:
+/// - *A*: buying 1.0 token 1 selling 2.0 token 2
+/// - *B*: buying 4.0 token 2 selling 1.0 token 3
+///
+/// We can define a transitive order *C* buying 1.0 token 1 selling 0.5 token 3
+/// by combining *A* and *B*. Note that the sell amount of token 3 is limited by
+/// the token 2 capacity for this transitive order.
+///
+/// Additionally, a transitive order over a single order is equal to that order.
+#[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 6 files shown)


nlordell:
> Do the Pricegraph native and transitive_orderbook methods share a lot of instance state? If not maybe it's a sign that they should indeed be separate sub-objects and the pricegraph is just a Facade around them.

They do not - the `Pricegraph` is a facade around `Orderbook` and `ReducedOrderbook` types. Each operation starts with a "fresh" graph (i.e. the `(Reduced)?Orderbook` is `clone()`-d).

> Otherwise I like the trait idea, but not sure it is needed yet.
>  don't think there is a need to [...] put them into traits.

:+1: I think the main goal currently is just to split each operation (or related group of operations) into their own modules because they were just dumped into the `src/orderbook.rs` module which became too beastly. I also like the trait idea, and if it keeps on growing I will give it a try. For now, I will just have `impl Pricegraph` blocks split into multiple files.

PR #1120, opened 2020-07-14
Allow price estimator to reuse the driver's default price source

e00E:
The price estimator needs to use the same price source as we use in the
driver to compute the rounding buffer based on this information.
This commit makes it easier to create this price source by moving it
into it's own function. The price estimator does not need a full price oracle, just the price source.

### Test Plan
No observable change.

### DIFF
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -48,6 +48,40 @@ pub struct PriceOracle {
     source: Box<dyn PriceSource + Send + Sync>,
 }
 
+pub fn default_price_source(
+    http_factory: &HttpFactory,
+    orderbook_reader: Arc<dyn StableXOrderBookReading>,
+    token_data: TokenData,
+    token_info_fetcher: Arc<dyn TokenInfoFetching>,
+    update_interval: Duration,
+) -> Result<Box<dyn PriceSource + Send + Sync>> {
+    let (kraken_source, _) = ThreadedPriceSource::new(
+        token_info_fetcher.clone(),
+        KrakenClient::new(http_factory, token_info_fetcher.clone())?,
+        update_interval,
+    );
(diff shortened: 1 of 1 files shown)


core/src/price_estimation.rs
```diff
@@ -48,6 +48,40 @@ pub struct PriceOracle {
     source: Box<dyn PriceSource + Send + Sync>,
 }
 
+pub fn default_price_source(
```
fleupold, line 51: nit: Should this method live in the price_source module directly?

fleupold (approved):
Nice putting this into a small, easy to review PR.

core/src/price_estimation.rs
e00E: The PriceOracle is what we use to give prices to the solver and this set up of average and priority is fairly specific to it so it makes sense here.

fleupold: Right, but if we want to reuse this setup in the price-estimator it's no longer specific to the PriceOracle but more to the PriceSource trait in general, or am I missing something?

e00E: In my mind, which price sources are used is decided by the price oracle. The price estimator just copies that. Looking at the modules the price source files just defines the trait and the implementations of it live in files on the same level. The price_estimator file is one level higher and uses them all. So that file is really the one bringing it all together.
It's not a big thing, just felt a little more appropriate that way.

e00E:
This PR turns out not that useful because we want more control over the price sources in the price estimator. For example the pricegraph estimator instead of being created from a StableXOrderBookReading, can be more optimized as we already have a pricegraph instance internally that we update.
I think the better approach will be manually recreate the same price source in price estimator for now and later see what the best way to avoid copying code is.

PR #1124, opened 2020-07-14
Rename methods for more consistent language

nlordell:
This PR renames some methods to have more consistent language. Specifically, in `pricegraph` there is the concept of a "limit price" which is a ratio of an order's buy amount over its sell amount. This is part of the order representation in the `BatchExchange` smart contract. These values implicitly include fees and are modeled internally by the `LimitPrice` type. Conversely, there is also the concept of an "exchange rate" which is an internal exchange rate that is used in the orderbook graph calculations for computing flow though the projection graph, these are modeled internally by the `ExchangeRate` type. However, the `Pricegraph` API only deals with `BatchExchange` "limit prices", so fix the method names to reflect this.

### Test Plan

Just method renames, so `rustc` is the test plan!


bh2smith (approved):

fleupold (approved):
🧹

e00E (approved):

PR #1129, opened 2020-07-15
Make pricegraph price source threaded

e00E:
This is nice to have for the price estimator. It also makes sense here
as the calculation can be expensive. It does mean that its prices could
be slightly out of date compared to before this commit but considering
that we didn't have this price source at all previously this should be
fine.
While doing this I noticed a problem that we already had when all price
sources are threaded: technically we could start the solver without
having fetched prices at least once. The threaded price source updates
itself immediately but it is still possible this could take longer than
the code path that runs the solver in the first batch that we handle
after startup.

### Test Plan
CI, new tests for the wait function.

### DIFF
--- a/core/src/price_estimation/price_source.rs
+++ b/core/src/price_estimation/price_source.rs
@@ -36,6 +36,48 @@ impl PriceSource for NoopPriceSource {
     }
 }
 
+/// Wait for the price source to have a price for this token.
+/// Useful for ThreadedPriceSource to ensure that it has updated at least once.
+pub async fn wait_for_price(price_source: &dyn PriceSource, token_id: TokenId) {
+    loop {
+        let has_price = match price_source.get_prices(&[token_id]).await {
+            Ok(prices) => prices.get(&token_id).is_some(),
+            Err(_) => false,
+        };
+        if has_price {
+            return;
+        }
+        async_std::task::sleep(std::time::Duration::from_secs(1)).await;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn wait_for_price_returns() {
+        let mut price_source = MockPriceSource::new();
+        price_source
+            .expect_get_prices()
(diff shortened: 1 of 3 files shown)


core/src/price_estimation/price_source.rs
```diff
@@ -36,6 +36,48 @@ impl PriceSource for NoopPriceSource {
     }
 }
 
+/// Wait for the price source to have a price for this token.
+/// Useful for ThreadedPriceSource to ensure that it has updated at least once.
+pub async fn wait_for_price(price_source: &dyn PriceSource, token_id: TokenId) {
```
nlordell, line 41: Maybe this should be implemented as an `initialize` method part of the `PriceSource` trait with a default no-op implementation (kind of how we do it with the `OrderbookReading`). This way the `ThreadedPriceSource` can use something like a `oneshot` channel to properly notify initialization instead of polling for token 0's price, which seems a bit incorrect to me, since its not part of the `PriceSource` contract that a price for token 0 *must* be provided. In fact, since token 0 has a known price, I wouldn't be surprised if a `PriceSource` never provided a price estimate for it.

nlordell (changes requested):

e00E:
I have to think more about this. Neither way of initialization felt right.

fleupold (commented):
I'm not sure I fully understand why the PriceGraph should be threaded? Doesn't it only take a couple of ms to compute prices? 

I'm usually a bit sceptical of introducing threaded components as it makes the control flow harder to reason about (e.g. if we try to debug a weird instance it might be because price estimates were based on a block that didn't contain all orders).

As for the initialisation I see how this is an issue. However, I wonder if this could be kept internal to the ThreadedPriceSource and basically make `get_prices` blocking if it has never set its `price_map`?

fleupold:
I think I see now why we need it to be threaded (for fast access in the price-estimation service). In that case maybe the setup over there should wrap the default source into a threaded price source?

PR #1130, opened 2020-07-15
Use solver based rounding buffer

e00E:
and removed the generics from the orderbook struct in price estimator because this is easier to read.

The rounding buffer has been properly implemented for `estimate_buy_amount` but it's not clear what correct means in other cases (see the TODO comments in filter.rs).

### Test Plan
Run the price estimator from master, the price estimator from this PR and the price estimator from this PR with rounding-buffer-factor set to 0.0. Compare results. We see that in the estimate-buy-amount route the buy amount is the lowest for the master price estimator which is lower than the this PR's default price estimator which is lower than this PR's price estimator with no rounding buffer. Example:

```
curl "localhost:8080/api/v1/markets/4-7/estimated-buy-amount/1?atoms=false"
{"baseTokenId":4,"quoteTokenId":7,"buyAmountInBase":"1.0080441574595185","sellAmountInQuote":"1"}
curl "localhost:8081/api/v1/markets/4-7/estimated-buy-amount/1?atoms=false"
{"baseTokenId":4,"quoteTokenId":7,"buyAmountInBase":"1.008854374703835","sellAmountInQuote":"1"}
curl "localhost:8082/api/v1/markets/4-7/estimated-buy-amount/1?atoms=false"
{"baseTokenId":4,"quoteTokenId":7,"buyAmountInBase":"1.0090532106701886","sellAmountInQuote":"1"}
```

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -14,25 +14,16 @@ use warp::{
 };
 
 /// Handles all supported requests under a `/api/v1` root path.
-pub fn all<T: Send + Sync + 'static>(
-    orderbook: Arc<Orderbook<T>>,
+pub fn all(
+    orderbook: Arc<Orderbook>,
     token_info: Arc<dyn TokenInfoFetching>,
-    price_rounding_buffer: f64,
 ) -> impl Filter<Extract = impl Reply, Error = Infallible> + Clone {
     warp::path!("api" / "v1" / ..)
         .and(
             markets(orderbook.clone())
-                .or(estimated_buy_amount(
-                    orderbook.clone(),
-                    token_info.clone(),
-                    price_rounding_buffer,
-                ))
-                .or(estimated_amounts_at_price(
-                    orderbook.clone(),
-                    token_info,
-                    price_rounding_buffer,
-                ))
-                .or(estimated_best_ask_price(orderbook, price_rounding_buffer)),
+                .or(estimated_buy_amount(orderbook.clone(), token_info.clone()))
+                .or(estimated_amounts_at_price(orderbook.clone(), token_info))
+                .or(estimated_best_ask_price(orderbook)),
         )
         .recover(handle_rejection)
 }
@@ -67,8 +58,8 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
 /// Validate a request of the form
 /// `/markets/<baseTokenId>-<quoteTokenId>`
 /// and answer it.
-fn markets<T: Send + Sync + 'static>(
-    orderbook: Arc<Orderbook<T>>,
+fn markets(
+    orderbook: Arc<Orderbook>,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     markets_filter()
         .and(warp::any().map(move || orderbook.clone()))
@@ -79,13 +70,11 @@ fn markets<T: Send + Sync + 'static>(
 /// Validate a request of the form
 /// `/markets/<baseTokenId>-<quoteTokenId>/estimated-buy-amount/<sellAmountInQuoteToken>`
 /// and answer it.
-fn estimated_buy_amount<T: Send + Sync + 'static>(
-    orderbook: Arc<Orderbook<T>>,
+fn estimated_buy_amount(
+    orderbook: Arc<Orderbook>,
     token_info: Arc<dyn TokenInfoFetching>,
-    price_rounding_buffer: f64,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     estimated_buy_amount_filter()
-        .and(warp::any().map(move || price_rounding_buffer))
         .and(warp::any().map(move || orderbook.clone()))
         .and(warp::any().map(move || token_info.clone()))
         .and_then(estimate_buy_amount)
@@ -95,13 +84,11 @@ fn estimated_buy_amount<T: Send + Sync + 'static>(
 /// Validate a request of the form:
 /// `/markets/<baseTokenId>-<quoteTokenId>/estimated-amounts-at-price/<exchangeRate>`
 /// and answer it.
-fn estimated_amounts_at_price<T: Send + Sync + 'static>(
-    orderbook: Arc<Orderbook<T>>,
+fn estimated_amounts_at_price(
+    orderbook: Arc<Orderbook>,
     token_info: Arc<dyn TokenInfoFetching>,
-    price_rounding_buffer: f64,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     estimated_amounts_at_price_filter()
-        .and(warp::any().map(move || price_rounding_buffer))
         .and(warp::any().map(move || orderbook.clone()))
         .and(warp::any().map(move || token_info.clone()))
         .and_then(estimate_amounts_at_price)
@@ -111,12 +98,10 @@ fn estimated_amounts_at_price<T: Send + Sync + 'static>(
 /// Validate a request of the form:
 /// `/markets/<baseTokenId>-<quoteTokenId>/estimated-amounts-at-price/<exchangeRate>`
 /// and answer it.
-fn estimated_best_ask_price<T: Send + Sync + 'static>(
-    orderbook: Arc<Orderbook<T>>,
-    price_rounding_buffer: f64,
+fn estimated_best_ask_price(
+    orderbook: Arc<Orderbook>,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     estimated_best_ask_price_filter()
-        .and(warp::any().map(move || price_rounding_buffer))
         .and(warp::any().map(move || orderbook.clone()))
         .and_then(estimate_best_ask_price)
         .with(warp::log("price_estimator::api::estimate_best_ask_price"))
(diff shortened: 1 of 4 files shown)


price-estimator/src/filter.rs
```diff
@@ -161,13 +146,15 @@ fn estimated_best_ask_price_filter(
...
 
-async fn get_markets<T>(
+async fn get_markets(
     market: Market,
     _query: QueryParameters,
-    orderbook: Arc<Orderbook<T>>,
+    orderbook: Arc<Orderbook>,
 ) -> Result<impl Reply, Infallible> {
+    // TODO: should this route make use of the rounding buffer? This wasn't the case with the
+    // previous rounding buffer.
```
e00E, line 155: TODO

price-estimator/src/filter.rs
```diff
@@ -256,29 +229,29 @@ async fn estimate_amounts_at_price<T>(
...
+        .map(|order| (order.buy, order.sell))
         .unwrap_or_default();
+    // TODO: What to do about the rounding buffer? The order we get with this exchange rate
+    // is too optimistic because the solver will see it as if its sell amount had been reduced by
+    // the rounding buffer.
+    // If we increase the sell amount here to make up for this then buy_amount / sell_amount is no
+    // longer equal to price_in_quote.
+    // This might not be a problem in practice because the orderbook has already been
+    // conservatively worsened with double the rounding buffer than what the solver actually uses.
+    // Or we might be okay with the ratio equality no longer being exactly true.
```
e00E, line 252: TODO

price-estimator/src/filter.rs
```diff
@@ -287,35 +260,34 @@ async fn estimate_amounts_at_price_atoms<T>(
...
-async fn estimate_best_ask_price<T>(
+async fn estimate_best_ask_price(
     token_pair: TokenPair,
     _query: QueryParameters,
-    price_rounding_buffer: f64,
-    orderbook: Arc<Orderbook<T>>,
+    orderbook: Arc<Orderbook>,
 ) -> Result<impl Reply, Infallible> {
+    // TODO: Similar rounding buffer issue as above. Here we only have an exchange rate but the true
+    // exchange rate depends on the amounts.
```
e00E, line 269: TODO

price-estimator/src/main.rs
```diff
@@ -125,7 +155,7 @@ async fn update_orderbook_forever(orderbook: Arc<Orderbook>, update_interval: Du
     loop {
         time::delay_for(update_interval).await;
         if let Err(err) = orderbook.update().await {
-            log::warn!("error updating orderbook: {:?}", err);
+            log::error!("error updating orderbook: {:?}", err);
```
e00E, line 158: We probably do want to get an alert when this fails, so error.

e00E:
It's orthogonal but the dyn change is small and uncontroversial so I don't think it's worth the time to extract out into another PR.

price-estimator/src/filter.rs
fleupold: Given that rounding went from a factor to a summand I think it's fine to not consider it for an epsilon amount.

fleupold:
Also I forgot to mention that this is great progress and I'm really excited to see this coming along!

price-estimator/src/filter.rs
```diff
@@ -198,13 +184,12 @@ async fn estimate_buy_amount<T>(
...
-        .order_for_sell_amount(token_pair, sell_amount_in_quote_atoms);
-    let buy_amount_in_base_atoms = transitive_order
-        .map(|order| apply_rounding_buffer(order.buy, price_rounding_buffer))
-        .unwrap_or_default();
+    let rounding_buffer = orderbook.rounding_buffer(token_pair).await;
+    let transitive_order = orderbook.pricegraph().await.order_for_sell_amount(
+        token_pair,
+        f64::max(sell_amount_in_quote_atoms - rounding_buffer, 0.0),
+    );
+    let buy_amount_in_base_atoms = transitive_order.map(|order| order.buy).unwrap_or_default();
```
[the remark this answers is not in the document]
e00E: Before we were for first calling `order_for_sell_amount` then decreasing the buy amount of the resulting order.
Now we are first reducing the sell amount and then leaving the resulting order untouched.
We are doing it this way because this reduced sell amount is what the solver would see after applying the rounding buffer.

price-estimator/src/filter.rs
e00E: If we change the buy amount by the same ratio then the result is no longer correct as in reality the solver will see a slightly lower sell amount which could mean that the order based on this won't get matched.

price-estimator/src/orderbook.rs
```diff
@@ -1,35 +1,83 @@
...
+    pub async fn rounding_buffer(&self, token_pair: TokenPair) -> f64 {
+        let fee_token = TokenId(0);
+        let sell_token = TokenId(token_pair.sell);
+        let buy_token = TokenId(token_pair.buy);
+        let prices = self
+            .price_source
+            .get_prices(&[fee_token, sell_token, buy_token])
+            .await
+            .unwrap_or_default();
+        let get_price = |token_id| prices.get(&token_id).copied().unwrap_or(1) as f64;
```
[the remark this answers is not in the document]
e00E: Both here and `solver_rounding_buffer` we need to choose some value for the price if we don't have it. It is unclear what reasonable is. Now that you mention it more reasonable would probably be to reuse the fee price and if we also don't have the fee price then use 1 for that.

price-estimator/src/orderbook.rs
```diff
@@ -1,35 +1,83 @@
...
     /// Recreate the pricegraph orderbook.
     pub async fn update(&self) -> anyhow::Result<()> {
-        let (account_state, orders) = self
+        let (mut account_state, mut orders) = self
             .orderbook_reading
             .get_auction_data(current_batch_id() as u32)
             .await?;
 
+        let token_prices = self.price_source.get_prices(&self.all_tokens).await?;
+        solver_rounding_buffer::apply_rounding_buffer(
```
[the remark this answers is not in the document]
e00E: Why keep it around when it's not used? The question I now see related to this is that we are actually changing /markets because it is working on these adjusted orders which wasn't the case before.

e00E:
> This is quite a big PR. Not because the generics change, although I also prefer to have precisely small and uncontroversial changes to be separate PRs. This way there is less code I have to categorize when parsing such a meaty PR.
Also maybe the changes in the Pricegraph and in the filter could be 2 separate PRs (even if this means we are increasing the rounding buffer in the short term).

I'm aware of this. It's always a question of practicality because making separate PRs takes more time and our internal measurement when a PR should be split is not always the same. I'll try to move some parts into separate PRs.

e00E:
> Also, is there a way for us to run this setup against the historic data (using only the PriceGraph as a PriceSource) and compare how much effect on accuracy the change in rounding buffer has?

Not in this PR and it doesn't seem that useful when we have external price sources. Do we want this feature now? If so we should implement and merge it before this PR.

price-estimator/src/filter.rs
fleupold: I see. So in order to be completely correct we would have to pass the rounding buffer all the way down here: https://github.com/gnosis/dex-services/blob/04da5aa18a2bc790a21ac9d9b68a65f3911cd85e/pricegraph/src/api/price_estimation.rs#L93-L100

There we have the concrete amounts and could apply the rounding on both sides. But I'm not sure we want the rounding concept to be that intrusive.

We could also do an "iterative" approach where we apply the rounding buffer to the sell amount here in the top level method and then check via `estimate_buy_amount` if we can get a buyAmount at least as high as the desired ratio. Otherwise we need to reduce the amount until we do.

price-estimator/src/orderbook.rs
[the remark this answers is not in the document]
fleupold: Shouldn't this consider the decimals of the token? What would be a pessimistic default here? A high price or a low price? We could also just fall back to 0.1% rounding in case there is no price available.

price-estimator/src/orderbook.rs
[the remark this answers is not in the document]
fleupold: I don't have a strong feeling about this one, but my intuition would be that markets should just show the state of the current order book as is. With price estimates we are actually making a soft commitment that if you use these prices, you will get matched (thus we need to apply our internal rounding buffer). But I can also see how you could argue for consistency, thus applying it to both.

price-estimator/src/orderbook.rs
```diff
@@ -1,35 +1,83 @@
...
             orderbook_reading,
+            price_source,
+            all_tokens,
             pricegraph: RwLock::new(Pricegraph::new(std::iter::empty())),
+            rounding_buffer_factor,
         }
     }
 
-    pub async fn get_pricegraph(&self) -> Pricegraph {
+    pub async fn pricegraph(&self) -> Pricegraph {
```
fleupold, line 42: Can we make it clear from the name that this pricegraph has rounding buffers applied to it?

price-estimator/src/filter.rs
nlordell: I think we should be using the `order_for_limit_price` instead of `order_at_limit_price`. This will return true buy and sell amounts, such that `buy/sell > limit_price`. You can then use @fleupold suggestion and iterate to converge to proper volumes. In pseudo-code:
```rust
        let mut current_limit_price = limit_price;
        let mut order;
        while {
            order = pricegraph.order_for_limit_price(pair, current_limit_price);
            order.sell += rounding_buffer;
            order.limit_price() < limit_price
        } {
            current_limit_price = order.limit_price();
        }
        order.buy = order.sell * limit_price;
```

price-estimator/src/filter.rs
nlordell: The issue I see with providing an orderbook with rounding buffers applied is that you place an order for 1ETH for 200DAI it will show up as an order for 1ETH for 199.9999DAI which is a little counter intuitive.

price-estimator/src/orderbook.rs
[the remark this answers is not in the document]
nlordell: > I don't have a strong feeling about this one, but my intuition would be that markets should just show the state of the current order book as is. 

I agree with this - it would be counter-intuitive for a user to place an order and see it in the market orderbook page with a different price. Also, not applying the rounding buffer would keep it consistent with how it works today (to my knowledge).

Issue #1099, opened 2020-07-09
[Driver] make the parameter --min-avg-fee-per-order dependent on current gas prices


josojo:
Currently, we have the problem that sometimes, when the gas prices on ethereum are peaking, we are not submitting any solution, since our gas_cap is below the prices prevailing on ethereum.

In order to introduce higher gas caps, without being un-economic, I am proposing that the driver should estimate the gas_price needed to make a quick solution submission before it is starting the solver. With this gas_price estimate, we can calculate the minimum --min-avg-fee-per-order such that the order settlement is economically viable, given the gas costs per order settlement is around 120k gas.

At current gas prices of 40 Gwei, we will end up with a minimal economical order size of around 1200 $ (120000 * 40 * 10**9 * 240 *1000). As this number is very high and Gnosis is willing to subsidize this number, I think we should introduce a constant `subsidize_factor`, by which this calculated minimal economically viable --min-avg-fee-per-order is reduced.

Example:
```
subsidize_factor:=10
Gas price: 40 Gwei
=> min economical viable fee = 1.2 OWL
=> Handover to the solver: --min-avg-fee-per-order= 1.2/10
```
wdyt?

(Once this logic is introduced, we should also introduce it in the front-end @anxolin, to warn users that their orders might not get settled )

fleupold:
I think this is a very important issue and a step towards making us future proof and long term sustainable.

One thing to keep in mind is that we are doubling the gas if our solution doesn't get mined within 2 blocks of submission. We probably want to set the `min-avg-fee-per-order` param higher than the current gas price would specify to allow at least one doubling.

We also need to access the current price estimate of ETH in OWL (since a low ETH price will make the min viable OWL price smaller).

e00E:
> Going forward, we should not only implement #1099 but also make it that gas cap depends on the total fee that can be gained by the solver. In the example today we could have made $50 in fees. This should make us be willing to pay at least $50 in gas to get our tx mined.

nlordell:
Implemented by #1198

PR #1131, opened 2020-07-15
higher gas_cap limit

josojo:
For the ETH-DAI auctions it is important that we are getting our transactions into the batches, as the orders have only validity of 1 batch. With the current prevailing gas prices, this is not given with our current gas_cap.

Longterm solution is described here:https://github.com/gnosis/dex-services/issues/1099
### Test Plan
non


e00E (approved):

nlordell (approved):

bh2smith:
This is great! I guess the plan is to have the dev solvers running this update immediately!

PR #1135, opened 2020-07-16
Make price estimator orderbook use Box instead of generics

e00E:
This is easier to read and we don't need it to be generics. Refactored out from the big rounding buffer PR.

### Test Plan
CI, no logic change.


fleupold (approved):
👍 Thanks

nlordell (approved):

PR #1136, opened 2020-07-16
Refactor price estimator rounding buffer calculation

e00E:
We will need the separate method when integrating the rounding buffer
with the rest of the price estimator.

### Test Plan
CI, no logic change

### DIFF
--- a/price-estimator/src/solver_rounding_buffer.rs
+++ b/price-estimator/src/solver_rounding_buffer.rs
@@ -1,11 +1,11 @@
-use core::models::{AccountState, Order, TokenId, TokenInfo};
+use core::models::{AccountState, Order, TokenId};
 use ethcontract::{Address, U256};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 
 // This code is closely related to dex-solver/src/opt/process/Rounding.py .
 // Discussion of motivation happened in https://github.com/gnosis/dex-services/issues/970 .
 
-type Tokens = BTreeMap<TokenId, TokenInfo>;
+type TokenPrices = HashMap<TokenId, u128>;
 
 const MAX_ROUNDING_VOLUME: f64 = 100_000_000_000.0;
 const PRICE_ESTIMATION_ERROR: f64 = 10.0;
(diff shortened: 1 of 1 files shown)


price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -16,45 +16,50 @@ fn max_rounding_amount(token_price: f64, fee_token_price: f64) -> f64 {
...
-        let estimated_xrate = estimated_buy_token_price / estimated_sell_token_price;
-        let rounding_buffer = max_rounding_amount(estimated_buy_token_price, fee_token_price)
-            * estimated_xrate
-            * PRICE_ESTIMATION_ERROR.powi(2);
+
         // Multiply by an extra factor which the solver doesn't do, as added safety in case the
         // prices move.
-        let rounding_buffer = (pessimistic_factor * rounding_buffer) as u128;
+        let rounding_buffer = (rounding_buffer(fee_token_price, sell_token_price, buy_token_price)
+            * extra_factor) as u128;
```
fleupold, line 62: I suggest moving the extra_factor inside the public method, so we force consumers to remember it when using this method.

price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -16,45 +16,50 @@ fn max_rounding_amount(token_price: f64, fee_token_price: f64) -> f64 {
...
     // Apply rounding buffer to account balances and order sell amounts.
     for order in orders.iter_mut() {
-        // Compute rounding buffer.
         // If a token doesn't have a price it means it isn't connected to the fee. In this case we
         // use a price of 1 to get some reasonable default rounding buffer.
         let (sell_token, buy_token) = (TokenId(order.sell_token), TokenId(order.buy_token));
-        let estimated_buy_token_price = match tokens.get(&buy_token) {
-            Some(TokenInfo { external_price, .. }) if *external_price > 0 => *external_price as f64,
+        let buy_token_price = match token_prices.get(&buy_token) {
+            Some(price) if *price > 0 => *price as f64,
```
fleupold, line 51: Should the default consider the token's decimals? Depending on if a higher price leads to a larger or lesser rounding buffer we should probably choose a price of `10^^(decimals +/- 5 or the like)`. This would allow to cover prices from 0,0001 to 10000

fleupold: Just saw #1137, let's discuss there.

e00E: This is a good point. I've already changed this in another PR, so let's discuss it there.

price-estimator/src/solver_rounding_buffer.rs
e00E: Done

fleupold (approved):

PR #1137, opened 2020-07-16
Handle tokens without price in rounding buffer calculation

e00E:
The rounding buffer calculation takes token prices as input. If we don't
have a price for an order we need to pick some default or we can't
perform the calculation.
Previously we used `1` as the default but it seems more correct to reuse
the fee price (10e18) instead of 1. Either way is a guess but this is a
better heuristic.

### Test Plan
This function is not yet used.

### DIFF
--- a/price-estimator/src/solver_rounding_buffer.rs
+++ b/price-estimator/src/solver_rounding_buffer.rs
@@ -5,8 +5,6 @@ use std::collections::HashMap;
 // This code is closely related to dex-solver/src/opt/process/Rounding.py .
 // Discussion of motivation happened in https://github.com/gnosis/dex-services/issues/970 .
 
-type TokenPrices = HashMap<TokenId, u128>;
-
 const MAX_ROUNDING_VOLUME: f64 = 100_000_000_000.0;
 const PRICE_ESTIMATION_ERROR: f64 = 10.0;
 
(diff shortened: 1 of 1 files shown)


price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -45,15 +46,15 @@ pub fn apply_rounding_buffer(
     // Apply rounding buffer to account balances and order sell amounts.
     for order in orders.iter_mut() {
         // If a token doesn't have a price it means it isn't connected to the fee. In this case we
-        // use a price of 1 to get some reasonable default rounding buffer.
+        // reuse the fee token price to get some reasonable default rounding buffer.
         let (sell_token, buy_token) = (TokenId(order.sell_token), TokenId(order.buy_token));
         let buy_token_price = match token_prices.get(&buy_token) {
             Some(price) if *price > 0 => *price as f64,
-            _ => 1.0,
+            _ => fee_token_price,
```
fleupold, line 53: Should the default consider the token's decimals? Depending on if a higher price leads to a larger or lesser rounding buffer we should probably choose a price of 10^^(decimals +/- 5 or the like). This would allow to cover prices from 0,0001 to 10000

e00E: Assuming that tokens usually have decimals set so that one base unit is very roughly one USD this would make sense.
I'm thinking a good way to sidestep this here would be to instead of taking the HashMap with prices take a function `TokenId -> Price`, then the caller has to decide how they want to implement it.
In a future PR that has to call this function we can then do something like what you describe.

e00E: Implemented this idea as another commit in this PR. The discussion is still valid but this module won't have to make the decision.

price-estimator/src/solver_rounding_buffer.rs
```diff
@@ -27,34 +25,25 @@ pub fn rounding_buffer(fee_token_price: f64, sell_token_price: f64, buy_token_pr
 
 /// Perform the same rounding buffer calculation as our solvers in order to increase the correctness
 /// of our estimates.
-/// All token prices must be nonzero.
+/// token_prices returns the price of a token like in PriceSource::get_prices. All token prices
+/// must be nonzero.
 #[allow(dead_code)]
 pub fn apply_rounding_buffer(
-    token_prices: &TokenPrices,
+    token_prices: &dyn Fn(TokenId) -> u128,
```
nlordell, line 32:
```suggestion
    token_prices: impl Fn(TokenId) -> u128,
```

price-estimator/src/solver_rounding_buffer.rs
fleupold: We might want to have different default behaviour based on whether it's the buy or sell token. Given the buffer depends on 

`let estimated_xrate = estimated_buy_token_price / estimated_sell_token_price;`

we might want the buy_token_price fallback to something large (10^(decimals+5)) and the sell_token_price fall back to something small (10^(decimals-5))

I think to fine-tune this it would be nice to see what the impact of these changes would be on accuracy. But for now, I think it's fine assuming 10^decimals (we should document somewhere all the places on which we might fine-tune later).

fleupold (approved):

PR #1138, opened 2020-07-16
implement historic price estimates

### LINKED ISSUE #967: [Price Estimation] Get price estimate/orderbooks for historic blockNumber/BatchId
For various purposes it will be helpful to get a historical price quote based on the on-chain orderbook at that time. Some use cases are:

- Run a cronjob, that Identifies past orders placed with "reasonable prices" (prices that should have been matched according to the price estimation) but that did not result in a trade (one of our top level customer facing KPIs)
- Compare how good our prices to those of other DEXs by taking their past trades and comparing the achieved price with the price that GP would have offered at that time. 

I'd suggest to use the same endpoint as today, but provide an optional argument (e.g. `block` which could be a number or in the future even `pending` - for virtual order placement).

nlordell:
This PR adds initial support for computing historic prices. In a follow up PR, I will ensure that the `OrderbookReading` implementation doesn't re-query the node for events if the batch ID is sufficiently in the past (to speed up the request a bit).

Closes #967 

### Test Plan

Run the price estimator and check historic prices - note that it is currently not an error checking future batches.
```
$ curl -s 'http://localhost:8080/api/v1/markets/1-7/estimated-best-ask-price?atoms=true' | jq
230.43986139024014
$ curl -s 'http://localhost:8080/api/v1/markets/1-7/estimated-best-ask-price?atoms=true&batchId=5290000' | jq
150.08262567963578
$ curl -s 'http://localhost:8080/api/v1/markets/1-7/estimated-best-ask-price?atoms=true&batchId=9999999' | jq
230.43986139024014
```

### DIFF
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -4,7 +4,7 @@ All endpoints use the query part of the url with these key-values:
 
 * `atoms`: Required. If set to `true` all amounts are denominated in the smallest available unit (base quantity) of the token. If `false` all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. TODO: `false` is currently only implemented for estimated-buy-amount and estimated-amounts-at-price .
 * `hops`: Optional. TODO: document this once it has been implemented.
-* `batchId`: Optional. Specify a specific batch ID to compute the estimate for, only accounting orders that are valid at the specified batch.
+* `batchId`: Optional. Specify a specific batch ID to compute the estimate for, only accounting orders that are valid at the specified batch. If no batch ID is specified, the current batch that is collecting orders will be used.
 
 Example: `<path>?atoms=true`
 
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -4,6 +4,7 @@ All endpoints use the query part of the url with these key-values:
 
 * `atoms`: Required. If set to `true` all amounts are denominated in the smallest available unit (base quantity) of the token. If `false` all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. TODO: `false` is currently only implemented for estimated-buy-amount and estimated-amounts-at-price .
 * `hops`: Optional. TODO: document this once it has been implemented.
+* `batchId`: Optional. Specify a specific batch ID to compute the estimate for, only accounting orders that are valid at the specified batch.
 
 Example: `<path>?atoms=true`
 
--- a/price-estimator/src/error.rs
+++ b/price-estimator/src/error.rs
@@ -1,0 +1,14 @@
+//! Module implementing a custom warp rejection for internal service errors.
+
+use anyhow::Error;
+use warp::reject::{self, Reject, Rejection};
+
+#[derive(Debug)]
+pub struct InternalError(pub Error);
+
+impl Reject for InternalError {}
+
+/// Create a
+pub fn internal_server_rejection(err: Error) -> Rejection {
+    reject::custom(InternalError(err))
+}
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -1,10 +1,9 @@
-use crate::models::*;
-use crate::orderbook::Orderbook;
+use crate::{error, models::*, orderbook::Orderbook};
 use core::{
     models::TokenId,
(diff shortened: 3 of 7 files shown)


pricegraph/src/api/price_estimation.rs
```diff
@@ -33,7 +33,7 @@ impl Pricegraph {
             orderbook.fill_optimal_transitive_order_if(inverse_pair, |flow| {
                 last_exchange_rate = Some(flow.exchange_rate);
                 false
-            })?;
+            });
```
nlordell, line 36: **NOTE**: This is a fix to a regression introduced in the `Pricegraph` API during refactoring because of insufficient test coverage. I am working on a separate PR that addresses this issue that should be merged before this one.

price-estimator/src/filter.rs
```diff
@@ -202,8 +203,9 @@ async fn estimate_buy_amount<T>(
         sell_amount_in_quote * token_info.base_unit_in_atoms()
     };
     let transitive_order = orderbook
-        .get_pricegraph()
+        .get_pricegraph(query.batch_id)
         .await
+        .map_err(error::internal_server_rejection)?
```
fleupold, line 208: What error is displayed when we query a future batch?

price-estimator/src/orderbook.rs
```diff
@@ -16,18 +16,14 @@ impl<T> Orderbook<T> {
...
 
 impl<T: StableXOrderBookReading> Orderbook<T> {
-    /// Recreate the pricegraph orderbook.
-    pub async fn update(&self) -> anyhow::Result<()> {
+    /// Creates a `Pricegraph` instance at the given batch ID.
+    async fn create_pricegraph_at_batch(&self, batch_id: BatchId) -> Result<Pricegraph> {
         let (account_state, orders) = self
             .orderbook_reading
-            .get_auction_data(current_batch_id() as u32)
+            .get_auction_data(batch_id.into())
```
fleupold, line 26: This is collecting all orders/balances as of the last block (inclusive) of this batchId, right?

fleupold (approved):
🆒 

Can you also update the readme to document that option?

price-estimator/src/orderbook.rs
nlordell: AFAIK this is inclusive (i.e. the solving batch, as per the `StableXOrderbookReading`).

price-estimator/src/filter.rs
nlordell: I don't think it is an error if we query a future batch, but this can be implemented.

price-estimator/README.md
```diff
@@ -4,6 +4,7 @@ All endpoints use the query part of the url with these key-values:
 
 * `atoms`: Required. If set to `true` all amounts are denominated in the smallest available unit (base quantity) of the token. If `false` all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. TODO: `false` is currently only implemented for estimated-buy-amount and estimated-amounts-at-price .
 * `hops`: Optional. TODO: document this once it has been implemented.
+* `batchId`: Optional. Specify a specific batch ID to compute the estimate for, only accounting orders that are valid at the specified batch.
```
fleupold, line 7: ubernit: document the default behavior.

price-estimator/src/filter.rs
fleupold: Agree this can be a separate PR. I would personally prefer a tighter API that yells at you if you are doing something that doesn't make sense.

fleupold (approved):

price-estimator/src/error.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[derive(Debug)]
+pub struct InternalError(pub Error);
+
+impl Reject for InternalError {}
+
+/// Create a
+pub fn internal_server_rejection(err: Error) -> Rejection {
+    reject::custom(InternalError(err))
+}
```
e00E, line 14: Might want to integrate this with the reject handler in `filter.rs`. I think what happens now is that the handler goes into the "unexpected internal error" branch. Or that is what you intend, just mentioning.

e00E (approved):

price-estimator/src/error.rs
nlordell: #1141

PR #1139, opened 2020-07-16
Refactor orderbook based price source to be more generic

e00E:
which allows it to be used directly from a pricegraph without having to
go through a StableXOrderbookReading.
This is useful in the price estimator where we already have a
pricegraph.

### Test Plan
CI, tests are the same.


core/src/price_estimation/orderbook_based.rs
```diff
@@ -26,35 +32,19 @@ impl PriceSource for PricegraphEstimator {
             let pricegraph = Pricegraph::new(orders.iter().map(|order| {
                 order.to_element(account_state.read_balance(order.sell_token, order.account_id))
             }));
-            Ok(self.estimate_prices(tokens, pricegraph))
+            pricegraph.get_prices(tokens).await
```
fleupold, line 35: Is `impl PriceSource for PricegraphEstimator` still needed? As far as I understand we have `impl LimitPriceEstimating for Pricegraph` and `impl<T: LimitPriceEstimating> PriceSource for T` so shouldn't this work without this specific impl?

e00E: `PriceSource` is implemented for all `LimitPriceEstimating` but `PricegraphEstimator` is not `LimitPriceEstimating`, only `Pricegraph` is `LimitPriceEstimating`. Indeed when would the code from this function that creates the orderbook be executed if we didn't have this implementation.

fleupold: I see (so many structs with very similar names). I guess we could also `impl LimitPriceEstimating for PricegraphEstimator` but that would create the orderbook n times (once for each token). Makes sense the way it is...

fleupold (approved):

core/src/price_estimation/orderbook_based.rs
```diff
@@ -26,35 +32,19 @@ impl PriceSource for PricegraphEstimator {
...
-    }
-
-    fn estimate_prices(
-        &self,
-        tokens: &[TokenId],
-        pricegraph: impl PriceEstimating,
-    ) -> HashMap<TokenId, u128> {
-        tokens
+use inner::LimitPriceEstimating;
+impl<T: LimitPriceEstimating> PriceSource for T {
```
nlordell, line 42: Can't this `impl` block be inside the `mod inner` as well?

nlordell (approved):

core/src/price_estimation/orderbook_based.rs
e00E: I think it can but it doesn't have any effect, does it? The reason I put the trait in the module is so that you cannot import it from outside. This is the only purpose of the module so if it doesn't matter for the impl block then it makes more sense to me outside.

PR #1140, opened 2020-07-16
Fix limit price estimation regression

nlordell:
This PR fixes a limit price estimation regression introduced in the recent refactoring. This is because we were `?`-short circuiting on an `Option` that was always `None` (since we don't fill transitive orders when volume is `0.0`).

### Test Plan

New unit test introduced to catch similar regressions.

### DIFF
--- a/pricegraph/src/api/price_estimation.rs
+++ b/pricegraph/src/api/price_estimation.rs
@@ -33,7 +33,7 @@ impl Pricegraph {
             orderbook.fill_optimal_transitive_order_if(inverse_pair, |flow| {
                 last_exchange_rate = Some(flow.exchange_rate);
                 false
-            })?;
+            });
         }
 
         let mut remaining_volume = sell_amount;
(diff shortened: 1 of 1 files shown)


e00E (approved):

pricegraph/src/api/price_estimation.rs
```diff
@@ -33,7 +33,7 @@ impl Pricegraph {
             orderbook.fill_optimal_transitive_order_if(inverse_pair, |flow| {
                 last_exchange_rate = Some(flow.exchange_rate);
                 false
-            })?;
+            });
```
fleupold, line 36: Interesting. Why does the linter not yell at you to consume this Option?

fleupold (approved):
Unit tests ftw!

pricegraph/src/api/price_estimation.rs
e00E: `Result`s are `#[must_use]` but not Options. If this function was pure it might be a warning (not sure) but in this it but in this case it has side effects so it is reasonable to ignore the return value.

nlordell: I made the code a little more self-documenting.

PR #1142, opened 2020-07-16
Price Estimator scaffolding for pricegraph with rounding buffer applied

e00E:
Some routes will make use of the pricegraph with the rounding buffer
applied and some without it. To facilitate this we store another
pricegraph instance and keep it up to date.
In a future PR `apply_rounding_buffer_to_auction_data` will be
implemented and routes will moved to the new pricegraph where
appropriate.

Also turned `fn get_pricegraph` into `fn pricegraph` because that is the norm in Rust and I had to touch all calls to this function anyway.

### Test Plan
The new pricegraph with rounding buffer applied isn't used anywhere yet. Tested that price estimator still works by running it and curling.

### DIFF
--- a/price-estimator/src/orderbook.rs
+++ b/price-estimator/src/orderbook.rs
@@ -1,50 +1,98 @@
 use anyhow::Result;
-use core::{models::BatchId, orderbook::StableXOrderBookReading};
+use core::{
+    models::{AccountState, BatchId, Order},
+    orderbook::StableXOrderBookReading,
+};
 use pricegraph::Pricegraph;
 use tokio::sync::RwLock;
+
+#[derive(Debug)]
+pub enum PricegraphType {
+    // pricegraph instance with the original orders from the orderbook
+    Raw,
+    // pricegraph instance with the orders to which the rounding buffer has been applied
+    #[allow(dead_code)]
+    WithRoundingBuffer,
+}
 
 /// Access and update the pricegraph orderbook.
 pub struct Orderbook {
     orderbook_reading: Box<dyn StableXOrderBookReading>,
-    pricegraph: RwLock<Pricegraph>,
+    pricegraph_raw: RwLock<Pricegraph>,
+    pricegraph_with_rounding_buffer: RwLock<Pricegraph>,
 }
 
 impl Orderbook {
     pub fn new(orderbook_reading: Box<dyn StableXOrderBookReading>) -> Self {
         Self {
             orderbook_reading,
-            pricegraph: RwLock::new(Pricegraph::new(std::iter::empty())),
+            pricegraph_raw: RwLock::new(Pricegraph::new(std::iter::empty())),
+            pricegraph_with_rounding_buffer: RwLock::new(Pricegraph::new(std::iter::empty())),
         }
     }
 
-    /// Creates a `Pricegraph` instance at the given batch ID.
-    async fn create_pricegraph_at_batch(&self, batch_id: BatchId) -> Result<Pricegraph> {
-        let (account_state, orders) = self
-            .orderbook_reading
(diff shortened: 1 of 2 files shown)


price-estimator/src/orderbook.rs
```diff
@@ -1,50 +1,98 @@
...
+        }
+        .read()
+        .await
+        .clone()
+    }
+}
+
+type AuctionData = (AccountState, Vec<Order>);
+
+fn pricegraph_from_auction_data(auction_data: &AuctionData) -> Pricegraph {
```
fleupold, line 91: Why not use a From/Into trait here? Seems like a pretty standard conversion (we might do a similar thing in the core lib)

price-estimator/src/orderbook.rs
```diff
@@ -1,50 +1,98 @@
...
+    // pricegraph instance with the orders to which the rounding buffer has been applied
+    #[allow(dead_code)]
+    WithRoundingBuffer,
+}
+
 /// Access and update the pricegraph orderbook.
 pub struct Orderbook {
     orderbook_reading: Box<dyn StableXOrderBookReading>,
-    pricegraph: RwLock<Pricegraph>,
+    pricegraph_raw: RwLock<Pricegraph>,
```
fleupold, line 21: ubernit: To keep the struct more focussed the two pricegraphs could be behind a common `cache` struct and the same lock (given they are only ever written to at the same time). Right now the Orderbook struct is still small so maybe not needed.

fleupold (approved):
One bite-size PR at a time - 👍 Thanks!

price-estimator/src/orderbook.rs
e00E: Here this is not possible because you cannot implement a trait from another crate for a type that is also from another crate.
 We could do this in the core crate but even then to me it doesn't feel appropriate. This feeling is  vague but what I mean is that while it's true that you happen to be able to turn `AuctionData` into `Pricegraph` these types are so unrelated that I don't think about this as a conversion to which the conversion traits apply. It seems more incidental that this works out but it could easily be the case that the pricegraph constructor needed a little more information that isn't part of AuctionData.

Technically any function that takes one type and returns another could be implemented as a conversion trait but semantically it doesn't always fit imo.

e00E: Oh I forgot that we are actually creating the type AuctionData in this file. Argument above still holds and when we are only doing this in this file then I also prefer not to go through the trait. It is simpler to just have a function. The trait makes sense if you want to provide this convenience for other users of your type.

price-estimator/src/orderbook.rs
```diff
@@ -1,50 +1,109 @@
...
-use core::{models::BatchId, orderbook::StableXOrderBookReading};
+use core::{
+    models::{AccountState, BatchId, Order},
+    orderbook::StableXOrderBookReading,
+};
 use pricegraph::Pricegraph;
 use tokio::sync::RwLock;
 
+#[derive(Debug)]
+pub enum PricegraphType {
```
nlordell, line 10: uber-naming-nit: personally I prefer  `PricegraphKind` to be more Rust-y

nlordell (approved):
Nice! Sorry for some reason I thought I had already reviewed this PR! Feel free to ping me if I'm taking too long to review something.

price-estimator/src/orderbook.rs
nlordell: I think the value of using `Into` and `From` is mostly is situations where you want to accept either an AuctionData or Pricegraph by value. This is not the case in our situation since we need to typically use the Pricegraph by reference and construct it before hand.

I also slightly prefer using a function here.

PR #1144, opened 2020-07-17
Estimate price based on available buy token liquidity

### LINKED ISSUE #1086: Improved Price Estimation Algorithm
Counter intuitively, it is possible for estimated buy amounts for larger sell amounts to be lower than if the sell amounts were smaller. This is because the pricegraph works by computing the **marginal exchange rate** in order to completely sell a certain volume. However, this is not entirely correct and can be modified so that this doesn't happen.

Consider the following scenario:
- Order 1: buy 100 OWL for 100 DAI
- Order 2: buy 1000 OWL for 100 DAI

Currently, when estimating the xrate for selling 150 OWL, the pricegraph algorithm will determine that 100 DAI can be bought for 100 OWL from order 1, and then the remaining 50 OWL must use order 2. Therefore, the marginal price will be 10 OWL for 1 DAI, and the the pricegraph will estimate the order "buy 15 DAI for 150 OWL".

However, the xrate can be capped by the previous buy amount. Specifically the order "buy 100 DAI for 150 OWL" also overlaps with order 1 and is a much better xrate estimate that is also matchable and less counter intuitive.

nlordell:
This PR fixes #1086 by implementing price estimates based on the total available liquidity at a given exchange rate, and by iteratively adding liquidity at worse exchange rates until the full sell amount can be sold.

A full description of the new approach for estimating prices is found here: https://github.com/gnosis/dex-services/issues/1086#issuecomment-660941928

The PR is quite large, but it is ~2/3 unit tests. I do however strongly encourage going through the new annotated unit tests to understand what is going on.

### Test Plan

Comprehensive unit tests added - old unit tests mostly unmodified (except where the new algorithm produces different results in low liquidity situations).


pricegraph/src/api/price_estimation.rs
```diff
@@ -26,30 +24,60 @@ impl Pricegraph {
...
-            last_exchange_rate = Some(flow.exchange_rate);
-            remaining_volume -= flow.capacity;
+        // NOTE: Iteratively compute the how much total buy volume (liquidity)
+        // is available at successively "worse" exchange rates until all the
+        // specified sell amount can be used to buy the available liquidity at
+        // the marginal exchange rate.
+        let mut total_buy_volume = 0.0;
+        let mut maximum_sell_volume = 0.0;
+        while let Some(flow) = orderbook.fill_optimal_transitive_order_if(inverse_pair, |flow| {
+            let current_exchange_rate = match ExchangeRate::new(total_buy_volume / sell_amount) {
```
fleupold, line 55: nit: the naming `sell_amount` & `maximum_sell_amount` was confusing me here. It's important that this amount is the input into the function and not the `sell_amount` so far. Maybe we can rename `total_buy_amount` and `maximum_sell_amount` to something that reflects that (`xxx_amount_so_far` for the lack of a better suggestion).

pricegraph/src/api/price_estimation.rs
```diff
@@ -26,30 +24,60 @@ impl Pricegraph {
...
+            // transitive order at its exchange rate makes the estimated
+            // exchange rate worse, and we are better off just buying off all
+            // the previously discovered liquidity instead of including new
+            // liquidity from this transitive order.
+            current_exchange_rate < flow.exchange_rate.inverse()
+        }) {
+            total_buy_volume += flow.capacity / flow.exchange_rate.value();
+            maximum_sell_volume = total_buy_volume * flow.exchange_rate.value();
+
+            if sell_amount <= maximum_sell_volume {
```
fleupold, line 70: This statement also reads weird. Why would you break if something is smaller than the maximum of that something (we should break when something becomes larger than its max).

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FavfcpYUjFWWvS%2Fgiphy.gif&token=rocket" data-gitmeme-token="rocket"></a>

Really impressive work and well done explaining everything with enough context. 

Could we also do a backtest with this change in logic to see that it doesn't have an effect on our correctness metric (how often the solver found a price within the estimates)?

nlordell:
> Could we also do a backtest with this change in logic to see that it doesn't have an effect on our correctness metric (how often the solver found a price within the estimates)?

Sure, but that is super unlikely, since the code path for `sell_volume == 0.0` hasn't changed. I'll run it and let you know.

pricegraph/src/api/price_estimation.rs
nlordell: How about `cumulative_{buy,sell}_amount`?

pricegraph/src/api/price_estimation.rs
nlordell: Yeah, its because `maximum_sell_volume` is a bit misnamed. Its more of the "maximum sell amount for the current cumulative buy amount".

pricegraph/src/api/price_estimation.rs
```diff
@@ -207,12 +235,197 @@ mod tests {
...
-        let price = pricegraph.estimate_limit_price(TokenPair { buy: 1, sell: 2 }, 1_000_000_000.0);
-        assert!(price.is_none());
+        // NOTE: For the next ~1_000_000 sell amount, we continue to fully use
+        // liquidity from the first order, and not include the second order
+        // because its limit price is too high.
+        assert_approx_eq!(
+            pricegraph
+                .order_for_sell_amount(
+                    TokenPair { buy: 2, sell: 1 },
+                    1_000_000.0 * FEE_FACTOR.powi(2),
```
fedgiac, line 353: Is this assertion part of the "story"? I'm not sure I understand what it's supposed to tell me.
Maybe you multiply by an extra `FEE_FACTOR` here to nudge up the amount sold just a little and see that nothing changes?

fedgiac (approved):
Nice PR! The tests in particular are very clear.

pricegraph/src/api/price_estimation.rs
nlordell: The `FEE_FACTOR^2` was that nudge :laughing:. I modified the test so that the "nudge" is more obvious.

pricegraph/src/api/price_estimation.rs
fleupold: I like you suggestion. I would still name sell_amount max_sell_amount to make it more obvious that the `current_exchange_rate` is not computed between the cumulative amounts.

nlordell:
@fleupold 

Indeed, it looks like it does not effect the current error rate:
```
$ node ./cmp.js -- spreads.improve-price-estimates.csv
7.11%
$ node ./cmp.js -- spreads.master.csv                 
7.13%
```

These small differences are expected since we use `sort_unstable` and `HashMap`s and the order in which the orders appear in the final sorted array can change from run to run.

PR #1152, opened 2020-07-20
Turn retry_with_gas_price_increase into a mockable trait

e00E:
So that StableXSolutionSubmitter can be tested independently.

### Test Plan
CI

### DIFF
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -13,8 +13,7 @@ use ethcontract::web3::types::TransactionReceipt;
 use ethcontract::U256;
 use futures::future::{BoxFuture, FutureExt as _};
 use log::info;
-#[cfg(test)]
-use mockall::automock;
+use retry::SolutionTransactionSending;
 use std::time::{Duration, SystemTime};
 use thiserror::Error;
 
@@ -32,7 +31,7 @@ const GAS_PRICE_CAP: u64 = 200_000_000_000;
 // least 12.5%.
 const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
 
-#[cfg_attr(test, automock)]
+#[cfg_attr(test, mockall::automock)]
 pub trait StableXSolutionSubmitting {
     /// Return the objective value for the given solution in the given
     /// batch or an error.
@@ -98,7 +97,7 @@ impl From<Error> for SolutionSubmissionError {
 
 pub struct StableXSolutionSubmitter<'a> {
     contract: &'a (dyn StableXContract + Sync),
-    gas_price_estimating: &'a (dyn GasPriceEstimating + Sync),
+    retry_with_gas_price_increase: Box<dyn SolutionTransactionSending + Send + Sync + 'a>,
 }
 
 impl<'a> StableXSolutionSubmitter<'a> {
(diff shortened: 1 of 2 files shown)


core/src/solution_submission.rs
```diff
@@ -99,16 +98,32 @@ impl From<Error> for SolutionSubmissionError {
...
+        gas_price_estimating: &'a (dyn GasPriceEstimating + Sync),
+    ) -> Self {
+        Self {
+            contract,
+            gas_price_estimating,
+            retry_with_gas_price_increase: Box::new(retry::RetryWithGasPriceIncrease {}),
+        }
+    }
+
+    #[cfg(test)]
```
nlordell, line 117: nit: Instead of conditional compilation we can have a `new` method and a `with_retrying` method. I find it a bit confusing that the same method `new` has multiple signatures. Additionally, for people unfamiliar with the codebase that may use the unit test code as a reference, it would also be unnecessarily counterintuitive.

core/src/solution_submission/retry.rs
```diff
@@ -6,9 +6,60 @@ use ethcontract::{
     errors::{ExecutionError, MethodError},
     U256,
 };
+use futures::future::{BoxFuture, FutureExt as _};
 
 use super::MIN_GAS_PRICE_INCREASE_FACTOR;
 
+pub struct Args<'a> {
```
nlordell, line 13: I like this! That way arguments aren't positional 🎉

core/src/solution_submission/retry.rs
```diff
@@ -6,9 +6,60 @@ use ethcontract::{
...
+
+pub trait RetryingWithGasPriceIncrease {
+    fn retry<'a>(&self, args: Args<'a>) -> BoxFuture<'a, Result<(), MethodError>>;
+}
+
+pub struct RetryWithGasPriceIncrease {}
+
+impl RetryingWithGasPriceIncrease for RetryWithGasPriceIncrease {
+    fn retry<'a>(&self, args: Args<'a>) -> BoxFuture<'a, Result<(), MethodError>> {
+        retry(
```
nlordell, line 31: Any reason this doesn't take an `Args` parameter? If the reason is to not change the `retry` method than cant you just destructure the args parameter?

For example this is valid Rust:
```
struct Foo {
    bar:i32,
    baz:i64,
}
fn foo(Foo { bar, baz }: Foo) {
    dbg!((bar, baz));
}
```

nlordell (approved):

core/src/solution_submission/retry.rs
fleupold: Is there a reason for the non-dynamic per instance state (contract, gas_price_estimating) to be passed on every invocation? I can see how the gas_price_cap is already passed as we want to make it dependent on each solution going forward.

Also, I'm wondering if the abstraction should be built on to of a MethodBuilder instead of the concrete set of parameters for a solution submission. This way the function would just know how to retry some ethereum tx up to a  certain gas cap.

core/src/solution_submission/retry.rs
```diff
@@ -6,9 +6,51 @@ use ethcontract::{
...
+    pub contract: &'a (dyn StableXContract + Sync),
+    pub batch_index: u32,
+    pub solution: Solution,
+    pub claimed_objective_value: U256,
+    pub gas_price_estimating: &'a (dyn GasPriceEstimating + Sync),
+    pub gas_price_cap: U256,
+    pub nonce: U256,
+}
+
+pub trait RetryingWithGasPriceIncrease {
```
fleupold, line 23: I would argue that the trait itself is really just capable of submitting a transaction (it's only invoked once from the caller) and therefore suggest to name it `TransactionSending` or `SolutionTransactionSending` if you want to keep the logic tx specific.

We then have a concrete implementation that does so by retrying with gas increase.

fleupold (approved):
Very nice refactoring! Looking at what solutionSubmission already does without the retry logic (verifying objective value, interpreting the tx result) it makes a lot of sense to handle the actual transaction submission in a separate component.

core/src/solution_submission/retry.rs
e00E: This is a bigger task that I find hard to estimate how complex it will be so I don't want to undertake it now. In a future pr this logic is also going to change to fix the nonce errors. We can look it again after that.

core/src/solution_submission/retry.rs
e00E: For now the intention is specific to the solution as is seen in the parameters of the one function the trait has. So I don't think it makes sense to give it a more generic name.

fleupold: I don't think the fact that it is retrying is specific to the arguments. Specifically because the gas_price_estimating does not need to part of Args even if we keep it tx specific.

e00E: Implemented rename and moved the two args into the struct. Has nice side effect that solution submission no longer has to store the gas price estimating.

PR #1156, opened 2020-07-20
Don't consider dust trades when enumerating transitive orders

### LINKED ISSUE #1000: Unmatchable dust orders are being considered,
I noticed that dust orders that can't actually be part of a solution are being considered in price estimates and transitive orderbooks. Some extra checks are needed on the transitive order path that we currently don't consider:
- All trades must be > min amount
- The fee token trade must be > min amount

Currently we only consider orders individually, and not the estimated trade that would happen.

nlordell:
Fixes #1000 

This PR fills flows for "dust" trades (that trade an amount below the minimum amount) so that they are not considered in price estimates. Note that they are still considered for the transitive orderbook computation as traded amounts depend on closing prices.

### Test Plan

New unit test.

### DIFF
--- a/pricegraph/src/api/price_estimation.rs
+++ b/pricegraph/src/api/price_estimation.rs
@@ -678,25 +678,25 @@ mod tests {
 
         // NOTE: This would trade ~10_000 -> ~1_000 -> ~100_000 -> ~1_000_000
         assert_eq!(
-            pricegraph.estimate_limit_price(TokenPair { buy: 0, sell: 3 }, 10_000.0),
+            pricegraph.estimate_limit_price(TokenPair { buy: 0, sell: 3 }, 10_001.0),
             None,
         );
 
         // NOTE: This would trade ~1_000 -> ~100_000 -> ~1_000_000
         assert_eq!(
-            pricegraph.estimate_limit_price(TokenPair { buy: 1, sell: 3 }, 10_000.0),
+            pricegraph.estimate_limit_price(TokenPair { buy: 1, sell: 3 }, 10_001.0),
             None,
         );
 
         // NOTE: This would trade ~9_000 -> ~90_000
         assert_eq!(
-            pricegraph.estimate_limit_price(TokenPair { buy: 0, sell: 4 }, 10_000.0),
+            pricegraph.estimate_limit_price(TokenPair { buy: 0, sell: 4 }, 10_001.0),
             None,
         );
 
         // NOTE: This would trade ~10_000 -> ~10_000 -> ~100
         assert_eq!(
-            pricegraph.estimate_limit_price(TokenPair { buy: 5, sell: 7 }, 10_000.0),
+            pricegraph.estimate_limit_price(TokenPair { buy: 5, sell: 7 }, 10_001.0),
             None,
         );
     }
--- a/pricegraph/src/api/price_estimation.rs
+++ b/pricegraph/src/api/price_estimation.rs
@@ -65,6 +65,10 @@ impl Pricegraph {
             // liquidity from this transitive order.
             current_exchange_rate < flow.exchange_rate.inverse()
         }) {
+            if flow.is_dust_trade() {
+                continue;
+            }
+
             cumulative_buy_volume += flow.capacity / flow.exchange_rate.value();
             cumulative_sell_volume = cumulative_buy_volume * flow.exchange_rate.value();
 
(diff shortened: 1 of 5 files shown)


e00E (approved):

fleupold (approved):
Very nice unit tests. Here again it would be nice to see the impact of that change in terms of backtested accuracy if possible.

nlordell:
@fleupold 

I ran the historic prices again, and the dust orders don't make much of a difference. It seems to *slightly* improve the accuracy but I'm not certain if that is due to some uncertainty caused by orders (as in exchange limit orders) being considered in a different order (as in ordering). Intuitively it should slightly improve the error rate as we consider fewer orders and are therefore more pessimistic.
```
$ node ./cmp.js -- spreads.master.csv
7.13%
$ node ./cmp.js -- spreads.filter-dust-orders.csv       
7.03%
```

pricegraph/src/api/price_estimation.rs
```diff
@@ -125,6 +129,10 @@ impl Pricegraph {
         while let Some(flow) = orderbook
             .fill_optimal_transitive_order_if(inverse_pair, |flow| flow.exchange_rate <= max_xrate)
         {
+            if flow.is_dust_trade() {
+                continue;
```
bh2smith, line 133: Is it my impression that anyplace where a "continue" is needed, the logic could be expressed in a different way. Like filtering out the dust trades before entering the loop or even, 
including the filter here:
```
fill_optimal_transitive_order_if(inverse_pair, |flow| !flow.is_dust_trade() && flow.exchange_rate <= max_xrate)
```

Of course, this is fine as is.

pricegraph/src/api/price_estimation.rs
```diff
@@ -65,6 +65,10 @@ impl Pricegraph {
             // liquidity from this transitive order.
             current_exchange_rate < flow.exchange_rate.inverse()
         }) {
+            if flow.is_dust_trade() {
+                continue;
+            }
```
bh2smith, line 70: Seems like this block appears twice... could this be taken care of inside the while condition?

pricegraph/src/num.rs
```diff
@@ -77,6 +78,12 @@ pub fn is_strictly_positive_and_finite(value: f64) -> bool {
     value > 0.0 && value < f64::INFINITY
 }
 
+/// Returns true if an amount is considered a dust amount. See `MIN_AMOUNT`
+/// documentation for more details.
+pub fn is_dust_amount(amount: f64) -> bool {
+    amount < MIN_AMOUNT
+}
+
```
bh2smith, line 86: The minimum amount is incredible low for most tokens. Do we know how effective this filtration will be for tokens with 18 decimal places? I mean it seems that even values like 0.00000000000001 ETH wouldn't be considered "dust orders"

bh2smith (commented):
Looks pretty good overall. Just a bit worried about how effective this is expected to be.

pricegraph/src/api/price_estimation.rs
nlordell: The issue is if we do that, then we will `break` when we encounter a dust order instead of skipping it. Alternatively, one cool way we could solve this is another helper method returning an `Iterator` over transitive orders and then use the `filter` method.

pricegraph/src/api/price_estimation.rs
nlordell: Not in this case either - since it would cause us to stop looping which is not what we want.

pricegraph/src/num.rs
nlordell: This is the `MIN_AMOUNT` as specified by the smart contract.

bh2smith: Yes, but min amount is the WEI amount of a token, so the example above with such a little eth value is not considered a dust order.

nlordell: Pricegraph only uses amounts in WEI

bh2smith: Perhaps there is some misunderstanding here. The goal is to exclude dust orders, but my claim is that what constitutes a dust order could be much larger than the smart contract's `MIN_AMOUNT`. For example, I would say that even `0.00001 ETH` is a dust order, but this is several factors of 10 larger than the `MIN_AMOUNT`.

nlordell: Ah I see. Maybe "dust order" is the incorrect name for this then. The purpose of excluding these orders is that matching them is incorrect (since the SC does not accept trades with these small amounts) so we have to make sure that the `pricegraph` doesn't find "ring" trades where one of the trades is tiny and too small to be considered by the SC.

Because of economic viability constraint, I would even say its OK to ignore orders of around `0.01 ETH` and below. But that would be a separate change.

PR #1161, opened 2020-07-21
Adjust price source error messages.

### LINKED ISSUE #1160: DexAG price parsing error.
The DexAG API is failing to parse the JSON for the DexAG prices:
```
2020-07-21T09:32:35.043Z WARN [core::price_estimation::clients::generic_client] error retrieving core::price_estimation::clients::dexag::api::DexagHttpApi prices for token ID 1 (ETH): failed to parse price json from dexag
```

nlordell:
Fixes #1160 

This PR modifies the warning messages when price source API calls fail to:
- not use the word `'error'` since we use this for searching log messages with the slack alert links and pollutes the log output.
- Include the full `anyhow` context trace
- Adjust errors to reflect that the error is not necessarily an error parsing JSON.

### Test Plan

Rustc.


e00E (approved):
I don't mind this but is it possible to instead adjust our kibana regex instead? That would be a cleaner solution. We should be able to do this because all error logs follow the format `<timestamp> ERRO <module> <message>`.

nlordell:
> I don't mind this but is it possible to instead adjust our kibana regex instead? 

I'll look into it!

PR #1163, opened 2020-07-21
Introduce mockable trait for async sleeping

e00E:
This is useful in tests and will be used in another PR soon.

### Test Plan
No test needed.


nlordell (approved):
I think this is a great idea - it also allows us to further abstract over async std in case we want to change it.

It's currently not being used though, I'm assuming that will come in follow up PRs

PR #1164, opened 2020-07-21
Use more explicit return type for noop transaction

e00E:
This allows inspecting the error in more detail with going through
downcasting which will be useful in a future PR.

### Test Plan
No test needed.

### DIFF
--- a/core/src/contracts/stablex_contract.rs
+++ b/core/src/contracts/stablex_contract.rs
@@ -84,8 +84,8 @@ pub struct FilteredOrderPage {
 pub enum NoopTransactionError {
     #[error("no account")]
     NoAccount,
-    #[error("execution error")]
-    ExecutedOrder(ExecutionError),
+    #[error("execution error: {0}")]
+    ExecutedOrder(#[from] ExecutionError),
 }
 
 #[cfg_attr(test, automock)]
@@ -359,10 +359,7 @@ impl StableXContract for StableXContractImpl {
                 .nonce(nonce)
                 .value(U256::zero())
                 .resolve(ResolveCondition::Confirmed(ConfirmParams::mined()));
-            transaction
-                .send()
-                .await
-                .map_err(NoopTransactionError::ExecutedOrder)
+            transaction.send().await.map_err(From::from)
         }
         .boxed()
     }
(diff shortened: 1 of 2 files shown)


nlordell:
What's the issue with downcasting? I personally prefer downcasting over using the `NoLocalAccount` variant incorrectly (that error indicates the node has no local accounts, not that the web3 provider is not configured with one).

Another cleaner approach IMO would be to define a `NoopError` type instead of using anyhow error type.

e00E:
The issue with downcasting is that you have to read the documentation comment to find out what the valid errors are. It is better to have this done with the type system. I didn't realize the no local account error was wrong. Will use an enum instead.

core/src/contracts/stablex_contract.rs
```diff
@@ -80,6 +80,14 @@ pub struct FilteredOrderPage {
     pub next_page_user_offset: u16,
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum NoopTransactionError {
+    #[error("no account")]
+    NoAccount,
+    #[error("execution error")]
+    ExecutedOrder(ExecutionError),
```
nlordell, line 88: nit: You can derive the `from` implementation so that you don't have to `map_err`:
```suggestion
    ExecutedOrder(#[from] ExecutionError),
```

core/src/contracts/stablex_contract.rs
```diff
@@ -80,6 +80,14 @@ pub struct FilteredOrderPage {
     pub next_page_user_offset: u16,
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum NoopTransactionError {
+    #[error("no account")]
+    NoAccount,
+    #[error("execution error")]
```
nlordell, line 87: nit: Include the underlying error in the error message
```suggestion
    #[error("execution error: {0}")]
```

core/src/contracts/stablex_contract.rs
e00E: I added this but still need map_err because we're not using `?`. Makes sense to have it anyway.

nlordell (approved):

PR #1165, opened 2020-07-21
Make sure gas station gas price does not decrease

e00E:
The solution submission logic enforces that gas prices increase by at
least 12% each resubmission. However with the current logic it is
possible that the gas station returns significantly lower prices in the
next iteration which could cause this to be violated.
This commit ensures that we only ever increase the gas price estimate.

### Test Plan
Added new test for this.

### DIFF
--- a/core/src/solution_submission/retry.rs
+++ b/core/src/solution_submission/retry.rs
@@ -42,11 +42,8 @@ impl<'a> InfallibleGasPriceEstimator<'a> {
     async fn estimate(&mut self) -> U256 {
         match self.gas_price_estimating.estimate_gas_price().await {
             Ok(gas_estimate) => {
-                let price = gas_estimate.fast;
                 // `retry` relies on the gas price always increasing.
-                if price >= self.previous_gas_price {
-                    self.previous_gas_price = gas_estimate.fast;
-                }
+                self.previous_gas_price = self.previous_gas_price.max(gas_estimate.fast);
             }
             Err(ref err) => {
                 log::warn!(
--- a/core/src/solution_submission/retry.rs
+++ b/core/src/solution_submission/retry.rs
@@ -215,7 +215,7 @@ mod tests {
                 always(),
                 always(),
                 always(),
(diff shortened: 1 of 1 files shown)


core/src/solution_submission/retry.rs
```diff
@@ -39,7 +41,11 @@ impl<'a> InfallibleGasPriceEstimator<'a> {
     async fn estimate(&mut self) -> U256 {
         match self.gas_price_estimating.estimate_gas_price().await {
             Ok(gas_estimate) => {
-                self.previous_gas_price = gas_estimate.fast;
+                let price = gas_estimate.fast;
```
nlordell, line 44: nit: Any reason not to use `cmp::max`?
```suggestion
                let price = cmp::max(gas_estimate.fast, self.previous_gas_price);
```

nlordell (approved):

bh2smith:
In the original implementation, the gas station was only required for the first estimation and every re-submission simply doubled the gas price. I am somewhat confused how we continue to regress from what was not originally broken and keep falling back on simpler solutions.

bh2smith (approved):

e00E:
> In the original implementation, the gas station was only required for the first estimation and every re-submission simply doubled the gas price. I am somewhat confused how we continue to regress from what was not originally broken and keep falling back on simpler solutions.

I don't think this is true. When I refactored this file for the first time it was already updating the gas estimate every iteration (unless you mean something going back further). I agree that it would be nice if this was simpler. Querying the gas price only once at the start is a good suggestion and would simplify the code so I'll make a PR where we can discuss it but it is a change to the existing behavior.

nlordell:
I think the reasoning behind querying the gas price every iterations is to better account for gas spikes of new blocks.

fleupold:
Once we get close to the deadline, I would even argue that a more sophisticated implementation of the solution submission should look at the most recent (and potentially pending) block to decide at what gas price the tx needs to be sent in order to make it in in time.

PR #1166, opened 2020-07-21
Increase gas price slower but higher

fleupold:
We missed our 50k ETH/DAI auction today because the gas price was too high. This PR increases the cap to 200GWEI but makes it so that we only increase in 50% increments (doubling 90 GWEI on first attempt seems a bit harsh).

Going forward, we should not only implement #1099 but also make it that gas cap depends on the total fee that can be gained by the solver. In the example today we could have made $50 in fees. This should make us be willing to pay at least $50 in gas to get our tx mined.

### Test Plan
Unit tests

### DIFF
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -26,7 +26,7 @@ const POLL_TIMEOUT: Duration = Duration::from_secs(5);
 #[cfg(test)]
 const POLL_TIMEOUT: Duration = Duration::from_secs(0);
 
-const GAS_PRICE_CAP: u64 = 90_000_000_000;
+const GAS_PRICE_CAP: u64 = 200_000_000_000;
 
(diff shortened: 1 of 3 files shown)


bh2smith (approved):
Looks good. Is it safe to assume that the gas station is only queried once for the initial gas estimation?

fleupold:
> Is it safe to assume that the gas station is only queried once for the initial gas estimation?

No it will be queried on every attempt. What's your concern?

fedgiac (approved):

josojo (approved):
Thanks!

bh2smith:
If the gas price changes (i.e. goes down) from one attempt to the next, this could result in not increasing by the required amount.

I believe there is a PR out for this #1165, but I am surprised we are querying on every attempt and not simply increasing from our initial starting point.

PR #1171, opened 2020-07-22
Adjust gas price limit based on earned fees

e00E:
With the current hardcoded maximum gas price limit we can run into
problems when gas prices are so high that the limit is not enough - even
when the submitting the solution would pay out more than the gas cost.
This commit changes the maximum gas prices to be a function of the gas
use of the solution, the fee payout and the cost of ether in the fee
(owl).

Later we probably want to estimate the price for ether in owl instead of hardcoding it.

The solver also outputs the fee in the solution json but I found it cleaner to calculate it.

### Test Plan
CI and added logging so we can see what is chosen as the limit.


core/src/driver/stablex_driver.rs
```diff
@@ -189,17 +216,49 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
...
+fn gas_price_cap(burnt_fees: U256, num_trades: usize) -> U256 {
+    // This is approximate. Depends on reversion of previous solution.
+    const GAS_PER_TRADE: f64 = 120_000.0;
+    const PRICE_OF_ETHER_IN_OWL: f64 = 200.0;
+    // The previous two values are approximations and we do not need to be economically viable at
+    // this point in time. So be more lenient with respect go gas prices.
+    const EXTRA_FACTOR: f64 = 10.0;
+    let burnt_fees = pricegraph::num::u256_to_f64(burnt_fees);
+    let gas_use = GAS_PER_TRADE * (num_trades as f64);
+    let result = (burnt_fees / (PRICE_OF_ETHER_IN_OWL * gas_use)) * EXTRA_FACTOR;
```
e00E, line 247: Are the formula and these constants ok?

fleupold: I think we should get the price of Ether in OWL from our price source. Prices of ether go out of date quickly and we already have this data available somewhere. Maybe better for a separate PR though.

core/src/driver/stablex_driver.rs
```diff
@@ -189,17 +216,49 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
...
+        .ok_or_else(|| anyhow!("invalid burnt fee amount: {}", fee_token_imbalance))
+}
+
+fn gas_price_cap(burnt_fees: U256, num_trades: usize) -> U256 {
+    // This is approximate. Depends on reversion of previous solution.
+    const GAS_PER_TRADE: f64 = 120_000.0;
+    const PRICE_OF_ETHER_IN_OWL: f64 = 200.0;
+    // The previous two values are approximations and we do not need to be economically viable at
+    // this point in time. So be more lenient with respect go gas prices.
+    const EXTRA_FACTOR: f64 = 10.0;
```
fleupold, line 244: nit: Let's call this `subsidy_factor` or the like (extra sounds very vague). It should probably also be passed in via a config so that we can adjust it without code changes.

core/src/driver/stablex_driver.rs
```diff
@@ -118,9 +126,28 @@ impl<'a> StableXDriverImpl<'a> {
...
+            let gas_price_cap = {
+                let num_trades = solution.executed_orders.len();
+                match burnt_fees(&orders, &solution) {
+                    Ok(fee) => {
+                        let cap = gas_price_cap(fee, num_trades);
+                        log::info!("Using gas price cap {} based on fee {}", cap, fee);
+                        cap
+                    }
+                    Err(err) => {
+                        const DEFAULT_GAS_PRICE_CAP: u64 = 200_000_000_000;
```
fleupold, line 138: Maybe we should collect all these constants in a common place?

core/src/driver/stablex_driver.rs
```diff
@@ -118,9 +126,28 @@ impl<'a> StableXDriverImpl<'a> {
         };
 
         let submitted = if let Some(objective_value) = verified {
+            let gas_price_cap = {
+                let num_trades = solution.executed_orders.len();
+                match burnt_fees(&orders, &solution) {
```
fleupold, line 131: It's too bad that this method can fail (we could get rid of the hardcoded cap otherwise). Is this really a practical concern? I don't really see how a solution could be found for orders we don't have. But then asserting this here seems a bit far away from where this invariant is actually guaranteed. Maybe solution could have a burnt_fees field on it that we can populate when we get it and just read here?

core/src/driver/stablex_driver.rs
```diff
@@ -586,4 +645,68 @@ mod tests {
             .unwrap()
             .is_ok());
     }
+
+    #[test]
+    fn burnt_fees_ok() {
```
fleupold, line 650: nit: Since there are multiple ways how burnt fees could be computed (we could also price all buy volumes in OWL and take 0.1% of it) I think it would be nice if the test could document in its name what ok means.

E.g. `burnt_fees_is_half_imbalance_of_reference_token_sold_and_bought`

core/src/driver/stablex_driver.rs
```diff
@@ -189,17 +216,49 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
...
+        }
+        if order.buy_token == 0 {
+            fee_token_imbalance -= executed_order.buy_amount;
+        }
+    }
+    bigint_u256::bigint_to_u256(&fee_token_imbalance)
+        .ok_or_else(|| anyhow!("invalid burnt fee amount: {}", fee_token_imbalance))
+}
+
+fn gas_price_cap(burnt_fees: U256, num_trades: usize) -> U256 {
```
fleupold, line 238: Maybe we can add a unit test that documents how much GWEI we would be willing to pay e.g. for our 50k market order.

fleupold (approved):
Looks good. I think the logic for computing the fee from a solution could also live closer to where the solution is created.

core/src/driver/stablex_driver.rs
e00E: Some part of the code will have to do unwrap for this which I am ok with. I'll see if I find a more natural place for the calculation, that also has access to the original orders.

core/src/driver/stablex_driver.rs
e00E: I agree and I also want to do this as part of a separate PR.

core/src/driver/stablex_driver.rs
```diff
@@ -1,12 +1,27 @@
...
+use std::{
+    collections::HashMap,
+    time::{Duration, Instant},
+};
+
+// Used when we cannot calculate the amount of burned fees.
+const DEFAULT_GAS_PRICE_CAP: u64 = 200_000_000_000;
+// This is approximate. Depends on reversion of previous solution.
+const GAS_PER_TRADE: f64 = 120_000.0;
+// Will be removed in a future PR where we are going to fetch the current price.
```
nlordell, line 23: Just a suggestion, we already get price estimates, and we can re-use the WETH price estimate for this.

core/src/driver/stablex_driver.rs
```diff
@@ -1,12 +1,27 @@
...
-use std::time::{Duration, Instant};
+use num::{BigInt, Zero as _};
+use std::{
+    collections::HashMap,
+    time::{Duration, Instant},
+};
+
+// Used when we cannot calculate the amount of burned fees.
+const DEFAULT_GAS_PRICE_CAP: u64 = 200_000_000_000;
+// This is approximate. Depends on reversion of previous solution.
```
nlordell, line 21: Maybe (in a future PR) we can estimate the gas of a transaction and use that to compute the gas cap.

nlordell (approved):
Looks great! Just some suggestions for future imrpovements to keep in mind.

core/src/driver/stablex_driver.rs
nlordell: I was under the impression that the solver result output contains burnt fees. Maybe we can reuse that (if it actually does)?

e00E: Yes it does but I am unsure whether we should be relying on that information since the solver output format is not documented. If we use it then we say the solver always has to output it. But it's not much work to calculate on our side.

nlordell: Another way I thought of you make this method not fail is to just add together **all** sell amounts and subtract **all** buy amounts. I think you don't need to look at the order buy and sell tokens for this to work.

Additionally the BigInt to U256 conversion can be asserted or be saturating.

e00E: Good idea. Implemented this.

driver/src/main.rs
```diff
@@ -185,6 +185,12 @@ struct Options {
 
     #[structopt(long, env = "ORDERBOOK_FILE", parse(from_os_str))]
     orderbook_file: Option<PathBuf>,
+
+    /// We calculate the maximum gas price cap based on the amount of earned fees from solution
+    /// submission. This factor is multiplied with the final result because we do not need to be
+    /// economically viable at the moment.
+    #[structopt(long, env = "GAS_PRICE_CAP_SUBSIDY_FACTOR", default_value = "2.0")]
```
e00E, line 192: Still the question what the default value should be.

nlordell: I think we can leave it as is. We can always re-configure the deployment if needed.

core/src/bigint_u256.rs
```diff
@@ -10,6 +10,14 @@ pub fn bigint_to_u256(n: &BigInt) -> Option<U256> {
     }
 }
 
+pub fn bigint_to_u256_saturating(n: &BigInt) -> U256 {
```
nlordell, line 13: :metal:

nlordell (approved):

e00E:
@fleupold let me know if you still approve after the code review changes then we can merge

josojo (approved):
Cool thanks. I think it makes sense the way you were choosing the constants(subsidisation factors etc.)

fleupold (approved):
Looks great! Really good idea with computing the fee based on all token balances.

e00E:
> Looks great! Really good idea with computing the fee based on all token balances.

That was Nick's idea btw.

PR #1176, opened 2020-07-22
Write to pricegraph without rounding buffer on update

nlordell:
This PR fixes an issue with the `Pricegraph` instance without the rounding buffer applied was not being updated with exchange events.

### Test Plan

Run the price estimator and get a price estimate:
```
$ curl -s 'http://localhost:8080/api/v1/markets/1-7/estimated-best-ask-price?atoms=true' | jq                
235.51801773466883
```

### DIFF
--- a/price-estimator/src/orderbook.rs
+++ b/price-estimator/src/orderbook.rs
@@ -61,10 +61,7 @@ impl Orderbook {
 
         // TODO: Move this cpu heavy computation out of the async function using spawn_blocking.
         let pricegraph = pricegraph_from_auction_data(&auction_data);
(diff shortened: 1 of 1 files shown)


fedgiac (approved):
Yeah, it looks like the current code has a typo.

e00E (approved):

e00E:
My bad, thanks for catching it.

Velenir (approved):

fleupold:
Glad this was caught so quick in staging but also agree we should have #1177...

anxolin:
Thanks for fixing so fast!

PR #1180, opened 2020-07-23
Use AsyncSleep trait in solution submission

e00E:
Previously we introduced a `handle_submit_result` function in order to
be able to test is separately because `submit_solution` was using
`async_std::timeout`.
Now that we have the async sleep trait we no longer need the extra
function by using `future::select` with the sleep instead of `timeout`.
This is also a step towards handling the nonce already used race
condition correctly which will be done in another PR.

### Test Plan
Unit tests adapted but this isn't change in observable behavior.

### DIFF
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -1,20 +1,20 @@
 mod retry;
 
 use crate::{
-    contracts::stablex_contract::StableXContract,
+    contracts::stablex_contract::{NoopTransactionError, StableXContract},
     gas_station::GasPriceEstimating,
     models::{BatchId, Solution},
+    util::AsyncSleeping,
 };
 
 use anyhow::{anyhow, Error, Result};
-use async_std::future::TimeoutError;
 use ethcontract::{
     errors::{ExecutionError, MethodError},
     jsonrpc::types::Error as RpcError,
     web3::{error::Error as Web3Error, types::TransactionReceipt},
     U256,
 };
-use futures::future::{BoxFuture, FutureExt as _};
+use futures::future::{self, BoxFuture, Either, FutureExt as _};
 use log::info;
 use pricegraph::num;
 use retry::SolutionTransactionSending;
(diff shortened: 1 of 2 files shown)


core/src/solution_submission.rs
```diff
@@ -142,40 +146,27 @@ impl<'a> StableXSolutionSubmitter<'a> {
...
-                "solution submission transaction not confirmed in time"
-            )))
-        }
+    ) -> Result<(), NoopTransactionError> {
+        // Add some extra time in case of desync between real time and ethereum node current block time.
+        let deadline = BatchId::from(batch_index).solve_end_time() + Duration::from_secs(30);
+        let remaining = deadline
+            .duration_since(SystemTime::now())
+            .unwrap_or(Duration::from_secs(0));
+        self.async_sleep.sleep(remaining).await;
```
fleupold, line 159: Should we still attempt the cancellation even if the other future has already completed? Or should we have some conditional variable that we signal when the tx gets confirmed to avoid causing unnecessary failure on the node?

fleupold (approved):
I like the use of the async sleeping trait a lot! Wondering if we can implement a `cancellable_exec_after` style logic where we don't send the noop tx in case it was already mined.

core/src/solution_submission.rs
e00E: The way this works is that both futures are created and then "race" in the call to `select`. But the cancel future only sends the noop transaction after sleeping until the deadline. So if the solution submission future completes first the cancel futures is dropped and will never send the noop transaction.

nlordell: Yup! The sleep + cancel future will stop being `poll`-ed once the submit future completes and therefore stop progressing.

There is a subtle case where the cancel transaction will just have been sent when the submit solutions transaction completes. However I believe this case will also be gracefully handled as the node will deny the transaction (since it had the same nonce as a mined tx) and the deadline future won't produce an error as it will not be `poll`-ed further, causing it not to handle the nodes error response.

fleupold: That makes sense and sounds good. I was just worried that we will always send a noop tx on every submission which doesn't seem to be the case. I don't know if the "drop" behavior can be easily tested now that we control the sleep. If so it would be nice to add, but otherwise never mind.

nlordell (approved):

Issue #1143, opened 2020-07-17
Solution Submission error: Transaction nonce is too low.


e00E:
TLDR: Solution submission transaction got mined right when we submitted a new transaction with higher gas price so the new transaction failed but the solution was already correctly submitted.

We saw this error in the [logs](https://logs.gnosis.io/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:'2020-07-17T10:10:00.000Z',mode:absolute,to:'2020-07-17T10:15:00.000Z'))&_a=(columns:!(log),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'63bc0b10-5f93-11e9-bf17-fbe383fc7c9c',key:kubernetes.pod_name,negate:!f,params:(query:standard-solver-mainnet,type:phrase),type:phrase,value:standard-solver-mainnet),query:(match:(kubernetes.pod_name:(query:standard-solver-mainnet,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!t,index:'63bc0b10-5f93-11e9-bf17-fbe383fc7c9c',key:log,negate:!t,params:(query:transport,type:phrase),type:phrase,value:transport),query:(match:(log:(query:transport,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'63bc0b10-5f93-11e9-bf17-fbe383fc7c9c',key:log,negate:!f,params:!(driver,core),type:phrases,value:'driver,%20core'),query:(bool:(minimum_should_match:1,should:!((match_phrase:(log:driver)),(match_phrase:(log:core))))))),index:'63bc0b10-5f93-11e9-bf17-fbe383fc7c9c',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc))).

Relevant parts:

1. > 2020-07-17T10:10:49.899Z INFO [core::driver::stablex_driver] Computed solution for batch 531660
2. > 2020-07-17T10:10:50.104Z DEBG [core::transport] [id:3990] sending request: {"jsonrpc":"2.0","method":"eth_getTransactionCount"
3. > 2020-07-17T10:10:50.109Z DEBG [core::transport] [id:3990] received response: '{"jsonrpc":"2.0","result":"0x6ed","id":3990}'
4. > 2020-07-17T10:10:50.175Z INFO [core::solution_submission] solution submission try 0 with gas price 78100000001
[...]

PR #1182, opened 2020-07-23
Fix race condition when cancelling solution submission transaction

e00E:
When the solution submission transaction and the cancellation
transaction execute at the same time it is possible for us to observe
the failure first.
Previously we did not detect this case and could emit an error message
when there was not actually an error.
This commit handles this situation by waiting on and returning the
result of the other transaction if the first one completes with a "nonce
already used error".

This is part #1143 . In another PR a similar case in `retry.rs` will
be handled.

### Test Plan
Unit tests for the new logic.

### DIFF
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -133,7 +133,7 @@ impl<'a> StableXSolutionSubmitter<'a> {
     }
 
     /// Turn a method error from a solution submission into a SolutionSubmissionError.
-    async fn make_error(
+    async fn convert_submit_error(
         &self,
         batch_index: u32,
         solution: Solution,
@@ -166,6 +166,18 @@ impl<'a> StableXSolutionSubmitter<'a> {
             }
         }
         SolutionSubmissionError::Unexpected(err.into())
+    }
+
+    async fn convert_submit_result(
+        &self,
+        batch_index: u32,
+        solution: Solution,
+        result: Result<(), MethodError>,
+    ) -> Result<(), SolutionSubmissionError> {
+        match result {
+            Ok(()) => Ok(()),
+            Err(err) => Err(self.convert_submit_error(batch_index, solution, err).await),
+        }
     }
 
     async fn cancel_transaction_after_deadline(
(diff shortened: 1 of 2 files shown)


core/Cargo.toml
```diff
@@ -16,6 +16,7 @@ ethcontract = "0.7.2"
 futures = { version = "0.3.5", features = ["compat"] }
 futures-timer = "3.0.2"
 isahc = { version = "0.9.6", features = ["json"] }
+jsonrpc-core = "14.2.0"
```
nlordell, line 19: This is not needed and re-exported through `ethcontract` as `use ethcontract::jsonrpc`. I would prefer to use the re-export to ensure that the jsonrpc-core version always matches the version needed by `ethcontract` and `web3`.

core/src/solution_submission.rs
```diff
@@ -246,18 +266,67 @@ fn extract_transaction_receipt(err: &MethodError) -> Option<&TransactionReceipt>
...
+
+impl IsNonceError for ExecutionError {
+    fn is_nonce_error(&self) -> bool {
+        // This is the error as we've seen it on openethereum nodes.
+        // TODO: check how this looks on geth and infura. Not recognizing the error is not a serious
+        // problem but it will make us sometimes log an error when there actually was no problem.
+        match self {
+            ExecutionError::Web3(Web3Error::Rpc(RpcError { code, message, .. }))
+                if code.code() == -32010
+                    && message.find("Transaction nonce is too low.").is_some() =>
```
nlordell, line 292: I think this might be node dependant (especially the message).

core/src/solution_submission.rs
```diff
@@ -417,4 +486,80 @@ mod tests {
             .unwrap();
         assert!(result.is_err());
     }
+
+    #[test]
+    fn submission_completes_during_cancellation() {
```
nlordell, line 491: :rocket:

core/src/solution_submission.rs
```diff
@@ -211,27 +226,32 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
             });
             let cancel_future = self.cancel_transaction_after_deadline(batch_index, nonce);
 
+            // Run both futures at the same time. When one of them completes check whether the
+            // result is a "nonce already used error". If this is the case then the other future's
+            // transaction must have gone through so return that one instead.
+            // We need to handle this error because exactly one of the transactions will go through
+            // but we might observe the other transaction failing first.
             futures::pin_mut!(cancel_future);
             match future::select(submit_future, cancel_future).await {
```
nlordell, line 235: Is there a reason why you prefer this over the `futures::select!` macro? I find it a bit more ergonomic so you don't have to deal with `Either`. I think it just requires the futures to be `.shared()` so that they are fused and pinned, and then they can be reused after the `select!` invocation.

nlordell (commented):
Just some concerns about the nonce error detection may be node specific.

core/src/solution_submission.rs
e00E: I agree. There is a comment about this right above.

core/src/solution_submission.rs
e00E: I find function + Either more natural than a macro.  With select! I am less sure about what control flow happens internally. In this case we also need access to the future that is still running when the other one completes. It isn't clear from the select! documentation whether that is possible.

core/src/solution_submission.rs
nlordell: 🙈

core/src/solution_submission.rs
nlordell: To be clear, I fine with how it is now, I was just curious.

fleupold: Why not reuse the `select_ignoring_nonce_error` function you wrote in the other PR? With the current logic, I'd be slightly worried that there is still a chance the cancellation tx failed due to "nonce too low", yet the solution submission tx might have gone through (I know we wait 30s before we send the tx, but let there be some sync issue between the nodes or a small reorg). I think it would be better to actually wait on the result of the submit future.

fleupold (approved):
Really nice unit test! Wondering if we could reuse the select future helper from the other PR?

core/src/solution_submission.rs
e00E: I don't understand. If the cancellation tx has nonce too low then we do wait on the submit tx.

fleupold: You are right, the logic seems correct the way it is. 
I'm still wondering why the `select_ignoring_nonce_error` would not be applicable here?

e00E: That function is only introduced in the next PR and is specific to the retry module because of the return types and because of the fact that it takes a vector instead of working on two futures. I can see that a generalization might be possible so that the same function can be used in both cases but I'd rather do that after both PRs are done.

fleupold: Doing it in a separate PR makes sense. My reasoning for reusing it is that the logic itself here is rather complicated (cf. my misunderstanding of how the thing worked) and thus could benefit from being abstracted and tested away in a common logic part.

nlordell (approved):

core/src/solution_submission.rs
nlordell: Created https://github.com/gnosis/dex-services/issues/1196

PR #1183, opened 2020-07-23
Implement token price estimation in pricegraph

nlordell:
This PR implements token price estimation in `pricegraph`. Additionally this PR modifies the `historic_prices` E2E script to use this method instead of the bid ask spread. This significantly modifies what the E2E script was testing, so a new E2E script has been introduced in #1218 to produce a more valuable metric to the one that previously existed.

Regarding the imprecision, the average error of the price estimate is 28000x :warning: (mostly because of some large outliers and non-liquid tokens). In recent batches for a token like WETH this is down to around 4% since the orderbook is more dense and therefore the estimates are much more accurate. Note that the previous promising analysis on price accuracy was using a different mechanism to estimate **exchange rates** between tokens and not token prices (in OWL reference token). 

### Test Plan

New unit tests in `pricegraph`, e2e historic prices now uses the new estimation API:
```
cargo run --release -p e2e --bin historic_prices -- --orderbook-file target/orderbook-file-mainnet
    Finished release [optimized] target(s) in 2.82s
54285 / 54285 [===============================================] 100.00 % 222.25/s
Processed 13200 token prices: bad estimates 46.73%, average error 28770.95%.
```

### DIFF
--- a/e2e/src/bin/historic_prices.rs
+++ b/e2e/src/bin/historic_prices.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use core::{history::Settlement, models::BatchId};
 use e2e::cmd;
-use pricegraph::{Element, Pricegraph};
+use pricegraph::{Element, Pricegraph, TokenId};
 use std::{fs::File, io::Write, path::PathBuf};
 use structopt::StructOpt;
 
@@ -24,10 +24,9 @@ struct Options {
 
 fn main() -> Result<()> {
     let options = Options::from_args();
+    let mut report = Report::new(File::create(&options.output_dir.join("prices.csv"))?);
 
-    let mut spreads = File::create(&options.output_dir.join("spreads.csv"))?;
-    writeln!(&mut spreads, "batch,token,price,estimate")?;
-
+    report.header()?;
     cmd::for_each_batch(&options.orderbook_file, |history, batch| {
         let settlement = match history.settlement_for_batch(batch) {
             Some(value) => value,
@@ -35,18 +34,23 @@ fn main() -> Result<()> {
         };
 
         let auction_elements = history.auction_elements_for_batch(batch)?;
-        check_batch_prices(batch, &auction_elements, &settlement, &mut spreads)?;
+        check_batch_prices(&mut report, batch, &auction_elements, &settlement)?;
 
         Ok(())
-    })
+    })?;
+
+    report.summary()
 }
 
-fn check_batch_prices(
+fn check_batch_prices<T>(
+    report: &mut Report<T>,
     batch: BatchId,
     auction_elements: &[Element],
     settlement: &Settlement,
-    mut output: impl Write,
-) -> Result<()> {
+) -> Result<()>
+where
+    T: Write,
+{
     let token_ids = &settlement.solution.token_ids_for_price;
     let prices = &settlement.solution.prices;
     debug_assert_eq!(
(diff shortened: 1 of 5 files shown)


pricegraph/src/api/price_source.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// is no transitive order buying the fee token for the specified token).
+    ///
+    /// The fee token is defined as the token with ID 0.
+    pub fn estimate_token_price(&self, token: TokenId) -> Option<f64> {
+        if token == FEE_TOKEN {
+            return Some(OWL_BASE_UNIT);
+        }
+
+        // Estimate price of selling 1 unit of the reference token for each
+        // token. We sell rather than buy the reference token because volume is
```
e00E, line 21: "for each token" sounds strange because there is only one token. (In the pricegraph based price source this made sense.)

pricegraph/src/api/price_source.rs
```diff
@@ -0,0 +1,89 @@
+//! This module implements price source methods for the `Pricegraph` API so that
+//! it can be used for OWL price estimates to the solver.
+
+use crate::encoding::{TokenId, TokenPair};
+use crate::{Pricegraph, FEE_TOKEN};
+
+const OWL_BASE_UNIT: f64 = 1_000_000_000_000_000_000.0;
+
+impl Pricegraph {
+    /// Estimates the fee token price in WEI for the specified token. Returns
```
e00E, line 10: Maybe atoms instead of wei. Not sure what the standard ethereum term is but I thought wei was specifically for ether.

pricegraph/src/api/price_source.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// The fee token is defined as the token with ID 0.
+    pub fn estimate_token_price(&self, token: TokenId) -> Option<f64> {
+        if token == FEE_TOKEN {
+            return Some(OWL_BASE_UNIT);
+        }
+
+        // Estimate price of selling 1 unit of the reference token for each
+        // token. We sell rather than buy the reference token because volume is
+        // denominated in the sell token, for which we know the number of
+        // decimals.
```
e00E, line 23: In our case we know it and I don't think it's a problem to hardcode this but in general we do not know this either since you can pick a different fee token. Mentioning this in case it's an indicator that the algorithm should be changed to not make this assumption but I don't know how possible that is.

e00E (approved):
No problem code wise.

pricegraph/src/api/price_source.rs
nlordell: Hehe just copy pasted comment over, nice catch!

pricegraph/src/api/price_source.rs
nlordell: This is a very valid point! I think this is something we can cleanup in the Pricegraph without much work. I would rather leave that as a separate issue/PR though as this one was really about just moving the logic to a different crate.

e2e/src/bin/historic_prices.rs
```diff
@@ -1,10 +1,13 @@
 use anyhow::Result;
 use core::{history::Settlement, models::BatchId};
 use e2e::cmd;
-use pricegraph::{Element, Pricegraph, TokenPair};
+use pricegraph::{Element, Pricegraph, TokenId};
 use std::{fs::File, io::Write, path::PathBuf};
 use structopt::StructOpt;
 
+/// Threshold at which a price estimate is considered "bad"; currently 10%.
+const BAD_ESTIMATE_THRESHOLD: f64 = 0.1;
```
fleupold, line 9: In practice this is 10000 for the current solver, right?

pricegraph/src/api/price_source.rs
```diff
@@ -0,0 +1,93 @@
...
+use crate::{Pricegraph, FEE_TOKEN};
+
+const OWL_BASE_UNIT: f64 = 1_000_000_000_000_000_000.0;
+
+impl Pricegraph {
+    /// Estimates the fee token price in atoms for the specified token.
+    /// Specifically, this price repesents the number of OWL atoms required to
+    /// buy 1e18 atoms of the specified token.
+    ///
+    /// Returns `None` if there token is not connected to the fee token (that
```
fleupold, line 14: nit `the token`

fleupold (approved):
Looks good...

e2e/src/bin/historic_prices.rs
nlordell: Yeah, this is just to get an idea for "close-ish" estimates.

PR #1193, opened 2020-07-27
Only submit solution if gas price cap is at least fast gas price

### LINKED ISSUE #1186: Noop transaction did not go through because max gas price was low
Today we noticed in the logs that the noop transaction used to cancel the solution submission did not go through which caused a future solution submission to fail.

This was likely caused by the change that makes the gas price limited by calculated based on the earned fee of the solution. If this is low then the solution transaction does not go through and we send a cancel transaction with 12.5% higher gas price. If that price still isn't enough then the the transaction does not get cancelled.

One way to fix this is to always send the the noop transaction with some minimum gas price (like the "fast" estimate). On the other hand we could also argue that we do not need to fix this as this can only be a problem if the next solution also earns so little fees that the maximum gas price has not increased, so that solution cannot be worth submitting either.
Looking at it from this perspective the best solution imo is to not even attempt to submit the solution in the first place when the maximum gas price is below our estimate for "fast".

e00E:
Now that the max gas price is calculated based on the earned fees of the
solution it is possible that we have caps that are so low that not even
the cancellation transaction completes
This commit only attempts solution submission if the gas price is at
least "fast".

Fixes #1186.

This might be obsoleted by  #1099.

### Test Plan
Ci, new unit test.


fleupold (commented):
Can we use standard for the cutoff (I think fast might prevent us from sending some solutions which could be going through).

Also wouldn't this logic fit more naturally in the `SolutionTransactionSending` implementation? It already has access to the gas estimator and seems like it could also just refuse to send a tx that is obviously not going to go through?

core/src/solution_submission.rs
```diff
@@ -213,6 +217,19 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
         gas_price_cap: U256,
     ) -> BoxFuture<Result<(), SolutionSubmissionError>> {
         async move {
+            // If the gas price cap is not at least the fast gas price then submitting the solution
+            // is not feasible because it would take too long.
+            match self.gas_price_estimating.estimate_gas_price().await {
+                Ok(gas_price) => {
+                    if gas_price_cap < gas_price.fast {
+                        return Err(SolutionSubmissionError::Benign(format!(
+                            "not submitting solution because gas price cap {} < fast gas price {}",
```
nlordell, line 226: Maybe we can reword this error message a bit. I think it should indicate that the solution does not generate enough fees for the current network gas price estimate.

nlordell (approved):
Just a comment on the error message.

e00E:
> Can we use standard for the cutoff (I think fast might prevent us from sending some solutions which could be going through).

The first solution transaction we send already starts at price fast (in retry.rs) so it makes sense to check for the same limit here.

> Also wouldn't this logic fit more naturally in the `SolutionTransactionSending` implementation? It already has access to the gas estimator and seems like it could also just refuse to send a tx that is obviously not going to go through?

That trait only returns MethodError but can't decide whether the result should be benign or not. Changing that requires more refactoring and creates more conflicts with the "fix race condition" PRs. Seeing as min-avg-be-per-order might replace this change I'd rather not do this.

josojo:
> > Can we use standard for the cutoff (I think fast might prevent us from sending some solutions which could be going through).
> 
> The first solution transaction we send already starts at price fast (in retry.rs) so it makes sense to check for the same limit here.

I am fine with the fast price. I think if we are missing some submission due to the difference between standard and fast speed, then these solutions are not that valuable ( the more critical factor to influence the decision will anyways the subsidiary factor)

I think we can merge the PR today and then I would do a new release tomorrow

fleupold (approved):

nlordell:
> >Can we use standard for the cutoff (I think fast might prevent us from sending some solutions which could be going through)
>
> The first solution transaction we send already starts at price fast (in retry.rs) so it makes sense to check for the same limit here.

Now that I think about it a bit more, this makes sense since the transaction might still be mined with a "standard" gas fee, but we don't want to re-submit with a higher one.

I think it would be nice to track with a prometheus metric how many solutions are being skipped because of this check.

PR #1194, opened 2020-07-27
Filter 0 prices

fleupold:
During the MTA sale we saw that e.g. 1Inch can list tokens with a price of 0. This should rather be seen as "no price". If this is the only price source, a price of 0 may get propagated to the solver which searches for solutions in a +/- 100x range. With a price estimate of 0 this range is invalid.

I was thinking about creating a ZeroFiltering price source by itself rather than putting it in the AveragePriceSoruce. Still happy to do so if someone else sees value (it felt a bit like over-using the price source chain we have).

### Test Plan

Adjusted unit test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2939,9 +2939,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.56"
+version = "1.0.57"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3433e879a558dde8b5e8feb2a04899cf34fdde1fafb894687e52105fc1162ac3"
+checksum = "164eacbdb13512ec2745fb09d51fd5b22b0d65ed294a1dcf7285a360c80a675c"
 dependencies = [
  "itoa",
  "ryu",


e00E (approved):
This is fine. If we end up wanting an individual FilteringPriceSource then we can refactor later.

core/src/price_estimation/average_price_source.rs
```diff
@@ -56,10 +56,22 @@ fn lossless_merge<T: Eq + Hash, U>(map_collection: Vec<HashMap<T, U>>) -> HashMa
...
     // Lossless merger of the collection of hash maps. That is, putting all
-    // available prices for each token into a list to be averaged at the end.
+    // available prices for each token into a list (ignorign 0 values) to be
+    // averaged at the end.
     lossless_merge(price_maps)
         .iter()
-        .map(|(token, prices)| (*token, prices.iter().sum::<u128>() / prices.len() as u128))
+        .filter_map(|(token, prices)| {
+            let filtered_prices: Vec<u128> =
+                prices.iter().filter(|&&p| p > 0u128).cloned().collect();
```
nlordell, line 65: ubernit: I would slightly prefer not collecting here if possible (I think it should be with `fold`). However, if it makes the code harder to read, then we should leave it as is.

nlordell (approved):

nlordell (commented):
Now that I think of it - I think the best approach here would be to modify the `PriceSource` trait to return `HashMap<TokenId, NonZeroU128>` [1] so we have type-level guarantees that we don't want 0 prices.

Filtering non-zero prices in the `AveragePriceSource` feels like its in the wrong place, don't we want the same behaviour in `PriorityPriceSource` or for all `PriceSource`s in general as well?

[1]: https://doc.rust-lang.org/std/num/struct.NonZeroU128.html

e00E:
That is a good argument, that we can make this part of the api guarantees that the PriceSource trait makes. Then we would only have to implement the check in the leaf price sources (not the combinators like Average). The question is whether a valid price can ever be 0 to which the answer is probably no.

PR #1197, opened 2020-07-27
Cache token infos at startup to prevent double fetch

### LINKED ISSUE #1155: GetTokenInfo fetched twice in parallel
Looking at some kibana logs: https://logs.gnosisdev.com/goto/81b52cdb0cd9c7630dad320bdb4d525a

Screen Shot 2020-07-20 at 16 28 57

We are sending the exact same RPC call twice with different request IDs. The reason might be similar as in #1154 that we check some cache, realize we need to go to network but do not block any parallel requests to also going to network.

This might also be causing the timeouts we see as we are putting a lot of pressure on the nodes (#1153)

e00E:
When the cached token info fetcher does not yet contain a token it must
fetch it. If multiple calls `get_token_info` happen in parallel then we
likely make multiple requests to retrieve token info from the node
because of the way the code around the mutex works out.
This happens when the PriceOracle is created because we have multiple
price sources using the token info cache, and performing their initial
update at the same time.
This commit explicitly caches all tokens before the cache is passed to
the price sources.

Fixes #1155 

### Test Plan
CI

### DIFF
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -50,7 +50,7 @@ pub struct PriceOracle {
 
 impl PriceOracle {
     /// Creates a new price oracle from a token whitelist data.
-    pub fn new(
+    pub async fn new(
         http_factory: &HttpFactory,
         orderbook_reader: Arc<dyn StableXOrderBookReading>,
         contract: Arc<StableXContractImpl>,
@@ -61,6 +61,7 @@ impl PriceOracle {
             contract,
             token_data.clone().into(),
         ));
+        token_info_fetcher.cache_all(10).await?;
         let (kraken_source, _) = ThreadedPriceSource::new(
             token_info_fetcher.clone(),
             KrakenClient::new(http_factory, token_info_fetcher.clone())?,
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -252,6 +252,7 @@ fn main() {
         options.token_data,
         options.price_source_update_interval,
     )
+    .wait()
     .unwrap();
 
     // Setup price.


nlordell (commented):
I think we should just fix the locking logic in the `get_token_info` to immediately upgrade the read lock to write lock, check the cache again and then hold the write lock while retrieving the new token info.

e00E:
At first I agreed with your comment there but then I read Felix response above again and I agree that we don't want to hold the lock while making the request because it would block other cached reads at the same time.

core/src/price_estimation.rs
```diff
@@ -61,6 +61,7 @@ impl PriceOracle {
             contract,
             token_data.clone().into(),
         ));
+        token_info_fetcher.cache_all(10).await?;
```
fleupold, line 64: Not really related to this PR and not a big issue but I didn't really get at first was this argument was. I guess it makes sense when in an IDE as it tells you what the argument name, but maybe we can use a reasonable default given there is only one call site right now.

fleupold (commented):
I'd prefer to not block the critical startup path for non-critical things such as token lookup for price estimates. Maybe querying twice is not so bad after all?

e00E:
> I'd prefer to not block the critical startup path for non-critical things such as token lookup for price estimates. Maybe querying twice is not so bad after all?

Good point. I was just looking at this issue and this seemed like an easy fix so I'm fine with closing this too.

nlordell:
> At first I agreed with your comment there but then I read Felix response above again and I agree that we don't want to hold the lock while making the request because it would block other cached reads at the same time.

Yes, this is a very good point. The only way around this I see is to have signals (a mutex or oneshot channel) for each token being updated, that way each token is individually synchronized and we only need to hold the write lock when writing the update.

PR #1198, opened 2020-07-27
implement variable minimum average fee

nlordell:
Fixes partially #1099 

This PR implements the minimum average fee per order computation in the driver. This allows us to dynamically specify the minimum average fee constraint based on current gas prices.

Note that this PR in its current form doesn't really do anything and requires implementing `get_eth_price` for `PriceOracle`, which currently always fails. This will be done as a separate PR. It should "just work" like before and fall back to the specified min average fee per order specified at runtime.

### Test Plan

Unit test now verifies the min average fee computation.


core/src/price_estimation.rs
```diff
@@ -37,7 +37,14 @@ type Tokens = BTreeMap<TokenId, Option<TokenInfo>>;
...
 #[cfg_attr(test, mockall::automock)]
 pub trait PriceEstimating {
+    /// Retrieves price estimates for all token in the specified orders on a
+    /// best-effort basis. Critically, it is not considered an error to not be
+    /// able to retrieve a price estimate for a given token.
     fn get_token_prices<'a>(&'a self, orders: &[Order]) -> BoxFuture<'a, Tokens>;
+
+    /// Retrieves a price estimate for ETH in OWL.
+    #[allow(clippy::needless_lifetimes)]
+    fn get_eth_price<'a>(&'a self) -> BoxFuture<'a, Option<u128>>;
```
e00E, line 47: This is the price of 1 ether (10e18 wei) in owl atoms, right? I'd put this in the comment for clarity.

core/src/price_finding/optimization_price_finder.rs
```diff
@@ -193,28 +193,59 @@ pub struct OptimisationPriceFinder {
...
+            Err(err) => {
+                log::warn!("failed to estimate gas {:?}", err);
+                return self.default_min_avg_fee_per_order;
+            }
+        };
+
+        let subsidized_gas_price = (gas_estimate * eth_owl_xrate) / self.min_avg_fee_subsidy_factor;
+
+        const GAS_PER_ORDER: f64 = 120_000.0;
+        const FEE_ADJUSTMENT: f64 = 1_000.0;
```
e00E, line 244: This is because of 1/1000, right? I was confused because the parameter is called "fee per order" so I thought it was already referring to the earned fee. It must mean something else, but then I'm also wondering if this takes into account that we only get half of the fee imbalance as a reward. Maybe @josojo can clarify.

nlordell: AFAIK fees get applied twice per trade (for buying and selling) of which half are burned. So we should collect 0.1% of each order.

core/src/price_finding/optimization_price_finder.rs
```diff
@@ -193,28 +193,59 @@ pub struct OptimisationPriceFinder {
     io_methods: Arc<dyn Io + Send + Sync>,
     fee: Option<Fee>,
     solver_type: SolverType,
-    price_oracle: Box<dyn PriceEstimating + Sync>,
-    min_avg_fee_per_order: u128,
+    price_oracle: Box<dyn PriceEstimating + Send + Sync>,
+    gas_station: Arc<dyn GasPriceEstimating + Send + Sync>,
+    min_avg_fee_subsidy_factor: f64,
+    default_min_avg_fee_per_order: u128,
```
fleupold, line 199: Given we need all these components, would it make sense to create a new "MinAvgFeeComputer"? I find it also slightly weird that the PriceFinder needs to know about gas estimation thingies.

driver/src/main.rs
```diff
@@ -154,10 +154,15 @@ struct Options {
     )]
     solver_time_limit: Duration,
 
-    /// Solver parameter: minimal average fee per order
-    /// Its unit is [OWL]
+    /// Subsidy factor used to compute the minimum average fee per order in a
+    /// solution.
+    #[structopt(long, env = "MIN_AVG_FEE_SUBSIDY_FACTOR", default_value = "10.0")]
```
fleupold, line 159: Could this not be the same as the other subsidy factor as they are directly related? I'm worried that if they are too far apart we might get a bunch of solutions which eventually don't pass as we are not willing to retry with a high enough gas price.

core/src/price_finding/optimization_price_finder.rs
```diff
@@ -193,28 +193,59 @@ pub struct OptimisationPriceFinder {
...
             internal_optimizer,
         }
     }
+
+    async fn compute_min_average_fee_per_order(&self) -> u128 {
+        let eth_owl_xrate = match self.price_oracle.get_eth_price().await {
+            Some(owl_price) => (owl_price as f64) / 10f64.powi(18),
+            None => {
+                log::warn!("failed to find ETH price estimate");
+                return self.default_min_avg_fee_per_order;
```
fleupold, line 230: How about not requiring a min avg fee in this very rare case?

core/src/price_finding/optimization_price_finder.rs
fleupold: I'm also super confused why we multiply by a thousand. Shouldn't we compute `tx_cost_in_OWL = GAS_PER_ORDER * gas_price * eth_owl_xrate` which gives us some price in OWL and this is the minimum fee (maybe it has to be multiplied by 2 as the solver might consider overall fee and not just what the submitter gets credited)?

I would then in the end divide by the subsidy factor to make it very clear what the "real" economically viable value is and what the value is that we are willing to accept in the short term.

fleupold (commented):
I'd vote for a separate component that can contain that logic... 
In the future we might even want to make the retrier reuse some of the information that lead to computing the minAvgFee in the first place.

core/src/price_finding/optimization_price_finder.rs
nlordell: This makes a lot of sense to me. I had taken the equation from @josojo's issue which also multiplied by 1000 which I assumed it was for fees.

Maybe @josojo or @marcovc can give some more details on the expected value for the `minAvgFeePerOrder` parameter.

core/src/price_finding/optimization_price_finder.rs
nlordell: I like this idea.

driver/src/main.rs
nlordell: This was also unclear to me since these factors have slightly different semantics. I'd be happy with using just one as well though.

core/src/price_finding/optimization_price_finder.rs
nlordell: This was more to keep the current behaviour until #1200 is implemented. At that point we can take it out. I guess when we can't get gas price estimates then it will be hard to reliably send a transaction anyway.

core/src/price_finding/optimization_price_finder.rs
nlordell: The computation was indeed incorrect. I have updated the test case to use @josojo's number from the issue and it seems to be OK now.

e00E (approved):

driver/src/main.rs
nlordell: Do you have any ideas how we can merge these two subsidy factors into one? Maybe they can be a multiple of each other?

core/src/price_finding/optimization_price_finder.rs
nlordell: I think this is now nicely implemented using the `MinAverageFeeComputing` abstraction where we have the default implementation that computes it based on the gas estimate and price estimate as well as a `FixedMinAverageFee` implementation and a `PriorityMinAverageFee` to combine them. This way we can fall back to using a fixed value if we want to, or remove the fixed value all together by changing how we initialize in `driver/src/main.rs`.

driver/src/main.rs
e00E: Can't they be the same factor?
Here we say that the solution should earn at least as much ether as it costs to submit it at fast gas price. This value is then divided by subsidy factor.
In the other place we say that the gas price cap is the amount earned through fees. This value is then multiplied by subsidy factor.
Either way we use the subsidy factor to effectively divide the gas price.

nlordell: Sounds reasonable to me.

core/src/price_finding.rs
```diff
@@ -1,20 +1,23 @@
...
 use log::info;
+use std::sync::Arc;
 
 pub fn create_price_finder(
     fee: Option<Fee>,
     solver_type: SolverType,
-    price_oracle: impl PriceEstimating + Sync + 'static,
-    min_avg_fee_per_order: u128,
+    price_oracle: Arc<dyn PriceEstimating + Send + Sync>,
+    min_avg_fee: Arc<dyn MinAverageFeeComputing>,
```
fleupold, line 20: The Arc is not yet needed, right? Just here for future proof component sharing?

core/src/price_finding/min_avg_fee.rs
```diff
@@ -0,0 +1,179 @@
...
+//! price estimates.
+
+use crate::{gas_station::GasPriceEstimating, price_estimation::PriceEstimating};
+use anyhow::{anyhow, Result};
+use futures::future::{BoxFuture, FutureExt as _};
+use std::sync::Arc;
+
+/// Trait that abstracts the minimum average fee computation.
+#[cfg_attr(test, mockall::automock)]
+pub trait MinAverageFeeComputing: Send + Sync + 'static {
```
fleupold, line 11: ❤️ this component and the abstractions you built around it.

driver/src/main.rs
fleupold: Agree it should be the same. Maybe we could even move the logic to have `ApproximateMinAverageFee` also be able to tell you the gas price given some solution.

fleupold (approved):

core/src/price_finding.rs
nlordell: It just makes the `MinAverageFeeComputing` to share later, for example if we share it with the solution submission component for economic viability computation.

driver/common.env
```diff
@@ -3,4 +3,3 @@ NETWORK_ID=5777
 PRIVATE_KEY=4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d
 SOLVER_TYPE=naive-solver
 SCHEDULER=evm
-GAS_PRICE_CAP_SUBSIDY_FACTOR=100
```
nlordell, line 6: @e00E Is it OK to remove this? And if not should the default value be updated in `driver/src/main.rs`?

PR #1199, opened 2020-07-27
Filter 0 prices

fleupold:
During the MTA sale we saw that e.g. 1Inch can list tokens with a price of 0. This should rather be seen as "no price". If this is the only price source, a price of 0 may get propagated to the solver which searches for solutions in a +/- 100x range. With a price estimate of 0 this range is invalid.

This PR is a reimplementation of #1194 that uses `std::num::NonZeroU128` to achieve this.

Test Plan
Adjusted unit test.


nlordell (approved):
Looks good.

e00E (approved):

bh2smith (approved):

Issue #1111, opened 2020-07-13
Return value for `atoms=true` should be whole numbers


fleupold:
> https://price-estimate-mainnet.dev.gnosisdev.com/api/v1/markets/7-4/estimated-amounts-at-price/0.000000000001019854?atoms=true
{"baseTokenId":"7","quoteTokenId":"4","buyAmountInBase":"25323216136017510000000","sellAmountInQuote":"25825983269.182007"}

Note, that sellAmountInQuote is returned as a fraction of atoms. Semantically, I don't think it makes sense to receive a fraction of atoms (which by definition should be the smallest indivisible unit). I'd therefore suggest to floor this number and make sure we are returning "whole" numbers.

fleupold:
@e00E mentioned on slack:

> Since we are returning json I prefer this being a json number which can be fractional. Even with atoms=true with things like the rounding buffer a fractional value is sensible.

fleupold:
I don't see how the rounding buffer makes fractional atoms more sensible. As a consumer of this API I want to know how many atoms I can sell or buy (in most cases to interact with the Smart Contract) and I shouldn't have to know anything about what or how rounding is applied (all I know is I can somewhat "trust" this number to give me a trade).
If I receive a fractional result I don't see any way other than rounding down on the client on how to convert this information into something I can actually make use of.

e00E:
In euros the atom is a cent. If I'm selling something at a price of 0.5 cents and you ask me how much you need to pay to receive one item then answer is 1 cent. Yet if I answer 0.5 this helps you realize that you could also buy 2 for the same price.

Regardless of the answer to whether we should round when we're parsing the number in another application we still won't be able to do in rust `u128::from_str` because we encode the number according to json which means it could look like `10.0` or `1e1`.

fleupold:
I agree that the price can remain in floats, however the amounts you receive/sell in atoms shouldn't be fractional. Note, that due to slippage the doubling of sell amounts doesn't mean a doubling of buy amounts per se.

As for the implementation, I think the string should never have float notation (scientific notation or trailing .0). Maybe we cannot use `display_fromstr` in that case although it seems to work for now. To be more robust we could have an opaque field atoms on `EstimatedOrderResult` struct that doesn't get serialized but can be used in a custom serialization to convert to uint before converting to string.

e00E:
> As for the implementation, I think the string should never have float notation

[...]

PR #1202, opened 2020-07-28
Add enum for modeling amounts so that atoms are never fractional

nlordell:
Related to #1111 

This PR defines a new `Amount` enum for representing token amounts in either atoms or base units. This has the effect of guaranteeing that atom amounts are never fractional. Note that the format is otherwise unchanged and base units are serialized as strings ("4.2" for example - see unit test) as there is still some discussion about how this should be done in the above mentioned issue.

Note that the transitive orderbook route does not use this type as it would be a "breaking" API change and we would need to evaluate with the FE team whether or not this can be changed. This should be easy to change in a follow up PR.

### Test Plan

Updated unit test and added new one. Additionally, run the price estimator and check that the results are in sync:
```
$ curl -s 'http://localhost:8080/api/v1/markets/7-1/estimated-buy-amount/1000000000000000000?atoms=true' | jq
{
  "baseTokenId": 7,
  "quoteTokenId": 1,
  "buyAmountInBase": "296069471770719551488",
  "sellAmountInQuote": "1000000000000000000"
}
$ curl -s 'http://localhost:8080/api/v1/markets/7-1/estimated-buy-amount/1?atoms=false' | jq                 
{
  "baseTokenId": 7,
  "quoteTokenId": 1,
  "buyAmountInBase": "296.06947177071953",
  "sellAmountInQuote": "1"
}
```

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -214,18 +214,17 @@ async fn estimate_buy_amount(
         .await
         .map_err(error::internal_server_rejection)?
         .order_for_sell_amount(token_pair, sell_amount_in_quote_atoms);
-    let buy_amount_in_base_atoms = Amount::Atoms(
+
+    let mut buy_amount_in_base = Amount::Atoms(
         transitive_order
             .map(|order| apply_rounding_buffer(order.buy, price_rounding_buffer))
             .unwrap_or_default() as _,
     );
-
-    let buy_amount_in_base = if query.atoms {
-        buy_amount_in_base_atoms
-    } else {
+    if !query.atoms {
         let token_info = get_token_info(token_pair.buy, token_infos.as_ref()).await?;
-        buy_amount_in_base_atoms.into_base_units(&token_info)
+        buy_amount_in_base = buy_amount_in_base.into_base_units(&token_info)
     };
+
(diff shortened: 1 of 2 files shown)


e00E (approved):

price-estimator/src/filter.rs
```diff
@@ -199,25 +199,32 @@ async fn estimate_buy_amount(
...
+            .map(|order| apply_rounding_buffer(order.buy, price_rounding_buffer))
+            .unwrap_or_default() as _,
+    );
+
     let buy_amount_in_base = if query.atoms {
         buy_amount_in_base_atoms
     } else {
         let token_info = get_token_info(token_pair.buy, token_infos.as_ref()).await?;
-        buy_amount_in_base_atoms / token_info.base_unit_in_atoms()
+        buy_amount_in_base_atoms.into_base_units(&token_info)
```
fleupold, line 227: just noticing this now, but for the naming it's somewhat unfortunate that we have prices in base (currency in) atoms and base units. Maybe we don't need the atoms in the variable name anymore now that it's enforced by the type system.

fleupold (approved):
Looks good. Typesafety for the win!

PR #1206, opened 2020-07-28
Update infallible price source with prices

e00E:
In the main orderbook update loop also update the infallible price source.
Routes are not yet using the rounding buffer.

### Test Plan
Tested that price estimator still works.

### DIFF
--- a/core/src/price_estimation/average_price_source.rs
+++ b/core/src/price_estimation/average_price_source.rs
@@ -34,8 +34,8 @@ impl PriceSource for AveragePriceSource {
 /// Get the price from each price source and apply `average_prices` to them.
 /// Errors if all price sources fail. If some but not all fail then the failure is logged and the
 /// failures are not part of the average but no error is returned.
-pub async fn average_price_sources(
-    sources: impl Iterator<Item = &dyn PriceSource>,
+pub async fn average_price_sources<'a>(
+    sources: impl Iterator<Item = &'a (impl PriceSource + 'a + ?Sized)>,
     tokens: &[TokenId],
 ) -> Result<HashMap<TokenId, NonZeroU128>> {
     let futures = future::join_all(sources.map(|s| s.get_prices(tokens))).await;
--- a/price-estimator/src/orderbook.rs
+++ b/price-estimator/src/orderbook.rs
@@ -1,12 +1,10 @@
-use crate::{infallible_price_source::InfalliblePriceSource, solver_rounding_buffer};
+use crate::{infallible_price_source::UpdatingInfalliblePriceSource, solver_rounding_buffer};
 use anyhow::Result;
 use core::{
     models::{AccountState, BatchId, Order, TokenId},
     orderbook::StableXOrderBookReading,
-    price_estimation::{average_price_source, price_source::PriceSource},
 };
 use pricegraph::{Pricegraph, TokenPair};
-use std::collections::HashMap;
 use tokio::sync::RwLock;
 
 #[derive(Debug)]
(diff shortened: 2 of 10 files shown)


core/src/price_estimation/average_price_source.rs
```diff
@@ -54,7 +54,7 @@ fn lossless_merge<T: Eq + Hash, U>(map_collection: Vec<HashMap<T, U>>) -> HashMa
     gathered_maps
 }
 
-fn average_prices(price_maps: Vec<HashMap<TokenId, u128>>) -> HashMap<TokenId, u128> {
+pub fn average_prices(price_maps: Vec<HashMap<TokenId, u128>>) -> HashMap<TokenId, u128> {
```
e00E, line 57: We use this function directly because we do not want to store just an AveragePriceSource because we update the pricegraph based price source manually.

nlordell: Maybe this means we need to refactor the `AveragePriceSource` abstraction a bit? Alternatively, maybe we need to implement an `UpdateingOrderbookBased` price source in `price-estimator` that can integrate with `AveragePriceSource` but be updated by external triggers?

e00E: I had the same thought about AveragePriceSource but the abstractions I came up with were all overly complex for what I want to accomplish here.
Your second idea is simpler but it still have the same feeling that the amount of extra code needed is not worth it.
To me there is no problem having this function public. This is the most generic form of calculating the average because it does not care about how the prices are fetched or who owns the price sources.

e00E:
Making a separate PR to discuss AveragePriceSource in.

core/src/price_estimation/average_price_source.rs
nlordell: > takes ownership of the inner price sources

One idea is we can:
```
impl<T: PriceSource> for Arc<T> { ... }
```

Then we can `Box<Arc<T>>` to share the inner price source? Alternatively, we can just make the `AveragePriceSource` take `Arc`.

fleupold: Don't we already have a ThreadedPriceSource, which is basically a cache that can get updated externally (by a timeout running in a separate thread). Maybe we can make that more generic that it can also be updated by a synchronous trigger.

core/src/price_estimation/average_price_source.rs
```diff
@@ -34,8 +34,8 @@ impl PriceSource for AveragePriceSource {
 /// Get the price from each price source and apply `average_prices` to them.
 /// Errors if all price sources fail. If some but not all fail then the failure is logged and the
 /// failures are not part of the average but no error is returned.
-pub async fn average_price_sources(
-    sources: impl Iterator<Item = &dyn PriceSource>,
+pub async fn average_price_sources<'a>(
+    sources: impl Iterator<Item = &'a (impl PriceSource + 'a + ?Sized)>,
```
e00E, line 38: This allows send and sync bounds to be preserved

price-estimator/src/orderbook.rs
```diff
@@ -40,8 +43,12 @@ impl Orderbook {
             }),
             // TODO: pass this from command line argument
             extra_rounding_buffer_factor: 2.0,
-            // TODO: pass real token infos, and update it with a real price source in a future PR
-            infallible_price_source: InfalliblePriceSource::new(HashMap::new()),
+            // TODO: pass real token infos
+            infallible_price_source: RwLock::new(InfalliblePriceSource::new(HashMap::new())),
```
fleupold, line 47: Looking at all the dependencies needed for the the InfalliblePriceSource, I'm wondering if the Infallible price source should be in charge of updating itself given a pricegraph (and own the external prices sources and all_tokens itself). 
Looking at the TODOs here it might be worthwhile to extract the "creation" logic somewhere outside the orderbook and make the pricesource a parameter for new.

price-estimator/src/orderbook.rs
```diff
@@ -63,12 +70,15 @@ impl Orderbook {
         }
     }
 
-    /// Recreate the pricegraph orderbook.
+    /// Recreate the pricegraph orderbook and update the infallible price source.
```
fleupold, line 73: nit: we could add a unit test that the infallible price source is actually updated on `orderbook.upate`

fleupold (approved):
Looks good!

price-estimator/src/orderbook.rs
e00E: Implemented this suggestion.

nlordell (approved):

price-estimator/src/orderbook.rs
e00E: Added test

price-estimator/src/orderbook.rs
```diff
@@ -131,3 +143,50 @@ fn pricegraph_from_auction_data(auction_data: &AuctionData) -> Pricegraph {
...
+mod tests {
+    use super::*;
+    use core::{
+        models::TokenId, orderbook::NoopOrderbook, price_estimation::price_source::PriceSource,
+    };
+    use futures::future::{BoxFuture, FutureExt as _};
+    use std::{collections::HashMap, num::NonZeroU128};
+
+    #[test]
+    fn updates_infallible_price_source() {
```
nlordell, line 157: Nice!

PR #1207, opened 2020-07-28
Make price oracle's external price sources available

e00E:
This is going to be used by the price estimator.

### Test Plan
No logic change.

### DIFF
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -61,10 +61,10 @@ impl PriceOracle {
             contract,
             token_data.clone().into(),
         ));
-        let mut external_price_sources =
+        let mut price_sources =
             external_price_sources(http_factory, token_info_fetcher.clone(), update_interval)?;
-        external_price_sources.push(Box::new(PricegraphEstimator::new(orderbook_reader)));
-        let averaged_source = Box::new(AveragePriceSource::new(external_price_sources));
+        price_sources.push(Box::new(PricegraphEstimator::new(orderbook_reader)));
+        let averaged_source = Box::new(AveragePriceSource::new(price_sources));
         let prioritized_source = Box::new(PriorityPriceSource::new(vec![
             Box::new(token_data),
             averaged_source,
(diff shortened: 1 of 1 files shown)


core/src/price_estimation.rs
```diff
@@ -160,6 +143,30 @@ impl PriceEstimating for PriceOracle {
...
+    token_info_fetcher: Arc<dyn TokenInfoFetching>,
+    update_interval: Duration,
+) -> Result<Vec<Box<dyn PriceSource + Send + Sync>>> {
+    let kraken = KrakenClient::new(http_factory, token_info_fetcher.clone())?;
+    let dexag = DexagClient::new(http_factory, token_info_fetcher.clone())?;
+    let oneinch = OneinchClient::new(http_factory, token_info_fetcher.clone())?;
+    Ok(vec![
+        thread_and_box(kraken, token_info_fetcher.clone(), update_interval),
+        thread_and_box(dexag, token_info_fetcher.clone(), update_interval),
+        thread_and_box(oneinch, token_info_fetcher.clone(), update_interval),
```
fleupold, line 158: nanonit: last clone not necessary.

core/src/price_estimation.rs
```diff
@@ -61,27 +61,10 @@ impl PriceOracle {
...
-            OneinchClient::new(http_factory, token_info_fetcher.clone())?,
-            update_interval,
-        );
-        let averaged_source = Box::new(AveragePriceSource::new(vec![
-            Box::new(kraken_source),
-            Box::new(dexag_source),
-            Box::new(oneinch_source),
-            Box::new(PricegraphEstimator::new(orderbook_reader)),
-        ]));
+        let mut external_price_sources =
```
fleupold, line 64: ubernit: since we add an internal one, I'd call these just price_sources.

fleupold (approved):
How is the price estimator using these?

e00E:
> How is the price estimator using these?

For the rounding buffer we need to know roughly the prices the solver uses. These prices sources are used by the solver so we also want them in price estimator.

nlordell (approved):

fleupold:
> These prices sources are used by the solver so we also want them in price estimator.

But then we want to add the pricegraph one manually because of the way the threading works?

e00E:
Yes

PR #1209, opened 2020-07-28
Simplify select_all future in solution submission retry

e00E:
It turns out that we accomplished with the function has already been
mostly implemented in the futures crate in the form of
FuturesUnordered.

### Test Plan
Previous unit tests still pass.


nlordell (approved):
:tada: Nice find!

fleupold (approved):
Great simplification!

PR #1215, opened 2020-07-29
Allow average price source to be used with references

e00E:
The struct AveragePriceSource forces ownership semantics onto the user.
We extract the average computation out of the struct so that it can be
used with references.
This is going to be useful for the price estimator where it won't be possible for AveragePriceSource to own the pricegraph based price source but we nonetheless want to calculate the average by reusing the existing code.

### Test Plan
No logic change, existing tests still work.

---

I also tried to make a more advanced version that makes the average price source generic over whether it owns the price source or has references. 

```rust
struct AveragePriceSource_<T0, T1: ?Sized, T2> {
    t: T0,
    _t1: PhantomData<T1>,
    _t2: PhantomData<T2>,
}

impl<'a, T0, T1, T2> AveragePriceSource_<T0, T1, T2>
where
    T0: AsRef<T1>,
    T1: 'a + ?Sized,
    &'a T1: IntoIterator<Item = T2>,
    T2: AsRef<dyn PriceSource + 'a>,
{
    fn new(t: T0) -> Self {
        Self {
            t,
            _t1: PhantomData,
            _t2: PhantomData,
        }
    }

    async fn get_prices(&'a self, tokens: &[TokenId]) -> Result<HashMap<TokenId, NonZeroU128>> {
        // For brevity we return the result of the first price source but in the real code we would average.
        let t1: &'a T1 = self.t.as_ref();
        let mut iter = t1.into_iter();
        let first = iter.next().unwrap();
        let source: &dyn PriceSource = first.as_ref();
        source.get_prices(tokens).await
    }
}

#[test]
fn compiles() {
    type BoxDyn = Box<dyn PriceSource>;

    let owned: Vec<BoxDyn> = Vec::new();
    let _ = AveragePriceSource_::<_, [BoxDyn], _>::new(owned).get_prices(&[]);

    let slice: &[BoxDyn] = &[];
    let _ = AveragePriceSource_::<_, [BoxDyn], _>::new(slice).get_prices(&[]);

    let slice_of_references: &[&BoxDyn] = &[];
    let _ = AveragePriceSource_::<_, [&BoxDyn], _>::new(slice_of_references).get_prices(&[]);
}
```

This works but there are limitations:

* You have to manually specify one of the generic types.
* I was unable to make this work with BoxFuture probably because there is a compiler [bug](https://github.com/rust-lang/rust/issues/64552) around how `Send` is propagated.
* Even ignoring the `Send` issue I was unable to get the lifetimes to work when implementing the real `trait PriceSource`. I feel this should be possible but gave up for now.


nlordell (approved):

nlordell:
What about using `Box<Arc<SomePriceSourceImplementation>>` to share the inner price source implementation?

e00E:
> What about using `Box<Arc<SomePriceSourceImplementation>>` to share the inner price source implementation?

It doesn't feel right to have to do this workaround because of the ownership design choice. But also in price estimator specifically the pricegraph is behind a mutex in a struct so then the custom price source would need a reference to the mutex and do locking even though locking shouldn't be needed when doing it another way.

nlordell:
> because of the ownership design choice

I think this is the caller's choice right? It wants an `AveragePriceSource` with a shared internal price source. The idea would be to `impl<T: PriceSource> PriceSource for Arc<T>` and then intantiate the `AveragePriceSource` with a `Box<Arc<PrigraphPriceSourceUsedByTHePriceEstimator>>`.

This way `AveragePriceSource` doesn't care that it has a boxed `Arc`, but the price estimator can use this boxed `Arc` to get at the `Pricegraph` instance for updating and API handlers.

core/src/price_estimation/average_price_source.rs
```diff
@@ -6,6 +6,7 @@ use std::collections::HashMap;
 use std::hash::Hash;
 use std::num::NonZeroU128;
 
+/// Owns inner price sources and applies `average_price_sources` to them in `get_prices`.
```
fleupold, line 9: ubernit: I'm not a fan of this comment as I don't see how it gives additional non-obvious information and is likely going to get outdated if anything about the methods below changes.

core/src/price_estimation/average_price_source.rs
```diff
@@ -21,41 +22,44 @@ impl PriceSource for AveragePriceSource {
...
+    if !acquired_prices.is_empty() {
+        Ok(average_prices(acquired_prices))
+    } else {
+        Err(anyhow!("All price sources failed!"))
     }
-    gathered_maps
 }
 
-fn average_prices(price_maps: Vec<HashMap<TokenId, NonZeroU128>>) -> HashMap<TokenId, NonZeroU128> {
+pub fn average_prices(
```
fleupold, line 60: Does this still need to be public?

core/src/price_estimation/average_price_source.rs
```diff
@@ -21,41 +22,44 @@ impl PriceSource for AveragePriceSource {
...
         .boxed()
     }
 }
 
-/// Lossless merger of a collection of maps puts all available values into a list for each available key
-fn lossless_merge<T: Eq + Hash, U>(map_collection: Vec<HashMap<T, U>>) -> HashMap<T, Vec<U>> {
-    let mut gathered_maps: HashMap<T, Vec<U>> = HashMap::new();
-    for (key, value) in map_collection.into_iter().flatten() {
-        gathered_maps.entry(key).or_default().push(value);
+/// Get the price from each price source apply `average_prices` to them.
```
fleupold, line 35: ubernit: missing some word like and

fleupold (approved):
This works. Any reason to not change the Box to and Arc to allow shared instead of sole ownership?

e00E:
> This works. Any reason to not change the Box to and Arc to allow shared instead of sole ownership?

I tried to highlight this in my generics example in the PR message which shows how I'd like it to work. The Arc + custom price source in price estimator would work but I feel is more convoluted than this solution because this way I do not need to work around the AveragePriceSource limitations.

core/src/price_estimation/average_price_source.rs
e00E: I don't think it *has* to be but it still makes sense for it. Again this is more control in case you already have the retrieved the hash maps yourself for whatever reason so you are not forced to use the function that also awaits the price sources.

PR #1216, opened 2020-07-29
Retry submitting solution during network congestion

### LINKED ISSUE #1187: Retry submitting solution during network congestion
Currently, when we receive a `There are too many transactions in the queue. You're transaction was dropped due to limit. Try increasing the fee` error from the node we stop trying to submit a solution but instead should probably retry submission with a higher gas amount (treat this error as a confirmation timeout error).

e00E:
Currently we only check for "nonce already used" errors specifically to
determine when we should retry sending a transaction with a higher gas
price and when we should cancel a solution submission transaction.
This commit changes this to include all what openethereum calls
transaction errors. These errors indicate that the transaction did not
execute which can happen for several reasons.
This makes us more flexible with regards to the exact error string. It
has the side effect that we will try to resubmit the transaction with a
higher gas price in cases where this won't help but it also will not
hurt as we will still end up with the same error eventually.
openethereum transaction errors that share this error code (-32010):

```rust
AlreadyImported => "Transaction with the same hash was already imported.".into(),
Old => "Transaction nonce is too low. Try incrementing the nonce.".into(),
TooCheapToReplace { prev, new } => {
        format!("Transaction gas price {} is too low. There is another transaction with same nonce in the queue{}. Try increasing the gas price or incrementing the nonce.",
                        new.map(|gas| format!("{}wei", gas)).unwrap_or("supplied".into()),
                        prev.map(|gas| format!(" with gas price: {}wei", gas)).unwrap_or("".into())
        )
}
LimitReached => {
        "There are too many transactions in the queue. Your transaction was dropped due to limit. Try increasing the fee.".into()
}
InsufficientGas { minimal, got } => {
        format!("Transaction gas is too low. There is not enough gas to cover minimal cost of the transaction (minimal: {}, got: {}). Try increasing supplied gas.", minimal, got)
}
InsufficientGasPrice { minimal, got } => {
        format!("Transaction gas price is too low. It does not satisfy your node's minimal gas price (minimal: {}, got: {}). Try increasing the gas price.", minimal, got)
}
InsufficientBalance { balance, cost } => {
        format!("Insufficient funds. The account you tried to send transaction from does not have enough funds. Required {} and got: {}.", cost, balance)
}
GasLimitExceeded { limit, got } => {
        format!("Transaction cost exceeds current gas limit. Limit: {}, got: {}. Try decreasing supplied gas.", limit, got)
}
InvalidSignature(ref sig) => format!("Invalid signature: {}", sig),
InvalidChainId => "Invalid chain id.".into(),
InvalidGasLimit(_) => "Supplied gas is beyond limit.".into(),
SenderBanned => "Sender is banned in local queue.".into(),
RecipientBanned => "Recipient is banned in local queue.".into(),
CodeBanned => "Code is banned in local queue.".into(),
NotAllowed => "Transaction is not permitted.".into(),
TooBig => "Transaction is too big, see chain specification for the limit.".into(),
InvalidRlp(ref descr) => format!("Invalid RLP data: {}", descr),
```
Fixes #1187 

### Test Plan
Tests still pass.

### DIFF
--- a/core/src/solution_submission.rs
+++ b/core/src/solution_submission.rs
@@ -254,7 +254,7 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
             futures::pin_mut!(cancel_future);
             match future::select(submit_future, cancel_future).await {
                 Either::Left((submit_result, cancel_future)) => {
-                    if submit_result.is_nonce_error() {
+                    if submit_result.is_transaction_error() {
                         log::info!("solution submission transaction is nonce error");
                         Err(convert_cancel_result(cancel_future.await))
                     } else {
@@ -264,7 +264,7 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
                     }
                 }
                 Either::Right((cancel_result, submit_future)) => {
-                    if cancel_result.is_nonce_error() {
+                    if cancel_result.is_transaction_error() {
                         log::info!("cancel transaction is nonce error");
                         self.convert_submit_result(batch_index, solution, submit_future.await)
                             .await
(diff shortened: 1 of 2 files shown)


nlordell:
I'm not the biggest fan of just checking error code -32010 since it feels a bit hacky (for example we could try submitting the solution multiple times when it has an `InsufficientBalance`). However, it is a very pragmatic and practical solution to this issue.

I did a quick search and it doesn't look like Geth shares this error code either :( BOO! I also wasn't able to find and error that is emitted in low gas price situations (admittedly, the search was very shallow).

However, from https://eth.wiki/json-rpc/json-rpc-error-codes-improvement-proposal it looks like 101 error code seems like it should be the "correct standard" for gas price too low errors, so maybe that should also be checked?

core/src/solution_submission.rs
```diff
@@ -297,40 +297,38 @@ fn convert_cancel_result(result: Result<(), NoopTransactionError>) -> SolutionSu
...
-            ExecutionError::Web3(Web3Error::Rpc(RpcError { code, message, .. }))
-                if code.code() == -32010
-                    && message.find("Transaction nonce is too low.").is_some() =>
-            {
-                true
-            }
-            _ => false,
-        }
+        let code = match self {
+            ExecutionError::Web3(Web3Error::Rpc(RpcError { code, .. })) => code.code(),
```
nlordell, line 312: You could use the new `matches!` macro:
```
matches!(self, ExecutionError::Web3(Web3Error::Rpc(RpcError { code, .. })) if code == -32010)
```

core/src/solution_submission.rs
```diff
@@ -297,40 +297,38 @@ fn convert_cancel_result(result: Result<(), NoopTransactionError>) -> SolutionSu
...
+            _ => return false,
+        };
+        code == -32010
     }
 }
 
-impl IsNonceError for Result<(), MethodError> {
-    fn is_nonce_error(&self) -> bool {
+impl IsTransactionError for Result<(), MethodError> {
+    fn is_transaction_error(&self) -> bool {
```
nlordell, line 320: naming-nit: This implementation seems very specific to OpenEthereum, so maybe this trait should be (or call internally) a `is_open_ethereum_transaction_error`?

nlordell (approved):
Looks good! A very pragmatic approach to a very hard problem. It feels icky, but now that I thought about it, I think it is probably the "cleanest" solution.

fleupold (approved):
I think checking the error code is better than checking a string and quickly double checked kibana that in the past 6 months we have only seen the three flavors of nonce errors.

PR #1217, opened 2020-07-29
[Easy] Use rinkeby gas staton on rinkeby

fleupold:
The DIA team was trying to test their sale on rinkeby and is not getting any trades settled, because:

> Solution does not generate enough fees for the transaction to execute quickly enough: price cap 239628589 < fast gas price 93500000001

The gas price on rinkeby is usually around 1GWEI but we are using the mainnet estimates for determining economic viability. This PR makes it so that we use the appropriate URL.

### Test Plan
Run locally, see gas estimates for mainnet and rinkeby be different.


core/src/gas_station.rs
```diff
@@ -6,8 +6,19 @@ use isahc::http::uri::Uri;
 use serde::Deserialize;
 use uint::FromDecStrErr;
 
-/// The default uri at which the gas station api is available under.
-pub const DEFAULT_URI: &str = "https://safe-relay.gnosis.io/api/v1/gas-station/";
+/// The default uris at which the gas station api is available under.
+const DEFAULT_MAINNET_URI: &str = "https://safe-relay.gnosis.io/api/v1/gas-station/";
+const DEFAULT_RINKEBY_URI: &str = "https://safe-relay.rinkeby.gnosis.io/api/v1/gas-station/";
+
+pub fn api_url_from_network_id(network_id: u64) -> &'static str {
```
nlordell, line 13: ubernit: I would slightly prefer returning an `Option<&'static str>` and then `.unwrap()`-ing at the call-site.

nlordell (approved):
Luks gud 2 me, just one minor nit.

PR #1218, opened 2020-07-29
Historic trades

### LINKED ISSUE #1022: Generate historic pricegraph estimates metric.
The idea is to historically run through the entire exchange history and check all orders against price estimates and see if the ones with reasonable xrates (i.e. with an xrate that is smaller than the pricegraph estimate) were matched.

nlordell:
Fixes #1022 

This PR implements historic trade estimates to compute how accurate the price graph is relative to historic batches.

This PR is rather big, I can try and split it into smaller pieces if it becomes too much to review.

### Test Plan

Run the historic trades and see that around 6% of pricegraph estimates are off (This takes a very long time).
```
cargo run --release -p e2e --bin historic_trades -- --orderbook-file target/orderbook-file-mainnet
   Compiling pricegraph v0.1.0 (/var/home/nlordell/Developer/dex-services/pricegraph)
   Compiling core v0.1.0 (/var/home/nlordell/Developer/dex-services/core)
    Finished release [optimized] target(s) in 30.55s
     Running `target/release/historic_trades --orderbook-file target/orderbook-file-mainnet`
54083 / 54083 [=============================================================] 100.00 % 28.00/s
Processed 87872 orders: 93.03% correct, 2.04% failed, 4.93% skipped.
```

### DIFF
--- a/e2e/src/bin/historic_trades.rs
+++ b/e2e/src/bin/historic_trades.rs
@@ -9,7 +9,7 @@ use std::{fs::File, io::Write, path::PathBuf};
 use structopt::StructOpt;
 
 const FULLY_FILLED_THREASHOLD: f64 = 0.95;
-const MIN_AMOUNT: u128 = 10_000;
+const MIN_AMOUNT: u128 = pricegraph::MIN_AMOUNT as _;
 
 /// Common options for analyzing historic batch data.
 #[derive(Debug, StructOpt)]
--- a/e2e/src/bin/historic_trades.rs
+++ b/e2e/src/bin/historic_trades.rs
@@ -4,11 +4,12 @@ use core::{
     models::BatchId,
 };
 use e2e::cmd;
-use pricegraph::{Element, Pricegraph, TokenPair};
+use pricegraph::{Element, Pricegraph, TokenPair, U256};
 use std::{fs::File, io::Write, path::PathBuf};
 use structopt::StructOpt;
 
 const FULLY_FILLED_THREASHOLD: f64 = 0.95;
+const MIN_AMOUNT: u128 = 10_000;
 
 /// Common options for analyzing historic batch data.
 #[derive(Debug, StructOpt)]
@@ -29,7 +30,9 @@ struct Options {
 
 fn main() -> Result<()> {
     let options = Options::from_args();
-    let mut report = Report::new(File::create(options.output_dir.join("trades.csv"))?)?;
+
+    let mut report = Report::new(File::create(options.output_dir.join("trades.csv"))?);
+    report.header()?;
 
     cmd::for_each_batch(&options.orderbook_file, |history, batch| {
         let auction_elements = history.auction_elements_for_batch(batch)?;
@@ -38,8 +41,10 @@ fn main() -> Result<()> {
         let new_orders = auction_elements
             .iter()
             .filter(|order| batch == order.valid.from)
-            .filter(|order| order.price.numerator != 0 && order.price.denominator != 0);
+            .filter(|order| order.price.numerator != 0 && order.price.denominator > MIN_AMOUNT)
+            .filter(|order| order.balance > U256::from(MIN_AMOUNT));
         for order in new_orders {
+            let settlement = settlement.as_ref();
             let pricegraph = Pricegraph::new(
                 auction_elements
                     .iter()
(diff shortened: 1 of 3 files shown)


e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,276 @@
...
+    order: &Element,
+    meta: &OrderMetadata,
+) -> TradeResult {
+    let settlement = match settlement {
+        Some(value) => value,
+        None => return TradeResult::NoSolution,
+    };
+
+    match meta.fill_ratio() {
+        Some(trade_ratio) if trade_ratio > FULLY_FILLED_THREASHOLD => TradeResult::FullyMatched,
```
e00E, line 124:
```suggestion
        Some(trade_ratio) if trade_ratio > FULLY_FILLED_THRESHOLD => TradeResult::FullyMatched,
```

e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,276 @@
...
+use core::{
+    contracts::stablex_contract::batch_exchange::event_data::Trade, history::Settlement,
+    models::BatchId,
+};
+use e2e::cmd;
+use pricegraph::{Element, Pricegraph, TokenPair, U256};
+use std::{fs::File, io::Write, path::PathBuf};
+use structopt::StructOpt;
+
+const FULLY_FILLED_THREASHOLD: f64 = 0.95;
```
e00E, line 11:
```suggestion
const FULLY_FILLED_THRESHOLD: f64 = 0.95;
```

e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,276 @@
...
+        OrderMetadata {
+            effective_sell_amount,
+            limit_price,
+            estimated_limit_price,
+            settled_xrate,
+            trade,
+        }
+    }
+
+    fn is_resonable_order(&self) -> bool {
```
e00E, line 99: Add doc comment explaining what "reasonable" means here. Looking at the other code I think this refers to that the order should reasonably be matched based on price.

e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// The events registry file store containing past exchange events.
+    #[structopt(long, env = "ORDERBOOK_FILE", parse(from_os_str))]
+    orderbook_file: PathBuf,
+
+    /// The output directory for the computed results.
+    #[structopt(long, env = "OUTPUT_DIR", default_value = "target", parse(from_os_str))]
+    output_dir: PathBuf,
+}
+
+fn main() -> Result<()> {
```
e00E, line 31: Maybe document what this binary does in a readme or as some form of help string or module comment.

e00E (approved):

e2e/src/bin/historic_trades.rs
nlordell: :see_no_evil:

e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,304 @@
...
+    let settlement = match settlement {
+        Some(value) => value,
+        None => return TradeResult::NoSolution,
+    };
+
+    match meta.fill_ratio() {
+        Some(trade_ratio) if trade_ratio > FULLY_FILLED_THRESHOLD => TradeResult::FullyMatched,
+        Some(_) => TradeResult::PartiallyMatched,
+        None => match find_settled_exchange_rate(settlement, order.pair) {
+            Some(xrate) if xrate > meta.limit_price => TradeResult::SkippedMatchableOrder,
```
fleupold, line 85: Do we have to adjust for fees here?

e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,304 @@
...
+) -> TradeResult {
+    let settlement = match settlement {
+        Some(value) => value,
+        None => return TradeResult::NoSolution,
+    };
+
+    match meta.fill_ratio() {
+        Some(trade_ratio) if trade_ratio > FULLY_FILLED_THRESHOLD => TradeResult::FullyMatched,
+        Some(_) => TradeResult::PartiallyMatched,
+        None => match find_settled_exchange_rate(settlement, order.pair) {
```
fleupold, line 84: Can we not use meta.settled_xrate here?

e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,304 @@
...
+            .as_ref()
+            .map(|trade| trade.executed_sell_amount as f64 / self.effective_sell_amount)
+    }
+}
+
+fn find_trade(settlement: &Settlement, order: &Element) -> Option<Trade> {
+    settlement
+        .trades
+        .iter()
+        .find(|trade| (trade.owner, trade.order_id) == (order.user, order.id))
```
fleupold, line 155: Very reasonable assumption that an order is only traded once per settlement, just want to mention this is not enforced by the smart contract. Theoretically you could first clear an order to 10% and then to 90% in the same batch.

e2e/src/bin/historic_trades.rs
```diff
@@ -0,0 +1,304 @@
...
+        .token_ids_for_price
+        .iter()
+        .position(|&id| id == pair.buy)?;
+    let sell_token_index = settlement
+        .solution
+        .token_ids_for_price
+        .iter()
+        .position(|&id| id == pair.sell)?;
+
+    // NOTE: This is unintuitive, but the echange rate is computed as the sell
```
fleupold, line 171: Maybe we can also argue with the "unit coherence" argument.

Prices p(x) are denoted in OWL "per" token (OWL/t_x)
The exchange rate we are looking for is expressed as t_buy per t_sell (t_buy/t_sell).

Therefore p(sell)/p(buy) = (OWL/t_sell) / (OWL/t_buy) = (OWL/t_sell) * (t_buy/OWL) = t_buy/t_sell

fleupold (approved):
Very nice script. This can be a great metric to improve going forward!

e2e/src/bin/historic_trades.rs
nlordell: Good point, exchange prices are before fees.

e2e/src/bin/historic_trades.rs
nlordell: Yes, that was a refactoring booboo.

e2e/src/bin/historic_trades.rs
nlordell: Ahh, good point, we should collapse trades then.

nlordell:
## Updates

So, after incorporating @fleupold's suggestion for correcting settled exchange rates to account for fees (which it needs to, since the limit price must be respected **after** fees are applied, but the price vector are prices **before** fees), then the results change to:

```
Processed 87872 orders: 93.03% correct, 2.04% failed, 4.93% skipped.
```

Since only "reasonable" orders can produce failures, this means that the incorrectly estimated trades that became considered "skipped" come from this condition:
```
Some(xrate) if xrate > meta.limit_price => TradeResult::SkippedMatchableOrder,
```

This leads me to believe that many most of the previous incorrect estimates were just within 1 fee factor of the settled prices. I can imagine that this might have something to do with the rounding buffer **which is currently not being applied** to the `Pricegraph` for this metric, so that the solver considers the limit price to be non-overlapping but the `Pricegraph` doesn't.

Either way, down to 2% failure :tada: 

I think that if we apply the rounding buffer some of these trades that are considered "skipped" might become successful estimates and fall into the `UnreasonableOrderNotMatched` case.

PR #1222, opened 2020-07-30
Implement get_eth_price

e00E:
So that it will be used by the min avg fee calculation and in a future
PR by the gas price cap calculation.

### Test Plan
CI and and ran locally to double check price makes sense:
```
>>> 316238461452908429312 / 1e18
316.23846145290844
```

### DIFF
--- a/core/src/macros.rs
+++ b/core/src/macros.rs
@@ -38,6 +38,6 @@ macro_rules! immediate {
 #[cfg(test)]
 macro_rules! nonzero {
     ($expression:expr) => {
-        NonZeroU128::new($expression).unwrap()
+        std::num::NonZeroU128::new($expression).unwrap()
     };
 }
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -17,7 +17,7 @@ use crate::{
     models::{Order, TokenId, TokenInfo},
     orderbook::StableXOrderBookReading,
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use average_price_source::AveragePriceSource;
 use futures::future::{BoxFuture, FutureExt as _};
 use log::warn;
@@ -51,7 +51,7 @@ pub trait PriceEstimating {
     /// same format as the above method, so the amount of OWL in atoms to
     /// purchase 1.0 ETH (or 1e18 wei).
     #[allow(clippy::needless_lifetimes)]
-    fn get_eth_price<'a>(&'a self) -> BoxFuture<'a, Option<u128>>;
+    fn get_eth_price<'a>(&'a self) -> BoxFuture<'a, Option<NonZeroU128>>;
 }
 
 pub struct PriceOracle {
@@ -115,6 +115,16 @@ impl PriceOracle {
             }
         }
     }
+
+    async fn eth_token_id(&self) -> Result<TokenId> {
+        for token_id in self.token_info_fetcher.all_ids().await? {
+            let info = self.token_info_fetcher.get_token_info(token_id).await?;
+            if info.symbol() == "ETH" {
+                return Ok(token_id);
+            }
+        }
+        Err(anyhow!("no token with ETH symbol found"))
+    }
 }
 
 impl PriceEstimating for PriceOracle {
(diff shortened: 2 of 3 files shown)


core/src/price_estimation.rs
```diff
@@ -51,7 +51,7 @@ pub trait PriceEstimating {
     /// same format as the above method, so the amount of OWL in atoms to
     /// purchase 1.0 ETH (or 10e18 wei).
     #[allow(clippy::needless_lifetimes)]
-    fn get_eth_price<'a>(&'a self) -> BoxFuture<'a, Option<u128>>;
+    fn get_eth_price<'a>(&'a self) -> BoxFuture<'a, Option<NonZeroU128>>;
```
nlordell, line 54: :+1:

core/src/price_estimation.rs
```diff
@@ -156,9 +166,13 @@ impl PriceEstimating for PriceOracle {
         .boxed()
     }
 
-    fn get_eth_price(&self) -> BoxFuture<'_, Option<u128>> {
-        // TODO: Implement ETH price estimation.
-        immediate!(None)
+    fn get_eth_price(&self) -> BoxFuture<'_, Option<NonZeroU128>> {
+        async move {
+            let token_id = self.eth_token_id().await.ok()?;
```
e00E, line 171: We need to determine the eth token id before running the solver do termine min_avg_fee, which involves fetching fetching all token infos into the cache.  Thus the previous argument that we shouldn't populate the token info cache on startup because it would introduce delay, doesn't hold anymore.
Imo we should in the constructor call cache all and also directly store the eth_token_id in PriceOracle so that we only need to call `eth_token_id` once.

nlordell (approved):
Looks great! Beat me to it.

nlordell:
> and in a future PR by the gas price cap calculation.

I think we should take @fleupold 's suggestion and refactor the `MinAverageFeeComputing` component to a `EconomicViabilityComputing` component that can both compute the min average fee as well as the total solution gas price. This way the same component is shared between the price finding and solution submitting components.

e00E:
Agree. I'll do that next.

core/src/price_estimation.rs
```diff
@@ -115,6 +115,16 @@ impl PriceOracle {
             }
         }
     }
+
+    async fn eth_token_id(&self) -> Result<TokenId> {
+        for token_id in self.token_info_fetcher.all_ids().await? {
+            let info = self.token_info_fetcher.get_token_info(token_id).await?;
+            if info.symbol() == "ETH" {
```
fleupold, line 122: Should we allow "WETH" too here? 
Also just a note to ourselves that this is one of the places that will need to be adjusted if we deploy on the xDAI chain where the native gas token is DAI.

In the Ganache tests this falls back to the hardcoded prices since we don't have an ETH token there I assume?

core/src/price_estimation.rs
fleupold: I would really like to make the cache_all use a single BatchTransport RPC call for this to be just one roundtrip instead of 80 at the moment. I started this some time last week, but got blocked by suboptimal support for it in ethcontracts (https://github.com/gnosis/ethcontract-rs/issues/355).

We could for now also hardcode the ID for WETH since it's always 1.

core/src/macros.rs
```diff
@@ -38,6 +38,6 @@ macro_rules! immediate {
 #[cfg(test)]
 macro_rules! nonzero {
     ($expression:expr) => {
-        NonZeroU128::new($expression).unwrap()
+        std::num::NonZeroU128::new($expression).unwrap()
```
fleupold, line 41: 👍

fleupold (approved):

core/src/price_estimation.rs
e00E: In token_info.rs we have an override so that WETH is turned into ETH.

fleupold: > we have an override so that WETH is turned into ETH.

Right, maybe it's still a bit safer to not rely on this here?

core/src/price_estimation.rs
e00E: It's <num of tokens> roundtrips but it does have parallelization even without batched transport.
I think I do like hardcoding it (and making the hardcoded value configurable if needed). Will change it to that.

PR #1224, opened 2020-07-30
Combine gas price cap and min avg fee computation

e00E:
In the next PR the file will be moved and stablex_driver will make use
of it to calculate the max gas price.

### Test Plan
CI, adapted old unit tests and added new.

### DIFF
--- a/core/src/price_finding/min_avg_fee.rs
+++ b/core/src/price_finding/min_avg_fee.rs
@@ -1,64 +1,94 @@
 //! Module implementing minimum average fee computation based on reference token
 //! price estimates.
 
-use crate::{gas_station::GasPriceEstimating, price_estimation::PriceEstimating};
+use crate::{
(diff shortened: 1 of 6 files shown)


core/src/price_finding/min_avg_fee.rs
```diff
@@ -1,64 +1,94 @@
 //! Module implementing minimum average fee computation based on reference token
```
nlordell, line 1: nit: Can we move the module to, say, `core/src/economic_viability.rs`?

nlordell (approved):
Looks great! Just one nit with the module name.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2Fl0HTYUmU67pLWv1a8%2Fgiphy.gif&token=very_nice" data-gitmeme-token="very_nice"></a>

PR #1225, opened 2020-07-30
Move economic viability trait and make use of gas price cap

e00E:
Second part of #1224 : move the file and remove the duplicated max gas price calculation.

### Test Plan
CI


core/src/driver/stablex_driver.rs
```diff
@@ -254,11 +244,21 @@ mod tests {
         }
     }
 
+    /// Always returns 1 max gas price cap.
+    fn default_economic_viability() -> Arc<dyn EconomicViabilityComputing + Sync> {
```
nlordell, line 248: I'm not a huge fan of this method, since it makes the unit tests a little harder to follow (you have to check the implementation of the method to know what value is being set) and would prefer if that were inlined in the unit test.

Also, if the goal is to simplify construction of this, you could always use `FixedEconomicViabilityComputer` instead of `MockEconomicViabilityComputing`. Alternatively, maybe the `max_gas_price` value could be passed in as a parameter to the method.

nlordell (approved):
Looks good, just a nit with the unit test `default_economic_viability` method.

core/src/driver/stablex_driver.rs
e00E: Good point about using Fixed directly. Will change it to that.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FZBNLpSsFjHPeo%2Fgiphy-downsized-medium.gif&token=like_a_boss" data-gitmeme-token="like_a_boss"></a>

PR #1226, opened 2020-07-30
Make solver input/output files directory agnostic

fleupold:
In order to make the driver run under the current binary setup (which is rooted under `/` but invoken from `/app`) and the mac compatible dockerized setup where we `cargo build` inside the container (rooted under /app/dex-services), we need to have the driver take working directory agnostic paths for the input and output directories.

This PR changes the paths from relative to absolute one.

Note the path for where we find the solver is still hardcoded to be under /app/batchauctions (or open-solver).

### Test Plan
Being able to run the solver from within the docker container (next PR). For now don't expect any difference, once merge check that AWS is still showing result files.


core/src/price_finding/optimization_price_finder.rs
```diff
@@ -260,8 +261,9 @@ impl PriceFinding for OptimisationPriceFinder {
             // We are solving the batch before the current one
             let batch_id = (now.timestamp() / 300) - 1;
             let date = now.format("%Y-%m-%d");
+            let pwd = env::current_dir()?;
```
nlordell, line 264: naming-nit: I would prefer `wd` or even better `working_dir`. I think `pwd` stands for "Print Working Directory".

nlordell (approved):
Love it!!!!! :tada:

e00E (approved):

PR #1228, opened 2020-07-31
Remove unuse dependencies

e00E:
Went through our dependencies and checked which features we don't need.
Not every change results in actual removal of dependency because we can
still have other dependencies depending on the same package.

### Test Plan
CI


fleupold (approved):
Would be nice if we could get a lint warning for having unused dependencies/features.

PR #1229, opened 2020-07-31
Parallelize batch processing in E2E scripts

nlordell:
This PR uses the `rayon` library to parallelize processing each batch to speed things up. This takes advantage of the fact that the auction state and `pricegraph` can be computed "on-the fly" from an immutable event collection in order to take full advantage of all CPU cores :rocket: 

Additionally, the nasty threading bits and dealing with synchronization has been abstracted behind a `Reporting` trait to make things easier and reduce boiler plate.

:warning: There is one disadvantage in that the output is **no longer in order**. I don't think that is a huge issue, since the data can be ordered when it is being processed and analyzed.

### Test Plan

Before `233.85s` :turtle: 
```
$ time -p target/release/historic_prices --orderbook-file target/orderbook.mainnet.bin
54678 / 54678 [==============================================] 100.00 % 230.39/s
Processed 13743 token prices: bad estimates 45.19%, average error 27643.61%.
real 237.53
user 233.85
sys 2.78
```

After `53.27s` :rocket: 
```
$ time -p target/release/historic_prices --orderbook-file target/orderbook.mainnet.bin
54678 / 54678 [=============================================] 100.00 % 1033.13/s
Processed 13743 token prices: bad estimates 45.22%, average error 27938.76%.
real 53.27
user 397.93
sys 2.05
```

A 4.5x increase on an 8-core machine. Notice how the `user` time is almost double. We are losing a lot of time dealing with thread synchronization, and might be able to speed it up even a bit more.

### DIFF
--- a/e2e/src/bin/historic_trades.rs
+++ b/e2e/src/bin/historic_trades.rs
@@ -34,7 +34,7 @@ fn main() -> Result<()> {
     let mut report = Report::new(File::create(options.output_dir.join("trades.csv"))?);
     report.header()?;
 
-    cmd::for_each_batch(&options.orderbook_file, |history, batch| {
+    cmd::for_each_batch_sync(&options.orderbook_file, |history, batch| {
         let auction_elements = history.auction_elements_for_batch(batch)?;
         let settlement = history.settlement_for_batch(batch);
 
--- a/e2e/src/cmd.rs
+++ b/e2e/src/cmd.rs
@@ -1,12 +1,14 @@
 //! Module containing command line helpers for e2e scripts.
 
-use anyhow::Result;
+use anyhow::{Error, Result};
 use core::{history::ExchangeHistory, models::BatchId};
+use crossbeam::thread;
 use pbr::ProgressBar;
-use std::{io, path::Path};
+use rayon::prelude::*;
+use std::{io, path::Path, sync::mpsc};
 
-/// Runs a closure for each batch for the specified event history.
-pub fn for_each_batch(
+/// DEPRECATED: Runs a closure for each batch for the specified event history.
+pub fn for_each_batch_sync(
     orderbook_file: impl AsRef<Path>,
     mut f: impl FnMut(&ExchangeHistory, BatchId) -> Result<()>,
 ) -> Result<()> {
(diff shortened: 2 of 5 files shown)


e2e/src/bin/historic_trades.rs
```diff
@@ -34,7 +34,7 @@ fn main() -> Result<()> {
     let mut report = Report::new(File::create(options.output_dir.join("trades.csv"))?);
     report.header()?;
 
-    cmd::for_each_batch(&options.orderbook_file, |history, batch| {
+    cmd::for_each_batch_sync(&options.orderbook_file, |history, batch| {
```
nlordell, line 37: I will migrate this script in a follow up PR.

e2e/src/cmd.rs
```diff
@@ -22,3 +24,101 @@ pub fn for_each_batch(
...
+                .try_for_each_with(sampler.clone(), move |sampler, batch| {
+                    handler(&sampler, &history, batch)?;
+                    sampler.batch_complete()
+                })
+            {
+                sampler.error(err);
+            }
+        });
+
+        while let Ok(result) = samples_rx.recv() {
```
e00E, line 60: This should probably be in a `scope.spawn` too because the intention is to have this run in parallel.

e00E (approved):

e2e/src/cmd.rs
nlordell: So, the idea here was to have the main thread update the progress bar iterator and record the samples while a background thread processes the parallel . I chose this for a couple of reasons:
1. Not require another thread for this, we can just use the main thread and then join once we finish collecting
2. Not require `report` to also be `Send`

e2e/src/cmd.rs
```diff
@@ -22,3 +24,101 @@ pub fn for_each_batch(
...
+
+/// A trait for reporting samples collected from each batch run.
+pub trait Reporting {
+    /// The sample type.
+    type Sample: Send;
+    /// The summary type.
+    type Summary;
+
+    /// Record a sample.
+    fn sample(&mut self, sample: Self::Sample) -> Result<()>;
```
fleupold, line 120: ubernanonit: _to sample_ for me is to select a record from a set of candidates not to actually so maybe record_sample would be a better verb here?

fleupold (approved):
Very nice abstraction and impressive speedup

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F26uf9QPzzlKPvQG5O%2Fgiphy.gif&token=warpspeed" data-gitmeme-token="warpspeed"></a>

e2e/src/cmd.rs
nlordell: Makes more sense that way. To "sample" is more about reading and less about recording as you pointed it.

PR #1236, opened 2020-08-03
Create infallible price source with real token data and price sources

e00E:
And turn warn logs into errors because we want to be informed when
this fails.
The rounding buffer is still not used by any route but the internal orderbook with rounding buffer is now correctly created.

### Test Plan
Tested that the price estimator still works and saw in logs that external price sources are queried.

### DIFF
--- a/price-estimator/src/infallible_price_source.rs
+++ b/price-estimator/src/infallible_price_source.rs
@@ -2,29 +2,26 @@ use anyhow::Result;
 use core::{
     models::TokenId,
     price_estimation::{average_price_source, price_source::PriceSource},
-    token_info::TokenBaseInfo,
+    token_info::{TokenBaseInfo, TokenInfoFetching},
 };
 use pricegraph::Pricegraph;
-use std::{collections::HashMap, num::NonZeroU128};
+use std::{collections::HashMap, num::NonZeroU128, sync::Arc};
 use tokio::sync::{RwLock, RwLockReadGuard};
 
 /// Roughly like `PriceSource` but is updated externally and cannot fail.
 #[derive(Debug, Default)]
-pub struct InfalliblePriceSource {
-    token_infos: HashMap<TokenId, TokenBaseInfo>,
+pub struct PriceCache {
+    tokens: HashMap<TokenId, TokenBaseInfo>,
     prices: HashMap<TokenId, NonZeroU128>,
 }
 
-impl InfalliblePriceSource {
-    fn new(token_infos: HashMap<TokenId, TokenBaseInfo>) -> Self {
-        Self {
-            token_infos,
-            prices: HashMap::new(),
-        }
+impl PriceCache {
+    fn update_prices(&mut self, prices: &HashMap<TokenId, NonZeroU128>) {
+        self.prices.extend(prices.iter());
     }
 
-    fn update(&mut self, prices: &HashMap<TokenId, NonZeroU128>) {
-        self.prices.extend(prices.iter());
+    pub fn update_tokens(&mut self, tokens: HashMap<TokenId, TokenBaseInfo>) {
+        self.tokens.extend(tokens.into_iter());
     }
 
     /// Tries to use the current price from the first source, if that fails one base unit of the
@@ -33,7 +30,7 @@ impl InfalliblePriceSource {
     pub fn price(&self, token_id: TokenId) -> NonZeroU128 {
         match self.prices.get(&token_id) {
             Some(price) => *price,
-            None => match self.token_infos.get(&token_id) {
+            None => match self.tokens.get(&token_id) {
                 Some(token_info) => token_info.base_unit_in_atoms(),
                 None => self
                     .prices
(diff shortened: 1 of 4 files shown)


core/src/token_info/cached.rs
```diff
@@ -37,14 +37,20 @@ impl TokenInfoCache {
...
+    pub async fn cache(&self) -> HashMap<TokenId, TokenBaseInfo> {
+        self.cache.read().await.clone()
+    }
+
     /// Attempt to retrieve and cache all token info that is not already cached.
     /// Fails if `all_ids` fails. Does not fail if individual token infos fail.
-    pub async fn cache_all(&self, number_of_parallel_requests: usize) -> Result<()> {
-        stream::iter(self.all_ids().await.context("failed to get all ids")?)
+    /// On success returns the result of the internal `all_ids` call.
+    pub async fn cache_all(&self, number_of_parallel_requests: usize) -> Result<Vec<TokenId>> {
```
e00E, line 47: As you can see below, we need the full cache and all_ids so I think it's a good idea to reuse this data when we already have it for faster startup and less stress test and on the node.

price-estimator/src/main.rs
```diff
@@ -105,8 +116,17 @@ fn main() {
...
-    let orderbook = Arc::new(Orderbook::new(Box::new(orderbook), Default::default()));
+
+    let external_price_sources = core::price_estimation::external_price_sources(
+        &http_factory,
+        token_info.clone(),
+        options.price_source_update_interval,
+    )
+    .expect("failed to create external price sources");
+    let infallible_price_source =
+        UpdatingInfalliblePriceSource::new(all_tokens, token_infos, external_price_sources);
```
fleupold, line 127: Why does the infallible price source need concrete `TokenInfos` and cannot use a `TokenInfoFetching` trait? In prod we might run for a week straight. Is it right that tokens that are added might unnecessarily fall back to OWL estimates in that case (which might also be hard to debug).

core/src/token_info/cached.rs
fleupold: I'm not sure I understand the suggestion. I'd not be  a fan of making `cache` part of the `TokenInfoFetching` trait sine that would be leaking behavior of one specific implementation to the general abstraction. 

Which data do you want to reuse - the call to `num_tokens`?

e00E: I mean a function that would return `HashMap<TokenId, TokenBaseInfo>` for all token ids, instead of just a single TokenBaseInfo. This function could have a default implementation in the trait and be overriden in the cache.
We reuse both all_token_ids and the full map.

price-estimator/src/main.rs
e00E: Good point. I'll change it.

core/src/token_info/cached.rs
fleupold: Good idea, this way we can also use batched calls in the onchain implementation once ethcontracts has better support for it.
I think there are some use cases (e.g. getting info for all tokens in the current orderbook), where we don't want all but only a subset of tokens.
How about making `get_token_info` plural and have it take a vector and return a vec or hash map? This way both a subset as well as all_ids can be fetched.

price-estimator/src/infallible_price_source.rs
```diff
@@ -2,29 +2,26 @@ use anyhow::Result;
...
 use pricegraph::Pricegraph;
-use std::{collections::HashMap, num::NonZeroU128};
+use std::{collections::HashMap, num::NonZeroU128, sync::Arc};
 use tokio::sync::{RwLock, RwLockReadGuard};
 
 /// Roughly like `PriceSource` but is updated externally and cannot fail.
 #[derive(Debug, Default)]
 pub struct InfalliblePriceSource {
-    token_infos: HashMap<TokenId, TokenBaseInfo>,
+    tokens: HashMap<TokenId, TokenBaseInfo>,
```
fleupold, line 14: no need to change, but I think a somewhat non-obvious design aspect about this component is that it allows immediate access to this information whereas the other PriceSources have an async interface (arguably the interface makes this "obvious" but the name doesn't). 
Therefore, maybe calling this component `InfalliblePrice**Cache**` and the other component `InfalliblePriceCacheUpdater` could be slightly more expressive.

price-estimator/src/orderbook.rs
```diff
@@ -98,8 +96,13 @@ impl Orderbook {
 
     /// Update the infallible price source with the averaged prices of the external price sources
     /// and the pricegraph prices.
-    async fn update_infallible_price_source(&self, pricegraph: &Pricegraph) -> Result<()> {
-        self.infallible_price_source.update(pricegraph).await
+    async fn update_infallible_price_source(&self, pricegraph: &Pricegraph) {
```
fleupold, line 99: Would this method make more sense on the UpdatingInfalliblePriceSource or do you foresee a reason why one would like to update one or the other? Also, should we join the two futures?

fleupold (approved):

price-estimator/src/infallible_price_source.rs
nlordell: I also agree the name makes it seem like a `PriceSource` implementation that never fails. Maybe it should just be called `PriceCache` (and `PriceCacheUpdater`)?

price-estimator/src/orderbook.rs
nlordell: Maybe this could be `UpdatingInfalliblePriceSource::update(&self, pricegraph: &Pricegraph)`? That way `update_tokens` and `update_prices` could remain private (unless of course, as @fleupold pointed out they will eventually be called separately).

nlordell (approved):

price-estimator/src/orderbook.rs
e00E: I was thinking we might update this in different intervals and they perform different actions to which we might react differently like the different error being logged. I don't think it's important either way but I like this slightly more. (Added the join).

PR #1237, opened 2020-08-03
Upgrade ethcontract

e00E:
The new ethcontract version is not yet released but I wanted to see whether there would be in any problems in this repo. So I upgraded to master. Once the new ethcontract version has been released we can change this PR to use the released version.

### Test Plan
CI


pricegraph/Cargo.toml
```diff
@@ -11,7 +11,7 @@ harness = false
 [dependencies]
 arbitrary = { version = "0.4", optional = true, features = ["derive"] }
 petgraph = "0.5"
-primitive-types = "0.6.2"
+primitive-types = "0.7"
```
e00E, line 14: Looks like this isn't being upgraded by dependabot.

Cargo.toml
```diff
@@ -14,8 +14,3 @@ default-members = [
     "price-estimator",
     "pricegraph",
 ]
-
-[patch.crates-io]
-# Needed to make coverage work on MacOS (https://github.com/servo/core-foundation-rs/pull/357)
-core-foundation = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
-core-foundation-sys = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
```
e00E, line 21: Removed this because I got a warning when building
> ... was not used in the crate graph.

This probably means the dependency is no longer used so overriding it doesn't make sense.

pricegraph/Cargo.toml
nlordell: I think this is because it was explicitly downgraded to be compatible with `core.`

core/Cargo.toml
```diff
@@ -12,8 +12,8 @@ blocking = "0.5.0"
 byteorder = "1.3.4"
 chrono = { version = "0.4.13", default-features = false  }
 crossbeam-utils = "0.7"
-ethcontract = "0.7.2"
-futures = { version = "0.3.5", features = ["compat"] }
+ethcontract = { git = "https://github.com/gnosis/ethcontract-rs.git", default-features = false }
+futures = { version = "0.3.5" }
```
nlordell, line 16: nit:
```suggestion
futures = "0.3.5"
```

core/src/contracts/stablex_contract.rs
```diff
@@ -219,7 +219,7 @@ impl StableXContract for StableXContractImpl {
                 previous_page_user_offset,
                 page_size,
             );
-            builder.block = block_number;
+            builder.block = block_number.map(BlockId::Number);
```
nlordell, line 222: I think you can just `.into()` right?

core/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -194,7 +196,7 @@ impl UpdatingOrderbook {
                     .block_timestamp(meta.block_hash)
                     .await?;
                 context.orderbook.handle_event_data(
-                    data,
+                    EventStatus::Added(data),
```
nlordell, line 199: I suspect that we don't need match on `EventStatus::{Added, Removed}` anymore. Maybe we can remove it?

core/src/contracts/stablex_contract.rs
```diff
@@ -12,10 +12,10 @@ use crate::{
 };
```
nlordell, line 12: At the top of this file there is a `#![allow(...)]` that should be removed.

e2e/src/common.rs
```diff
@@ -33,60 +28,6 @@ pub trait FutureWaitExt: Future + Sized {
 
 impl<F> FutureWaitExt for F where F: Future {}
 
-pub trait FutureBuilderExt: Sized {
-    type Future: Future;
```
nlordell, line 37: We can remove the `type Future: Future` and used `BoxFuture` instead. I'm not sure I understand why this was removed altogether.

core/src/orderbook/streamed/updating_orderbook.rs
e00E: The way we use the past events, yes we could remove this. On the other I hand since we already have this code it might make sense to keep this flexibility.

nlordell: Personally, I think the price of maintaining dead code is usually not worth it. We can always roll back these changes if we need it again (since the original implementation with `EventData::{Added, Removed}` will be in version control.

core/src/contracts/stablex_contract.rs
e00E: BlockId does implement From<BlockNumber> but I still can't `into`, not sure why.

e2e/src/common.rs
e00E: To get this to work I had to change for example

```rust
impl<T, R> FutureBuilderExt for MethodBuilder<T, R>
where
    T: Transport + 'static + Send + Sync,
    <T as Transport>::Out: Send,
    R: Detokenizable + 'static + Send,
{
    type Future = BoxFuture<'static, Result<TransactionResult, MethodError>>;

    fn into_future(self) -> Self::Future {
        self.send().boxed()
    }
}
```

It felt like that wasn't worth it just so we can call `wait` directly instead of `send().wait`.

nlordell: I suggest using `.local_boxed()` instead so you don't need the `Send` bounds, which shouldn't be needed here in any case.

PR #1241, opened 2020-08-03
Single fetch per token

### LINKED ISSUE #1155: GetTokenInfo fetched twice in parallel

nlordell:
Fixes #1155

This PR adds code to track pending token info requests for the cache. The new caching procedure will first check the cache to see if the token info was already retrieved and then create a future and put it into a "pending" hash map so that other simultaneous requests for the same token info block on the same future instead of fetching it again.

This PR is a draft as it does add some additional complexity around the token info caching, so if the general sentiment is that it isn't worth it, then we can abandon this and close the issue for it. 

### Test Plan

Added new unit test. I plan on adding an additional test for the error paths for concurrent requests if the general consensus is to keep this change.

### DIFF
--- a/core/src/token_info/cached.rs
+++ b/core/src/token_info/cached.rs
@@ -284,6 +284,8 @@ mod tests {
     #[test]
     fn token_infos_fetched_once() {
         let mut inner = MockTokenInfoFetching::new();
+        // NOTE: Use `return_once` to ensure this test panics if there is
+        // more than one request.
         inner.expect_get_token_info().return_once(|_| {
             immediate!(Ok(TokenBaseInfo {
                 alias: "FOO".to_string(),
--- a/core/src/token_info/cached.rs
+++ b/core/src/token_info/cached.rs
@@ -1,27 +1,35 @@
 use super::{TokenBaseInfo, TokenId, TokenInfoFetching};
 
-use anyhow::{Context as _, Result};
-use async_std::sync::RwLock;
+use anyhow::{anyhow, Context as _, Result};
+use async_std::sync::{Mutex, RwLock};
 use futures::{
-    future::{BoxFuture, FutureExt},
+    future::{BoxFuture, FutureExt, Shared},
     stream::{self, StreamExt as _},
 };
 use std::collections::HashMap;
 use std::sync::Arc;
 
-/**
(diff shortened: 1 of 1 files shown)


core/src/token_info/cached.rs
```diff
@@ -1,27 +1,35 @@
...
 use std::sync::Arc;
 
-/**
- * Implementation of TokenInfoFetching that stores previously fetched information in an in-memory cache for fast retrieval.
- * TokenIds will always be fetched from the inner layer, as new tokens could be added at any time.
- */
+/// Implementation of TokenInfoFetching that stores previously fetched information in an in-memory cache for fast retrieval.
+/// TokenIds will always be fetched from the inner layer, as new tokens could be added at any time.
 pub struct TokenInfoCache {
-    cache: RwLock<HashMap<TokenId, TokenBaseInfo>>,
```
e00E, line 17: I wonder if you could store something like MaybeDone<Shared<BoxFuture<TokenBaseInfo>>>, so there is only one hashmap. Then when trying to get, return result if ready or wait on shared future if pending.

e00E (commented):
With https://github.com/gnosis/dex-services/pull/1245 the only time this would help is on startup when the external price sources query tokens in parallel. I find the added complexity not worth the benefit.

core/src/token_info/cached.rs
nlordell: Yes we can. I was thinking about doing that but then decided not to until I get a feel for whether or not we want to include the change.

fleupold (commented):
I'm also not sure if this added complexity is worth it. I think if we convert `get_token_info` into a plural that uses batched RPC calls, we will already have much fewer duplicate calls.

nlordell:
Closing because the added complexity is not worth the benefit. Also closing the issue.

PR #1245, opened 2020-08-04
Do not keep retrying to fetch token infos if call reverts

### LINKED ISSUE #1242: Don't keep retrying to fetch token info if there was a method error
There are a large number of tokens where we fail to retrieve the ERC20 info. On failure we do not cache anything so every time we update the a price source we will attempt to retrieve the info and fail. This is especially noticeable with the large number of "fake" tokens that got added today.
We should check whether the error is the result of not being able to reach the ethereum node, in which case we can retry. Or the result of an evm error. In that case it is unlikely to ever succeed so we should mark this token as failed and not attempt to retrieve it again.

e00E:
There are many tokens who do not implement the ERC20 standard and for
which fetching the token base info fails. In these cases it does not
make sense keep retrying. Instead we cache the error and return it back.

Fixes #1242 

### Test Plan
New unit test, manually tested that calling cache all twice does not fetch errors again on second try.

### DIFF
--- a/core/src/token_info/cached.rs
+++ b/core/src/token_info/cached.rs
@@ -1,7 +1,8 @@
 use super::{TokenBaseInfo, TokenId, TokenInfoFetching};
 
-use anyhow::{Context as _, Result};
+use anyhow::{anyhow, Context as _, Error, Result};
 use async_std::sync::RwLock;
+use ethcontract::errors::{ExecutionError, MethodError};
 use futures::{
     future::{BoxFuture, FutureExt},
     stream::{self, StreamExt as _},
@@ -14,8 +15,16 @@ use std::sync::Arc;
  * TokenIds will always be fetched from the inner layer, as new tokens could be added at any time.
  */
 pub struct TokenInfoCache {
-    cache: RwLock<HashMap<TokenId, TokenBaseInfo>>,
+    cache: RwLock<HashMap<TokenId, CacheEntry>>,
     inner: Arc<dyn TokenInfoFetching>,
+}
+
+#[derive(Debug)]
+enum CacheEntry {
+    TokenBaseInfo(TokenBaseInfo),
+    /// For contract calls that revert. In this case we are unlikely to ever be able to get the
+    /// token info so it does not make sense to keep retrying.
+    UnretryableError(String),
 }
 
 impl TokenInfoCache {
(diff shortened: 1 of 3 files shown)


core/src/token_info/cached.rs
```diff
@@ -60,14 +74,31 @@ impl TokenInfoCache {
...
+            match info {
+                Ok(info) => {
+                    self.cache
+                        .write()
+                        .await
+                        .insert(id, CacheEntry::TokenBaseInfo(info.clone()));
+                    Ok(info)
+                }
+                Err(err) if is_revert(&err) => {
+                    self.cache
```
nlordell, line 94: nit: Can we `log::debug!("{:?}")` the error here so we print the stack trace and context. Mostly because we demote the error into a `String` and lose this information.

core/src/token_info/cached.rs
```diff
@@ -60,14 +74,31 @@ impl TokenInfoCache {
...
+                        .write()
+                        .await
+                        .insert(id, CacheEntry::UnretryableError(format!("{}", err)));
```
nlordell, line 97: ubernit: `err.to_string()`.

core/src/token_info/cached.rs
```diff
@@ -33,7 +42,12 @@ impl TokenInfoCache {
     ) -> Self {
```
nlordell, line 42: nit: Can we have `cache: impl IntoIterator<Item = (TokenId, TokenBaseInfo)>`?

core/src/token_info/cached.rs
```diff
@@ -60,14 +74,31 @@ impl TokenInfoCache {
 impl TokenInfoFetching for TokenInfoCache {
     fn get_token_info<'a>(&'a self, id: TokenId) -> BoxFuture<'a, Result<TokenBaseInfo>> {
         async move {
-            if let Some(info) = self.cache.read().await.get(&id).cloned() {
-                return Ok(info);
+            match self.cache.read().await.get(&id) {
+                Some(CacheEntry::TokenBaseInfo(info)) => return Ok(info.clone()),
+                Some(CacheEntry::UnretryableError(reason)) => {
+                    return Err(anyhow!("cached error: {}", reason));
```
nlordell, line 80: Maybe it makes sense to use `Context` here?
```suggestion
                    return Err(
                        anyhow!(reason)
                            .context("cached error retrieving token info"),
                    );
```

I'm also fine with not using it, just thought it was a good application for it.

nlordell (approved):
Looks good. I think for #1241 we can extend `CacheEntry` to also have a `Pending` variant with a `Shared<...>` future.

fleupold (approved):
Very nice change! Theoretically, this could also be implemented as an orthogonal `TokenInfoFetching` implementation that "filters" requests for which it has seen unretriable errors before. This would not couple it to the cache. But it's also fine the way it is.

e00E:
> Very nice change! Theoretically, this could also be implemented as an orthogonal `TokenInfoFetching` implementation that "filters" requests for which it has seen unretriable errors before. This would not couple it to the cache. But it's also fine the way it is.

Since this is very cache related and it's overall less work to do it here I prefer this way.

PR #1246, opened 2020-08-04
Use slightly higher min_avg_fee compared to max_gas_price

e00E:
Currently we calculate the min avg fee and max gas price so that if the
solution earns exactly min avg fee it can be submitted at exactly fast
gas price while not losing money.
This is a problem when the gas price or eth price moves slightly between
the start of solution creation and submitting. To be more robust this
commit adds 10% to the min average fee.

Context for this is that I saw a lot of " Solution does not generate enough fees for the transaction to execute quickly enough" in the logs with the open solver and I think this is the reason. Also added more debug logs so I can check the logs in more detail in the future.

### Test Plan
Adjusted unit tests and added some debug prints .

### DIFF
--- a/core/src/economic_viability.rs
+++ b/core/src/economic_viability.rs
@@ -5,7 +5,7 @@ use crate::{
     gas_station::GasPriceEstimating, models::solution::EconomicViabilityInfo,
     price_estimation::PriceEstimating,
 };
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context as _, Result};
 use ethcontract::U256;
 use futures::future::{BoxFuture, FutureExt as _};
 use std::sync::Arc;
@@ -30,6 +30,10 @@ pub struct EconomicViabilityComputer {
     price_oracle: Arc<dyn PriceEstimating + Send + Sync>,
     gas_station: Arc<dyn GasPriceEstimating + Send + Sync>,
     subsidy_factor: f64,
+    /// We multiply the min average fee by this amount to ensure that if a solution has this minimum
+    /// amount it will still be end up economically viable even when the gas or eth price moves
+    /// slightly between solution computation and submission.
+    min_avg_fee_factor: f64,
 }
 
 impl EconomicViabilityComputer {
@@ -37,11 +41,13 @@ impl EconomicViabilityComputer {
         price_oracle: Arc<dyn PriceEstimating + Send + Sync>,
         gas_station: Arc<dyn GasPriceEstimating + Send + Sync>,
         subsidy_factor: f64,
+        min_avg_fee_factor: f64,
     ) -> Self {
         EconomicViabilityComputer {
             price_oracle,
             gas_station,
             subsidy_factor,
+            min_avg_fee_factor,
         }
     }
 
@@ -54,7 +60,12 @@ impl EconomicViabilityComputer {
     }
 
     async fn gas_price(&self) -> Result<f64> {
-        let fast = self.gas_station.estimate_gas_price().await?.fast;
+        let fast = self
+            .gas_station
+            .estimate_gas_price()
+            .await
+            .context("failed to get gas price")?
+            .fast;
         Ok(pricegraph::num::u256_to_f64(fast))
     }
 }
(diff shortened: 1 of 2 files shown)


nlordell (commented):
Should we make this value configurable right away? Otherwise looks good.

e00E:
> Should we make this value configurable right away? Otherwise looks good.

Yes makes sense. Implemented this.

core/src/economic_viability.rs
```diff
@@ -88,6 +98,10 @@ impl EconomicViabilityComputing for EconomicViabilityComputer {
             let eth_price = self.eth_price_in_owl().await?;
             let cap = gas_price_cap(eth_price, earned_fee, num_trades);
             let subsidized = cap * self.subsidy_factor;
+            log::debug!(
+                "computer max gas price to be {} subsidized to {} based on earned fee {} num trades {} eth price {}",
```
fleupold, line 102:
```suggestion
                "computed max gas price to be {} subsidized to {} based on earned fee {} num trades {} eth price {}",
```

fleupold (approved):
Good observation.

Is the assumption that the solver can actually find solutions that generate a higher average fee in these cases? Otherwise I'd argue it's better to have some solution (even if we decide it's not viable) than getting the trivial solution (so that in case the gas price doesn't move against us we still have something to submit).

Another option would be to "store" the fast gas price given to the solver for solution submission (given that it's already quite high). Even if it moves slightly against us chances are we can still submit at the previous fast price.

e00E:
> Is the assumption that the solver can actually find solutions that generate a higher average fee in these cases? Otherwise I'd argue it's better to have some solution (even if we decide it's not viable) than getting the trivial solution (so that in case the gas price doesn't move against us we still have something to submit).

We have a large subsidy factor at the moment so this 10% is like changing the subsidy factor from 10 to 9. Whether the solver could find a solution at factor 9 but not 10 is minor. It seems like the open solver generally gets very close to the min avg fee which motivated this change. Not sure yet if that is a bug in the solver.

> Another option would be to "store" the fast gas price given to the solver for solution submission (given that it's already quite high). Even if it moves slightly against us chances are we can still submit at the previous fast price.

That would be intentionally using less accurate information so that we can't "see" the issue. (We would maybe see it in terms of timed out transactions later).

fleupold:
> That would be intentionally using less accurate information so that we can't "see" the issue.

I'd say w'd be trading off complexity (now we are using different viability values at different parts of the process which is another thing to keep in mind when trying to understand e.g. why solutions were not generated or submitted) for slightly more recent information (the gas estimate itself is also based on pretty outdated historic information), but it's a good enough argument.

core/src/economic_viability.rs
```diff
@@ -54,7 +60,12 @@ impl EconomicViabilityComputer {
     }
 
     async fn gas_price(&self) -> Result<f64> {
-        let fast = self.gas_station.estimate_gas_price().await?.fast;
+        let fast = self
+            .gas_station
+            .estimate_gas_price()
+            .await
+            .context("failed to get gas price")?
```
nlordell, line 67: :+1:

core/src/economic_viability.rs
```diff
@@ -218,12 +232,18 @@ mod tests {
                 ..Default::default()
             }))
         });
-        let economic_viability =
-            EconomicViabilityComputer::new(Arc::new(price_oracle), Arc::new(gas_station), 10.0);
+        let subsidy = 10.0f64;
+        let min_avg_fee_factor = 1.0f64;
```
nlordell, line 236: nit: I don't like using a `min_avg_fee_factor` of `1.0`, since theoretically this test would have passed before the modification to the `EconomicViabilityComputing::min_avg_fee` modification.

nlordell (approved):
Just a minor nit about the unit test.

core/src/economic_viability.rs
e00E: Thanks. This was meant to be 1.1.

e00E:
> I'd say w'd be trading off complexity (now we are using different viability values at different parts of the process which is another thing to keep in mind when trying to understand e.g. why solutions were not generated or submitted) for slightly more recent information (the gas estimate itself is also based on pretty outdated historic information), but it's a good enough argument.

Since this is already reviewed and I would like to have the log messages with more info now I will merge.
Your point is valid too so I'm not against changing it in that direction either.

PR #1249, opened 2020-08-05
Add get_token_infos to TokenInfoFetching

e00E:
This is more convenient when code needs multiple infos and can be
specialized in trait implementors to be more performant for example by
using a batched transport.

I also had the idea to be more generic and take `ids: &'a [TokenId]` by `impl IntoIterator<Item = &TokenId>`. It turns out that if a trait has generic methods it isn't "object safe" anymore and we wouldnt' be able to use it as `dyn &Trait` anymore. There are workarounds like requiring `Self: Sized` for the methods and then `impl Trait for Box<dyn Trait> {...}`. I didn't feel that was worth it, just wanted to note for the future.

### Test Plan
CI, new unit test.

### DIFF
--- a/core/src/token_info.rs
+++ b/core/src/token_info.rs
@@ -1,8 +1,6 @@
 use anyhow::Result;
-use futures::future::BoxFuture;
+use futures::future::{BoxFuture, FutureExt as _};
 use lazy_static::lazy_static;
-#[cfg(test)]
-use mockall::automock;
 use std::{collections::HashMap, num::NonZeroU128};
 
 use crate::models::TokenId;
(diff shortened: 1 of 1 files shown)


core/src/token_info.rs
```diff
@@ -1,23 +1,74 @@
...
+        .boxed()
+    }
+
     /// Returns a vector with all the token IDs available
     fn all_ids<'a>(&'a self) -> BoxFuture<'a, Result<Vec<TokenId>>>;
 }
+
+// mockall workaround https://github.com/asomers/mockall/issues/134
+#[cfg(test)]
+mod mock {
```
fleupold, line 42: I know we had this discussion before, but just re-checking that we feel more comfortable maintaining this workaround than making the trait take TokenId (which is an u16) by value.

core/src/token_info.rs
```diff
@@ -102,4 +154,33 @@ mod tests {
         assert_eq!(TokenBaseInfo::new("", 1).base_unit_in_atoms().get(), 10);
         assert_eq!(TokenBaseInfo::new("", 2).base_unit_in_atoms().get(), 100);
     }
+
+    #[test]
+    fn default_get_token_infos_forwards_calls_and_ignores_errors() {
+        // Not using mockall because we want to test the default impl.
+        struct TokenInfo {};
```
fleupold, line 161: nit: could use the hardcoded `TokenData` here, but not depending on a concrete implementation is fine too.

fleupold (approved):

core/src/token_info.rs
e00E: It's `&[TokenId]` vs `Vec<TokenId>`.

nlordell (approved):

PR #1251, opened 2020-08-05
Specialize get_token_infos for TokenInfoCache

e00E:
Several reasons:
1. It is more efficient.
2. It prevents the warning log message "failed to get token info" from
   showing up repeatedly for token's whose errors we have cached in
   TokenInfoCache. This currently happens because the default
   implementation of get_token_infos logs the error for each individual
   token and doesn't know about TokenInfoCache's unretryable errors.
3. It is also a step toward being able to use batch transport. But we
   would have to change the signature of the function for that to work
   (see the comment in the code).

### Test Plan
New unit test. CI.


core/src/token_info/cached.rs
```diff
@@ -104,11 +109,49 @@ impl TokenInfoFetching for TokenInfoCache {
...
+            }
+
+            let cache = self.cache.read().await;
+            let result = ids
+                .iter()
+                .filter_map(|id| {
+                    let entry = cache.get(id)?;
+                    let result = cache_entry_to_result(entry);
+                    let info = result.ok()?;
+                    Some((*id, info))
```
fleupold, line 133: nit: why not call `self.get_token_info` here?

fleupold (approved):
Do you envision the revised `get_token_infos` signature being a `Result<HashMap<TokenId, Result<TokenBaseInfo>>>` or only a `HashMap<TokenId, Result<TokenBaseInfo>>`?

core/src/token_info/cached.rs
e00E: Functionally it's the same but this way we avoid locking the mutex for every token.

e00E:
> Do you envision the revised `get_token_infos` signature being a `Result<HashMap<TokenId, Result<TokenBaseInfo>>>` or only a `HashMap<TokenId, Result<TokenBaseInfo>>`?

I'm not sure. If I think about the batch transport then Result<HashMap<Result>> makes more sense because we can have individual token failures or failure of the whole transport.

nlordell (approved):

PR #1253, opened 2020-08-05
Fix how min average fee is passed to solver

e00E:
Solver expects total amount of paid fee while we were passing *earned*
fees, which is half of the total. This lead to many solutions failing
with benign error: max gas price lower than fast price.

See https://github.com/gnosis/dex-open-solver/blob/593d2022568b669a4f6fe65a333e05e8b3f01971/src/match.py#L49.
I found this through checking a recent solution and seeing that the fee in the solution.json was double than what we printed in the logs and that the min-avg-fee-per-order was only upheld assuming the doubled fee.

### Test Plan
CI

### DIFF
--- a/core/src/economic_viability.rs
+++ b/core/src/economic_viability.rs
@@ -17,6 +17,8 @@ const GAS_PER_TRADE: f64 = 120_000.0;
 #[cfg_attr(test, mockall::automock)]
 pub trait EconomicViabilityComputing: Send + Sync + 'static {
     /// Used by the solver so that it only considers solution that are economically viable.
+    /// This is the minimum average amount of earned fees per order. The total amount of paid fees
+    /// is twice this because half of the fee is burnt.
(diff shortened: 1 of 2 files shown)


e00E:
I'm assuming merge when green will still only merge when there has been one reviewer.

nlordell (approved):
:rofl:

fleupold (commented):
Nice catch! I also stumbled across this behaviour today when building a dune widget for min tradable order size but didn't notice this requires a code change in the driver (https://gnosisinc.slack.com/archives/CC3C5SMB5/p1596628176215500?thread_ts=1596613420.201000&cid=CC3C5SMB5). 

I think there are three different things with slightly different names:

1) Trading fee, 
2) protocol fee
3) solver reward

where 2) == 3) and 1) == 2) + 3)

Maybe we can refactor the methods/variables to reflect what is the protocol fee and what is the trading fee or even call it solver reward (since this is what the driver is mostly interested in).

PR #1256, opened 2020-08-05
Reorganize implementation of Bellman-Ford algorithm

fedgiac:
The Bellman-Ford implementation is restructured in preparation for the restriction in the number of hops allowed in the search.

The implementation on master returns some technical data that is, after some extra steps, used to either find shortest paths or negative cycles. With these changes, the function returns data structures that can be directly used to retrieve shortest paths and negative cycles. This is the role of the types `ShortestPathGraph` and (a modified version of the current) `NegativeCycle`.

Technically, the predecessor information that is currently output of Bellman-Ford is moved to the `PredecessorStore` struct as part of `ShortestPathGraph`. This type will be extended in another PR to use the same algorithm with a limited number of hops.

The global behavior of the code after this PR does not change. A new test (`shortest_path_graph_finds_connected_nodes`) is added to check that a new functionality coming from some previous isolated code works as expected.


pricegraph/src/graph/bellman_ford.rs
```diff
@@ -2,40 +2,170 @@
...
+type Distance<G> = Vec<<G as Data>::EdgeWeight>;
+
+/// Stores the information needed to manage the predecessor list.
+/// For each node index, this type contains its predecessor node and
+/// distance in the graph from a source node.
+struct PredecessorStore<G: GraphBase + Data>(Vec<Option<G::NodeId>>, Distance<G>);
+
+impl<G: GraphBase + Data> PredecessorStore<G> {
+    fn distance(&self, node_index: usize) -> &G::EdgeWeight {
+        let PredecessorStore(_, distance) = self;
```
e00E, line 21: Seeing that you always destructure `self` back into a tuple and give the fields names, it makes more sense to have `PredecessorStore` be a struct with named fields in the first place like

```rust
struct PredecessorStore<G: GraphBase + Data> {
    predecessor: Vec<Option<G::NodeId>>,
    distance: Distance<G>,
}
```
Or is there another reason for the tuple struct?
You can also access the fields without naming them first btw:

```rust
&self.1[node_index]
```

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -2,40 +2,170 @@
...
+        let mut node_indices: Vec<_> = predecessor
+            .iter()
+            .enumerate()
+            .filter_map(|(i, &pre)| pre.map(|_| i))
+            .collect();
+        // if the source was not connected to anything, no node would have it as its
+        // predecessor
+        node_indices.push(self.graph.to_index(self.source));
+        node_indices.sort();
+        node_indices.dedup();
```
e00E, line 93: I think the purpose of the sort and dedup is to make sure that `to_index(source)` doesn't appear twice in case it was already a predecessor. No other index can appear twice because we have only added elements from the enumeration sequence.
Thus it is faster to do
```rust
if !node_indicices.contains(&self.source) {
    node_indices.push(self.graph.to_index(self.source));
}
```

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -2,40 +2,170 @@
...
+}
+
+impl<G> ShortestPathGraph<G>
+where
+    G: IntoNodeIdentifiers + IntoEdges + NodeIndexable,
+    G::EdgeWeight: FloatMeasure,
+{
+    /// Initializes a shortest path graph that will be later built with the
+    /// Bellman-Ford algorithm.
+    fn init(g: G, source: G::NodeId) -> ShortestPathGraph<G> {
```
e00E, line 108: Could call it `new` which is more common in Rust but if init has a specific meaning different from new then that is fine too.

In this function signature and body you can replace `ShortestPathGraph<G>` and `ShortestPathGraph` with `Self` for less repetition.

pricegraph/src/graph/path.rs
```diff
@@ -1,46 +1,84 @@
...
+#[derive(Debug)]
+/// An ordered collection of nodes that form a cycle of negative weight.
+/// The first node of the cycle coincides with the last.
+pub struct NegativeCycle<N>(pub Vec<N>);
+
+impl<N: PartialEq> NegativeCycle<N> {
+    /// Returns the negative cycle changing its starting and terminating
+    /// node to be the given node. If the given node is not part of the
+    /// cycle, it returns an error containing the original cycle.
+    pub fn with_starting_node(self, start: N) -> Result<NegativeCycle<N>, NegativeCycle<N>> {
```
e00E, line 24: I think this can be written more efficiently as:

```rust
    pub fn with_starting_node(mut self, start: N) -> Result<NegativeCycle<N>, NegativeCycle<N>> {
        match self.0.iter().position(|i| *i == start) {
            None => Err(self),
            Some(pos) if pos == 0 => Ok(self),
            Some(pos) => {
                self.0.pop();
                self.0.rotate_right(pos);
                self.0.push(start);
                Ok(self)
            }
        }
    }
```

pricegraph/src/graph/path.rs
```diff
@@ -1,46 +1,84 @@
...
+                cycle.pop(); // the last element of self.0 is identical to the first
+                cycle.append(&mut cycle_end);
+                cycle.push(start);
+                return Ok(NegativeCycle(cycle));
+            } else {
+                cycle_end.push(node);
+            }
+        }
+        Err(NegativeCycle(cycle_end))
+    }
```
e00E, line 41: I think this can be written more efficiently:
```suggestion
    pub fn with_starting_node(mut self, start: N) -> Result<NegativeCycle<N>, NegativeCycle<N>> {
        match self.0.iter().position(|i| *i == start) {
            None => Err(self),
            Some(pos) if pos == 0 => Ok(self),
            Some(pos) => {
                self.0.pop();
                self.0.rotate_right(pos);
                debug_assert!(self.0[0] == start);
                self.0.push(start);
                Ok(self)
            }
        }
    }
```

fedgiac: Nice!

pricegraph/src/graph/bellman_ford.rs
fedgiac: True, I'll change it to a struct.
I wrote it this way because I plan to make `PredecessorStore` into an `enum`, so with this destructuring the diff would be smaller when the change happen. On the other hand, it's better to use a `struct` with named fields here to keep this PR cleaner.

pricegraph/src/graph/bellman_ford.rs
fedgiac: Good point. Now that I think more about it, `node_indices` will never contain `self.source` because if the source gets a predecessor then its updated distance must be smaller than the initial zero, which means that there is a negative cycle.
So I'll just add `self.source` to `node_indices` and add a test for good measure.

pricegraph/src/graph/bellman_ford.rs
e00E: An enum can be struct-like with named fields too btw, if you prefer that.

pricegraph/src/graph/path.rs
fleupold: Or maybe use the rotate method on slice (https://doc.rust-lang.org/std/primitive.slice.html#method.rotate_left)?

pricegraph/src/orderbook.rs
```diff
@@ -156,45 +166,37 @@ impl Orderbook {
...
-                    continue;
+        while let Err(cycle) = bellman_ford::search(&self.projection, quote) {
+            let cycle_not_in_path = match cycle.with_starting_node(quote) {
+                Err(cycle) => cycle,
+                Ok(cycle) => {
+                    let base_index_search = cycle.0.iter().position(|node| *node == base);
+                    match base_index_search {
+                        None => cycle,
+                        Some(base_index) => {
+                            let (ask, bid) = (&cycle.0[0..base_index + 1], &cycle.0[base_index..]);
```
fleupold, line 177: maybe the split logic should also be on the NegativeCycle struct?

pricegraph/src/orderbook.rs
```diff
@@ -156,45 +166,37 @@ impl Orderbook {
...
-
-            return Some(Ring { ask, bid });
+            // NOTE: Skip negative cycles that are not along the
+            // specified market.
+            self.fill_path(&cycle_not_in_path.0).unwrap_or_else(|| {
+                panic!(
+                    "failed to fill path along detected negative cycle {}",
+                    format_path(&cycle_not_in_path.0),
+                )
+            });
```
fleupold, line 199: Can we move this into the Err() arm of the match? It took me a while to understand how we would get here and I think it would clearer to treat it right next to the condition.

It could be something like:

```rust
match cycle.with_starting_node(quote).map(|cycle| cycle.0.iter().position(|node| *node == base)) {
  Err(cycle_not_in_path) => ...
  Some(base_index) => ...
}
```

fleupold (commented):
This review is based on a now outdated commit, will double check recent changes.

pricegraph/src/orderbook.rs
```diff
@@ -249,26 +251,26 @@ impl Orderbook {
...
             Some(path) => path,
             None => return Ok(None),
         };
 
-        let flow = self.find_path_flow(&path).unwrap_or_else(|| {
+        let flow = self.find_path_flow(&path.0).unwrap_or_else(|| {
             panic!(
                 "failed to fill detected shortest path {}",
-                format_path(&path),
+                format_path(&path.0),
```
fleupold, line 263: Agree with @e00E that unpacking here is not very readable in particular here. I think a named struct would be better.

fleupold (commented):
This is a very nice and advanced refactor. 

Unfortunately the PR became very large which made it hard to follow all the movements of the code. No need to break this up any more now (I realize you already broke up your original change into 2 PRs), but next time it would be good to do even smaller refactors (e.g. split NegativeCycle struct out before introducing the ShortestPath logic).

This way we can give a more thorough review and have less risks for unnoticed changes.

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -14,20 +14,20 @@ type Distance<G> = Vec<<G as Data>::EdgeWeight>;
 /// Stores the information needed to manage the predecessor list.
 /// For each node index, this type contains its predecessor node and
 /// distance in the graph from a source node.
-struct PredecessorStore<G: GraphBase + Data>(Vec<Option<G::NodeId>>, Distance<G>);
+struct PredecessorStore<G: GraphBase + Data> {
+    predecessor: Vec<Option<G::NodeId>>,
```
fleupold, line 18: nit: plural predecessor**s**

fleupold (approved):
Looks good to me. Still some open suggestions to move more logic into the new structs but also fine as is.

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -2,40 +2,170 @@
...
+            .filter_map(|(i, &pre)| pre.map(|_| i))
+            .collect();
+        // if the source was not connected to anything, no node would have it as its
+        // predecessor
+        node_indices.push(self.graph.to_index(self.source));
+        node_indices.sort();
+        node_indices.dedup();
+        node_indices
+            .into_iter()
+            .map(|i| self.graph.from_index(i))
```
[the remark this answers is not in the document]
fedgiac: I wanted to sort `node_indices` but `G::NodeId` is not necessarily `Ord` while the node index is `usize`. Anyway, this step is not needed anymore.

pricegraph/src/graph/bellman_ford.rs
```diff
@@ -44,9 +174,9 @@ where
                 let i = edge.source();
                 let j = edge.target();
                 let w = *edge.weight();
-                if distance[ix(i)] + w < distance[ix(j)] {
-                    distance[ix(j)] = distance[ix(i)] + w;
-                    predecessor[ix(j)] = Some(i);
+                if *shortest_path_graph.distance(i) + w < *shortest_path_graph.distance(j) {
```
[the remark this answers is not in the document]
fedgiac: As I see it, the function `bellman_ford::search` is the real constructor of a shortest path graph. The exports of this file are structured in a way to only expose shortest path graphs that are valid. (So, for example, `ShortestPathGraph::new()` (previously called `init`) is not `pub`.)
So, I think the current construction makes sense from the point of view of using it in other files. It is true that this file could be changed in a way that breaks this assumption however, I'm not sure how to evaluate the risk.

pricegraph/src/orderbook.rs
fedgiac: The suggestion doesn't work as it is because `cycle` is consumed in the process and i need it inside the `Some` branch, but I'll think of something.
If you think it's more readable I can probably make it look more like it was before my changes, I remember I was quite confused by how it was written before, but maybe my changes didn't improve the situation.

pricegraph/src/orderbook.rs
fedgiac: The thing here is that `format_path` should take a path as input instead of a vector, but I didn't want to add even more changes to the code so I simply extracted the vector from the path and used the old function. Will change this!

e00E (approved):

pricegraph/src/graph/bellman_ford.rs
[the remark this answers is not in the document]
fleupold: Is there an advantage of not moving it inside?

pricegraph/src/orderbook.rs
```diff
@@ -156,45 +166,37 @@ impl Orderbook {
...
-                            format_path(&path),
-                        )
-                    });
-
-                    continue;
+        while let Err(cycle) = bellman_ford::search(&self.projection, quote) {
+            let cycle_not_in_path = match cycle.with_starting_node(quote) {
+                Err(cycle) => cycle,
+                Ok(cycle) => {
+                    let base_index_search = cycle.0.iter().position(|node| *node == base);
```
fleupold, line 173: ubernit: Maybe we can implement the Iterable trait on cycle directly in order to avoid the .0 here.

pricegraph/src/orderbook.rs
fleupold: Would this become possible if cycle implemented `split` directly and would give you an `Option<(before, after)>` instead of `Option<base_index>`? Then you shouldn't need to access the original cycle in the some block

fedgiac: I did something similar, you can check if you find the result more readable.

pricegraph/src/graph/bellman_ford.rs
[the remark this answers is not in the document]
fedgiac: Addressed this comment with commit `59f54dd8fceb277831627df1fa44a8e25c4c7e02`. Do you prefer these changes?

pricegraph/src/graph/path.rs
```diff
@@ -34,6 +52,21 @@ impl<N: PartialEq> NegativeCycle<N> {
             }
         }
     }
+
+    /// Returns an iterator over the nodes in the cycle in the order
+    /// they appear.
+    pub fn iter(&self) -> Iter<N> {
```
fleupold, line 58: I was thinking about the [iter trait](https://doc.rust-lang.org/std/iter/index.html), but that works too.

pricegraph/src/orderbook.rs
```diff
@@ -166,37 +166,36 @@ impl Orderbook {
...
-
-                            return Some(Ring { ask, bid });
-                        }
-                    }
+        while let Err(cycle) = ShortestPathGraph::new(&self.projection, quote) {
+            let paths_base_quote = match cycle.with_starting_node(quote) {
+                Err(cycle) => Err(cycle),
+                Ok(cycle) => match cycle.iter().position(|node| *node == base) {
+                    None => Err(cycle),
+                    Some(base_index) => Ok(cycle.split_at(base_index)),
```
fleupold, line 174: split_at could take a node instead of an index to avoid the extra match arm, but that's fine too.

fleupold (approved):
Looks good. Just to make sure it doesn't actually change any logic, would you mind running @nlordell's benchmark to make sure the numbers are similar before merging?

pricegraph/src/graph/path.rs
fedgiac: Implementing `Iterator` for `NegativeCycle` means that the cycle itself can be iterated, so for every call to `.next()` the cycle would become smaller. This seems strange to me, while having something in the cycle that returns an iterator would be more intuitive.

pricegraph/src/orderbook.rs
```diff
@@ -118,22 +121,29 @@ impl Orderbook {
...
-                        .expect("negative cycle not found after being detected");
-
-                    self.fill_path(&path).unwrap_or_else(|| {
-                        panic!(
-                            "failed to fill path along detected negative cycle {}",
-                            format_path(&path),
-                        )
-                    });
-                }
+        let mut remaining_tokens: BTreeSet<_> = self.projection.node_indices().collect();
```
nlordell, line 124: I think this can be implemented with the `Subgraph` abstraction if instead of accepting a `ShortestPathGraph` it accepts a `Vec<NodeIndex>` as it did before and return `shortest_path_graph.connected_nodes()` from the closure.

pricegraph/src/orderbook.rs
```diff
@@ -156,45 +166,32 @@ impl Orderbook {
...
-                // even need to search for the base token.
-                None
-            };
-            let base_index = match base_index_search {
-                Some(value) => value,
-                None => {
+        while let Err(cycle) = ShortestPathGraph::new(&self.projection, quote) {
+            let paths_base_quote = cycle
+                .with_starting_node(quote)
+                .and_then(|cycle| cycle.split_at(base));
```
nlordell, line 172: Very nice! Its like reading English.

pricegraph/src/orderbook.rs
```diff
@@ -106,10 +109,10 @@ impl Orderbook {
...
     pub fn is_overlapping(&self) -> bool {
         // NOTE: We detect negative cycles from each disconnected subgraph.
-        Subgraphs::new(self.projection.node_indices().skip(1))
+        Subgraphs::<&OrderbookGraph>::new(self.projection.node_indices().skip(1))
             .for_each_until(
-                |token| match bellman_ford::search(&self.projection, token) {
-                    Ok((_, predecessor)) => ControlFlow::Continue(predecessor),
+                |token| match ShortestPathGraph::new(&self.projection, token) {
+                    Ok(shortest_path_graph) => ControlFlow::Continue(shortest_path_graph),
                     Err(NegativeCycle(..)) => ControlFlow::Break(true),
```
nlordell, line 116: Something that was there from before that I just noticed now:
```suggestion
                    Err(_) => ControlFlow::Break(true),
```

pricegraph/src/orderbook.rs
```diff
@@ -118,22 +121,29 @@ impl Orderbook {
...
-                }
+        let mut remaining_tokens: BTreeSet<_> = self.projection.node_indices().collect();
+
+        while let Some(&token) = remaining_tokens.iter().next() {
+            remaining_tokens.remove(&token);
+
+            let shortest_path_graph = loop {
+                match ShortestPathGraph::new(&self.projection, token) {
+                    Ok(shortest_path_graph) => break shortest_path_graph,
+                    Err(NegativeCycle(cycle)) => {
```
nlordell, line 132: Since `NegativeCycle: Deref<Target = [N]>`, can't we just do:
```suggestion
                    Err(cycle) => {
```

Without having to destructure the `NegativeCycle`?

pricegraph/src/orderbook.rs
```diff
@@ -548,6 +545,10 @@ mod tests {
         let ReducedOrderbook(orderbook) = orderbook.reduce_overlapping_orders();
         // NOTE: We expect user 1's 2->1 order to be completely filled as well
         // as user 2's 5->3 order and user 4's 6->7 order.
+        // User 3's 7->6 order may be filled or not depending on the order in
+        // which the loop {6, 7} is cleared:
+        // 6->7->6: 100_000 T6 -> 1_000_000 T7 -> 1_000_000 T6, both orders cleared
+        // 7->6->7: 100_000 T7 -> 100_000 T6 -> 1_000_000 T7, only 6->7 cleared
         assert_eq!(orderbook.num_orders(), 6);
```
nlordell, line 552: Maybe we should make it `assert!(orderbook.num_orders() == 5 || orderbook.num_orders() == 6);`?

pricegraph/src/graph/subgraph.rs
```diff
@@ -1,44 +1,39 @@
...
-    pub fn new(nodes: impl Iterator<Item = NodeIndex>) -> Self {
-        Subgraphs(nodes.collect())
-    }
-
-    /// Iterate through each subgraph with the provided closure returning the
-    /// predecessor vector for the current node indicating which nodes are
-    /// connected to it.
-    pub fn for_each(self, mut f: impl FnMut(NodeIndex) -> Vec<Option<NodeIndex>>) {
-        self.for_each_until(|node| <ControlFlow<()>>::Continue(f(node)));
+pub struct Subgraphs<G: GraphBase>(BTreeSet<G::NodeId>);
```
nlordell, line 13: I think this can be kept as is if we return `shortest_path_graph.connected_nodes()` instead of the `ShortestPathGraph` so that the `Subgraphs` does not need to have a lifetime tied to a graph reference.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -0,0 +1,311 @@
...
+                if *shortest_path_graph.distance(i) + w < *shortest_path_graph.distance(j) {
+                    shortest_path_graph.update_distance(j, *shortest_path_graph.distance(i) + w);
+                    shortest_path_graph.update_predecessor(j, Some(i));
+                    did_update = true;
+                }
+            }
+        }
+        if !did_update {
+            break;
+        }
```
nlordell, line 203: I find this implementation nice to read. Maybe we should measure the performance impact of this.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -0,0 +1,311 @@
...
+    G::EdgeWeight: FloatMeasure,
+{
+    /// Creates a representation of all shortest paths from the given source
+    /// to any other node in the graph.
+    ///
+    /// Shortest paths are well defined if and only if the graph does not
+    /// contain any negative weight cycle reachabe from the source. If a
+    /// negative weight cycle is detected, it is returned as an error.
+    pub fn new(g: G, source: G::NodeId) -> Result<Self, NegativeCycle<G::NodeId>> {
+        bellman_ford(g, source)
```
nlordell, line 167: Very nice! We can now completely abstract away the actual search algorithm and instead just provide an interface to "finding the shortest path in a graph" :+1:

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -0,0 +1,311 @@
...
+struct PredecessorStore<G: GraphBase + Data> {
+    predecessors: Vec<Option<G::NodeId>>,
+    distance: Distance<G>,
+}
+
+impl<G: GraphBase + Data> PredecessorStore<G> {
+    fn distance(&self, node_index: usize) -> &G::EdgeWeight {
+        &self.distance[node_index]
+    }
+    fn update_distance(&mut self, node_index: usize, updated_distance: G::EdgeWeight) {
```
nlordell, line 26: ubernit:
```suggestion

    fn update_distance(&mut self, node_index: usize, updated_distance: G::EdgeWeight) {
```

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -0,0 +1,311 @@
...
+        self.distance[node_index] = updated_distance;
+    }
+    fn update_predecessor(&mut self, node_index: usize, updated_predecessor: Option<G::NodeId>) {
```
nlordell, line 29:
```suggestion

    fn update_predecessor(&mut self, node_index: usize, updated_predecessor: Option<G::NodeId>) {
```

nlordell (approved):
Very nice. I think the abstractions turned out great!

One thing in general I'm not a fan of is that `ShortestPathGraph` needs to retain a reference to the original graph, when to me it should just be the "result" data (the information for getting the shortest path). This, however, seems to be related to how the `petgraph` abstraction works where you need the graph in order get node IDs and I don't see a way around it. It worked fine before because we ensured that `node_id == node_index == token_id` but its not necessarily an assumption that we want to carry into the `shortest_path` module. So, I don't think anything needs to change.

The other thing that we discussed is around the `Subgraphs` abstraction. I think we should keep it as before, so that the `reduce_overlapping_orders` method does not need to change so drastically. While it is a pity that `Subgraphs` can't operate on the new `ShortestPathGraph` abstraction because of type lifetime issues, I think repeating the call to `connected_nodes` is better than repeating the loop logic.

One thing I would be curious would be to measure the performance benchmarks to see just how "zero-cost abstraction" Rust is :wink:.

pricegraph/src/graph/subgraph.rs
```diff
@@ -48,11 +43,11 @@ impl Subgraphs {
...
 
 /// An enum for representing control flow when iterating subgraphs.
-pub enum ControlFlow<T> {
+pub enum ControlFlow<N, T> {
     /// Continue the iterating through the subgraphs with the provided
-    /// predecessor vector indicating which nodes are connected to the current
-    /// subgraph.
-    Continue(Vec<Option<NodeIndex>>),
+    /// shortest path graph indicating which nodes are connected to the
+    /// current subgraph.
```
nlordell, line 49: This comment should be updated.

pricegraph/src/orderbook.rs
```diff
@@ -106,10 +108,12 @@ impl Orderbook {
     /// this ring trade.
     pub fn is_overlapping(&self) -> bool {
         // NOTE: We detect negative cycles from each disconnected subgraph.
-        Subgraphs::new(self.projection.node_indices().skip(1))
+        Subgraphs::<NodeIndex>::new(self.projection.node_indices().skip(1))
```
nlordell, line 111: The explicit type may no longer be needed.

pricegraph/src/graph/path.rs
```diff
@@ -1,46 +1,116 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<N: Clone + PartialEq> NegativeCycle<N> {
+    /// Returns the negative cycle changing its starting and terminating
+    /// node to be the given node. If the given node is not part of the
+    /// cycle, it returns an error containing the original cycle.
+    pub fn with_starting_node(mut self, start: N) -> Result<Self, Self> {
```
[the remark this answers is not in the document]
fedgiac: No specific reason, I think it reads well.

fedgiac:
Benchmarking results: no noticeable performance degradation. All results are within a ±2% range of current master.

<details>
<summary>Full results.</summary>

```
test result: ok. 0 passed; 0 failed; 32 ignored; 0 measured; 0 filtered out

Running target/release/deps/pricegraph-ee18f1da024dc9d3
Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 5.3s, enable flat sampling, or reduce sample count to 30.
Pricegraph::read        time:   [4.1214 ms 4.1253 ms 4.1296 ms]                             
                        change: [-2.1392% -1.3673% -0.5981%] (p = 0.00 < 0.05)
  2 (4.00%) high mild
  2 (4.00%) high severe

Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 5.2s, enable flat sampling, or reduce sample count to 30.
                        time:   [3.9292 ms 3.9448 ms 3.9603 ms]
                        change: [-2.7227% -1.8073% -0.4889%] (p = 0.00 < 0.05)
Found 2 outliers among 50 measurements (4.00%)
  2 (4.00%) high severe

                        time:   [58.150 us 58.180 us 58.216 us]
                        change: [-1.5059% -0.9630% -0.4083%] (p = 0.00 < 0.05)
Found 9 outliers among 50 measurements (18.00%)
  1 (2.00%) high mild
  8 (16.00%) high severe
                        time:   [61.570 us 61.753 us 61.895 us]
                        change: [-1.1039% -0.6145% -0.0736%] (p = 0.01 < 0.05)
Found 10 outliers among 50 measurements (20.00%)
  5 (10.00%) low severe
  2 (4.00%) high mild
  3 (6.00%) high severe
                        time:   [105.16 us 105.69 us 106.24 us]
                        change: [-0.7944% -0.1209% +0.5550%] (p = 0.74 > 0.05)
  3 (6.00%) high mild
  1 (2.00%) high severe
                        time:   [194.26 us 194.91 us 195.60 us]
                        change: [-1.6276% -1.0938% -0.5459%] (p = 0.00 < 0.05)
                        time:   [432.04 us 433.18 us 434.53 us]
                        change: [-2.6096% -0.7769% +0.4164%] (p = 0.45 > 0.05)
  3 (6.00%) high mild
  4 (8.00%) high severe

                        time:   [52.061 us 52.212 us 52.398 us]
                        change: [+0.5385% +1.1706% +1.8478%] (p = 0.00 < 0.05)
Found 2 outliers among 50 measurements (4.00%)
  1 (2.00%) high mild
  1 (2.00%) high severe
                        time:   [130.54 us 131.31 us 132.29 us]
                        change: [-3.2159% -1.1228% +0.3879%] (p = 0.29 > 0.05)
  1 (2.00%) high severe
                        time:   [207.78 us 208.59 us 209.47 us]
                        change: [+0.1242% +0.6585% +1.2315%] (p = 0.03 < 0.05)
                        time:   [423.35 us 423.92 us 424.80 us]
                        change: [-1.4281% -0.7668% -0.0428%] (p = 0.03 < 0.05)
Found 9 outliers among 50 measurements (18.00%)
  7 (14.00%) high mild
  2 (4.00%) high severe
                        time:   [459.03 us 460.39 us 462.69 us]
                        change: [-0.0439% +0.4632% +1.0004%] (p = 0.12 > 0.05)
Found 11 outliers among 50 measurements (22.00%)
  2 (4.00%) high mild
  9 (18.00%) high severe

```
</details>

nlordell (approved):

nlordell:
All the changes were either "Change within noise threshold" or "No change in performance detected". Looks like Rust's "zero cost abstraction" promise held up well :smile:

PR #689, opened 2020-04-06
Implement contract deployment in Rust

nlordell:
This PR adds a new `contract` crate with a `deploy` script to deploy exchange contracts to a development network.

This an initial step towards #846 where we need an alternative to `scripts/setup_contracts.sh` which uses Truffle. The next steps in order to no longer require NodeJS are:
- Update `core` and `e2e` to use the new `contracts` crate.
- Update the Travis build file to use `cargo run -p contracts --bin deploy` instead of performing a Truffle migration.
- Use the `@gnosis.pm/dex-contracts` npm package for retrieving the contract artifact instead of building `dex-contracts` with Truffle.
- Update the README with new build instructions.

### Test Plan

Run the deploy script:
```
$ (cd contracts; cargo run -p contracts --bin deploy --features bin)
   Compiling contracts v0.1.0 (/var/home/nlordell/Developer/dex-services/contracts)
    Finished dev [unoptimized + debuginfo] target(s) in 12.14s
     Running `/var/home/nlordell/Developer/dex-services/target/debug/deploy`
[2020-08-05T12:43:07Z INFO  deploy] checking connection to local test node http://localhost:8545
[2020-08-05T12:43:07Z INFO  deploy] deploying library contracts
[2020-08-05T12:43:07Z INFO  deploy] deployed IdToAddressBiMap to 0x65dbc7c034b644401ad2a1f1ed8b284ae41e56bc
[2020-08-05T12:43:07Z INFO  deploy] deployed IterableAppendOnlySet to 0xc27e5a197f7f1c794bcd348f673d80687eff83a6
[2020-08-05T12:43:07Z INFO  deploy] deploying fee token contracts
[2020-08-05T12:43:07Z INFO  deploy] deployed TokenOWL to 0x8f534a8084d9306527ff15c919f7fa691c0ca856
[2020-08-05T12:43:07Z INFO  deploy] deployed TokenOWLProxy to 0x445d64d63f42401c8c7ff846d2a92edd2b31417b
[2020-08-05T12:43:07Z INFO  deploy] deploying exchange and viewer contracts
[2020-08-05T12:43:07Z INFO  deploy] deployed BatchExchange to 0x5355d6d088ddc1a99b29114976a6ca0321d71fa6
[2020-08-05T12:43:07Z INFO  deploy] deployed BatchExchangeViewer to 0x63f7170efe984d40d1a435a917b99c71ad645901
```

Then check that the `BatchExchange` contract address matches in the generated code:
```
[...]

PR #1257, opened 2020-08-05
Use npmjs registry instead of building contracts 

nlordell:
Fixes #689 

This PR removes the dex-contracts submodule and any NodeJS/Truffle requirements. Note that this change makes the `contracts` crate retrieve the Truffle JSON artifact from npm packges. Alternatively, we could vendor the JSON artifacts so they aren't retrieved on build.

### Test Plan

CI!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,6 @@
 target
 !target/debug/driver
 !target/release/driver
-dex-contracts
-!dex-contracts/build
 **/node_modules
 .git
 .vscode
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "dex-contracts"]
-	path = dex-contracts
-	url = https://github.com/gnosis/dex-contracts
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,6 @@ env:
     - OPEN_SOLVER_VERSION=v0.0.11
     - PRIVATE_SOLVER_VERSION=v0.8.2
 
-install:
-  - nvm install 12 && nvm alias default 12 && npm install -g yarn@latest && yarn --version
-  - (cd dex-contracts && yarn --frozen-lockfile && yarn run prepack)
-
 jobs:
   fast_finish: true
   allow_failures:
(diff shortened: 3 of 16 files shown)


e00E (commented):
I do think we should vendor them. I dislike making network requests during the build. This can randomly fail and it is unclear how to fix that as a developer in that case. I've seen with the etherscan contract example in ethcontract that a lot of requests seem to be generated just from my editor compiling the code.

If they are part of the repo we do not have these problems. We can add a script that updates them and either way an update requires some manual intervention (changing the version number).

nlordell:
These are very good points. I think we can make a new `contracts/src/vendor.rs` script that does just that.

nlordell:
One thing to note though, is that with #1258 we shouldn't rebuild often at all and unlike the etherscan example in ethcontract we're using a dependency registry (npmjs and unpkg.io - comparable to cargo) that is capable of handling a large volume of requests and doesn't rate limit you of you don't use an API key.

e00E (approved):
I'm okay with leaving it for now but I'd still want to change it in the future. It's true that we already depend crates.io but it would be nice if that was the only online dependency.
Also crates.io is cached so that I can still compile without internet even after `cargo clean` but this wouldn't be the case for the contracts and as far as I can tell there is no easy way to do this for the contracts (is there?).

nlordell:
> Also crates.io is cached so that I can still compile without internet even after cargo clean but this wouldn't be the case for the contracts and as far as I can tell there is no easy way to do this for the contracts 

:+1: I think you're right and vendoring is the way to go. I'll do it in a followup PR.

> (is there?).

Not at the moment.

fleupold (approved):
> Also crates.io is cached so that I can still compile without internet even after cargo clean but this wouldn't be the case for the contracts and as far as I can tell there is no easy way to do this for the contracts (is there?).

We could create a crate that only contains the contract artifacts, but I guess that's the same as vendoring it directly.

nlordell:
> We could create a crate that only contains the contract artifacts, but I guess that's the same as vendoring it directly.

I think we should vendor it in `contracts` crate directly - the artifacts will be accessible from `contract::BatchExchange::artifact()`.

PR #1261, opened 2020-08-06
Use new rounding buffer in estimate_buy_amount route

e00E:
Finally :)

### Test Plan
Run the price estimator from master, the price estimator from this PR and the price estimator from master with rounding-buffer-factor set to 0.0. Compare results. We see that in the estimate-buy-amount route the buy amount is the lowest for the master price estimator which is lower than the this PR's default price estimator which is lower than this PR's price estimator with no rounding buffer. Example:

```
# No rounding buffer
curl "localhost:8082/api/v1/markets/4-7/estimated-buy-amount/1?atoms=false"
{"baseTokenId":4,"quoteTokenId":7,"buyAmountInBase":"1.010907","sellAmountInQuote":"1"}
# Solver based rounding buffer
curl "localhost:8080/api/v1/markets/4-7/estimated-buy-amount/1?atoms=false"
{"baseTokenId":4,"quoteTokenId":7,"buyAmountInBase":"1.010707","sellAmountInQuote":"1"}
# Old rounding buffer
curl "localhost:8081/api/v1/markets/4-7/estimated-buy-amount/1?atoms=false"
{"baseTokenId":4,"quoteTokenId":7,"buyAmountInBase":"1.009896","sellAmountInQuote":"1"}
```

Also checked that /markets route is still the same.

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -16,26 +16,21 @@ use warp::{
     Filter, Rejection, Reply,
 };
 
+// TODO: There are some routes that still use this because adapting them to work with the new
+// solver rounding buffer is more complicated. These routes will be changed in future PRs.
+const OLD_ROUNDING_BUFFER: f64 = 0.001;
+
 /// Handles all supported requests under a `/api/v1` root path.
 pub fn all(
     orderbook: Arc<Orderbook>,
     token_info: Arc<dyn TokenInfoFetching>,
-    price_rounding_buffer: f64,
 ) -> impl Filter<Extract = impl Reply, Error = Infallible> + Clone {
     warp::path!("api" / "v1" / ..)
         .and(
             markets(orderbook.clone())
-                .or(estimated_buy_amount(
-                    orderbook.clone(),
-                    token_info.clone(),
-                    price_rounding_buffer,
-                ))
-                .or(estimated_amounts_at_price(
-                    orderbook.clone(),
-                    token_info,
-                    price_rounding_buffer,
-                ))
-                .or(estimated_best_ask_price(orderbook, price_rounding_buffer)),
+                .or(estimated_buy_amount(orderbook.clone(), token_info.clone()))
+                .or(estimated_amounts_at_price(orderbook.clone(), token_info))
+                .or(estimated_best_ask_price(orderbook)),
         )
         .recover(handle_rejection)
 }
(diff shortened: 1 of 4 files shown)


price-estimator/src/filter.rs
```diff
@@ -213,17 +203,19 @@ async fn estimate_buy_amount(
...
-
-    let mut buy_amount_in_base = Amount::Atoms(
-        transitive_order
-            .map(|order| apply_rounding_buffer(order.buy, price_rounding_buffer))
-            .unwrap_or_default() as _,
+        .map_err(error::internal_server_rejection)?;
+    // This reduced sell amount is what the solver would see after applying the rounding buffer.
+    let transitive_order = pricegraph.order_for_sell_amount(
+        token_pair,
+        f64::max(sell_amount_in_quote_atoms - rounding_buffer, 0.0),
```
nlordell, line 214: I think this might be backwards:
```suggestion
        f64::max(sell_amount_in_quote_atoms + rounding_buffer, 0.0),
```

Since we want to estimate an order such that when placed, it has the amount specified by the parameter correct? I think if we reduce the sell amount here, we will get "better" prices, which is the opposite of what we want as far as I can tell.

e00E: Tested:
With minus: 1.010709
With plus   : 1.011105
So the user will get more buy amount with plus than with minus, so minus is more conservative which is what we want. The question that the user wants answered is what they should put as the buy amount and still get matched. Internally the solver will see their order as having slightly less sell amount than they have set so the expected buy amount is lower too. This seems inline with the test, or am I switching something up?

nlordell: Oh duh... The price might be better but the total amount will not be. I think you are correct and we need to minus. Sorry for the confusion.

nlordell (approved):

fleupold (approved):
Very nice. I think it would be nice to quantify the impact of this change somehow. Do you have any ideas how we could measure the hopefully notable impact on spreads, while not affecting accuracy much? 
Maybe @nlordell has an idea too?

e00E:
We can measure how selling one non atom amount changes the buy amount for a bunch of token pairs. But the accuracy we will only see in the future, won't we?

nlordell:
> Maybe @nlordell has an idea too?

Sorry! Forgot to answer this. This is pretty hard to measure with just the events since we don't have access to historic price estimates.

I think if we wanted to do something fancy, we would have to provide the computed price estimates for each batch (from the `instance.json` file) and then add them to the estimate in the `historic_trades` script. It will probably be a lot of work, but might be worth it to see how it impacts the estimates.

PR #1270, opened 2020-08-10
Fix order_for_limit_price not returning None

e00E:
According to the doc comment it should return None when there are no
overlapping orders at this price.

### Test Plan
New unit test. CI.


nlordell (approved):
Nice catch!

Issue #1272, opened 2020-08-10
Allow to get price estimates at certain block numbers


fleupold:
To debug e.g. why EFrontier has or has not placed orders, it would be helpful if we can re-create price estimates not only at a certain batchId but even at a certain time/block within this batch Id.

PR #1279, opened 2020-08-10
Refactor query parameter parsing to allow mutually exclusive values

nlordell:
This PR is a precursor to #1272 to enable mutually exclusive query parameters (`?batch=123&blockNumber=456&timestamp=789` for example, where each exactly one or none of these parameters should be present).

### Test Plan

Unit tests, they check query parameter parsing.

### DIFF
--- a/price-estimator/src/models/query.rs
+++ b/price-estimator/src/models/query.rs
@@ -45,7 +45,6 @@ pub enum Generation {
 #[serde(rename_all = "camelCase")]
 struct RawQuery {
     atoms: Option<bool>,
-    unit: Option<Unit>,
     hops: Option<usize>,
     batch_id: Option<BatchId>,
 }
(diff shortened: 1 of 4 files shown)


e00E (approved):

price-estimator/src/models/query.rs
```diff
@@ -0,0 +1,69 @@
...
+pub enum Unit {
+    /// Atoms, smallest unit for a token.
+    Atoms,
+    /// Base units, such that `1.0` is `10.pow(decimals)` atoms.
+    BaseUnits,
+}
+
+/// When to perform a price estimate.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Generation {
```
fleupold, line 33: nit: I'm not a huge fan of this name, but can see how it's hard to find a good one. How about `PointInTime` or `EstimationTime`?

fleupold (approved):
Really like the new enums, just not entirely happy with the Generation name.

price-estimator/src/models/query.rs
nlordell: :rofl: I had no idea what to name this so I literally googled "time synonym" and found it under "a portion of time in history or characterized by particular events or circumstances.". I also am not a huge fan of this name. I think I like `EstimationTime` the most.

PR #1282, opened 2020-08-10
Shared filter context

nlordell:
This PR introduces a shared filter context for price estimation routes. The main motivation behind this is a starting point to add new filters for amounts so that there are less `if query.atoms { ... }` everywhere to do query conversions and move some of that into filters to make the actual handlers a little more `warp` agnostic.

I'm not 100% what the best way to structure this is at the moment, so I will leave this in a Draft PR for now.

### Test Plan

Unit tests.

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -16,26 +16,25 @@ use warp::{
     Filter, Rejection, Reply,
 };
 
+/// Main filter context containing an orderbook and token information.
+#[derive(Clone)]
+pub struct Context {
+    /// Shared access to the current orderbook.
+    pub orderbook: Arc<Orderbook>,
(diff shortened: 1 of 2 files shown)


fleupold:
So far this looks good. For now a shared context object makes a lot of sense (we need to be a bit careful that we don't keep adding random stuff to it that is only needed for a subset of routes which can easily happen with these "toolbox/context" structs).

Let me know if you want to merge this change by itself already.

nlordell:
> Let me know if you want to merge this change by itself already.

I want to think about this a bit more. Currently we have what feels like a lot of duplicated boilerplate code to convert to and from amounts, getting the pricegraph instance, etc. and it would be nice to have it all be more automagical by combining some filters.

PR #1283, opened 2020-08-10
Shared logger in price estimator.

### LINKED ISSUE #1280: [Price Esitmation] Use same logger as in the driver
For alerts it's useful to have the stderr (for ERR log level), stdout (for all other levels), so that we can easily configure our slack alerts to only report these type of statements (rather than any log level that happens to contain the word "error").

We are already doing that in the driver, but not yet in the price estimation service.

nlordell:
Fixes #1280 

Used the shared logging component.

### Test Plan

Run the price estimator and see logs:
```
$ cargo run -p price-estimator -- --orderbook-file target/orderbook-file-mainnet                                                  
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/price-estimator --orderbook-file target/orderbook-file-mainnet`
2020-08-10T16:03:42.612Z INFO [price_estimator] Starting price estimator with runtime options: Options {
...
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2290,7 +2290,6 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "core",
- "env_logger",
  "ethcontract",
  "futures 0.3.5",
  "log 0.4.11",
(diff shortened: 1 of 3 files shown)


e00E (approved):

fleupold:
Great, thanks a lot!

PR #1285, opened 2020-08-11
Allow minimum solution submit time to be set

### LINKED ISSUE #1230: Start calculating early with open-solver, but submit only late
Unfortunately, the running time of the open solver is no longer constant. It varies heavily with the instances. Especially during sales, we see it.

Due to this reason, we have to start the open-solver earlier than in the past, in order to guarantee that a solution is found in time.
But in order to avoid unnecessary submission from the open-solver, which will likely be overwritten by other solvers, it would be great if the open solver would start calculating the solutions early, but only submit the solution if the other solvers don't find a solution within the first 3 mins. It will save us quite some gas costs.

e00E:
This was requested for the open solver.

Fixes #1230.

### Test Plan
CI.

### DIFF
--- a/core/src/driver/scheduler.rs
+++ b/core/src/driver/scheduler.rs
@@ -24,7 +24,11 @@ pub struct AuctionTimingConfiguration {
 
     /// The offset from the start of the batch to cap the solver's execution
     /// time.
-    solver_time_limit: Duration,
+    latest_solution_submit_time: Duration,
+
+    /// The earliest offset from the start of a batch in seconds at which point we should submit the
+    /// solution.
+    earliest_solution_submit_time: Duration,
 }
 
 impl AuctionTimingConfiguration {
@@ -36,13 +40,18 @@ impl AuctionTimingConfiguration {
     /// invariants must hold:
     /// - `target_start_solve_time < solver_time_limit`
     /// - `solver_time_limit < SOLVING_WINDOW`
+    /// - `min_solution_submit_time < SOLVING_WINDOW`
     ///
     /// Where `SOLVING_WINDOW` represents the amount of time within a batch in
     /// which a solution is accepted. There is an amount of time at the end of a
     /// batch where solutions are no longer accepted, this is done to allow
     /// traders time to make decisions after the previous batch has already
     /// finalized.
-    pub fn new(target_start_solve_time: Duration, solver_time_limit: Duration) -> Self {
+    pub fn new(
+        target_start_solve_time: Duration,
+        solver_time_limit: Duration,
+        min_solution_submit_time: Duration,
+    ) -> Self {
         assert!(
             solver_time_limit < SOLVING_WINDOW,
             "The solver time limit must be within the solving window",
@@ -51,17 +60,26 @@ impl AuctionTimingConfiguration {
             target_start_solve_time < solver_time_limit,
             "the target solve start time must be earlier than the solver time limit",
         );
+        assert!(
+            min_solution_submit_time < SOLVING_WINDOW,
+            "The min solution submit time must be within the solving window",
+        );
 
         AuctionTimingConfiguration {
             target_start_solve_time,
-            solver_time_limit,
+            latest_solution_submit_time: solver_time_limit,
+            earliest_solution_submit_time: min_solution_submit_time,
         }
     }
 }
 
 impl Default for AuctionTimingConfiguration {
     fn default() -> Self {
-        AuctionTimingConfiguration::new(Duration::from_secs(30), Duration::from_secs(180))
+        AuctionTimingConfiguration::new(
+            Duration::from_secs(30),
+            Duration::from_secs(180),
+            Duration::from_secs(0),
+        )
     }
 }
 
--- a/core/src/driver/stablex_driver.rs
+++ b/core/src/driver/stablex_driver.rs
@@ -1,17 +1,18 @@
 use crate::{
     economic_viability::EconomicViabilityComputing,
     metrics::StableXMetrics,
-    models::{account_state::AccountState, order::Order, Solution},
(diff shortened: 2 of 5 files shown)


core/src/driver/stablex_driver.rs
```diff
@@ -127,16 +154,26 @@ impl<'a> StableXDriverImpl<'a> {
         };
 
         let submitted = if let Some(objective_value) = verified {
+            if let Ok(duration) = (batch_to_solve.solve_start_time() + min_solution_submit_time)
+                .duration_since(SystemTime::now())
+            {
+                self.sleep.sleep(duration).await;
```
e00E, line 160: I know Nick suggested removing the sleep from SolutionSubmission and instead I've added another sleep here.
I attempted that refactoring but it is more controversial and complex so I want this PR to focus on this change and overall. I am not sure whether that refactoring makes sense and it's not needed now.

nlordell (commented):
The change in itself looks good. I'm just wondering if we should refactor the driver first to split `run` into `solve` and `submit` so that the sleeping can all be done in the scheduler? This way we can get rid of the in `solution_submission` to wait for the batch to start accepting solutions and not add a new sleep here.

e00E:
As I wrote in the code comment above, I want to get this change in first because the refactor is more work and it's not clear to me yet whether it's the correct way to go.

nlordell (approved):

core/src/driver/scheduler.rs
```diff
@@ -36,13 +40,18 @@ impl AuctionTimingConfiguration {
...
     /// Where `SOLVING_WINDOW` represents the amount of time within a batch in
     /// which a solution is accepted. There is an amount of time at the end of a
     /// batch where solutions are no longer accepted, this is done to allow
     /// traders time to make decisions after the previous batch has already
     /// finalized.
-    pub fn new(target_start_solve_time: Duration, solver_time_limit: Duration) -> Self {
+    pub fn new(
+        target_start_solve_time: Duration,
+        solver_time_limit: Duration,
+        min_solution_submit_time: Duration,
```
fleupold, line 53: Conceptually, I'd find it easier to understand if we had an `earliest_solution_submission_time` and a `latest_solution_submission_time`. The `solver_time_limit` then follows from `latest_solution_submission_time`.

core/src/driver/stablex_driver.rs
```diff
@@ -127,16 +154,32 @@ impl<'a> StableXDriverImpl<'a> {
         };
 
         let submitted = if let Some(objective_value) = verified {
+            if min_solution_submit_time > Duration::from_secs(0) {
+                if let Ok(duration) = (batch_to_solve.solve_start_time() + min_solution_submit_time)
```
fleupold, line 158: Is there a reason for the nested if? Could we just have the `if let Ok()` statement?

fleupold (approved):
Looks really nice! I think it would be nicer if `time_limit` and `min_solution_submission_time` could be mapped into the same cognitive concept (e.g. min vs. max submission time). This would make the params we provide more coherent and intuitive.

core/src/driver/stablex_driver.rs
e00E: I added a comment to explain this:

            // In the e2e test batch ids don't always correspond to the correct batch id based on
            // real time (we fake advancement of time) so without the outer if we would be waiting
            // a long time. There will likely be a refactor that lets the scheduler decider when
            // solutions should be submitted which fixes this problem in a better way.

core/src/driver/scheduler.rs
e00E: I agree with this. Renamed it in AuctionTimingConfiguration and in the Options struct. I did leave the env variable as SOLVER_TIME_LIMIT to not cause issues with our configuration. Let me know if you're ok with these changes @fleupold .

fleupold: Yes changes look good. I would however prefer if the env variable could also be renamed (as this is what consumers of our binary need to understand when running it). As for configuration changes, we can create a staging PR that has both values (old and new) set to the same value and then remove the old one once the PR is merged.

For prod, we usually make sure when we deploy that we adjust the configs to have only the "expected" diff to staging (e.g. gurobi for standard solver).

Maybe we can do this in a separate PR and tag it with [ConfigChange] so that whoever does the next release is reminded about this?

core/src/driver/stablex_driver.rs
```diff
@@ -25,7 +26,12 @@ pub enum DriverResult {
 pub trait StableXDriver {
     // mockall needs the lifetimes but clippy warns that they are not needed.
     #[allow(clippy::needless_lifetimes)]
-    fn run<'a>(&'a self, batch_to_solve: u32, time_limit: Duration) -> BoxFuture<'a, DriverResult>;
+    fn run<'a>(
+        &'a self,
+        batch_to_solve: BatchId,
+        time_limit: Duration,
```
fleupold, line 32: ubernit: could be renamed here (and other occurrences) as well.

core/src/driver/scheduler.rs
nlordell: Created a [`config change` label](https://github.com/gnosis/dex-services/labels/config%20change). I would also prefer renaming the environment variable in this case.

fleupold: The issue with github labels is that they don't show up when compiling the changelog for a release (which is the point at which we should remember to update the prod env var as well). That's why I tend to prefer good old ASCI title prefixes as  labels.

e00E:
The staging PR setting the new env variable has been merged so now we can merge this PR.

core/src/driver/scheduler.rs
nlordell: Oh shoot good point! I like the GitHub labels cause they are so colourful (here the dark red really catches the attention).

PR #1288, opened 2020-08-11
Block number and timestamp query parameters

nlordell:
This PR is the first step to #1272 and introduces new query parameters. They are currently not supported by the service and return internal server errors.

### Test Plan

See that the parameter causes a `internal server error` instead of a `invalid url query`.
```
...
$ curl -s 'http://localhost:8080/api/v1/markets/7-1/estimated-buy-amount/1?blockNumber=1337' | jq
{
  "message": "internal server error"
}
```


e00E (approved):

fleupold (approved):
👍

PR #1290, opened 2020-08-11
Introduce HistoricOrderbookReading trait

nlordell:
This PR adds a new `HistoricOrderbookReading` trait for reading the orderbook at arbitrary batches and timestamps. This will be used for #1272 in order to construct historic `Pricegraph`s at specified block numbers and timestamps.

### Test Plan

Adds to new trait without any implementations - so Rustc!

### DIFF
--- a/core/src/orderbook/history.rs
+++ b/core/src/orderbook/history.rs
@@ -1,0 +1,98 @@
+//! Module for historic orderbook reading.
+
+use super::{NoopOrderbook, StableXOrderBookReading};
+use crate::models::{AccountState, Order};
+use anyhow::Result;
+use ethcontract::BlockNumber;
+use futures::future::BoxFuture;
+use std::time::SystemTime;
+
+/// Trait for reading historic orderbook data.
+pub trait HistoricOrderbookReading: StableXOrderBookReading {
+    /// Retrieves the open orderbook at the specified block number.
+    ///
+    /// The open orderbook is defined as the auction state for the current batch
+    /// at the specified block number. This implies that, assuming no futher
+    /// changes to the exchange, this will be the auction state passed to the
+    /// solver to be submitted in the next batch.
+    fn get_auction_data_for_block<'a>(
+        &'a self,
+        block_number: BlockNumber,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
+
+    /// Retrieves the open orderbook at the specified timestamp.
+    fn get_auction_data_for_timestamp<'a>(
+        &'a self,
+        timestamp: SystemTime,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
+}
+
+impl HistoricOrderbookReading for NoopOrderbook {
+    fn get_auction_data_for_block<'a>(
+        &'a self,
+        _: BlockNumber,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>> {
+        immediate!(Ok(Default::default()))
+    }
+
+    fn get_auction_data_for_timestamp<'a>(
+        &'a self,
+        _: SystemTime,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>> {
+        immediate!(Ok(Default::default()))
+    }
+}
+
+#[cfg(test)]
+mod mock {
+    use super::*;
+
+    /// Proxy trait used for mocking a `HistoricOrderbookReading` to work around
+    /// the fact that `mockall` doesn't support trait inheritance.
+    #[mockall::automock]
+    pub trait Proxy {
+        fn get_auction_data<'a>(
+            &'a self,
+            batch_id_to_solve: u32,
+        ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
+        fn initialize<'a>(&'a self) -> BoxFuture<'a, Result<()>>;
+        fn get_auction_data_for_block<'a>(
+            &'a self,
+            block_number: BlockNumber,
+        ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
+        fn get_auction_data_for_timestamp<'a>(
+            &'a self,
+            timestamp: SystemTime,
+        ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
+    }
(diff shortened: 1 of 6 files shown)


e00E (approved):

core/src/orderbook/history.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use super::{NoopOrderbook, StableXOrderBookReading};
+use crate::models::{AccountState, Order};
+use anyhow::Result;
+use ethcontract::BlockNumber;
+use futures::future::BoxFuture;
+use std::time::SystemTime;
+
+/// Trait for reading historic orderbook data.
+pub trait HistoricOrderbookReading: StableXOrderBookReading {
```
fleupold, line 11: Are there going to be OrderBookReading implementations that won't implement `HistoricOrderbookReading`? I find the name slightly confusing since the existing `StableXOrderBookReading` also works on historic batches.

Maybe we don't need to differentiate these traits here or if so maybe even have three traits ReadingByBatch|Block|Timestamp?

core/src/orderbook/history.rs
```diff
@@ -0,0 +1,98 @@
...
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>> {
+        immediate!(Ok(Default::default()))
+    }
+}
+
+#[cfg(test)]
+mod mock {
+    use super::*;
+
+    /// Proxy trait used for mocking a `HistoricOrderbookReading` to work around
```
fleupold, line 50: Maybe it's time to re-evaluate if there are more powerful rust mocking frameworks out there.

fleupold (commented):
Question on why we need the separation in batch and (block + timestamp) for the orderbook reading.

It looks like this PR is rebased weirdly and therefore much larger than it should. Would you mind rebasing so that only the relevant change shows up?

core/src/orderbook/history.rs
e00E: That's a good point. Stablexorderbook reading accepts a batch id parameter but it doesn't change the state of the contract to that batch, it just filters orders so they are different. Maybe it's enough to point that out in a comment.
I don't think we need to split it up into two traits but it can see situations where it would be useful in the future (like optimized orderbook reading that can't do historic). Three traits is more clunky to use because then you don't have one trait that can do everything. This trait requires StableXOrderBookReading so it doesn't have that problem.

core/src/orderbook/history.rs
```diff
@@ -0,0 +1,98 @@
...
+        &'a self,
+        block_number: BlockNumber,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
+
+    /// Retrieves the open orderbook at the specified timestamp.
+    fn get_auction_data_for_timestamp<'a>(
+        &'a self,
+        timestamp: SystemTime,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
+}
```
e00E, line 28: This could be one method taking an enum BlockNumberOrSystemTime (with a better name). Just an idea.

nlordell: I found the separate functions easier to compose when doing the implementation.

core/src/orderbook/history.rs
nlordell: > Are there going to be OrderBookReading implementations that won't implement HistoricOrderbookReading?

I didn't plan on implementing these methods for the `OnchainOrderbookReader` just because:
- It would require an archive node in order to execute the view methods on historic blocks
- Finding the block for a timestamp is fairly difficult and requires a binary search of some sort

This is why I did it this way for now, so that we didn't have to implement this on all of our existing orderbook reading implementations, just the event-based one. If we end up implementing this for the other orderbook readers then we can definitely merge it into one trait.

nlordell: How about `OrderbookReadingByBlockOrTimestamp`?

fleupold: I still find the trait inheritance a bit weird to be honest (since it also makes the mocking more complicated). The current fetching by batchId does implement historic lookup even for the onchain reader (up to the history limit of the [node](https://github.com/gnosis/dex-services/blob/8434889f5c969f5417abffa899908e56b153c4af/core/src/orderbook/onchain_filtered_orderbook.rs#L42-L45)).
I can understand that you don't want to implement it for no longer used orderbook fetching methods though.

Would it make sense to define a type that is `dyn OrderbookReadingByBatchId + OrderbookReadingByBlock + OrderbookReadingByTimestamp` so that it can be used in the price estimator but we still don't need to inherit traits which seem only separated for the convenience of not supporting the new methods in legacy implementations?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FfQoxwZBVWq5jhLXRty%2Fgiphy.gif&token=fine" data-gitmeme-token="fine"></a> , 

although I still find the separation between OrderbookReadingByBlockOrTimestamp and OrderbookReadingByBatchId a bit arbitrary.

core/src/orderbook/history.rs
nlordell: Hmm, good point. Maybe we should just have a single trait and then leave some implementations `todo!()`?

fleupold: > you can't create a trait object for Trait1 + Trait2 + ...

That's very unfortunate, here is the related [issue with an explanation](https://github.com/rust-lang/rust/issues/32220#issuecomment-195836224)?

I'm not sure what is better. Leaving them unimplemented may lead to runtime errors rather than compile time errors which is usually not great. Maybe it's not too much work to implement them afterall given they already operate on block numbers? But that doesn't need to be in this PR.

nlordell: > Leaving them unimplemented may lead to runtime errors rather than compile time errors which is usually not great

Yeah, if we make it one trait, then we should commit to implementing it for all currrent implementations.

> Maybe it's not too much work to implement them afterall given they already operate on block numbers?

The biggest issue is searching for the block with a specified timestamp. But we should be able to reuse @fedgiac 's work with the binary search for a block with the closest timestamp.

nlordell:
Marking as `on hold` in order to explore adding these methods to `StableXOrderBookReading` trait.

PR #1291, opened 2020-08-11
Implement HistoricOrderbookReading for the event-based orderbook

nlordell:
This PR implements the new `HistoricOrderbookReading` for the event-based orderbook.

This PR depends on #1292

### Test Plan

Added unit test for constructing an orderbook by a timestamp for the event registry.

### DIFF
--- a/core/src/history/events.rs
+++ b/core/src/history/events.rs
@@ -1,26 +1,31 @@
 use crate::{
     models::{AccountState, BatchId, Order},
-    orderbook::{streamed::State, StableXOrderBookReading},
+    orderbook::{history::HistoricOrderbookReading, streamed::State, StableXOrderBookReading},
+    time::SystemTimeExt as _,
 };
 use anyhow::{Context, Result};
 use contracts::batch_exchange;
-use ethcontract::{contract::EventData, H256};
-use futures::future::{BoxFuture, FutureExt as _};
+use ethcontract::{contract::EventData, BlockNumber, H256};
+use futures::future::BoxFuture;
 use log::info;
 use serde::{Deserialize, Serialize};
-use std::collections::BTreeMap;
-use std::convert::TryFrom;
-use std::fs;
-use std::fs::File;
-use std::io::{Read, Write};
-use std::path::Path;
+use std::{
+    collections::BTreeMap,
+    convert::TryFrom,
+    fs,
+    fs::File,
+    io::{Read, Write},
+    ops::Bound,
+    path::Path,
+    time::SystemTime,
+};
 
 // Ethereum events (logs) can be both created and removed. Removals happen if the chain reorganizes
 // and ends up not including block that was previously thought to be part of the chain.
 // However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
 // list of all events based on which the state is built.
 
-#[derive(Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(Debug, Default, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
 struct EventSortKey {
     block_number: u64,
     /// Is included to differentiate events from the same block number but different blocks which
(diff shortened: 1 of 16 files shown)


nlordell:
This PR depends on #1292

core/src/orderbook/streamed/block_timestamp_reading.rs
```diff
@@ -117,12 +117,17 @@ impl<T: BlockTimestampBatchReading + Send> CachedBlockTimestampReader<T> {
 }
 
 impl<T: BlockTimestampReading + Send> BlockTimestampReading for CachedBlockTimestampReader<T> {
-    fn block_timestamp(&mut self, block_hash: H256) -> BoxFuture<Result<u64>> {
+    fn block_timestamp(&mut self, block: BlockId) -> BoxFuture<Result<u64>> {
         async move {
+            let block_hash = match block {
+                BlockId::Hash(hash) => hash,
+                _ => return self.inner.block_timestamp(block).await,
```
fleupold, line 124: I assume we don't want to cache the timestamps of block numbers, since they can change? Maybe we can later implement a cache if the timestamp is sufficiently long in the pas to be considered "final". I'm a little bit worried that consumers of a `CachedBlockTimestampReader` would expect the implementation to cache all types of blockIds.

fleupold (commented):
As for versioning, can the `EventRegistry` struct maybe have a version number field? Then in `try_from(bytes: &[u8]` we can check that the loaded version from disk is equal to the one currently hardcoded in the code.

fleupold:
> As for versioning, can the EventRegistry struct maybe have a version number field?

I swear I didn't see #1292 🤦

core/src/history/events.rs
```diff
@@ -199,7 +232,46 @@ impl StableXOrderBookReading for EventRegistry {
...
     ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>> {
-        async move { self.auction_state_for_batch(batch_id_to_solve) }.boxed()
+        immediate!(self.auction_state_for_batch(batch_id_to_solve))
+    }
+}
+
+impl HistoricOrderbookReading for EventRegistry {
+    fn get_auction_data_for_block<'a>(
+        &'a self,
+        block_number: BlockNumber,
```
e00E, line 242: Maybe it makes sense to change the trait so that it is always a u64 and cannot be Earliest, Latest, Pending since those values probably do not make sense to be used. Earliest is always empty, if you want latest or pending you wouldn't use the **Historic**OrderbookReading.

core/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -215,6 +215,44 @@ impl StableXOrderBookReading for UpdatingOrderbook {
...
+}
+
+impl HistoricOrderbookReading for UpdatingOrderbook {
+    fn get_auction_data_for_block<'a>(
+        &'a self,
+        block_number: BlockNumber,
+    ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>> {
+        // NOTE: Get the timestamp for the block, this will give more accurate
+        // results as to what batch the orderbook is retrieved for.
+        self.do_with_context(move |context| {
```
e00E, line 229: We can solve this by exposing the block number -> timestamp conversion in EventRegistry and using this from here. That way no changes to the Cache are needed.
EventRegistry will give us the timestamp of the closest lower block it has which is enough since that means it doesn't have any events for other blocks up to this number anyway.

core/src/orderbook/streamed/block_timestamp_reading.rs
nlordell: The issue with caching by block number is that in reorgs the cache may become invalidated.

core/src/orderbook/streamed/updating_orderbook.rs
nlordell: This was done intentionally, since we want a more accurate timestamp to use for determining the current batch ID. The reason for this is, lets say that there are no events for some blocks and we request the orderbook at a block in the middle. It could be that the block is in a different batch than the what the current batch was at the block where the last event was received.

core/src/orderbook/streamed/block_timestamp_reading.rs
nlordell: I think we can fix this if we are sensitive to reorgs by having a number of confirmation blocks before committing the timestamp to the cache.

nlordell:
Closing as this is superseded by #1356

PR #1293, opened 2020-08-11
Handle block queries

### LINKED ISSUE #1272: Allow to get price estimates at certain block numbers

nlordell:
Fixes #1272 

This PR implements retrieving prices at specified block numbers and timestamps.

### Test Plan

TODO: currently waiting on event orderbook to finish syncing.

### DIFF
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -3,9 +3,11 @@
 All endpoints use the query part of the URL with these key-values:
 
 * `unit`: Either `atoms` or `baseunits` (default). If set to `atoms` all amounts are denominated in the smallest available unit (atom) of the token. If `baseunits` all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. TODO: `baseunits` is currently not implemented for all URLs.
-* `atoms`: Deprecated. An boolean alias for `unit` parameter where `atoms=true` is equivalent to `unit=atoms` and `atoms=false` is equivalent to `unit=baseunits`.
+* `atoms`: Deprecated. An boolean alias for `unit` parameter where `atoms=true` is equivalent to `unit=atoms` and `atoms=false` is equivalent to `unit=baseunits`. Note that this parameter may not be specified together with the `unit` parameter.
 * `hops`: TODO: document this once it has been implemented.
-* `batchId`: Specify a specific batch ID to compute the estimate for, only accounting orders that are valid at the specified batch. If no batch ID is specified, the current batch that is collecting orders will be used.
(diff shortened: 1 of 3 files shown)


fleupold (approved):
🚀

price-estimator/README.md
```diff
@@ -3,9 +3,11 @@
...
 * `hops`: TODO: document this once it has been implemented.
-* `batchId`: Specify a specific batch ID to compute the estimate for, only accounting orders that are valid at the specified batch. If no batch ID is specified, the current batch that is collecting orders will be used.
+* `batchId`: Specify a specific batch ID to compute the estimate for, only accounting orders that are valid at the specified batch. If no batch ID is specified, the current batch that is collecting orders will be used. This parameter cannot be specified together with the `blockNumber` and `timestamp` parameters.
+* `blockNumer`: Specify a specific block number to compute the estimate for. This will use th open orderbook at that block (i.e. orders that will be considered for solving the current batch at the block number). This parameter cannot be specified together with the `batchId` and `timestamp` parameters.
+* `timestamp`: Specify a specific Unix epoch timestamp **in seconds** at which point in time to compute the estimate for. This will use th open orderbook at that time. This parameter cannot be specified together with the `batchId` and `blockNumber` parameters.
```
e00E, line 10: Just throwing it out there that the timestamp could be a float (in seconds) instead of u64. SystemTime should be creatable from float too. That's more general so maybe preferable but not important anyway.

e00E (approved):

price-estimator/README.md
nlordell: I don't think this has much value since batches are delimited on whole seconds and block timestamps on the EVM are in whole seconds as well. I don't think there would be any situation where the fractional bit of the timestamp changes the result. But maybe it would be nice to parse fractional timestamps for client convenience?

e00E: I'm fine either way. Just wanted to mention it because I remembered how in python the unix time is `time.time()` is a float in seconds. As you say the only advantage is client convenience. The way it's written in the readme now doesn't specify whether `10.0` would be accepted.

nlordell: :+1: Lets implement it then for convenience.

nlordell: It turns out the the query deserializer is not a self-describing serializer so `serde(untagged)` doesn't work. I'll do this as a separate PR since it might be a bit more involved.

nlordell:
Superseded by #1357

Issue #1275, opened 2020-08-10
Query orderbook/price estimates by tokenSymbol & address


fleupold:
For convenience, it would be nice to query all routes not just by tokenId but also by symbol and contract address.

E.g.

https://dex-price-estimator.gnosis.io/api/v1/markets/DAI-WETH?atoms=true

https://dex-price-estimator.gnosis.io/api/v1/markets/0x6b175474e89094c44da98b954eedeac495271d0f-0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2?atoms=true

I'm not sure how to proceed in the case of duplicate symbols. Should we return an error or should we use the first token with this symbol (to make sure integrations can not break later, e.g. if someone lists a fake "DAI" token.

PR #1294, opened 2020-08-11
Token by symbol or address

nlordell:
One step towards #1275 

This PR implements handling of token symbols and addresses for incoming requests. Note that there is still no implementation for mapping token symbols and addresses to token IDs, that will come in a follow up PR.

### Test Plan

```
$ cargo run -p price-estimator -- --orderbook-file target/orderbook.mainnet.bin
...
$ curl -s 'http://localhost:8080/api/v1/markets/WETH-OWL/estimated-buy-amount/1?atoms=true' | jq 
{
  "message": "token symbol or address not found"
}
```

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -8,7 +8,7 @@ use core::{
     models::TokenId,
     token_info::{TokenBaseInfo, TokenInfoFetching},
 };
-use pricegraph::{Pricegraph, TokenPair, TransitiveOrder};
+use pricegraph::{Market, Pricegraph, TokenPair, TransitiveOrder};
 use std::convert::Infallible;
 use std::sync::Arc;
 use warp::{
@@ -24,10 +24,13 @@ pub fn all(
 ) -> impl Filter<Extract = impl Reply, Error = Infallible> + Clone {
     warp::path!("api" / "v1" / ..)
         .and(
-            markets(orderbook.clone())
+            markets(orderbook.clone(), token_info.clone())
                 .or(estimated_buy_amount(orderbook.clone(), token_info.clone()))
-                .or(estimated_amounts_at_price(orderbook.clone(), token_info))
-                .or(estimated_best_ask_price(orderbook)),
+                .or(estimated_amounts_at_price(
+                    orderbook.clone(),
+                    token_info.clone(),
+                ))
+                .or(estimated_best_ask_price(orderbook, token_info)),
         )
         .recover(handle_rejection)
 }
@@ -35,6 +38,10 @@ pub fn all(
 #[derive(Debug)]
 struct NoTokenInfo;
 impl Reject for NoTokenInfo {}
+
+#[derive(Debug)]
+struct TokenNotFound;
+impl Reject for TokenNotFound {}
 
 async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
     let code;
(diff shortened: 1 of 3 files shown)


price-estimator/src/filter.rs
```diff
@@ -44,6 +48,10 @@ pub fn all(
 struct NoTokenInfo;
 impl Reject for NoTokenInfo {}
 
+#[derive(Debug)]
+struct TokenNotFound;
+impl Reject for TokenNotFound {}
```
e00E, line 53: Maybe merge it with the struct above that we already have `NoTokenInfo`. Or do we think it's important to give the user a different error message?

price-estimator/src/models/currency_pair.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+/// A token reference, either by token ID, address, or symbol.
+#[derive(Clone, Debug, PartialEq)]
+pub enum TokenRef {
+    Id(TokenId),
+    Address(Address),
+    Symbol(String),
+}
```
e00E, line 58: One problem I see now is that it is possible for a token symbol to also be a valid id or address. We should make sure that Id has the highest priority, then Address then Symbol. In this PR it is already correct but I'd add a comment there that the order of deserialization attempts matters.

e00E (approved):

price-estimator/src/filter.rs
```diff
@@ -54,6 +62,9 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
     } else if let Some(NoTokenInfo) = err.find() {
         code = StatusCode::BAD_REQUEST;
         message = "request with atoms=true for token we don't have erc20 info for";
```
fleupold, line 64: Unrelated: It seems like this message is a) inverted and b) referencing the deprecated parameter

price-estimator/src/filter.rs
fleupold: I think in one case we have found the token but don't have decimal information (all routes in atoms still work) and in the other case the token might not even be listed (no routes will work). I think it's worth while differentiating the two.

price-estimator/src/models/currency_pair.rs
fleupold: Good observation. I think this behaviour is already guaranteed/documented by the unit tests (but a comment can't hurt).

fleupold (approved):

price-estimator/src/filter.rs
nlordell: > Maybe merge it with the struct above that we already have NoTokenInfo

I agree, we should probably only have one error type for rejections.

> I think it's worth while differentiating the two.

I think we can use an enum with different variants so we can differentiate between the errors.

e00E: These structs are only needed so we can downcast and set the correct error message. I don't think we need to create one enum that holds both NoTokenInfo and TokenNotFound but it is an idea to create a big enum that holds all errors (but not in this PR).

nlordell: I implemented this in #1299, so I'll leave the error as is for now.

price-estimator/src/models/currency_pair.rs
nlordell: Added a comment in the `FromStr` implementation.

PR #1295, opened 2020-08-12
Fuzz order for limit price

e00E:
This method wasn't being fuzzed.

### Test Plan
Let fuzzing run on my machine.

### DIFF
--- a/pricegraph/src/api/price_estimation.rs
+++ b/pricegraph/src/api/price_estimation.rs
@@ -122,6 +122,10 @@ impl Pricegraph {
             buy: pair.sell,
             sell: pair.buy,
         };
+
+        if limit_price < 1e-15 {
+            return None;
+        }
         let max_xrate = LimitPrice::new(limit_price)?.exchange_rate().inverse();
 
         let mut total_buy_volume = 0.0;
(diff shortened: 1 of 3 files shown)


pricegraph/src/api/price_estimation.rs
```diff
@@ -122,6 +122,10 @@ impl Pricegraph {
             buy: pair.sell,
             sell: pair.buy,
         };
+
+        if limit_price < 1e-15 {
+            return None;
+        }
         let max_xrate = LimitPrice::new(limit_price)?.exchange_rate().inverse();
```
e00E, line 129: TODO: How do we properly solve this? The problem is that a very low limit price

(the fuzzer found
```
`0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000059755401472367
```
)

causes the inverse to become infinite which fails the debug assert without this check.

Since this is an extreme edge cache I am okay with forcing the fuzzer to have more reasonable limit prices but I am not sure whether it is better to have some kind of check here instead. I chose `1e-15`  just to illustrate the point.

nlordell: We should create an issue. I think this is definitely the case of an incorrect assumption about floating point math in the `pricegraph` that we should fix.

nlordell (approved):

Issue #1264, opened 2020-08-07
Create interactive api docs


e00E:
The safe has this kind of https://safe-relay.gnosis.io/#v1/gas-station/ api documentation which is explorable in the browser and has a "try it out" button.
Would be nice to have this for the price estimator too.

Potential todo:
- [ ] Host the interactive api under a gnosis url using the docker image from the price estimator readme.

PR #1296, opened 2020-08-12
Add openapi documentation

e00E:
Part of #1264 .

### Test Plan
Load the file in https://editor.swagger.io/ and see the api docs.

### DIFF
--- a/price-estimator/openapi.yml
+++ b/price-estimator/openapi.yml
@@ -1,0 +1,150 @@
+openapi: 3.0.3
+info:
+  version: 0.0.1
+  title: Price Estimator OpenAPI
+paths:
+  /api/v1/markets/{market}/estimated-buy-amount/{sell amount in quote}:
+    get:
+      summary: Estimated Buy Amount
+      description: Estimate the buy amount (in buy tokens) a user can set as a limit order while still expecting to be completely matched when selling the given amount of quote token.
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AmountResponse"
+      parameters:
+        - $ref: "#/components/parameters/Market"
+        - name: sell amount in quote
+          required: true
+          in: path
+          schema:
+            $ref: "#/components/schemas/NumberParameter"
+          example: 1
+        - $ref: "#/components/parameters/Unit"
+        - $ref: "#/components/parameters/BatchId"
+  /api/v1/markets/{market}/estimated-amounts-at-price/{price}:
+    get:
+      summary: Estimated Amounts At Price
+      description: Estimate largest buy and sell amounts for a limit at given price that would get matched. Note that it might be possible to use a higher buy amount for the same returned sell amount and still likely get completely matched by the solver. This buy amount can be computed with a subsequent estimate buy amount API call using the returned sell amount in quote value.
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AmountResponse"
+      parameters:
+        - $ref: "#/components/parameters/Market"
+        - name: price
+          required: true
+          in: path
+          schema:
+            $ref: "#/components/schemas/NumberParameter"
+          example: 400
+        - $ref: "#/components/parameters/Unit"
+        - $ref: "#/components/parameters/BatchId"
+  /api/v1/markets/{market}/estimated-best-ask-price:
+    get:
+      summary: Estimated Best Ask Price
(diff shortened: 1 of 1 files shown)


price-estimator/openapi.yml
```diff
@@ -0,0 +1,81 @@
...
+      example: 1-7
+  UnitParameter:
+    name: unit
+    in: query
+    description: If `baseunits` (the default) all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. If set to `atoms` all amounts are denominated in the smallest available unit (atom) of the token.
+    required: false
+    schema:
+      type: string
+      enum: [baseunits, atoms]
+      example: baseunits
```
e00E, line 63: I am assuming Nick's PR that changes the unit parameter to work like this instead of `atoms=true|false` gets merged. That PR deprecates the atoms syntax so we shouldn't document it.

price-estimator/src/filter.rs
```diff
@@ -40,13 +39,12 @@ pub fn all(
...
 
-async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
+async fn handle_rejection(err: Rejection) -> Result<impl Reply, Rejection> {
     let code;
     let message;
 
     if err.is_not_found() {
-        code = StatusCode::NOT_FOUND;
-        message = "invalid url path";
+        return Err(reject::not_found());
```
e00E, line 47: The reason for this is that rejections should be used to signal that a filter cannot handle a request but another filter might. So we shouldn't return a response in the case of the path not matching so that the next filter can try to match the path.

price-estimator/src/openapi.rs
```diff
@@ -0,0 +1,5 @@
+use warp::{Filter, Rejection, Reply};
+
+pub fn openapi() -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    warp::path("openapi").and(warp::fs::dir("price-estimator/openapi"))
```
e00E, line 4: Will make the path on disk configurable in another PR.

price-estimator/openapi.yml
```diff
@@ -0,0 +1,136 @@
...
+      type: string
+      example: 1-7
+  UnitParameter:
+    name: unit
+    in: query
+    description: If `baseunits` (the default) all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. If set to `atoms` all amounts are denominated in the smallest available unit (atom) of the token.
+    required: false
+    schema:
+      type: string
+      enum: [baseunits, atoms]
```
nlordell, line 92: Should you add `default: baseunits`?

price-estimator/swagger-ui.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
nlordell, line 1: Personally, I'm kind of against having `price-estimator` host the swagger UI HTML. Its fairly easy to go to https://petstore.swagger.io/ and just input `http://localhost:8080/openapi/openapi.yaml` to get the UI. I think just having the OpenAPI definition is enough.

If we really want to include host the swagger UI, I think it should be added in a separate PR.

e00E: Ah so we would only host the one file but not the full ui. I'll have to check if this will work with CORS (setting allow origin header).
I didn't think of this possibility when I made the PR but I do like self hosting all the required files because it works without any external website. Why are you against this?

nlordell: > I'll have to check if this will work with CORS

Good point. We may need to enable CORS correctly.

> Why are you against this?

Not against self-hosting, I just don't think it should be the `price-estimator` that hosts this, and maybe this should be hosted in the dev portal instead of the service. Another concern of mine is that the zip bundled UI may contain Swagger branding that we maybe aren't allowed to distribute.

Another thing to consider is using `swagger-ui(-dist)?` NPM package and host that in a little HTML page of our own instead of downloading from a ZIP bundle.

fleupold (approved):
The creation script wasn't fully working for me (comments inline). Agree that we should maybe host it somewhere in a separate container (or for now just link the swagger hosted link in the dev portal).

e00E:
Changed PR to only include the yaml file on disk. No http route for it. Then we can decide the rest in another PR.

price-estimator/openapi.yml
```diff
@@ -0,0 +1,139 @@
...
+    Unit:
+      name: unit
+      in: query
+      description: If `baseunits` (the default) all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. If set to `atoms` all amounts are denominated in the smallest available unit (atom) of the token.
+      required: false
+      schema:
+        type: string
+        enum: [baseunits, atoms]
+        default: baseunits
+        example: baseunits
```
nlordell, line 139: I think the `batchId` query parameter is missing.

nlordell (approved):
Thanks. I think I agree with @fleupold that we can have a `Dockerfile` to build an image for hosting the Swagger UI.

PR #1298, opened 2020-08-12
Pass extra_rounding_buffer_factor as command line argument

e00E:
so that it can be configured instead of being hardcoded.

### Test Plan
CI


nlordell (approved):
easy peasy lemon squeezy

PR #1304, opened 2020-08-12
Health endpoint

nlordell:
This PR introduces a `health` filter so that the price estimator can signal to our deployment that it is ready and we can implement rolling updates for 0 downtime goodness.

This is implemented in the `core` crate since `driver` will need to have an endpoint running in the background as well.

### Test Plan

New unit test for filter, and
```
```

While it is starting up, check that the `health` endpoint does not respond:
```
$ curl -i 'http://localhost:8080/health'
curl: (7) Failed to connect to localhost port 8080: Connection refused
```

Wait for the `price-estimator` to be ready:
```
2020-08-12T19:56:34.396Z INFO [price_estimator] Server ready.
```

Then check that the `health` endpoint responds correctly:
```
$ curl -i 'http://localhost:8080/health'
HTTP/1.1 204 No Content
cache-control: no-store
content-length: 0
date: Wed, 12 Aug 2020 19:56:36 GMT
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3768,9 +3768,9 @@ checksum = "373c8a200f9e67a0c95e62a4f52fbf80c23b4381c05a17845531982fa99e6b33"
 
 [[package]]
 name = "uint"
-version = "0.8.4"
+version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "429ffcad8c8c15f874578c7337d156a3727eb4a1c2374c0ae937ad9a9b748c80"
+checksum = "9db035e67dfaf7edd9aebfe8676afcd63eed53c8a4044fed514c8cccf1835177"
 dependencies = [
  "byteorder",
  "crunchy",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -461,7 +461,6 @@ dependencies = [
  "thiserror",
  "uint",
  "url 2.1.1",
- "warp",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


e00E (approved):
For the price estimator we could use any route as the health check because it only starts running the server once it is ready but using a specific route is maybe even better in case the check gets more complicated in the future.

However, I don't think core should be using warp and tokio)because we already have a runtime (asyncstd) and web server (rouille) there. I don't like the added dependencies. That does mean we won't share the code for the filter but it is short anyway. Then the driver health check can probably live in the driver crate and reuse the existing dependencies.

We can discuss more in the other PR. I'll approve based on the assumption that we move the filter into price-estimator for this PR.

nlordell:
> However, I don't think core should be using warp and tokio)because we already have a runtime (asyncstd) and web server (rouille) there.

Fair points... I was just trying to make sure that our HTTP health response is the same in both the driver and the price estimator. That being said, you make a good point and we can simply re-use `rouille` or `tiny_http` (what `rouille` is built on) for the `driver`.

fleupold (approved):
Looks good for a start (agree with @e00E that warp should not become a core dependency). I'd envision the current `health` to be the ready check. For a liveness check should we check the timestamp of the last block that the orderbook has indexed? If that one is too far in the past it could be a sign that we are not "live" anymore.

e00E:
@fleupold  you mention a health and ready check, but do we really need two? Imo health means it is ready to do it's job so if it is healthy it is also ready. While it is starting up it is not yet healthy.

nlordell:
I agree with @e00E, I think for the driver and price estimator we can just have a `readiness` check since we should have prometheus metrics with alerts for `healthyness` checks.

nlordell:
Moved filter to `price-estaimtor` and changed path to `/health/readiness` so we can have a `/health/liveliness` path as well in the future.

PR #1305, opened 2020-08-12
introduce HTTP health reporting endpoint that runs in a backgroun thread

nlordell:
This PR adds a new `HttpHealthEndpoint` component for that handles `/health/readiness` requests in the bacground HTTP server. It is currently only used in the price estimator (the driver always responds with "service unavailable").

This is in the spirit of providing a simple readiness endpoint so that we can do rolling updates in order to stay with 100% uptime even across restarts because of new deployments.

### Test Plan

Added unit tests. Start the price estimator:
```
```

While it is starting up, check that the `health` responds with a `503: Service Unavailable`:
```
$ curl -i http://localhost:9586/health/readiness
HTTP/1.1 503 Service Unavailable
Server: tiny-http (Rust)
Date: Tue, 18 Aug 2020 08:45:41 GMT
Content-Type: text/plain; charset=utf8
Content-Length: 19

service unavailable
```

Wait for the `price-estimator` to be ready:
```
2020-08-18T08:45:42.414Z INFO [price_estimator] Server ready.
```

Then check that the `health` endpoint responds correctly:
```
$ curl -i http://localhost:9586/health/readiness
HTTP/1.1 204 No Content
Server: tiny-http (Rust)
Date: Tue, 18 Aug 2020 08:45:43 GMT
Content-Length: 0
```


fleupold (commented):
Lookging at https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ I'm not sure if blocking until we are ready is the right way to go. It feels like the probe should finish quickly and will then just be redone after some time if not yet ready. Maybe we can just share the AtomicBool and render an empty 200 response if it's ready and a 503 if not?

Also, any reason for using warp here? This feels conceptually closer to the metrics server we already expose and that is using rouille (maybe it could become part of that server instead of exposing a new port). I'm not against warp, but maybe we can then get rid of rouille.

core/src/health.rs
```diff
@@ -0,0 +1,83 @@
...
+use anyhow::Result;
+use rouille::{Request, Response};
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// Trait for asyncronously notifying health information.
+#[cfg_attr(test, mockall::automock)]
+pub trait HealthReporting: Send + Sync {
+    /// Notify that the service is ready.
+    fn notify_ready(&self);
+}
```
e00E, line 13: I don't think we need this trait because the only implementor is HttpHealthEndpoint and we don't need to mock it. Not a big deal to have it either or maybe you plan to have a second implementor.

e00E (approved):

core/src/health.rs
nlordell: Its mostly to test that the driver only notifies that it is healthy after running the first batch.

core/src/health.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl HttpHealthEndpoint {
+    /// Creates a new HTTP health enpoint.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns true if the service is ready, false otherwise.
+    fn is_ready(&self) -> bool {
+        self.ready.load(Ordering::Acquire)
```
fleupold, line 29: So fancy, why not just `SeqCst`?

price-estimator/src/main.rs
```diff
@@ -195,11 +195,16 @@ fn duration_secs(s: &str) -> Result<Duration, ParseIntError> {
     Ok(Duration::from_secs(s.parse()?))
 }
 
-fn setup_driver_metrics() -> HttpMetrics {
+fn setup_driver_metrics() -> (HttpMetrics, Arc<dyn HealthReporting>) {
```
fleupold, line 198: nit: rename as well/

fleupold (approved):

core/src/health.rs
e00E: Acquire, Release is more efficient and sufficient in this case even though for this operation the speed difference doesn't matter.

nlordell: So, from the docs:
>
> all previous writes become visible to all threads that perform an `Acquire` (or stronger) load of this value.

fleupold: No strong feelings. I felt that using the "right" flags here requires some thought and care which isn't really necessary given the marginal benefit. But if you prefer to be "optimal", that's fine.

nlordell: Switched it to `SeqCst` for lower developer overhead.

PR #1308, opened 2020-08-13
Estimate considered good if it is within 1 order of magnitude

nlordell:
Previously we were using the absolute error to determine if a price estimate was bad. This was not exactly correct, because its more important for the estimate to be within a certain x orders of magnitude of the real price.

### Test Plan

```
cargo run --release -p e2e --bin historic_prices -- --orderbook-file target/orderbook-file-mainnet
    Finished release [optimized] target(s) in 3.29s
58296 / 58296 [=====================================================================] 100.00 % 364.71/s
Processed 16548 token prices: average error 22941.12%, bad 2.91%, missed 9.39%.
```

### DIFF
--- a/e2e/src/bin/historic_prices.rs
+++ b/e2e/src/bin/historic_prices.rs
@@ -71,7 +71,7 @@ fn check_batch_prices(
         .filter(|(_, token)| *token != 0)
     {
         let price = prices[token_index];
-        let estimate = pricegraph.estimate_token_price(token).unwrap_or(0.0) as u128;
+        let estimate = pricegraph.estimate_token_price(token).map(|p| p as u128);
 
         samples.record_sample(Row {
             batch,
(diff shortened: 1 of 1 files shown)


e2e/src/bin/historic_prices.rs
```diff
@@ -5,8 +5,9 @@ use pricegraph::{Element, Pricegraph, TokenId};
 use std::{fs::File, io::Write, path::PathBuf};
 use structopt::StructOpt;
 
-/// Threshold at which a price estimate is considered "bad"; currently 10%.
-const BAD_ESTIMATE_THRESHOLD: f64 = 0.1;
+/// Threshold logarithmic distance from the actual price at which an estimate is
+/// considered "bad"; currently `0.1 * price < estimate < 10 * price`.
+const BAD_ESTIMATE_THRESHOLD_LOG_DISTANCE: f64 = 1.0;
```
e00E, line 10: Alternatively we could use a normal factor (10 in this case) but I don't mind logarithm either.

e00E (approved):

e2e/src/bin/historic_prices.rs
nlordell: I just thought it was easier to write with log and we usually say "within x orders of magnitude" for estimates that the base-10 log made sense here.

fleupold (approved):
📏

PR #1310, opened 2020-08-13
Remove non openapi documentation

e00E:
We should not duplicate the documentation and openapi is the better
format.

### Test Plan
Follow steps in readme and try interactive request.

### DIFF
--- a/price-estimator/openapi.yml
+++ b/price-estimator/openapi.yml
@@ -2,6 +2,13 @@ openapi: 3.0.3
 info:
   version: 0.0.1
   title: Price Estimator OpenAPI
+servers:
+- url: https://price-estimate-mainnet.dev.gnosisdev.com
+  description: Staging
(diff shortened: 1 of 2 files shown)


price-estimator/openapi.yml
```diff
@@ -137,10 +149,7 @@ components:
       description: If `baseunits` (the default) all amounts are denominated in the "natural" unit of the respective token given by the number of decimals specified through the ERC20 interface. If set to `atoms` all amounts are denominated in the smallest available unit (atom) of the token.
       required: false
       schema:
-        type: string
-        enum: [baseunits, atoms]
-        default: baseunits
-        example: baseunits
+        $ref: "#/components/schemas/Unit"
```
e00E, line 152: This is because I noticed that the online editor has an extra table for schemas so that they can be viewed separately from the full routes.

nlordell (approved):
Awesome!

PR #1311, opened 2020-08-13
Implement `GasPriceEstimating` for `Web3`

nlordell:
This PR adds am implementation of `GasPriceEstimating` for `Web3` so we can use it for CI. THis will give us more reproducible gas prices and [not cause CI failures when rinkeby gas prices are too high](https://travis-ci.com/github/gnosis/dex-services/jobs/371541072#L2346).

### Test Plan

CI should pass.


core/src/solution_submission.rs
```diff
@@ -227,10 +227,11 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
             // is not feasible because it would take too long.
             match self.gas_price_estimating.estimate_gas_price().await {
                 Ok(gas_price) => {
-                    if gas_price_cap < gas_price.fast {
+                    if gas_price_cap < gas_price {
                         return Err(SolutionSubmissionError::Benign(format!(
-                            "Solution does not generate enough fees for the transaction to execute quickly enough: price cap {} < fast gas price {}",
```
nlordell, line 232: This line was too long and cause `rustfmt` to skip this block.

e00E (approved):
Makes sense and it's nice that the trait only returns a number now instead of a struct where we select fast.

fleupold (commented):
I'm not a huge fan of adding another configuration for this. If I understand correctly the problem is this line:

https://github.com/gnosis/dex-services/blob/d60cd37bde3291dc2b389e2a43e1120b74b71ac9/core/src/gas_station.rs#L17-L18


Can we have a GasPriceEstimating factory method that in case of mainnet/rinkeby returns the SafeGasStation with mainnet/rinkeby URL and in case of 5777 returns the Web3 based one.

fleupold (approved):
Otherwise this looks good.

nlordell:
> I'm not a huge fan of adding another configuration for this

I think we might eventually need this when we deploy to xDAI anyway, but we can cross that bridge when we get there. Maybe revamp how we do configurations.

driver/src/main.rs
```diff
@@ -194,7 +194,7 @@ struct Options {
 
     /// The default maximum gas price. This is used when computing the maximum gas price based on
     /// ether price in owl fails.
-    #[structopt(long, env = "DEFAULT_MAX_GAS_PRICE", default_value = "10000000000")]
+    #[structopt(long, env = "DEFAULT_MAX_GAS_PRICE", default_value = "100000000000")]
```
nlordell, line 197: Was 10GWei, raised it to 100GWei

PR #1313, opened 2020-08-13
Document how to use openapi locally

e00E:
using the official swagger-ui docker container.

### Test Plan
Run command from the readme

### DIFF
--- a/price-estimator/README.md
+++ b/price-estimator/README.md
@@ -2,22 +2,27 @@
 
 The api is documented with [OpenAPI](https://www.openapis.org/) in `openapi.yml`. It can be accessed in a rendered and interactive fashion on https://editor.swagger.io/ through `File -> Import file`.
 
-TODO: Add a binary or docker image that can host the interactive ui without an external service.
+Alternatively locally with docker
+
+```
+docker run -p 80:8080 -e SWAGGER_JSON=/openapi.yml --mount type=bind,src="$PWD/openapi.yml",dst=/openapi.yml swaggerapi/swagger-ui
+```
+
+and open http://localhost:80 .
 
 ## Testing
 
-To test a locally running price estimator with the frontend at https://mesa.eth.link/ we need to set our browser to allow websites to access localhost and change the URL that the javascript uses for the price estimator. With chromium:
+To test a locally running price estimator with the frontend at https://mesa.eth.link/ we need to set our browser to allow websites to access localhost and change the URL that the javascript uses for the price estimator.
 
-1. `chromium --disable-web-security --user-data-dir=temp/`.
-2. Open the frontend.
-3. Open the developer tools with `F12`.
-4. In the browser console enter `dexPriceEstimatorApi.urlsByNetwork[1] = "http://localhost:8080/api/v1/"`.
-5. Induce a request by changing the sell amount and check that price estimator prints that it handled the request.
+* Open the frontend in a web browser.
+* Open the developer tools with `F12`.
+* In the browser console enter `dexPriceEstimatorApi.urlsByNetwork[1] = "http://localhost:8080/api/v1/"`.
+* Induce a request by changing the sell amount and check that price estimator prints that it handled the request.
 
 It is useful to start the price estimator with logging enabled, using the gnosis staging node URL and using a permanent orderbook file:
 
 ```
-env RUST_LOG=warn,price_estimator=info,core=info cargo run -p price-estimator -- --node-url https://staging-openethereum.mainnet.gnosisdev.com --orderbook-file ../orderbook-file-mainnet
+cargo run -p price-estimator -- --node-url https://staging-openethereum.mainnet.gnosisdev.com --orderbook-file ../orderbook-file-mainnet
 ```
 
 ## Benchmarking
--- a/price-estimator/src/main.rs
+++ b/price-estimator/src/main.rs
@@ -159,9 +159,17 @@ fn main() {
         options.orderbook_update_interval,
     ));
 
+    // We add the allow origin header so that requests from the interactive openapi documentation
+    // go through to locally running instance. This does mean we set the header for non openapi
+    // requests too. This doesn't have security implications because this is a public,
+    // unauthenticated api anyway.
     let filter = health::filter()
         .or(filter::all(orderbook, token_info))
-        .with(warp::log("price_estimator"));
+        .with(warp::log("price_estimator"))
+        .with(warp::reply::with::header(
+            "Access-Control-Allow-Origin",
+            "*",
+        ));
     let serve_task = runtime.spawn(warp::serve(filter).run(options.bind_address));
 
     log::info!("Server ready.");


price-estimator/src/main.rs
```diff
@@ -158,7 +158,14 @@ fn main() {
         options.orderbook_update_interval,
     ));
 
-    let filter = filter::all(orderbook, token_info).with(warp::log("price_estimator"));
+    // We add the allow origin header so that requests from the interactive openapi documentation
+    // go through to locally running instance.
```
fleupold, line 162: Is this still up to date? It seems like we don't need this for openapi anymore.

price-estimator/README.md
```diff
@@ -2,7 +2,13 @@
 
 The api is documented with [OpenAPI](https://www.openapis.org/) in `openapi.yml`. It can be accessed in a rendered and interactive fashion on https://editor.swagger.io/ through `File -> Import file`.
 
-TODO: Add a binary or docker image that can host the interactive ui without an external service.
+Alternatively locally with docker
+
+```
+sudo docker run -p 80:8080 -e SWAGGER_JSON=/openapi.yml --mount type=bind,src="$PWD/openapi.yml",dst=/openapi.yml swaggerapi/swagger-ui
```
fleupold, line 8: nit: Maybe document pwd and I don't believe sudo is necessary.

fleupold (approved):
Nice, is there a link we can share with public people (e.g. using the raw github file of openapi) to make it convenient to give the latest documentation to users (e.g. EFrontier)

price-estimator/src/main.rs
e00E: It is needed or the interactive docs are unable to access the locally running price estimator, even if the interactive docs are also running from localhost.

e00E:
> Nice, is there a link we can share with public people (e.g. using the raw github file of openapi) to make it convenient to give the latest documentation to users (e.g. EFrontier)

We can tell them to go to https://petstore.swagger.io/ and input `https://raw.githubusercontent.com/gnosis/dex-services/master/price-estimator/openapi.yml` at the top. Alternatively we could host the docker container under some Gnosis url. The latter feels more professional but not sure if the overhead of managing another kubernetes container is worth it.

price-estimator/README.md
```diff
@@ -2,7 +2,13 @@
 
 The api is documented with [OpenAPI](https://www.openapis.org/) in `openapi.yml`. It can be accessed in a rendered and interactive fashion on https://editor.swagger.io/ through `File -> Import file`.
 
-TODO: Add a binary or docker image that can host the interactive ui without an external service.
+Alternatively locally with docker. Note that the bind mount src path has to be absolute so we use `$PWD`.
+
+```
+docker run -p 80:8080 -e SWAGGER_JSON=/openapi.yml --mount type=bind,src="$PWD/openapi.yml",dst=/openapi.yml swaggerapi/swagger-ui
```
nlordell, line 8: When inspecting the dockerfile, I noticed: `SWAGGER_JSON_URL`. Any change this would work with a github URL?

price-estimator/src/main.rs
nlordell, line 162: Not a fan of this comment as it could be made more general: the statement is true for any CORS request, not just specific to OpenAPI. This also means that the instructions for testing locally will no longer require starting chrome with security enabled. In production and staging, we just have CORS configured in nginx for our deployment, so we never had to configure it at an app level.

price-estimator/src/main.rs
```diff
@@ -158,7 +158,14 @@ fn main() {
...
+    let filter = filter::all(orderbook, token_info)
+        .with(warp::log("price_estimator"))
+        .with(warp::reply::with::header(
+            "Access-Control-Allow-Origin",
+            "*",
+        ));
```
nlordell, line 168: Maybe this should only be enabled in debug builds, just in case. I expect that nginx should set these headers and not the `price-estimator` in production.

nlordell (approved):

price-estimator/README.md
nlordell: Indeed!

https://github.com/swagger-api/swagger-ui/blob/d9f5691f657dc64842b892255fe4a67cb6246f1b/docker/run.sh#L33-L36

e00E: The reason I want to use the local is to make it easier to edit it. You change it and refresh the page and it will be updated.

price-estimator/src/main.rs
e00E: We also want the openapi to work with the staging and production and nodes and either way this has no security impact so it makes sense to me that this is always enabled.

price-estimator/README.md
nlordell: Makes sense! Just a cool feature that we can use if we want to build a Docker image to host the API doc that we can deploy to k8s. I think for development purposes running it with `SWAGGER_JSON` makes more sense.

price-estimator/src/main.rs
nlordell: > We also want the openapi to work with the staging and production

This is already the case because of the nginx configuration:
```
curl -i -X OPTIONS https://price-estimate-mainnet.dev.gnosisdev.com/                                                                  
HTTP/1.1 204 No Content
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization
Access-Control-Allow-Methods: OPTIONS, GET, HEAD
Access-Control-Allow-Origin: *
Access-Control-Max-Age: 1728000
Content-Length: 0
Date: Thu, 13 Aug 2020 18:45:49 GMT
Connection: keep-alive
```

> this has no security impact 

This isn't strictly true, it enables browsers to make arbitrary requests to the service. Its much easier to get a whole bunch of browsers at different IPs to make HTTP requests than a whole bunch of applications. I do agree that this is not a realistic threat.

> that this is always enabled.

My main argument is that should be a deployment configuration and not a hard-coded thing. But either way, I don't think its super important in this case.

price-estimator/README.md
```diff
@@ -2,22 +2,27 @@
 
 The api is documented with [OpenAPI](https://www.openapis.org/) in `openapi.yml`. It can be accessed in a rendered and interactive fashion on https://editor.swagger.io/ through `File -> Import file`.
 
-TODO: Add a binary or docker image that can host the interactive ui without an external service.
+Alternatively locally with docker
+
+```
+sudo docker run -p 80:8080 -e SWAGGER_JSON=/openapi.yml --mount type=bind,src="$PWD/openapi.yml",dst=/openapi.yml swaggerapi/swagger-ui
```
nlordell, line 8: `sudo` is only necessary if you aren't in the `docker` (or something like that) group. I would also suggest removing it.

price-estimator/src/main.rs
```diff
@@ -159,9 +159,17 @@ fn main() {
         options.orderbook_update_interval,
     ));
 
+    // We add the allow origin header so that requests from the interactive openapi documentation
+    // go through to locally running instance. This does mean we set the header for non openapi
+    // requests too. This doesn't have security implications because this is a public,
+    // unauthenticated api anyway.
```
nlordell, line 165: I still think this comment is too specific to OpenAPI. We enable CORS for OpenAPI and for testing with Mesa.

e00E: Afaik the intention of CORS is to protect against some other site making requests to your site's "private" api. So you can't send an authenticated requests with cookies to do something with your youtube account when you're on example.com. That's why I included that statement.

The ddos part is not usually seen a security concern. Especially because even without the allow origins header the ddos would still work because the request still has to be sent for the browser to look at  the response header.

nlordell: Ah OK now I understand.

nlordell (approved):

PR #1315, opened 2020-08-13
Implement BaseUnits for Market route

### LINKED ISSUE #1082: Handle `atoms=false` (should be default)

fleupold:
Second to last part of #1082

This PR implements base unit support for the markets route.

Thinking about how we can unify this type of logic, I'm wondering if it would make sense to introduce a trait `into_unit(unit, context)` that all our result types could implement (given the logic of how we convert atoms to base units likely depends on the operation). This would allow us moving logic away from the filters, but I'm not sure how much value there is in that.

### Test Plan
Added unit test + comparing the resulting book (atoms=false) with the visualization on Mesa. In particular picking some large orders and checking that volume and price correspond to the base unit values displayed on the web.


price-estimator/src/models/markets_results.rs
```diff
@@ -25,6 +26,45 @@ impl From<&pricegraph::TransitiveOrderbook> for MarketsResult {
...
+                self.bids,
+                &base_token_info,
+                &quote_token_info,
+            ),
+        };
+    }
+}
+
+fn into_transitive_orders_in_base_unit(
+    orders: Vec<TransitiveOrder>,
```
e00E, line 51: Instead of taking Vec<TransitiveOrder> and returning Vec<TransitiveOrder> this function could take &mut [TransitiveOrder] but if you prefer this it's also ok.

e00E (approved):

price-estimator/Cargo.toml
```diff
@@ -20,3 +20,6 @@ thiserror = "1.0"
 tokio = { version = "0.2", features = ["macros", "rt-threaded", "time"] }
 url = "2.1"
 warp = "0.2"
+
+[dev-dependencies]
+assert_approx_eq = "1"
```
nlordell, line 25: nit missing newline

price-estimator/src/models/markets_results.rs
nlordell: I agree, I would prefer in-place mutation here with `iter_mut()`.

nlordell (approved):
:tada: , some small nits

price-estimator/src/models/markets_results.rs
fleupold: From other programming languages I tend to prefer methods without side effects, but I guess since rust makes the mutation very explicit it's really the same here (or is there any benefit in not consuming it). Will change.

price-estimator/src/models/markets_results.rs
```diff
@@ -25,6 +26,34 @@ impl From<&pricegraph::TransitiveOrderbook> for MarketsResult {
     }
 }
 
+impl MarketsResult {
+    pub fn into_base_units(
+        self,
```
fleupold, line 31: Should this then also take a &mut self? What would be the naming convention for that?

price-estimator/src/models/markets_results.rs
nlordell: > From other programming languages I tend to prefer methods without side effects

In those cases, I would slightly prefer passing in an `impl IntoIterator<Item = TransitiveOrder>` and return an `impl Iterator<Item = TransitiveOrder>`. This way the caller doesn't necessarily need to a) privide a `Vec`, but any iterator over transitive orders and b) the caller isn't forced to collect a `Vec`, but can theoretically collect into a `HashMap` of price ranges (a contrived example). However, if you are forcing the caller to give ownership of a `Vec` to return a new one, then I would argue the extra allocation is unnecessary.

One long term advantage of the `fn (IntoIterator) -> Iterator` way is that we may be able to implement transitive order grouping as an iterator, then we wouldn't need any intermediate allocations, but instead compose these iterators and collect once at the end :smile:

price-estimator/src/models/markets_results.rs
nlordell: I think taking `self` here is fine, since it doesn't have any `Box`-ed data anyway. In that case you coud:
```
pub fn into_base_units(mut selt, ...) -> Self {
    transitive_orders_in_base_unit(&mut self.asks, ...);
    transitive_orders_in_base_unit(&mut self.bids, ...);
    self
}
```

price-estimator/src/models/markets_results.rs
e00E: It's definitely a preference which is why I said i'm fine with either way. Whether the function takes one mutable argument or returns one value is mostly equivalent but by taking a mut slice reference it is clear that the number of elements doesn't change.
On the other hand I'm not sure I like introducing generics until they are needed. It's good to keep it in mind that the function can be generict that way but if we know it's only called in one place and that place is a vector then I usually prefer it to not be generic yet.

nlordell: > Whether the function takes one mutable argument or returns one value is mostly equivalent

But... my poor memory allocator!

> It's good to keep it in mind that the function can be generict that way but if we know it's only called in one place and that place is a vector then I usually prefer it to not be generic yet.

This is a very good point! To be clear I was just trying to illustrate an example where the "pure functional" way to write this function could be beneficial and could produce equally optimized code as with the mutable reference and be more composable than with mutable references or fixed types.

For this case, I think the `&mut [_]` is the way to go.

PR #1316, opened 2020-08-13
Add method for reading orderbook at block number

nlordell:
This PR adds a new `StableXOrderbookReading::get_auction_data_for_block` trait method with additional implementations.

This would be an alternative to introducing a new trait for querying at a block number and timestamp in #1290. In a follow up PR I can add the method for querying at a specific timestamp as well.

### Test Plan

Added some new unit tests.

### DIFF
--- a/core/src/history/events.rs
+++ b/core/src/history/events.rs
@@ -4,23 +4,26 @@ use crate::{
 };
 use anyhow::{Context, Result};
 use contracts::batch_exchange;
-use ethcontract::{contract::EventData, H256};
-use futures::future::{BoxFuture, FutureExt as _};
+use ethcontract::{contract::EventData, BlockNumber, H256};
+use futures::future::BoxFuture;
 use log::info;
 use serde::{Deserialize, Serialize};
-use std::collections::BTreeMap;
-use std::convert::TryFrom;
-use std::fs;
-use std::fs::File;
-use std::io::{Read, Write};
-use std::path::Path;
+use std::{
+    collections::BTreeMap,
+    convert::TryFrom,
+    fs,
+    fs::File,
+    io::{Read, Write},
+    ops::Bound,
+    path::Path,
+};
 
 // Ethereum events (logs) can be both created and removed. Removals happen if the chain reorganizes
 // and ends up not including block that was previously thought to be part of the chain.
 // However, the orderbook state (`State`) cannot remove events. To support this, we keep an ordered
 // list of all events based on which the state is built.
 
-#[derive(Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
 struct EventSortKey {
     block_number: u64,
     /// Is included to differentiate events from the same block number but different blocks which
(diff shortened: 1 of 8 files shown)


e00E:
> My main concern with this PR is that it introduces a lot of dead code that is likely to never see the light of day (such as the `PaginatedOrderbookReader`, which will probably never be used again). Maybe as a prelude we can also remove some of these older `StableXOrderbookReading` implementations that have been deprecated?

I haven't yet read the PR but I agree that we should remove them and thought about suggesting that myself before.

fleupold:
I'd be fine with killing the old paginated orderbook reading (I don't see why would want to support it going forward given how poorly it scales) before.

This way there would be much less "dead" code to write, right?

e00E:
Then let's remove the unused orderbook reading structs first and then we can adapt the trait to however makes it most convenient for the only remaining orderbook implementation which is the event based orderbook.

core/src/history/events.rs
```diff
@@ -195,19 +219,54 @@ impl TryFrom<&Path> for EventRegistry {
...
+
+    fn get_auction_data_for_block(
+        &self,
+        block: BlockNumber,
+    ) -> BoxFuture<Result<(AccountState, Vec<Order>)>> {
+        let (batch_id, block_number) = match block {
+            BlockNumber::Earliest => (BatchId(0), 0),
+            BlockNumber::Latest | BlockNumber::Pending => (BatchId::now(), u64::MAX),
+            // NOTE: Approximate the timestamp of the block by finding the batch
+            // ID of the last event before the specified block.
```
e00E, line 237: Let's put in a doc comment for this method that this doesn't always match what you expect from the documentation on the Trait. If there has been some time without events before this block then the correct batch id might be later than the batch id internally used here.

e00E (approved):

core/src/history/events.rs
nlordell: I was under the impression that doc comments don't do anything for trait methods (but I may be wrong).

core/src/history/events.rs
```diff
@@ -195,19 +219,60 @@ impl TryFrom<&Path> for EventRegistry {
...
+            }
+        };
+        immediate!(self.auction_state_for_batch_at_block(batch_id, block_number))
+    }
+}
+
+fn bounds_until_end_of_block(block_number: u64) -> (Bound<EventSortKey>, Bound<EventSortKey>) {
+    (
+        Bound::Unbounded,
+        Bound::Excluded(EventSortKey {
```
fleupold, line 263: Any reason to not use Included without the +1?

core/src/orderbook.rs
```diff
@@ -22,6 +23,15 @@ pub trait StableXOrderBookReading: Send + Sync {
         batch_id_to_solve: u32,
     ) -> BoxFuture<'a, Result<(AccountState, Vec<Order>)>>;
 
+    /// Returns the state of the open orderbook at the specified block number.
+    ///
+    /// The open orderbook contains orders and balances that are valid for the
+    /// current batch at the given block.
+    fn get_auction_data_for_block<'a>(
```
fleupold, line 30: For consistency we could rename the other method to `get_auction_data_for_batch`

core/src/history/events.rs
fleupold: This is unfortunate since it might lead to inconsistencies e.g. when there are future withdraw requests or orders expiring in the future. E.g. if we have an order valid until t and there is no events in t-1, t, t+1. Then if I request the orderbook for a timestamp in t+1 I might still see this order as valid since the batch I'm getting the data for would be t-1 at which point the order was still valid.

Since we are already async here, could we not get the timestamp of this block number to get an exact batchID?

fleupold (approved):
Looks good, although I'm slightly worried that the batchID approximation might lead to inconsistencies.

PR #1319, opened 2020-08-14
Introduce trait for mocking current time

e00E:
and show one example of how it would be used.

We've talked about this kind of trait before as an alternative to the "action pattern". What do you think of this implementation? I would like to use it in my refactor to get rid of the sleeping in stablex_driver and solution_submission.

### Test Plan
CI

### DIFF
--- a/core/src/driver/scheduler/system.rs
+++ b/core/src/driver/scheduler/system.rs
@@ -1,7 +1,9 @@
 use super::{AuctionTimingConfiguration, Scheduler};
-use crate::driver::stablex_driver::{DriverResult, StableXDriver};
-use crate::models::BatchId;
-use crate::util::FutureWaitExt as _;
+use crate::{
+    driver::stablex_driver::{DriverResult, StableXDriver},
+    models::BatchId,
+    util::{AsyncSleep, AsyncSleeping, DefaultNow, FutureWaitExt as _, Now},
+};
 use anyhow::{Context, Result};
 use crossbeam_utils::thread::Scope;
 use log::error;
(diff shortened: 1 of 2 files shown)


core/src/util.rs
```diff
@@ -67,6 +70,22 @@ pub trait AsyncSleeping: Send + Sync {
 pub struct AsyncSleep;
 impl AsyncSleeping for AsyncSleep {}
 
+#[cfg_attr(test, mockall::automock)]
+pub trait Now: Send + Sync {
```
nlordell, line 74: I see how this can be useful for testing things that rely on specific times.

nlordell (approved):
I think this is a useful trait for improving test coverage.

core/src/util.rs
fleupold: Agree. In this case maybe a Fake implementation with a `set_time` method rather than a mock (that requires slightly more involved expectation setup) could be convenient, but we can add this when we think it's useful.

core/src/driver/scheduler/system.rs
```diff
@@ -48,16 +50,15 @@ impl<'a> SystemScheduler<'a> {
...
-                    DriverResult::Ok | DriverResult::Skip(_) => break,
-                }
-            }
+            solve(
+                batch_id,
+                solver_deadline,
+                min_solution_submit_time,
+                driver,
+                &DefaultNow {},
+                &AsyncSleep {},
```
fleupold, line 59: nit: is using default more rusty here?

core/src/util.rs
fleupold, line 74: Could this trait even implement a default method that returns the DefaultNow for convenience?

fleupold (approved):
Nice change!

core/src/util.rs
nlordell: I think the module could implement a `default_timing() -> (DefaultNow, AsyncSleep)` method.

e00E: I've added a default_now function but not yet for async sleep because not sure how we want to structure this long term. Maybe we want to have 1 trait that contains all io operations or maybe we want to keep it separated.

PR #1320, opened 2020-08-14
Split run into two functions internally

e00E:
in preparation for splitting the trait into two functions.

### Test Plan
CI, no logic change

### DIFF
--- a/core/src/driver/stablex_driver.rs
+++ b/core/src/driver/stablex_driver.rs
@@ -90,30 +90,34 @@ impl<'a> StableXDriverImpl<'a> {
         &self,
         batch_to_solve: BatchId,
         latest_solution_submit_time: Duration,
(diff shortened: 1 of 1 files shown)


nlordell (approved):
:+1:

core/src/driver/stablex_driver.rs
```diff
@@ -240,15 +244,21 @@ impl<'a> StableXDriver for StableXDriverImpl<'a> {
...
                     orders,
                 )
                 .await
+            {
+                Ok(solution) => solution,
+                Err(err) => return DriverResult::Skip(err),
+            };
+            match self
+                .submit(batch_to_solve, earliest_solution_submit_time, solution)
+                .await
```
fleupold, line 261: nanonit: could `.map` here to avoid mid-code return

fleupold (approved):
noice

PR #1322, opened 2020-08-14
Handle timing of solution submission in scheduler

e00E:
by splitting StableXDriver into two functions. This avoids sleeping in stablex_driver and is more appropriate in the scheduler.

PR is big because we get to delete a lot of test code. I wasn't able to split it up more and I already did not include handling earliest_solution_submit_time in the evm scheduler here.

### Test Plan
CI, new test


fleupold (approved):
Nice, it looks like the driver tests got a good amount simpler thanks to this. Is there even a point in having the same struct implement solve and submit? Maybe at some future point it makes sense breaking this up into separate components.

nlordell (approved):
:tada:

PR #1323, opened 2020-08-14
Search by symbol

nlordell:
This PR adds the ability of to use symbols for currency pairs. Note that the symbols are retrieved from the EVM, so it is possible that names with duplicates (such as `DAI`) run into collisions and require updating our deployment.

### Test Plan

```
$ cargo run -p price-estimator -- --token-data '{"T0009":{"alias":"sUSD depr","decimals":18},"T0012":{"alias":"SAI","decimals":18},"T0015":{"alias":"SNX depr","decimals":18}}'
...
$ curl -s 'http://localhost:8080/api/v1/markets/WETH-DAI/estimated-buy-amount/1000?atoms=false' | jq
{
  "baseTokenId": 1,
  "quoteTokenId": 7,
  "buyAmountInBase": "2.3573254925937595",
  "sellAmountInQuote": "1000"
}
$ curl -s 'http://localhost:8080/api/v1/markets/ETH-DAI/estimated-buy-amount/1000?atoms=false' | jq # Bonus!
{
  "baseTokenId": 1,
  "quoteTokenId": 7,
  "buyAmountInBase": "2.3573254925937595",
  "sellAmountInQuote": "1000"
}
```

### DIFF
--- a/core/src/token_info.rs
+++ b/core/src/token_info.rs
@@ -1,9 +1,9 @@
+use crate::models::TokenId;
 use anyhow::Result;
 use futures::future::{BoxFuture, FutureExt as _};
 use lazy_static::lazy_static;
-use std::{collections::HashMap, num::NonZeroU128};
+use std::{borrow::Borrow, collections::HashMap, num::NonZeroU128};
 
-use crate::models::TokenId;
 pub mod cached;
 pub mod hardcoded;
 pub mod onchain;
@@ -46,12 +46,23 @@ pub trait TokenInfoFetching: Send + Sync {
     ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
         async move {
             let infos = self.get_token_infos(&self.all_ids().await?).await?;
-            Ok(infos
-                .into_iter()
-                .find(|(_, info)| info.matches_symbol(symbol)))
+            Ok(search_for_token_by_symbol(infos, symbol))
         }
         .boxed()
     }
+}
+
+fn search_for_token_by_symbol<T>(
+    tokens: impl IntoIterator<Item = (TokenId, T)>,
+    symbol: &str,
+) -> Option<(TokenId, T)>
+where
+    T: Borrow<TokenBaseInfo>,
+{
+    tokens
+        .into_iter()
+        .filter(|(_, info)| info.borrow().matches_symbol(symbol))
+        .min_by_key(|(id, _)| *id)
 }
 
 // mockall workaround https://github.com/asomers/mockall/issues/134
@@ -214,5 +225,24 @@ mod tests {
         assert_eq!(result.get(&TokenId(0)).unwrap().alias, "0");
         assert_eq!(result.get(&TokenId(1)).unwrap().alias, "1");
         assert!(result.get(&TokenId(2)).is_none());
+    }
+
+    #[test]
+    fn search_prefers_symbol_of_lower_token_ids() {
+        let owl = TokenBaseInfo {
+            alias: "OWL".to_owned(),
+            decimals: 18,
+        };
+
+        let (id, info) = search_for_token_by_symbol(
+            vec![
+                (TokenId(42), owl.clone()),
+                (TokenId(1337), owl.clone()),
+                (TokenId(0), owl.clone()),
+            ],
+            "OWL",
+        )
+        .unwrap();
+        assert_eq!((id, info), (TokenId(0), owl));
     }
 }
--- a/core/src/token_info.rs
+++ b/core/src/token_info.rs
@@ -66,6 +66,10 @@ mod mock {
             ids: &[TokenId],
         ) -> BoxFuture<'a, Result<HashMap<TokenId, TokenBaseInfo>>>;
         fn all_ids<'a>(&'a self) -> BoxFuture<'a, Result<Vec<TokenId>>>;
+        fn find_token_by_symbol<'a>(
+            &'a self,
+            symbol: &str,
+        ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>>;
     }
 
     impl<T: TokenInfoFetching_ + Send + Sync> TokenInfoFetching for T {
@@ -80,6 +84,12 @@ mod mock {
         }
         fn all_ids<'a>(&'a self) -> BoxFuture<'a, Result<Vec<TokenId>>> {
             TokenInfoFetching_::all_ids(self)
+        }
+        fn find_token_by_symbol<'a>(
+            &'a self,
+            symbol: &'a str,
+        ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
+            TokenInfoFetching_::find_token_by_symbol(self, symbol)
         }
     }
 }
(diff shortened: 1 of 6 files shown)


e00E (approved):
Code is approved.
I wonder how we can ensure that users get the correct token by symbol instead of a fake token with the same symbol. The current default implementation would prefer the first token based on token id with that symbol but the cache implementation takes the first token in the hashmap iterator with that symbol. This is random because hash map does not guarantee any ordering.
In practice this is not a problem right now, just mentioning it. In your example we change the symbol of older "deprecated" tokens but we can't necessarily keep that up to date when a token can be listed at any time.

nlordell:
> I wonder how we can ensure that users get the correct token by symbol instead of a fake token with the same symbol. ...  In your example we change the symbol of older "deprecated" tokens but we can't necessarily keep that up to date when a token can be listed at any time.

Good point. Maybe we should prefer lower token IDs.

nlordell:
@e00E I modified the implementation for the cached token infos to also prefer lower token IDs. This way we hopefully won't run into any issues because someone decides to list a second "DAI" token.

e00E (approved):

core/src/token_info/cached.rs
```diff
@@ -75,8 +74,22 @@ impl TokenInfoCache {
         ids.into_iter()
             .copied()
             .filter(|id| !cache.contains_key(id))
+            // NOTE: Make sure to `collect` to not hold the `cache` lock.
```
fleupold, line 77: Wow, that seems like a nasty one (I assume you were trying to return an iterator and ran in some kind of lock?). Should we maybe clone the cache to before iterating?

core/src/token_info/cached.rs
```diff
@@ -141,6 +154,24 @@ impl TokenInfoFetching for TokenInfoCache {
...
+    ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
+        async move {
+            // NOTE: First check the cache directly before refetching. This
+            // allows us to exit early without checking to see if there are
+            // missing token infos that we need to cache.
+            if let Some(result) = self.find_cached_token_by_symbol(symbol).await {
+                return Ok(Some(result));
+            }
+
+            self.cache_all(1).await?;
```
fleupold, line 170: Why 1 here (I'm not a huge fan of this unnamed argument at least when reviewing)?

core/src/token_info.rs
```diff
@@ -34,7 +34,24 @@ pub trait TokenInfoFetching: Send + Sync {
...
+    /// resulting token for the specified symbol.
+    fn find_token_by_symbol<'a>(
+        &'a self,
+        symbol: &'a str,
+    ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
+        async move {
+            let infos = self.get_token_infos(&self.all_ids().await?).await?;
+            Ok(infos
+                .into_iter()
+                .find(|(_, info)| info.matches_symbol(symbol)))
```
fleupold, line 51: Don't we have to break ties here too?

core/src/token_info/cached.rs
```diff
@@ -141,6 +154,24 @@ impl TokenInfoFetching for TokenInfoCache {
...
+
+    fn find_token_by_symbol<'a>(
+        &'a self,
+        symbol: &'a str,
+    ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
+        async move {
+            // NOTE: First check the cache directly before refetching. This
+            // allows us to exit early without checking to see if there are
+            // missing token infos that we need to cache.
+            if let Some(result) = self.find_cached_token_by_symbol(symbol).await {
```
fleupold, line 166: Are we guaranteed that the cache will be updated in order or at least atomically? Otherwise we might get different results at different times of the request (admittedly a very rare race condition, but also one that seems impossible to debug).

fleupold (approved):
I'm still slightly worried about the on chain fetching implementation. I think it would be nice if we could guarantee that all tokens up to a certain point are already fetched and then only consider tokens up to that point (to make sure we didn't e.g. cache a single entry that happens to be the fake DAI token).

core/src/token_info/cached.rs
nlordell: I'm OK with just the comment here. A unit test caught the issue so it appears we are covered :smile:

core/src/token_info/cached.rs
nlordell: I'll name it! Its the number of concurrent requests.

core/src/token_info.rs
nlordell: :facepalm:

core/src/token_info/cached.rs
nlordell: > I'd be fine with killing the old paginated orderbook reading

I think we do a `cache_all` on startup, but we can guarantee it.

nlordell: Hmm, yes this is a tough one...

nlordell: I modified the implementation to ensure that all tokens up to the found token are cached. This makes it so the results are reproducible even as new tokens get added to the exchange while still having a "fast-path" so that we don't query the EVM once the cache is sufficiently warmed up.

core/src/token_info/cached.rs
```diff
@@ -141,6 +153,32 @@ impl TokenInfoFetching for TokenInfoCache {
...
     }
+
+    fn find_token_by_symbol<'a>(
+        &'a self,
+        symbol: &'a str,
+    ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
+        const CONCURRENT_REQUESTS: usize = 1;
+
+        async move {
+            if let Some((id, _)) = self.find_cached_token_by_symbol(symbol).await {
```
fleupold, line 164: Smart way to solve this tricky issue.

core/src/token_info/cached.rs
```diff
@@ -141,6 +153,32 @@ impl TokenInfoFetching for TokenInfoCache {
     fn all_ids<'a>(&'a self) -> BoxFuture<'a, Result<Vec<TokenId>>> {
         self.inner.all_ids()
     }
+
+    fn find_token_by_symbol<'a>(
+        &'a self,
+        symbol: &'a str,
+    ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
+        const CONCURRENT_REQUESTS: usize = 1;
```
fleupold, line 161: Any reason for using a different value than [here](https://github.com/gnosis/dex-services/blob/b60f00f8ddf517cab6ae12cd61e46f9e24d024d3/price-estimator/src/main.rs#L122)

fleupold (approved):
🔣

core/src/token_info/cached.rs
nlordell: No, in my head I thought "since we cache all tokens at startup anyway, I don't expect having concurrent requests will be much benefit" and "this way the implementation is similar to the default one".

But now that I think about it, maybe we can use a similar value to the code you linked.

nlordell: They use the same value now.

PR #1325, opened 2020-08-14
Replace crossbream scoped thread with asycnstd task

e00E:
This involves converting several structs to store Arcs instead of references. This is more code in terms of lines but also simplifies lifetimes so imo it's an improvement.

### Test Plan
CI and `cargo test -p core system::tests::test_real -- --ignored --nocapture`

### DIFF
--- a/core/src/driver/scheduler.rs
+++ b/core/src/driver/scheduler.rs
@@ -7,8 +7,7 @@ use crate::contracts::stablex_contract::StableXContract;
 use crate::driver::stablex_driver::StableXDriver;
 use crate::models::batch_id::SOLVING_WINDOW;
 use anyhow::{anyhow, Error, Result};
-use std::str::FromStr;
-use std::time::Duration;
+use std::{str::FromStr, sync::Arc, time::Duration};
 
 /// A scheduler that can be started in order to run the driver for each batch.
 pub trait Scheduler {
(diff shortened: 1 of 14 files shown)


core/src/driver/scheduler/system.rs
```diff
@@ -37,28 +37,21 @@ impl<'a> SystemScheduler<'a> {
...
-        'a: 'b,
-    {
-        let driver = self.driver;
+    fn start_solving_in_background(&self, batch_id: BatchId, solver_deadline: Instant) {
+        let driver = self.driver.clone();
         let earliest_solution_submit_time = self
             .auction_timing_configuration
             .earliest_solution_submit_time;
-        scope.spawn(move |_| {
+        async_std::task::spawn(async move {
```
e00E, line 45: This is the "real" change

nlordell:
> This is more code in terms of lines but also simplifies lifetimes so imo it's an improvement.

I would be in favour of just using `Arc`s everywhere as well. We have a lot of shared components so in this case it makes sense.

core/src/driver/scheduler.rs
```diff
@@ -94,12 +93,12 @@ pub enum SchedulerKind {
 
 impl SchedulerKind {
     /// Creates a new scheduler based on the parameters.
-    pub fn create<'a>(
+    pub fn create(
         &self,
-        exchange: &'a dyn StableXContract,
-        driver: &'a (dyn StableXDriver + Sync),
+        exchange: Arc<dyn StableXContract + Send + Sync>,
+        driver: Arc<dyn StableXDriver + Send + Sync>,
```
nlordell, line 99: Maybe all our components that need to be shared across threads should be `trait Component: Send + Sync` because we need them to be like this pretty much everywhere.

core/src/driver/scheduler/system.rs
nlordell, line 45: Does this task need to be `await`-ed anywhere or is it automatically driven by the `async_std` runtime?

nlordell:
> Apart from that, is there a functional reason for async-std tasks over crossbeam scopes?

They can be `await`-ed instead of blocking until all spawned threads finish. While it doesn't really have much of an effect here, it can make future refactoring easier. It also means everything is running in our `async-std` executor instead throwing a bit of threading in the mix.

core/src/driver/scheduler/system.rs
e00E: It's like spawning a thread so it doesn't need to be awaited. It can be awaited through a join handle but it doesn't have to.

Issue #949, opened 2020-06-03
[Solver-team] Building graphana solver dashboard


josojo:
The solver team started to write additional information into the solution json.
They look like this:
```
"objVals": {
        "volume": 0,
        "utility": 0,
        "utility_disreg": 0,
        "utility_disreg_touched": 0,
        "fees": 0,
        "orders_touched": 0
    },
    "solver": {
        "name": "standard",
        "args": [
            "/app/instances/2020-06-03/instance_5303807_2020-06-03T00:01:02.158704801+00:00.json",
            "--outputDir=/app/results/2020-06-03/instance_5303807_2020-06-03T00:01:02.158704801+00:00/",
            "--logging=WARNING",
            "--solverTimeLimit",
            "118",
            "--minAvgFeePerOrder=10000000000000000",
            "--solver=SCIP",
            "--useExternalPrices",
            "--solveBestCycle"
        ],
        "runtime": 51.208314418792725,
        "nr_variables": 788,
        "nr_bool_variables": 195,
        "optimality_gap": 1.958399072019926e-15,
        "solver_status": "ok",
        "termination_condition": "optimal",
        "exit_status": "completed",
        "obj_val": 0.0,
        "obj_val_sc": 0.0
    }
```

These parts of this information should be parsed in the driver so that we can build a dashboard showing the execution times, optimality gaps, and obj_val_sc.
cc @twalth3r

e00E:
Created initial dashboard.

PR #1328, opened 2020-08-16
Solver metrics: Dirty start

josojo:
This is the handover PR for #949 , since I am leaving for vacation.

My previous work is really just a rough start. Nothing is thought through and basically, I was just trying to get a very basic Prometheus logging started. The idea I followed was to parse the solver output into a "Solution" and "SolutionInfo" struct, in order to keep the usual processing on the "Solution" object for the solution submission and use the "SolutionInfo" struct to write the Prometheus metrics for the solver.

Probably, this work will not help anyone ;) It's really a pity that I never found the time to start this properly. I would have been such nice tasks.
 Feel free to close this PR any time...


e00E:
Closing because I'll recreate the PR.

PR #1332, opened 2020-08-18
remove paginated orderbook reader

nlordell:
This PR removes the paginated orderbook reader. This is because the on-chain filtered implementation is better in every way:
- Also EVM based
- Does not suffer from the same limitations as the paginated implementation

This way we no longer need to maintain an extra implementation that we arguably have no reason to use. Note that we also removed the `PaginatedAuctionDataReader` since it was no longer being used and clippy complained.

The question is, do you guys think we should also remove the:
1) Shadowed orderbook
2) Onchain filter orderbook

### Test Plan

Mostly removed code, adjusted unit tests where necessary.


e00E (approved):

e00E:
> The question is, do you guys think we should also remove the:

Imo we should remove both of them too because the onchain filtered orderbook is also too slow to be feasible and when it is gone there is no reason to have a shadowed orderbook.

nlordell:
> onchain filtered orderbook is also too slow to be feasible and when it is gone there is no reason to have a shadowed orderbook.

I agree, the only thing holding me back is if we decide to change the event-based orderbook implementation and want to verify that it works. Arguably, we should have unit tests to catch these mistakes but I fear they are hard to identify.

e00E:
Even in that case I don't think we will be able to verify it live as we used to with the shadowed orderbook anyway, because it takes too long. So it would be more of an e2e test binary and doesn't need to interact with the traits. Since it is history we can create that kind of test if we need it and remove them for now.

fleupold (approved):
🔴 

I'm fine with removing everything that we no longer need. It's not that we cannot bring it back from version control and since it was very close to the SC it shouldn't be much work to rebuild.

PR #1333, opened 2020-08-18
Print default backtrace in additon to error log on panic

e00E:
The log message does not include the backtrace (only the last location)
but it is useful to see a full backtrace.

This is probably not optimal because kibana might jumble the backtrace if it doesn't come from a logger but still better than no backtrace at all.

### Test Plan
`env RUST_BACKTRACE=1 cargo test -p core panic_is_printed -- --ignored --nocapture`

### DIFF
--- a/core/src/logging.rs
+++ b/core/src/logging.rs
@@ -1,10 +1,9 @@
 use chrono::Utc;
-use slog::Level;
-use slog::{o, Drain, Logger, OwnedKVList, Record};
+use slog::{b, o, record, Drain, Level, Logger, OwnedKVList, Record};
 use slog_async::{Async, OverflowStrategy};
 use slog_envlogger::LogBuilder;
 use slog_scope::GlobalLoggerGuard;
-use slog_term::{Decorator, TermDecorator};
+use slog_term::{Decorator, PlainDecorator, TermDecorator};
 use std::{
     panic::{self, PanicInfo},
     thread,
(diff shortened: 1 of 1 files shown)


core/src/logging.rs
```diff
@@ -32,14 +32,18 @@ pub fn init(filter: impl AsRef<str>) -> (Logger, GlobalLoggerGuard) {
...
-    fn hook(info: &PanicInfo) {
+    let default_hook = panic::take_hook();
+    let hook = move |info: &PanicInfo| {
         let thread = thread::current();
         let thread_name = thread.name().unwrap_or("<unnamed>");
 
+        // It is not possible for our custom hook to print a full backtrace on stable rust. To not
+        // lose this information we call the default panic handler which prints the full backtrace.
+        default_hook(info);
         log::error!("thread '{}' {}", thread_name, info);
```
nlordell, line 45: Maybe we should log first? This way I imagine that on Kibana the parser will attach the panic message the the error log document.

e00E: I don't think that guarantees anything because we print log messages at some point later in time in another thread. Maybe we could print a fake log message prefix / separator before the panic though so it's treated as one message in kibana.

nlordell: > I don't think that guarantees anything because we print log messages at some point later in time in another thread.

Oh yeah!

> Maybe we could print a fake log message prefix / separator before the panic though so it's treated as one message in kibana.

Great idea.

core/src/logging.rs
```diff
@@ -32,14 +32,21 @@ pub fn init(filter: impl AsRef<str>) -> (Logger, GlobalLoggerGuard) {
...
+    let default_hook = panic::take_hook();
+    let hook = move |info: &PanicInfo| {
         let thread = thread::current();
         let thread_name = thread.name().unwrap_or("<unnamed>");
 
+        // It is not possible for our custom hook to print a full backtrace on stable rust. To not
+        // lose this information we call the default panic handler which prints the full backtrace.
+        // We print a fake log message prefix so that kibana can identify that this is supposed to
+        // a single message.
+        eprint!("{} ERRO [core::logging] ", formatted_current_time());
```
nlordell, line 46: any chance we can reuse the `log_to_decorator` method? At least, I would say we should format the log level so we are bit more future proof:
```suggestion
        eprint!("{} {} [core::logging] ", formatted_current_time(), log::Level::Error);
```

nlordell: I think we should be able to use [`PlainDecorator<Vec<u8>>`](https://docs.rs/slog-term/2.6.0/slog_term/struct.PlainDecorator.html)

e00E: implemented

nlordell (approved):
Thanks! This looks really good.

fleupold (approved):

PR #1336, opened 2020-08-18
Fix panic caused by wait() instead of await

e00E:
The async fn should not be using wait(). Crashes because it is not
allowed to wait() in a future that is itself wait()ed.

[Optional] 
Fixes #issue

*<Context of what this change does, why it is needed and how it is accomplished>*

### Test Plan
*<Explain how you and a reviewer have/intend to test this change>*


e00E:
Since this is such a simple fix I will force merge before CI. I ran tests locally.

PR #1338, opened 2020-08-18
delay health ready notification until start of second batch

nlordell:
This PR notifies the health reporter that the driver is ready at the start of the second batch.

I thought about it for a while, and I couldn't think of a way of implementing this that I was 100% happy with. I think eventually, the scheduler should be responsible for notifying the health reporter that it is "ready" once the second batch has started. The issue there is that there seems to be quite a bit of duplicated code :cry:. I implemented it this quick and dirty way for now, but I'm thinking of moving this logic into the scheduler unless there are any objections.

### Test Plan

Added unit test for delayed message. Also, run the driver and check the `/health/readiness` endpoint.

TODO: add the commands and output.


fleupold (commented):
Should we even prevent it from solving the current batch? Otherwise I'm afraid it might find a solution and result in some weird race conditions with the other solver that is using the same PK and potentially solver licence.

e00E (approved):

e00E:
> Should we even prevent it from solving the current batch? Otherwise I'm afraid it might find a solution and result in some weird race conditions with the other solver that is using the same PK and potentially solver licence.

I think this makes sense. We can start the scheduler and signal readiness at this time point.

nlordell:
Closing this as @e00E is working on it.

PR #1339, opened 2020-08-18
Move solution submission batch id check to scheduler

### LINKED ISSUE #707: [Cleanup] Remove sleep in solution submission
In order to silence some warning when solutions are submitted right away, a sleep was added to wait for a block in the batch where the solution is being accepted.

Here is the offending line:
https://github.com/gnosis/dex-services/blob/a2589b5beb4b2cd9ebf45d272765192e417ce404/driver/src/solution_submission/mod.rs#L110

See here for some discussion on solutions to fix this:
https://github.com/gnosis/dex-services/pull/687#discussion_r404188080

> **nlordell**: I just had an idea. We can make "Solutions are no longer accepted for this batch" a benign error if we are in the previous block and have the scheduler do the sleep and re-run the driver.
> The downside here is that it will re-fetch the orderbook and look for a solution again, but I don't think this should be a big issue.

> **fleupold**: That's a nice idea. We briefly discussed it over slack. It would require fetching the last block before verifying and then return a benign retryable error.
> I'm somewhat concerned about running the solver again unnecessarily, but if we run into this it's already a sign that we were finishing too early (however this might become the norm if we start solving before the batch is finalized). We could also introduce a caching layer for the price finder that can quickly return the same result if the input files are the same.

> **nlordell**:
> "however this might become the norm if we start solving before the batch is finalized"
> I imagine at that point, the scheduler would be responsible for starting the solution submitter.
> "We could also introduce a caching layer for the price finder that can quickly return the same result if the input files are the same."
> That would be useful, especially when facing retryable errors.

e00E:
It is more appropriate there. The evm scheduler already knows that the
batch id is correct so no new code is added there.

[Optional] 
Fixes #707
*<Context of what this change does, why it is needed and how it is accomplished>*

### Test Plan
CI, new unit test

### DIFF
--- a/core/src/driver/scheduler.rs
+++ b/core/src/driver/scheduler.rs
@@ -100,7 +100,7 @@ impl SchedulerKind {
         config: AuctionTimingConfiguration,
     ) -> Box<dyn Scheduler> {
         match self {
-            SchedulerKind::System => Box::new(SystemScheduler::new(driver, config)),
+            SchedulerKind::System => Box::new(SystemScheduler::new(exchange, driver, config)),
             SchedulerKind::Evm => Box::new(EvmScheduler::new(exchange, driver, config)),
         }
     }
--- a/core/src/driver/scheduler/system.rs
+++ b/core/src/driver/scheduler/system.rs
@@ -1,5 +1,6 @@
 use super::{AuctionTimingConfiguration, Scheduler};
 use crate::{
+    contracts::stablex_contract::StableXContract,
     driver::stablex_driver::{DriverError, StableXDriver},
     models::{BatchId, Solution},
     util::{self, AsyncSleep, AsyncSleeping, Now},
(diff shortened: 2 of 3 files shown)


core/src/driver/scheduler.rs
```diff
@@ -100,7 +100,7 @@ impl SchedulerKind {
         config: AuctionTimingConfiguration,
     ) -> Box<dyn Scheduler> {
         match self {
-            SchedulerKind::System => Box::new(SystemScheduler::new(driver, config)),
+            SchedulerKind::System => Box::new(SystemScheduler::new(driver, exchange, config)),
             SchedulerKind::Evm => Box::new(EvmScheduler::new(exchange, driver, config)),
```
nlordell, line 104: uber-nano-nit: `driver` and `exchange` parameters are in switched orders for the two schedulers. Personally I think it would be nice if they were the same.

nlordell (approved):
1000 times yes!

fleupold (approved):
Yes, very nice change.

PR #1343, opened 2020-08-19
Handle earliest solution submit time in evm scheduler

e00E:
I decided to use mock async sleep because this makes the running code simpler.
The alternative is to make `step()` use a state machine which makes the tests simpler but the running code more complex.

### Test Plan
New unit test, CI

### DIFF
--- a/core/src/driver/scheduler/evm.rs
+++ b/core/src/driver/scheduler/evm.rs
@@ -2,10 +2,12 @@
 //! batch duration information directly from the EVM instead of system time.
 
 use super::{AuctionTimingConfiguration, Scheduler};
-use crate::contracts::stablex_contract::StableXContract;
-use crate::driver::stablex_driver::{DriverError, StableXDriver};
-use crate::models::batch_id::BATCH_DURATION;
-use crate::util::FutureWaitExt as _;
+use crate::{
+    contracts::stablex_contract::StableXContract,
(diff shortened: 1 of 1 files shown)


core/src/driver/scheduler/evm.rs
```diff
@@ -28,26 +31,34 @@ impl EvmScheduler {
...
+            sleep,
+            config: AuctionTimingConfiguration::default(),
+            last_batch: None,
+        }
     }
 
     /// Runs the scheduler for a single iteration.
-    fn step(&mut self) -> Result<()> {
-        let batch_id = self.current_solving_batch()?;
+    async fn step(&mut self) -> Result<()> {
```
nlordell, line 60: One other alternative is to make `step() -> Result<Option<Solution>>` and then move the sleep into the `Scheduler::start` loop. This would effectively split `step` into a `run` and `submit` methods. Also if we are going to use `AsyncSleeping` then maybe we should have `step` just wait for the next batch and wait to submit instead of sleeping in two different places with two different methods (`thread::sleep` and `AsyncSleeping::sleep`).

fleupold: I iike the proposal of splitting the functions and having the sleep in one place. This makes it a bit closer to a state machine while presumably makes the factoring simpler as we break the large step function into two.

PR #1345, opened 2020-08-20
Solver metrics

e00E:
The solver outputs some information that we would like to forward to prometheus.

Part of #949 

### Test Plan
CI, and queried metrics after startup to see that they are all 0. Will also run with open solvers to see that it works.


fleupold (approved):
👍 Looks very good. Thanks for tackling this long-time outstanding task...

nlordell (approved):
Awesome!

PR #1348, opened 2020-08-20
make /markets and /estimated-best-ask-price return consistent prices

### LINKED ISSUE #1344: /markets and markets/.../estimated-best-ask-price return inconsistent prices
Exaggerating on structure here since these price issues can easily be overwhelming...

## Issue

`/markets` returns prices in quote, e.g. https://price-estimate-mainnet.dev.gnosisdev.com/api/v1/markets/WETH-DAI shows prices in DAI

> ... {"price":404.67014059206866,"volume":246744853556.0} ...

`markets/.../estimated-best-ask-price` however shows prices in base, e.g. https://price-estimate-mainnet.dev.gnosisdev.com/api/v1/markets/WETH-DAI/estimated-best-ask-price?unit=atoms returns prices in WETH 

> 0.002295201341664352

despite the comment in the code saying otherwise:

https://github.com/gnosis/dex-services/blob/03075b0c419f6a69f6ef6f62e37db37e0f4eb625/price-estimator/src/models.rs#L64-L68

## Why is that?

I believe this comes from the fact that we use the bid pair to estimate the best ask price. This itself is correct as we want to "simulate" a person trying to buy a tiny amount of base from the people offering it to get the "best offer" (best ask):

https://github.com/gnosis/dex-services/blob/03075b0c419f6a69f6ef6f62e37db37e0f4eb625/price-estimator/src/filter.rs#L310-L315

In the snippet below we can see that the returned price is given in buy/sell token and thus in base/quote (in the example WETH per $) rather than quote/base (in the example $ per WETH). The latter is what the orderbook is showing.

https://github.com/gnosis/dex-services/blob/03075b0c419f6a69f6ef6f62e37db37e0f4eb625/pricegraph/src/api/price_estimation.rs#L14-L16

## Proposed Fix

I believe it is fine to just invert the price in the filter, as fees are accounted in a way to make the buy amount smaller (lowering best the best price available in base). Inverting this would lead to increasing the best ask. This makes the spread larger than in a no-fee model which seems reasonable.

@e00E @nlordell could you please double check if my thinking is correct and if it's safe to just invert the price before creating the result type:

https://github.com/gnosis/dex-services/blob/03075b0c419f6a69f6ef6f62e37db37e0f4eb625/price-estimator/src/filter.rs#L315-L318

Note, that I don't think this has anything to do with base-units or atoms, but I noticed this behavior when coding the base unit support for the best ask route.

fleupold:
Fixes #1344

For more context see discussion in the issue. I added a long comment in the updated unit test justifying why the logic is now correct if the price is in quote.

Alternatively we could also just invert the price after it has been turned in the PriceEstimateResult type. Generally it would probably be nice to have some kind of meta information about what is base and what is quote in the result type (similar to how we have it for 

https://dex-price-estimator.gnosis.io/api/v1/markets/11-4/estimated-buy-amount/100000000?atoms=true

> {"baseTokenId":11,"quoteTokenId":4,"buyAmountInBase":"837137","sellAmountInQuote":"100000000"}

### Test Plan
Adjusted unit test

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -312,7 +312,9 @@ async fn estimate_best_ask_price(
         .pricegraph(query.time, PricegraphKind::WithRoundingBuffer)
         .await
         .map_err(RejectionReason::InternalError)?
-        .estimate_limit_price(token_pair, 0.0);
+        .estimate_limit_price(token_pair, 0.0)
+        // The price above is in base, but we need to return it in quote.
+        .map(|p| 1.0 / p);
 
     let result = PriceEstimateResult(price);
     let result = match query.unit {
--- a/price-estimator/src/models.rs
+++ b/price-estimator/src/models.rs
@@ -74,7 +74,7 @@ impl PriceEstimateResult {
         quote_token_info: &TokenBaseInfo,
     ) -> Self {
         Self(self.0.map(|p| {
-            p * 10f64.powi(quote_token_info.decimals as i32 - base_token_info.decimals as i32)
+            p / 10f64.powi(quote_token_info.decimals as i32 - base_token_info.decimals as i32)
                 as f64
         }))
     }
@@ -144,12 +144,15 @@ mod tests {
             decimals: 6,
         };
 
-        let one_owl = 10_u64.pow(18);
-        let one_usdc = 10_u64.pow(6);
-
-        let price_estimate = PriceEstimateResult(Some(one_owl as f64 / one_usdc as f64));
+        // Since OWL has 18 decimals, and USDC 6, we get 10^12 OWL per USDC (counterintuitively denoted as USDC/OWL
+        // with OWL being the quote token) From https://www.investopedia.com/terms/c/currencypair.asp:
+        // > The quotation EUR/USD = 1.2500 means that one euro is exchanged for 1.2500 U.S. dollars. In this case,
+        // EUR is the base currency and USD is the quote currency (counter currency). This means that 1 euro can be
+        // exchanged for 1.25 U.S. dollars.
+        let owl_per_usdc = 10_u64.pow(12);
+        let price_estimate = PriceEstimateResult(Some(owl_per_usdc as f64));
         assert_approx_eq!(
-            price_estimate.into_base_units(&owl, &usdc).0.unwrap(),
+            price_estimate.into_base_units(&usdc, &owl).0.unwrap(),
             1.0f64
         )
     }


e00E (approved):

nlordell:
I'm still confused about using the bid pair to estimate an ask order price.

price-estimator/src/filter.rs
```diff
@@ -312,7 +312,9 @@ async fn estimate_best_ask_price(
         .pricegraph(query.time, PricegraphKind::WithRoundingBuffer)
         .await
         .map_err(RejectionReason::InternalError)?
-        .estimate_limit_price(token_pair, 0.0);
+        .estimate_limit_price(token_pair, 0.0)
+        // The price above is in base, but we need to return it in quote.
+        .map(|p| 1.0 / p);
```
nlordell, line 317: So one thing to watch out for is that inverting prices is risky because of fees.

nlordell (commented):
> I believe this comes from the fact that we use the bid pair to estimate the best ask price. This itself is correct as we want to "simulate" a person trying to buy a tiny amount of base from the people offering it to get the "best offer" (best ask):

Yes this makes sense. I guess we should be using the best-bid order here.

nlordell (approved):
I am convinced this is the correct fix! Thanks for the huge write-up with the issue - thinking about prices always mess with my mind.

So I also went a little far with testing this change. I applied the following to ensure that the `market` route was using the same orderbook as the `estimated-best-ask-price` one.
```diff
diff --git a/price-estimator/src/filter.rs b/price-estimator/src/filter.rs
index c258f43..c779b39 100644
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -309,7 +309,7 @@ async fn estimate_best_ask_price(
     let market = get_market(pair, &*token_infos).await?;
     let token_pair = market.bid_pair();
     let price = orderbook
-        .pricegraph(query.time, PricegraphKind::WithRoundingBuffer)
+        .pricegraph(query.time, PricegraphKind::Raw)
         .await
         .map_err(RejectionReason::InternalError)?
         .estimate_limit_price(token_pair, 0.0)
diff --git a/pricegraph/src/api/transitive_orderbook.rs b/pricegraph/src/api/transitive_orderbook.rs
index 9f36da3..0d3dd85 100644
--- a/pricegraph/src/api/transitive_orderbook.rs
+++ b/pricegraph/src/api/transitive_orderbook.rs
@@ -55,7 +55,7 @@ impl Pricegraph {
     ///
     /// This method panics if the spread is zero or negative.
     pub fn transitive_orderbook(&self, market: Market, spread: Option<f64>) -> TransitiveOrderbook {
-        let mut orderbook = self.full_orderbook();
+        let mut orderbook = self.reduced_orderbook().into_inner();
 
         let mut transitive_orderbook = TransitiveOrderbook::default();
         while let Some(Ring { ask, bid }) = orderbook.fill_market_ring_trade(market) {
```

Then checked service responses:
```
$ curl -s http://localhost:8080/api/v1/markets/1-7/estimated-best-ask-price | jq
377.45727709691675
$ curl -s http://localhost:8080/api/v1/markets/1-7 | jq '.asks[0].price' 
377.0798198198199
$ echo $((377.0798198198199 * 1.001))
377.45689963963963
```

`377.45727709691675 ~= 377.45689963963963`

And these numbers seem to agree. Note that the price for the transitive order is lower than the best ask price, this is expected because the transitive order's "price" is actually an exchange rate and not a limit price. So to match that order you need to add fees one more time. As for the small error, I am unsure why. I imagine it has to do with applying fees before inverting vs. after inverting.

Either way, thanks for figuring this out!

PR #1356, opened 2020-08-24
fetch batch for block number

nlordell:
This PR modifies the `UpdatingOrderbook` implementation to read the timestamp for the block number so that the batch ID is more accurate.

### Test Plan

`EventRegistry` is tested by unit tests. Unfortunately the `UpdatingOrderbook` does not have unit tests and it does not seem trivial to set one up. Fortunately, only very little logic was added there (`updating_orderbook.rs`) so please review with extra attention. Suggestions for a better test plan are welcome!

### DIFF
--- a/core/src/history/events.rs
+++ b/core/src/history/events.rs
@@ -161,8 +161,13 @@ impl EventRegistry {
     pub fn auction_state_for_batch_at_block(
         &self,
         batch_id: impl Into<BatchId>,
-        block_number: u64,
+        block: impl Into<BlockNumber>,
     ) -> Result<(AccountState, Vec<Order>)> {
+        let block_number = match block.into() {
+            BlockNumber::Earliest => 0,
+            BlockNumber::Latest | BlockNumber::Pending => u64::MAX,
+            BlockNumber::Number(block_number) => block_number.as_u64(),
+        };
         auction_state_for_batch_from_events(
             batch_id,
             self.events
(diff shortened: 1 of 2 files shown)


e00E (approved):

core/src/history/events.rs
```diff
@@ -236,24 +241,20 @@ impl StableXOrderBookReading for EventRegistry {
...
-                    .unwrap_or(BatchId(0));
-                (batch_id, block_number)
-            }
+            BlockNumber::Number(block_number) => self
+                .events
+                .range(bounds_until_end_of_block(block_number.as_u64()))
+                .rev()
+                .next()
+                .map(|(_, Value { batch_id, .. })| *batch_id)
+                .unwrap_or(BatchId(0)),
```
fleupold, line 255: We also want to replace this logic by looking up the batch number from the block-timestamp, right? But that will be a different PR I assume?

fleupold (approved):
I think a nice unit test would be the case where we want to get the orderbook for some block that has a timestamp far after the last recorded event. If that last event was e.g. placing an order that is valid for one batch, in the old system we would have seen the order in the returned state. 
With this change, we should no longer see this order in the results as it is no longer valid and we actually query the timestamp from the queried block.

But I can see how this test might be hard to set up.

core/src/history/events.rs
nlordell: That would require modifying the `EventRegistry` to have access to a `BlockTimestampReading` instance which is a pretty significant change. If we decide to do it, then it should be a separate PR.

That being said, I don't know how I feel about that change - currently the `EventRegistry` has a "pure" and nicely encapsulated logic for computing auction states just from past events, and adding a `BlockTimestampReading` would change that. I would argue that `UpdatingOrderbook` should be the "side-effect-y" implementation that updates the `EventRegistry` periodically and can query block timestamps and that `EventRegistry` doesn't necessarily need to implement `StableXOrderBookReading` as it has synchronous `auction_state_for_batch` and `acution_state_for_batch_at_block` methods that aren't async and build an auction state purely from past events.

nlordell:
> But I can see how this test might be hard to set up.

The issue is that tests are hard to setup for `UpdatingOrderbook` :cry:. I'll spend some time tomorrow and try to come up with something.

core/src/history/events.rs
fleupold: As discussed offline, maybe it's best if we unimplement this method for now, instead of having an incorrect implementation on which we might rely in the future...

PR #1360, opened 2020-08-25
Refactor evm scheduler

e00E:
Split up the step function into several parts that prepare it for the
health check and earliest solution submit time.

`step` no longer returns to the caller (`start`) so that sleeps are handled there. Instead it solves a full batch by waiting for the next batch to start, solving and submitting the solution.

### Test Plan
CI.

### DIFF
--- a/core/src/driver/scheduler/evm.rs
+++ b/core/src/driver/scheduler/evm.rs
@@ -24,7 +24,6 @@ pub struct EvmScheduler {
     driver: Arc<dyn StableXDriver>,
     config: AuctionTimingConfiguration,
     sleep: Box<dyn AsyncSleeping>,
-    last_batch: Option<u32>,
 }
 
 impl EvmScheduler {
@@ -39,7 +38,6 @@ impl EvmScheduler {
             exchange,
             config,
             sleep: Box::new(AsyncSleep),
-            last_batch: None,
         }
     }
 
@@ -55,7 +53,6 @@ impl EvmScheduler {
             exchange,
             config: AuctionTimingConfiguration::default(),
             sleep,
-            last_batch: None,
         }
     }
 
@@ -91,7 +88,7 @@ impl EvmScheduler {
             .checked_sub(batch_time))
     }
 
-    async fn solve(&mut self, batch_id: u32) -> Result<Option<Solution>> {
+    async fn solve(&self, batch_id: u32) -> Result<Option<Solution>> {
         while let Some(time_limit) = self.solver_time_limit(batch_id).await? {
             info!(
                 "solving for batch {} with time limit {}s",
@@ -128,17 +125,14 @@ impl EvmScheduler {
     }
 
     /// Wait for and solve the next batch.
-    async fn step(&mut self) -> Result<u32> {
-        let last_batch = match self.last_batch {
+    async fn step(&self, last_batch: Option<u32>) -> Result<u32> {
+        let last_batch = match last_batch {
             None => self.current_solving_batch().await?,
             Some(batch) => batch,
         };
         let new_batch = self.wait_for_batch_to_change(last_batch).await?;
         // TODO: signal healthiness
         let solution = self.solve(new_batch).await?;
-        // The batch is marked as handled so that the next call to step does not handle it again.
-        // In the next call `step` will sleep until the batch id changes.
-        self.last_batch = Some(new_batch);
         if let Some(solution) = solution {
             self.submit(new_batch, solution).await;
         }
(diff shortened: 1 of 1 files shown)


core/src/driver/scheduler/evm.rs
```diff
@@ -32,110 +38,119 @@ impl EvmScheduler {
...
-                batch_id, verify_batch_id,
-            );
-            return Ok(());
+    async fn wait_for_batch_to_change(&self, batch: u32) -> Result<u32> {
+        loop {
+            let current_batch = self.current_solving_batch().await?;
+            if current_batch != batch {
+                return Ok(current_batch);
+            }
+            self.sleep.sleep(POLL_TIMEOUT).await;
```
fleupold, line 76: Should this be a separate component that we can mock inside unit tests? Or can we drive it by mocking the contract directly?

fleupold (approved):

core/src/driver/scheduler/evm.rs
```diff
@@ -32,110 +38,119 @@ impl EvmScheduler {
...
+    /// Wait for and solve the next batch.
+    async fn step(&mut self) -> Result<u32> {
+        let last_batch = match self.last_batch {
+            None => self.current_solving_batch().await?,
+            Some(batch) => batch,
+        };
+        let new_batch = self.wait_for_batch_to_change(last_batch).await?;
+        // TODO: signal healthiness
+        let solution = self.solve(new_batch).await?;
+        self.last_batch = Some(new_batch);
```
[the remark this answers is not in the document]
e00E: I've added a comment in the code:
        // The batch is marked as handled so that the next call to step does not handle it again.
        // In the next call `step` will sleep until the batch id changes.

core/src/driver/scheduler/evm.rs
e00E: It's already being tested so it doesn't need to be a mockable trait. `scheduler_only_runs_next_batch_after_previous_has_finished`

core/src/driver/scheduler/evm.rs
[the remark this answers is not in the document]
fleupold: Would it be possible to return batchID from step and pass it back into the step function from above?

I'd prefer that over storing state on the instance as it would make it clear that no other part of the code relies on or mutates it.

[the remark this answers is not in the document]
e00E: I changed this. The reason for having it was that the code before this PR would retry solving from `start` if `step` returned an error before finishing solving but not during solution submission. Trying to keep this behavior while removing the field is awkward because we would have to return the batch id in both the ok and error case in which case we wouldn't be able to use `?` in `step`.

However, I now realize that the original behavior before splitting stablex_driver to have a solve and submit method was to not retry in this case. This is also how the system scheduler works. So I changed the behavior back to that which allows us to make this change here.

core/src/driver/scheduler/evm.rs
```diff
@@ -32,110 +37,118 @@ impl EvmScheduler {
...
-            warn!("skipping batch {}", batch_id);
-            return Ok(());
+    async fn solver_time_limit(&self, batch_id: u32) -> Result<Option<Duration>> {
+        let time_remaining = self.exchange.get_current_auction_remaining_time().await?;
+        // Without this check it would be appear as if the the time remaining increased when the
+        // batch changes.
+        if self.current_solving_batch().await? != batch_id {
+            return Ok(None);
         }
+        let batch_time = BATCH_DURATION - time_remaining;
```
nlordell, line 84: Maybe explicitly panic if the subtraction overflows just so we have the same behaviour in debug and release. Although this is super nit-picky and should never happen.

```suggestion
        let batch_time = BATCH_DURATION.checked_sub(time_remaining).expect("time remaining greater than batch duration");
```

core/src/driver/scheduler/evm.rs
```diff
@@ -32,110 +37,118 @@ impl EvmScheduler {
...
 
-    /// Gets the current solving batch ID.
-    ///
-    /// This is the current batch ID minus 1, as the current batch ID is the ID
-    /// of the batch that is currently accepting orders.
-    fn current_solving_batch(&self) -> Result<u32> {
-        Ok(self.exchange.get_current_auction_index().wait()? - 1)
+    /// Wait for and solve the next batch.
+    async fn step(&self, last_batch: Option<u32>) -> Result<u32> {
+        let last_batch = match last_batch {
```
nlordell, line 129: Is it possible to skip a batch if the previous run takes longer than the batch duration?

**edit**: NVM, I see the previous batch is stored in the run loop.

nlordell (approved):
Nice! I think it looks really good now.

PR #1361, opened 2020-08-25
Add address to TokenBaseInfo

e00E:
in preparation for allowing token infos to be looked up by address.

### Test Plan
Unit for deserialize address format. Will need to adjust staging config to include addresses in the override info before merging.

### DIFF
--- a/core/src/price_estimation.rs
+++ b/core/src/price_estimation.rs
@@ -203,13 +203,15 @@ mod tests {
     use super::*;
     use crate::token_info::hardcoded::{TokenData, TokenInfoOverride};
     use anyhow::anyhow;
+    use ethcontract::Address;
     use price_source::{MockPriceSource, NoopPriceSource};
 
     #[test]
     fn price_oracle_fetches_token_prices() {
+        let address = Address::from_low_u64_be(0);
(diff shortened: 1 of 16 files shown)


fleupold (approved):
Nice! In the longer term we could also start using these addresses for the 1Inch/dex.ag price estimation which at the moment still blindly trust the symbols.

core/src/price_estimation.rs
```diff
@@ -203,13 +203,15 @@ mod tests {
...
+    use ethcontract::Address;
     use price_source::{MockPriceSource, NoopPriceSource};
 
     #[test]
     fn price_oracle_fetches_token_prices() {
+        let address = Address::from_low_u64_be(0);
         let tokens = Arc::new(TokenData::from(hash_map! {
-            TokenId(1) => TokenInfoOverride::new("WETH", 18, None),
-            TokenId(2) => TokenInfoOverride::new("USDT", 6, None),
+            TokenId(1) => TokenInfoOverride::new(address, "WETH", 18, None),
```
nlordell, line 213: Maybe it makes sense to have a `TokenInfoOverride::for_test(symbol, decimals)` method? Just a suggestion, not a request for a change.

nlordell (approved):
Great!

PR #1367, opened 2020-08-26
Implement earliest solution submit time for evm scheduler

e00E:
This was only implemented for the system scheduler before.

### Test Plan
New unit test

### DIFF
--- a/core/src/driver/scheduler/evm.rs
+++ b/core/src/driver/scheduler/evm.rs
@@ -89,12 +89,12 @@ impl EvmScheduler {
 
     async fn solver_time_limit(&self, batch_id: u32) -> Result<Option<Duration>> {
         let batch_time = self.batch_time(batch_id).await?;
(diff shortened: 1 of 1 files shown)


core/src/driver/scheduler/evm.rs
```diff
@@ -84,10 +84,17 @@ impl EvmScheduler {
...
-        Ok(self
-            .config
-            .latest_solution_submit_time
-            .checked_sub(batch_time))
+        Ok(Some(batch_time))
+    }
+
+    async fn solver_time_limit(&self, batch_id: u32) -> Result<Option<Duration>> {
+        let batch_time = self.batch_time(batch_id).await?;
+        let time_limit = batch_time.map(|batch_time| {
```
fleupold, line 92: nit: if you use `and_then` here I believe you don't need to flatten below.

nlordell (approved):

fleupold (approved):
⏳

Issue #971, opened 2020-06-09
Handle `atoms` and `hops` parameters


e00E:
There are still two url parameters from the nodejs price estimator that we haven't implemented:
atoms=true and hops=<num>

`hops` doesn't apply to the pricegraph so I think we should remove it.

`atoms` always has to be set to true for the nodejs price estimator so we can either enforce this here too or ignore the parameter (currently it is being ignored). I'm not sure what the expected behavior is supposed to be with atoms=false for which use case we want this.

> From what I remember, the atoms parameter was supposed to specify the unit of the estimated buy amount. Since we currently only support atoms as the units, I think if atoms is false or missing, we should return an 501 Not Implemented HTTP error.

fleupold:
@nlordell we could enforce the maximum number of hops we consider for our price/orderbook estimation couldn't we?

If so, I think we should implement it, because it provides a very powerful visualisation of the pooled liquidity our protocol provides when slowly increasing in the number of hops considered available:

Direct Orderbook:
<img width="843" alt="Screen Shot 2020-06-09 at 16 55 55" src="https://user-images.githubusercontent.com/1200333/84163801-2b706000-aa72-11ea-89a2-fc06b82adcd6.png">

1 Hop
<img width="838" alt="Screen Shot 2020-06-09 at 16 55 48" src="https://user-images.githubusercontent.com/1200333/84163799-2ad7c980-aa72-11ea-88d3-c0e037b178bc.png">

2 Hops
<img width="838" alt="Screen Shot 2020-06-09 at 16 55 38" src="https://user-images.githubusercontent.com/1200333/84163795-28756f80-aa72-11ea-8922-0d8f8c993ca0.png">

fleupold:
`atoms` is a relict from a larger discussion with @anxolin that it would be more intuitive to expect and return base units by default.

That is, a request like `/8-1/estimateBuyAmount/1` would estimate the amount of sETH you get for selling 1ETH, whereas `/8-1/estimateBuyAmount/1?atoms=true` estimates the price for selling 1 wei (10^-18 ETH). Due to slippage the results are fundamentally different.

This requires information about the token decimals which wasn't easily available in the JS codebase so we skipped the "default" implementation at the time and focussed on the minimum amount of functionality needed for the launch.

nlordell:
> we could enforce the maximum number of hops we consider for our price/orderbook estimation couldn't we?

[...]

PR #1368, opened 2020-08-26
Extend Bellman-Ford to include number of hops

fedgiac:
Progress in fixing issue #971.

`ShortestPathGraph` handles a new `hops` parameter. If `None`, it behaves as it did before this PR. If `Some(n)`, then it bounds the size of a shortest path to `n`. (In particular, if identifying a negative cycle requires more than `n` steps, then the negative cycle is not detected and a shortest path graph is returned.)

The behavior of the price estimator and all surrounding services does not change, for now, since `ShortestPathGraph` is always called with `hops = None`.

### Logic
The standard ("unbounded") implementation of Bellman-Ford uses a single predecessor vector, which intuitively tells you which way to go to minimize the weight of the path to the source. All nodes are "relaxed" at each step, and at each step the predecessor vector is updated. There are as many steps as there are nodes. Negative cycles are found by executing one extra relaxation step for each node and checking if the predecessors were updated. A negative cycle exists if and only if there was an update.

When the number of hops is bounded, the number of relaxation steps is bounded by the number of hops.
Each step `s` has its own list of predecessors, which stores all predecessors that would be updated in this step by the unbounded algorithm.
The predecessors at step `s` represent the edge to take to reach the origin in a shortest path of length exactly `s`.
This property is used to find the shortest path taking at most `n` steps.
Negative cycles are found by iterating over all shortest paths from any node to the origin and explicitly looking for cycles.

### Performance, this PR
Interestingly, the Rust compiler doesn't prune unused code.
Benchmarking this PR shows a significant decrease in performance, up to 25% higher running time, even if no new code is ever executed.

<details>
  <summary>Benchmark results</summary>

```
$ git checkout master; cargo bench -p pricegraph; git checkout bellmanford-with-hops ; cargo bench -p pricegraph

<snip>
     Running target/release/deps/pricegraph-c4a70d9d337eb8b6
Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 6.8s, enable flat sampling, or reduce sample count to 30.
Pricegraph::read        time:   [5.2821 ms 5.3024 ms 5.3244 ms]                             
                        change: [+18.074% +25.719% +32.492%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (4.00%) high mild
  2 (4.00%) high severe

Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 6.6s, enable flat sampling, or reduce sample count to 30.
                        time:   [5.0365 ms 5.0622 ms 5.0950 ms]
                        change: [+22.521% +25.110% +26.916%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 50 measurements (4.00%)
  1 (2.00%) high mild
  1 (2.00%) high severe

                        time:   [59.383 us 59.674 us 59.972 us]
                        change: [+1.7435% +2.8315% +3.8814%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (2.00%) high severe
                        time:   [63.580 us 63.824 us 64.086 us]
                        change: [-0.6118% +1.6415% +3.2909%] (p = 0.10 > 0.05)
                        time:   [115.15 us 115.74 us 116.48 us]
                        change: [+11.340% +12.533% +13.866%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (2.00%) high mild
                        time:   [231.13 us 233.18 us 235.22 us]
                        change: [+12.554% +14.982% +16.843%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 50 measurements (4.00%)
  2 (4.00%) high mild
                        time:   [512.28 us 518.48 us 523.86 us]
                        change: [+18.505% +19.747% +20.994%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [54.159 us 54.729 us 55.259 us]
                        change: [+3.7326% +4.6799% +5.7450%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 50 measurements (4.00%)
  1 (2.00%) low mild
  1 (2.00%) high mild
                        time:   [142.05 us 142.66 us 143.26 us]
                        change: [+2.5520% +3.6196% +4.7113%] (p = 0.00 < 0.05)
                        Performance has regressed.
                        time:   [232.41 us 233.35 us 234.52 us]
                        change: [+9.4744% +10.347% +11.178%] (p = 0.00 < 0.05)
                        Performance has regressed.
                        time:   [484.72 us 486.56 us 488.59 us]
                        change: [+9.4196% +10.584% +11.661%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 50 measurements (10.00%)
  4 (8.00%) high mild
  1 (2.00%) high severe
                        time:   [525.35 us 527.94 us 531.17 us]
                        change: [+7.3940% +9.5152% +11.173%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (2.00%) high mild
  3 (6.00%) high severe
```

</details>

### Performance in context
It's hard to measure what would be the performance impact if the `hops` parameter was set without measuring its effect on some actual price estimator query. So I'm creating a draft PR with an implementation of `hops` which can be used for testing in which I give more details on the performance impact.
It appears that when the number of hops is very small, then the bounded path algorithm is more efficient than the current one. However, already for 30 hops the running times are worse than with the current algorithm by a factor of `2`.

In any case, we probably want to limit the size of `hops` since finding a shortest path graph with this method scales in the worst case at least quadratically in the number of hops for both memory and time.

(Note that a large number of hops is an issue only if the order graph is also very complex. Specifying a number of hops that is larger than the number of tokens has no meaningful impact on the performance of the algorithm.) 

### Test plan
New unit tests have been added.
Additional test plan in the linked PR.


pricegraph/src/graph/shortest_paths.rs
```diff
@@ -11,25 +11,53 @@ use petgraph::visit::{
...
-struct PredecessorStore<G: GraphBase + Data> {
-    predecessors: Vec<Option<G::NodeId>>,
-    distance: Distance<G>,
+/// For each node index, this type contains the information needed to
+/// compute predecessor nodes and the distance in the graph from a source
+/// node.
+enum PredecessorStore<G: GraphBase + Data> {
+    Unbounded(PredecessorVec<G>, Distance<G>),
+    Bounded(Vec<PredecessorVec<G>>, UpdatableDistance<G>),
 }
```
e00E, line 28: A lot of functions in this file have to match on this and do different things based on the case. While reading this I was wondering if there was a more readable way to implement this by having each algorithm (no hops vs hops) in its own file and connecting them in a third file either with match or through a trait.
However, it is not clear what the best approach or even if there is a better one than this so I think it's okay if we do that after merging the hops related PRs if we still feel the need then.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -80,34 +133,63 @@ where
...
+                nodes_from_predecessors(self.graph, &predecessors)
+            }
+            PredecessorStore::Bounded(predecessors_at_step, _) => {
+                let mut repeating_node_indices: Vec<_> = predecessors_at_step
+                    .iter()
+                    .map(|predecessors| nodes_from_predecessors(self.graph, &predecessors))
+                    .flatten()
+                    .collect();
+                repeating_node_indices.sort();
+                repeating_node_indices.dedup();
```
e00E, line 147: Maybe insert into HashSet instead

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -11,25 +11,53 @@ use petgraph::visit::{
...
 /// Stores the information needed to manage the predecessor list.
-/// For each node index, this type contains its predecessor node and
-/// distance in the graph from a source node.
-struct PredecessorStore<G: GraphBase + Data> {
-    predecessors: Vec<Option<G::NodeId>>,
-    distance: Distance<G>,
+/// For each node index, this type contains the information needed to
+/// compute predecessor nodes and the distance in the graph from a source
+/// node.
+enum PredecessorStore<G: GraphBase + Data> {
```
e00E, line 25: This file could use a textual explanation in a comment about how the two approaches differ and what the different fields in this enum mean.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -64,14 +93,38 @@ where
...
+                    step -= 1;
+                    if let Some(pred) = predecessors_at_step[step][self.graph.to_index(current)] {
+                        path.push(current);
+                        if pred == self.source {
+                            break path;
+                        }
+                        current = pred;
+                    }
+                }
+            }
```
e00E, line 124: In the description you say:
> The predecessors at step s represent the edge to take to reach the origin in a shortest path of length exactly s.

But then shouldn't the algorithm to find the shortest path iterate over all predecessor lists, starting at the lowest `s` and for each one do the same operation as we do in the `PredecessorStore::Unbounded` case? We have to check whether there is a path at length 0, if not then length 1, if not then length 2, ...
Whereas this algorithm starts at the highest `s`, takes one predecessor and decrements s. I don't understand this. How can it be correct to mix predecessors from different `s`.

e00E: Looking at this more I don't understand why need to keep the vector of predecessors at each step. Couldn't we keep the one vector as usual but just stop bellman ford after s steps?

fleupold: I think the issue is that in the nth step we might concatenate a paths that is already n edges long (because it was earlier in the loop) with another node (thus making it n+1 edge long). The correct implementation uses the previous vector and thus connects D with distance 4.

Therefore we look at the last predecessor vector, for which are are guaranteed to only have n-1 hops long paths.

E.g.
```
A -- 1 --> B --1--> C --1-->D
|                   ^
\-------3----------/
```

and hops two. In the first step from A we reach node B (1) and node C (3). In the second step we learn how to reach node C in 2 and thus node D in three 3. However the path from A to D in three requires 3 hops (more than we specified).

This might be a nice test case.

pricegraph/src/graph/shortest_paths.rs
fleupold: I think this is a great suggestion. It does look like the algorithms are quite different so maybe implementing them in separate traits is a good idea.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -117,25 +199,55 @@ where
...
+                        }
+                    }
+                }
+                None
+            }
+            PredecessorStore::Bounded(predecessors_at_step, _) => {
+                let steps = predecessors_at_step.len();
+                for end_node in self.graph.node_identifiers() {
+                    let mut node = end_node;
+                    for step in (0..steps).rev() {
```
fleupold, line 235: Can you give an example of where we have a cycle in some intermediate step but not in the last one? I'm asking to see if we could just store the last predecessor vector rather than having one for every step.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -200,6 +341,7 @@ where
                 }
             }
         }
+        shortest_path_graph.prepare_next_relaxation_step();
```
fleupold, line 344: I don't like that we rely on this side-effect heavy method here quite unrelated to where we actually read the predecessor vector. It would be nice if it could be computed on the fly if needed.

pricegraph/src/graph/shortest_paths.rs
fleupold: Why does it need to be sorted?

fleupold (commented):
Looks great so far. I agree with @e00E that it would be nice to split it up and have a PathFinding trait that can be implemented by a bounded and Unbounded bellman ford (but you might have experimented with that already). I'm not too worried about complexity as in the future we are likely going to limit the number of hops (even on the smart contract level) to something <10.

pricegraph/src/graph/shortest_paths.rs
e00E: In the current version of the code it's sorted so that `dedup` works correctly (see std docs on dedup).

pricegraph/src/graph/shortest_paths.rs
nlordell: This is exactly the reason, we might return paths that are in fact longer than the specified number of hops if we do not create a predecessor vector per hop.

pricegraph/src/graph/shortest_paths.rs
nlordell: The return type requires a `Vec` so we would have to go from `impl Iterator -> HashSet -> Vec`. I think maybe we should:
- measure the performance impact of that change (since this would make reducing the orderbook even slower, which is currently the bottleneck).
- If the impact is significant, then maybe just group the two statements so it is clear why they are together:
```rust
{
    // NOTE: `dedup` requires a sorted vector.
   repeating_node_indices.sort();
   repeating_node_indices.dedup();
}
```

pricegraph/src/graph/shortest_paths.rs
nlordell: This should be a unit test I think.

nlordell (commented):
I think the code is very nice and clear.

That being said, now that I think about it, I really like @e00E suggestion of splitting this into two types. I the `match` statements in a lot of the methods is what causes the slow down since there is a branch introduced at each method. If we split this at the type level, there would be just one `match` statement, so there would just be a single branch that would have a much smaller impact on performance.

Also at a conceptual level this makes sense to me, since the algorithms seem far enough apart. I think there is a good chunk of common code, which was nicely shared in the current implementation, that can be factored out and stay shared.

pricegraph/src/graph/shortest_paths.rs
nlordell: I'm OK with side-effecty code as private methods for now since this is the "hot-path" for `pricegraph`. I think if we do introduce nice abstractions we should measure their performance, and we can do that incrementally in future PRs.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -64,14 +93,38 @@ where
...
+                    path.push(current);
+                    current = predecessors[self.graph.to_index(current)]?;
+                }
+                path
+            }
+            PredecessorStore::Bounded(predecessors_at_step, _) => {
+                let steps = predecessors_at_step.len();
+                path = Vec::with_capacity(steps);
+                let mut step = steps;
+                loop {
```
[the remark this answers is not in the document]
nlordell: We could use reverse ranges 
```
for pred in (0..predecessors_at_step.len()).rev()
    .map(|step| predecessors_at_step[step][self.graph.to_index(current)])
{
    // ...
}
```

PTSD from the previous memory leak...

pricegraph/src/graph/shortest_paths.rs
nlordell: > separate traits

I agree minus that we should use one trait and separate implementations.

pricegraph/src/graph/shortest_paths.rs
nlordell: I think if we split this into two separate implementations, we should include the documentation there. Also, the modified "bounded Bellman-Ford" algorithm should have some module level documentation since it is sort of invented.

pricegraph/src/graph/shortest_paths.rs
e00E: I have another question for the algorithm here. Would it be possible to instead of storing on predecessor vector for each hop, store a single additional vector that contains for each node the number of hops to the start node. So we would have `predecessors: Vec<PredecessorVec<G>>, weight_distances: Vec<EdgeWeight>, hops_distances: Vec<usize>`. 
Then in bellman_ford we only update the predecessor of a node if we see that the hops doesn't go over the max number of hops.
I haven't fully thought this true, just an idea I had.

nlordell: An issue I see is that it could update the predecessor of a predecessor of the longest path, making the longest path potentially go over the hop limit.

fedgiac: A reason to use a predecessor vector for each hop is that finding shortest paths for different nodes may require you to know the shortest path from the same node for a different number of (remaining) hops.
Example (like the one by Felix, but with an extra node at the end):
```
 0 -10.0-> 1 --1.0-> 2 --1.0-> 3
 |         ∧
1.0        |
 v         |
 4 -------1.0
```
Assume `hops = 3`.
Then `shortestWeight(0->3) = weight(0->1->2->3) = 12`. This is using the fact that the shortest path from 0 to 1 with at most _one_ hop has weight 10.
On the other hand, `shortestWeight(0->2) = weight(0->1->4->2) = 3`. Here the shortest path from 0 to 1 with at most _two_ hops has weight 1.
This example could be made arbitrarily more complicated so that you would need the weight of the shortest path from 0 to 1 for any number of steps.

pricegraph/src/graph/shortest_paths.rs
fedgiac: Performance doesn't seem to change too much when using `HashSet`, worst performance decreases are in `Pricegraph::order_for_limit_price`, of about 4%.

Tested changes are some extra `Hash` trait bounds and:
```suggestion
                let mut unique_nodes = HashSet::new();
                for node in repeating_node_indices.into_iter() {
                   unique_nodes.insert(node)
                }
                unique_nodes.drain().collect()
```


<details>
<summary>Full results</summary>

```
Pricegraph::read        time:   [5.0849 ms 5.0906 ms 5.0985 ms]                              
                        change: [+1.8985% +2.0957% +2.3268%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 11 outliers among 500 measurements (2.20%)
  8 (1.60%) high mild
  3 (0.60%) high severe

                        time:   [4.7446 ms 4.7504 ms 4.7564 ms]
                        change: [-5.2606% -4.9262% -4.6681%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 500 measurements (0.20%)
  1 (0.20%) high mild

Benchmarking Pricegraph::estimate_limit_price/100000000000000000: Warming up for 3.0000 s
Warning: Unable to complete 500 samples in 5.0s. You may wish to increase target time to 7.5s, enable flat sampling, or reduce sample count to 280.
                        time:   [57.816 us 57.912 us 58.011 us]
                        change: [+1.3288% +1.8919% +2.4802%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 500 measurements (1.60%)
  5 (1.00%) high mild
  3 (0.60%) high severe
Benchmarking Pricegraph::estimate_limit_price/1000000000000000000: Warming up for 3.0000 s
Warning: Unable to complete 500 samples in 5.0s. You may wish to increase target time to 7.9s, enable flat sampling, or reduce sample count to 280.
                        time:   [60.097 us 60.179 us 60.265 us]
                        change: [-1.3895% -0.8448% -0.3410%] (p = 0.00 < 0.05)
Found 14 outliers among 500 measurements (2.80%)
  9 (1.80%) high mild
  5 (1.00%) high severe
                        time:   [106.58 us 106.68 us 106.77 us]
                        change: [+1.5183% +1.8297% +2.0602%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 142 outliers among 500 measurements (28.40%)
  106 (21.20%) low severe
  12 (2.40%) low mild
  10 (2.00%) high mild
  14 (2.80%) high severe
                        time:   [200.96 us 201.21 us 201.48 us]
                        change: [+0.4806% +0.7399% +0.9579%] (p = 0.00 < 0.05)
Found 7 outliers among 500 measurements (1.40%)
  6 (1.20%) high mild
  1 (0.20%) high severe
                        time:   [453.78 us 454.39 us 455.03 us]
                        change: [+0.7609% +1.0751% +1.3240%] (p = 0.00 < 0.05)
Found 18 outliers among 500 measurements (3.60%)
  11 (2.20%) high mild
  7 (1.40%) high severe

Benchmarking Pricegraph::order_for_limit_price/200: Warming up for 3.0000 s
Warning: Unable to complete 500 samples in 5.0s. You may wish to increase target time to 6.6s, enable flat sampling, or reduce sample count to 300.
                        time:   [50.740 us 50.821 us 50.903 us]
                        change: [-1.4101% -0.7140% -0.1848%] (p = 0.01 < 0.05)
Found 7 outliers among 500 measurements (1.40%)
  3 (0.60%) high mild
  4 (0.80%) high severe
                        time:   [136.99 us 137.15 us 137.32 us]
                        change: [+3.4241% +3.7101% +3.9312%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 136 outliers among 500 measurements (27.20%)
  89 (17.80%) low severe
  21 (4.20%) low mild
  9 (1.80%) high mild
  17 (3.40%) high severe
                        time:   [220.40 us 220.64 us 220.89 us]
                        change: [+4.4884% +4.6582% +4.8224%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 500 measurements (1.20%)
  4 (0.80%) high mild
  2 (0.40%) high severe
                        time:   [452.49 us 453.31 us 454.37 us]
                        change: [+2.9383% +3.2534% +3.5734%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 500 measurements (0.60%)
  1 (0.20%) high mild
  2 (0.40%) high severe
                        time:   [489.42 us 490.06 us 490.73 us]
                        change: [+1.9823% +2.1616% +2.3452%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 500 measurements (0.40%)
  2 (0.40%) high severe
```

</details>

pricegraph/src/graph/shortest_paths.rs
fedgiac: I'm not sure what you mean by "intermediate step".
If you mean that it finds a loop when `step != 0` then this should be already covered in the test `bounded_search_with_negative_cycle_not_involving_origin`.
If you mean a test where the cycle doesn't start with `end_node`, this should be impossible: `self.graph.node_identifiers()` should contain all nodes in the graph, and if there is a cycle then a node in the cycle will eventually be selected as the `end_node`, which means that the cycle occurs at the last step.

I wonder if there would be an efficiency improvement if I used `self.connected_nodes()` here instead of `self.graph.node_identifiers()`.

fleupold: I meant an example for when it finds a loop with `n` steps, but not with `max_steps` or `max_steps-1` steps. I was under the impression that any cycle that exists in the `n`-th predecessor list would also exist in the last/last but one predecessor list.

pricegraph/src/graph/shortest_paths.rs
nlordell: What if you use `into_iter` instead of `drain`?

fedgiac: It doesn't seem to make much of a difference, here is the bench against the current PR for comparison.

<details>
<summary>Full results</summary>

```
Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 6.4s, enable flat sampling, or reduce sample count to 30.
Pricegraph::read        time:   [5.0343 ms 5.0500 ms 5.0644 ms]                             
                        change: [-3.2447% -1.9703% -0.5280%] (p = 0.00 < 0.05)
Found 10 outliers among 50 measurements (20.00%)
  7 (14.00%) low mild
  2 (4.00%) high mild
  1 (2.00%) high severe

Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 6.4s, enable flat sampling, or reduce sample count to 30.
                        time:   [4.8342 ms 4.8490 ms 4.8639 ms]
                        change: [-4.1812% -2.8137% -1.6666%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 50 measurements (6.00%)
  2 (4.00%) high mild
  1 (2.00%) high severe

                        time:   [57.368 us 57.601 us 57.796 us]
                        change: [-7.8755% -5.8115% -4.5116%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 50 measurements (12.00%)
  5 (10.00%) low mild
  1 (2.00%) high mild
                        time:   [61.133 us 61.373 us 61.615 us]
                        change: [-5.0646% -4.3206% -3.5991%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [107.26 us 108.37 us 109.60 us]
                        change: [-1.0815% -0.6203% -0.0586%] (p = 0.02 < 0.05)
Found 2 outliers among 50 measurements (4.00%)
  2 (4.00%) high severe
                        time:   [203.57 us 204.02 us 204.44 us]
                        change: [+0.0362% +0.3729% +0.7059%] (p = 0.03 < 0.05)
Found 8 outliers among 50 measurements (16.00%)
  1 (2.00%) low severe
  6 (12.00%) low mild
  1 (2.00%) high severe
                        time:   [455.62 us 456.77 us 457.85 us]
                        change: [+1.2362% +1.6203% +2.0236%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (10.00%) low mild
  1 (2.00%) high mild
  1 (2.00%) high severe

                        time:   [51.123 us 51.361 us 51.641 us]
                        change: [-3.4017% -2.7420% -2.0780%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 50 measurements (24.00%)
  1 (2.00%) high mild
  11 (22.00%) high severe
                        time:   [132.39 us 132.64 us 132.95 us]
                        change: [-0.4205% -0.2325% -0.0333%] (p = 0.03 < 0.05)
Found 8 outliers among 50 measurements (16.00%)
  2 (4.00%) low severe
  1 (2.00%) high mild
  5 (10.00%) high severe
                        time:   [217.06 us 217.84 us 218.65 us]
                        change: [+0.0761% +0.4669% +0.8369%] (p = 0.02 < 0.05)
Found 10 outliers among 50 measurements (20.00%)
  3 (6.00%) low severe
  5 (10.00%) low mild
  2 (4.00%) high severe
                        time:   [449.34 us 450.96 us 452.58 us]
                        change: [+1.5642% +2.2870% +3.1362%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 50 measurements (4.00%)
  2 (4.00%) high severe
                        time:   [499.42 us 511.71 us 525.95 us]
                        change: [+3.0038% +4.3750% +5.8430%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 18 outliers among 50 measurements (36.00%)
  2 (4.00%) low severe
  5 (10.00%) low mild
  1 (2.00%) high mild
  10 (20.00%) high sever	
```

</details>

pricegraph/src/graph/shortest_paths.rs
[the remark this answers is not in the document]
fedgiac: I can't make Nick's suggestion work as it is because `current`, borrowed by the closure, is modified inside the loop.

pricegraph/src/graph/shortest_paths.rs
fedgiac: Not with the current implementation, at least. Example:
```
    #[test]
    fn bounded_search_with_negative_cycle_not_involving_origin_modified() {
        //   0 --1.0-> 5 --1.0-> 6 --1.0-> 1 --1.0-> 2
        //                                 ∧         |
        //                              -100.0      1.0
        //                                 |         v
        //                                 4 <-1.0-- 3
        let graph = Graph::<(), f64>::from_edges(&[
            (0, 5, 1.0),
            (5, 6, 1.0),
            (6, 1, 1.0),
            (1, 2, 1.0),
            (2, 3, 1.0),
            (3, 4, 1.0),
            (4, 1, -100.0),
        ]);
        let negative_cycle = ShortestPathGraph::new(&graph, 0.into(), Some(7));
    }
```
Add `println!("Loop found at step {} out of {}", steps - step, steps);` after `if pred == end_node {` to see when the cycle is found, in this case the output is `Loop found at step 5 out of 8`. The step at which the loop is found is always 5 even when the "tail" to zero becomes longer.

fleupold:
Superseeded by #1495

Issue #1297, opened 2020-08-12
Fix debug assert failing on very low price


e00E:
https://github.com/gnosis/dex-services/pull/1295 discovered that an almost zero price will change to infinite when inversed which fails the debug assertion in LimitPrice.
When this is fixed we can merge the PR.

PR #1370, opened 2020-08-26
OOM Error in Pricegraph

### LINKED ISSUE #1350: Price Estimator Out Of Memory
It seems like a good idea to track this in an issue. The summary is that we've seen a few cases where the staging container got killed due to exceeding the memory limit. The memory graph itself doesn't show any spike so it is likely that this is happens in a quick burst.

Felix checked the Kibana logs of the nginx ingress service for the first request that returned a 502 or 503 error around the point where the container got killed. `GET /api/v1/markets/4-1?atoms=true&hops=2` . This seems like a normal frontend request to me and when I ran it locally I couldn't see anything special about it in terms of memory usage.

One idea I had for the out of memory was many parallel requests since we use potentially a couple of megabytes during the handling of a single request but we would have seen that in the ingress logs so I don't think this happened.

The other idea is a bug in our code where we allocate too much memory. This could be reserving too much in a vector or an infinite loop that allocates.
For example [this](https://github.com/gnosis/dex-services/blob/b6064498b76089b60d565b859308f4aa6a06db20/pricegraph/src/graph/shortest_paths.rs#L70) could be a problem if we had a bug that would cause a loop in the predecessor list. I added a debug assert for this and let fuzzing run for a while but it didn't hit it (and really fuzzing should notice this even without a debug assert because it would be an infinite loop and run out of memory).

Or it could some kind of weird interaction with how kubernetes measures memory. For example it is possible for a process to request a large amount of virtual memory without any physical memory backing it. So if there is some condition in which we or a dependency does that it could trigger kubernetes even though the container does not really use the memory and we wouldn't notice this locally. I thought of this because I remembered this part of the [clang address sanitizer documentation](https://clang.llvm.org/docs/AddressSanitizer.html#limitations):

> On 64-bit platforms AddressSanitizer maps (but not reserves) 16+ Terabytes of virtual address space. This means that tools like ulimit may not work as usually expected.

fleupold:
Will fix #1350 

Right now I have only created a unit test that shows the issue. We can clearly demonstrate a cycle in the predecessor list between token ID 1 and 2 (due to floating point artihmetic), yet `graph.find_cycle()` returns `None`

Let's use this PR to discuss potential solutions.

### DIFF
--- a/pricegraph/src/graph/shortest_paths.rs
+++ b/pricegraph/src/graph/shortest_paths.rs
@@ -124,15 +124,26 @@ where
         // each node: if a node can be relaxed then a negative cycle exists and is
         // created in the predecessor store; otherwise no such cycle exists.
 
-        for i in self.graph.node_identifiers() {
-            for edge in self.graph.edges(i) {
-                let j = edge.target();
-                let w = *edge.weight();
-
-                if *self.distance(i) + w < *self.distance(j) {
-                    self.update_predecessor(j, Some(i));
-                    return Some(j);
+        let mut visited: Vec<Option<usize>> = vec![None; self.predecessor_store.predecessors.len()];
+        for current_generation in 0..visited.len() {
+            if visited[current_generation].is_some() {
+                continue;
+            }
+            visited[current_generation] = Some(current_generation);
+            let mut current_index = current_generation;
+            while let Some(next) = self.predecessor_store.predecessors[current_index] {
+                let next_index = self.graph.to_index(next);
+                match visited[next_index] {
+                    Some(visited_generation) => {
+                        if visited_generation < current_generation {
+                            break;
+                        } else {
+                            return Some(next);
+                        }
+                    }
+                    None => visited[next_index] = Some(current_generation),
                 }
+                current_index = next_index;
             }
         }
         None
@@ -185,8 +196,8 @@ where
 {
     let mut shortest_path_graph = ShortestPathGraph::empty(g, source);
 
-    // scan up to |V| - 1 times.
-    for _ in 1..g.node_count() {
+    // scan up to |V| times.
+    for _ in 0..g.node_count() {
         let mut did_update = false;
         for i in g.node_identifiers() {
             for edge in g.edges(i) {
@@ -217,19 +228,15 @@ pub mod tests {
     use petgraph::Graph;
 
     #[test]
-    fn floating_point_issue() {
+    fn negative_cycle_with_floating_point_precision() {
+        // In this example the path 0->1 should have the same distance (0.1) as the path 0->1->2->1 (0.1)
+        // However due to floating point precision 0.1 + 6.8 - 6.8 < 0.1. The latter path leaves the graph
+        // in a cyclical state, thus this test ensures that we cannot create such a graph.
         let graph = Graph::<(), f64>::from_edges(&[(0, 1, 0.1), (1, 2, 6.8), (2, 1, -6.8)]);
-        let mut graph = ShortestPathGraph::new(&graph, 0.into()).unwrap();
-        let predecessor_ids = graph
-            .predecessor_store
-            .predecessors
-            .iter()
-            .map(|item| item.map(|i| graph.graph.to_index(i)))
-            .collect::<Vec<_>>();
-        // There exists a cycle
-        assert!(predecessor_ids[1] == Some(2) && predecessor_ids[2] == Some(1));
-        // But we don't find it
-        assert!(graph.find_cycle().is_some());
+        let result = ShortestPathGraph::new(&graph, 0.into());
+        assert!(result.is_err());
+        let negative_cycle = result.err().unwrap();
+        assert_eq!(negative_cycle.0, vec![1.into(), 2.into(), 1.into()]);
     }
 
     #[test]
--- a/pricegraph/src/graph/shortest_paths.rs
+++ b/pricegraph/src/graph/shortest_paths.rs
@@ -217,6 +217,22 @@ pub mod tests {
     use petgraph::Graph;
 
     #[test]
+    fn floating_point_issue() {
+        let graph = Graph::<(), f64>::from_edges(&[(0, 1, 0.1), (1, 2, 6.8), (2, 1, -6.8)]);
+        let mut graph = ShortestPathGraph::new(&graph, 0.into()).unwrap();
+        let predecessor_ids = graph
+            .predecessor_store
+            .predecessors
+            .iter()
+            .map(|item| item.map(|i| graph.graph.to_index(i)))
+            .collect::<Vec<_>>();
+        // There exists a cycle
+        assert!(predecessor_ids[1] == Some(2) && predecessor_ids[2] == Some(1));
+        // But we don't find it
+        assert!(graph.find_cycle().is_some());
+    }
+
+    #[test]
     fn search_finds_negative_cycle() {
         // NOTE: There is a negative cycle from 1 -> 2 -> 3 -> 1 with a
         // transient weight of -1.


fleupold:
One suggestion on how we could possibly fix this would be to not use a different `mark_cycle` logic.

Instead of searching for nodes that can be reached more cheaply using some edge, we could also mark each node "visited" in pass n. We start at the first node and traverse all successor, marking them visited in the first pass. If we reach a node that has been visited in the same pass, we have found a cycle. If we find a node without successors or one that has been visited in an earlier pass we can stop and continue with the next node/pass.

This should find the cycle shown in the unit test while still running in O(|E|).

e00E:
I added debug prints to the algorithm to see what happens:

```
[pricegraph/src/graph/shortest_paths.rs:207] i = NodeIndex(0)
[pricegraph/src/graph/shortest_paths.rs:207] j = NodeIndex(1)
[pricegraph/src/graph/shortest_paths.rs:207] w = 0.1
[pricegraph/src/graph/shortest_paths.rs:207] left_distance = 0.1
[pricegraph/src/graph/shortest_paths.rs:207] right_distance = inf
[pricegraph/src/graph/shortest_paths.rs:207] smaller = true

[pricegraph/src/graph/shortest_paths.rs:207] i = NodeIndex(1)
[pricegraph/src/graph/shortest_paths.rs:207] j = NodeIndex(2)
[pricegraph/src/graph/shortest_paths.rs:207] w = 6.8
[pricegraph/src/graph/shortest_paths.rs:207] left_distance = 6.8999999999999995
[pricegraph/src/graph/shortest_paths.rs:207] right_distance = inf
[pricegraph/src/graph/shortest_paths.rs:207] smaller = true

[pricegraph/src/graph/shortest_paths.rs:207] i = NodeIndex(2)
[pricegraph/src/graph/shortest_paths.rs:207] j = NodeIndex(1)
[pricegraph/src/graph/shortest_paths.rs:207] w = -6.8
[pricegraph/src/graph/shortest_paths.rs:207] left_distance = 0.09999999999999964
[pricegraph/src/graph/shortest_paths.rs:207] right_distance = 0.1
[pricegraph/src/graph/shortest_paths.rs:207] smaller = true
...
```

We see that the cost of going from 0 to 1 to 2 is `0.1 + 6.8` which is rounded to `6.8999999999999995`. Then the cost of going back from 2 to 1 is `6.8999999999999995 - 6.8` which is rounded to `0.09999999999999964` which is less than the original `0.1` of going from 0 to 1.

nlordell:
> We start at the first node and traverse all successors

Bellman-Ford requires a predecessor list, since a node can theoretically have multiple successors, but only a single predecessor.

I think the only way would be to start from every node and mark nodes backwards in the predecessor list until:
- It reaches a node without a predecessor (this would mean that it reached the starting node *or* it is a node that is not connected to the starting node, meaning it is not connected to the start node).
- It reaches a previously marked node (this would mean its connected to a path to the start node without any cycles, so it does not have any cycles itself)
- It reaches itself (meaning it is part of a negative cycle).

**note**: This is basically @fleupold 's algorithm but with `s/successor/predecessor/g`.

I'm not sure what the runtime complexity of this is, but my guess is that it is `O(|N|)` and not `O(|E|)` Since we should visit at most 2x the number of nodes right?

Also, this would require a slight modification to the Bellman-Ford algorithm where we would have to run it for one extra iteration to make sure that we find the maximum negative cycle (i.e. going through all the nodes and ending at the starting node - this was previously guaranteed by the `find_cycle` method) and we would have to check for negative cycles on every search instead of being able to return early when have an iteration where we don't relax any nodes (`did_update == false`).

**edit**: It appears that `did_update == false` wasn't returning early previously when I think it was able to, so this won't need to change.

fleupold:
> I'm not sure what the runtime complexity of this is, but my guess is that it is O(|N|) and not O(|E|)

Yes, you are totally right. I was thinking that we touch each edge once, but that's not correct. We iterate over the predecessor list which is O(V).

e00E:
My internal understanding/summary of the issue is that the predecessor vector got out of sync with the weights due to floating point error. Bellman ford uses the weights to determine whether there is a negative cycle in the final relaxation step but in this case this does not show the cycle even though it exists in the predecessor vector.
The approach of using the predecessor vector itself to detect the negative cycle makes sense to me since that is exactly what we need to have hold later. It is still possible that we see a negative cycle when there is none or vice versa due to floating point accuracy but at least the result will always be consistent with the returned predecessor vector.

e00E:
Another idea just to throw it is that we could try using a bigint based rational number type which does not have precision problems which gets rid of this problem and all rounding errors. It is *probably* too slow but I'll probably try it out to see.

fleupold:
One issue with the solution I implemented, is that it might not be deterministic in the way that cycles are detected. When running the problematic block with this change I'm seeing:

> [core::logging] thread 'tokio-runtime-worker' panicked at 'overlapping orders in reduced orderbook: OverlapError(NegativeCycle([NodeIndex(1), NodeIndex(78), NodeIndex(1)]))', pricegraph/src/api/transitive_orderbook.rs:81:13

Will look closer into why this is the case.

e00E:
I'm guessing what happens is that whether a negative cycle is found depends on the starting nodes used for bellman ford. So `reduce_overlapping_orders` thinks it has removed all negative cycles but then when getting the transitive orderbook we discover new cycles even though the assumptions was that there are none. This is again the problem with the floating point accuracy.
Not sure how to fix it. One hacky way we would be to reduce those new found cycles again even though our assumption was that there are none.

fedgiac:
I think the discussed solution is good for us (that is, run Bellman-Ford as usual but check if there are negative cycles at the end anyway). There is still the issue that some positive cycles will be returned as negative and some negative cycles might sneak into the `ShortestPathGraph` because of floats, but `ShortestPathGraph` will always be guaranteed to find a path.

Some comments on the implementation:
1. I don't understand it. In particular, the check `visited_generation < current_generation` is perplexing. What is special about nodes that are after `visited_generation`? It seems to me that if the source node has a large index then two nodes with smaller index than that following different but legitimate paths to the source would be marked as a negative cycle. (If there are no negative cycles, the predecessor of the source is `None`.) 
2. Changing `1..g.node_count()` to `0..g.node_count()` takes care of the "mark cycle" step. At this point we don't need the function `mark_cycle` anymore and we can use this code directly in `find_cycle` to explicitly find the cycle.

I imagine the algorithm to have a `reaches_source` bool vector initialized to `false` except `reaches_source[source] = true`. Loop over all nodes `n`, then:
1. if `reaches_source[n] == true`, continue.
2. if `predecessor == None`, continue.
3. else follow predecessor path marking all nodes `n'` in the path as `reaches_source[n'] = true` until `reaches_source[n'] == true`. Panic if `predecessor[n'] == None`. (Should not happen with BF, not even with approximation issues.)

nlordell (commented):
The algorithm looks good, as you pointed out, this raises issues regarding unpredictable behaviour where a different starting node may cause a negative cycle that isn't observable from a different, but connected, starting node. Basically:
```
A --> B --> C
      ^----/
```

A negative cycle might be detected when searching from B but not from A, and this assumption is used in the `pricegraph`. Additionally, when there are multiple orders per token connection, where the better order doesn't cause a negative cycle, but a worse one does. This means that part way through finding orders, it may be that a negative cycle shows up in a reduced graph.

> Another idea just to throw it is that we could try using a bigint based rational number type which does not have precision problems which gets rid of this problem and all rounding errors. It is probably too slow but I'll probably try it out to see.

Maybe the `num` crate already has this for us? It would also fix #1297

fleupold:
> This means that part way through finding orders, it may be that a negative cycle shows up in a reduced graph.

@nlordell can you give an example of that? I'd assume that if there was no negative cycle left in the graph at some point, removing edges will not introduce a new negative cycle.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -124,15 +124,26 @@ where
...
+            if visited[current_generation].is_some() {
+                continue;
+            }
+            visited[current_generation] = Some(current_generation);
+            let mut current_index = current_generation;
+            while let Some(next) = self.predecessor_store.predecessors[current_index] {
+                let next_index = self.graph.to_index(next);
+                match visited[next_index] {
+                    Some(visited_generation) => {
+                        if visited_generation < current_generation {
```
fleupold, line 138: >  In particular, the check visited_generation < current_generation is perplexing. What is special about nodes that are after visited_generation?

@fedgiac visited generation can only be smaller or equal to current_generation. It it is smaller, it means we already visited this node in a previous path (and haven't found a cycle).
If it's equal it means we have been here in the current pass, thus we have found a cycle.

You are right that I should add some more comments into the logic.

fleupold:
Generally I think the suggestion of using rational is nice. We did this in the original js implementation but ran into issues when chaining many large numbers together (which is the case when we go over many hops) - numbers became extremely large. Maybe this is not an issue if we limit the hops to something reasonable like 5 or 6.

@fedgiac also pointed out that these float are not actually prices, but logarithms of prices. He suggested we could convert them into integer (by taking the top n significant decimals) and maybe perform Math on that basis.

nlordell:
> can you give an example of that?

Imagine the following graph:
```
  /---0.1--\
 /--0.0625--\
|            v
0             1 ---6.8---> 2
              ^            |
               \--(-6.8)--/
```

When trying to estimate the xrates between token 0 and 1 for some volume `x`. We sould use BF to first find that the shortest path to 1 from 0 is over the order with price `0.0625`, and in this example, the volume of that order is not enough to fully fill `x`. We would then run BF again and this time run into the negative cycle (as per the unit test). 0.0625 was chosen because it is a power of two, so multiplying by it only changes the exponent and not the mantissa and as such does not cause any precision errors:

```rust
        let predecessors = |edges: &[(u32, u32, f64)]| {
            ShortestPathGraph::new(&Graph::<(), f64>::from_edges(edges), 0.into())
                .map(|g| g.predecessor_store.predecessors)
        };
```

yields:
```
[pricegraph/src/graph/shortest_paths.rs:246] predecessors(&[(0, 1, 0.0625), (1, 2, 6.8), (2, 1, -6.8)]) = Ok(
    [
        None,
        Some(
            NodeIndex(0),
        ),
        Some(
            NodeIndex(1),
        ),
    ],
)
[pricegraph/src/graph/shortest_paths.rs:247] predecessors(&[(0, 1, 0.1), (1, 2, 6.8), (2, 1, -6.8)]) = Err(
    NegativeCycle(
        [
            NodeIndex(1),
            NodeIndex(2),
            NodeIndex(1),
        ],
    ),
)
```

```

fleupold:
Thanks for the example. I think this makes it pretty much "unusable" since we would have to do a full pass over all nodes on every step of our algorithms to check that the intermediate graphs are cycle free and if not reduce them. 

I think using some rounded prices that fit into rationals might be better in that case.

I also created https://github.com/petgraph/petgraph/issues/373 to see if the original authors have an idea how to fix the algorithm itself.

nlordell:
> @fedgiac also pointed out that these float are not actually prices, but logarithms of prices. He suggested we could convert them into integer (by taking the top n significant decimals) and maybe perform Math on that basis.

I think this is a super awesome idea. Order amounts are u128 meaning that the highest possible weight is ~128 (`log2(u128::max / 1)`) and the lowest if ~-128 (`log2(1 / u128::max)`), effectively limiting our range of weight to `(-128.0, 128.0)` which means we not only have a lot of wasted precision. If we compute the 2-based logarithm, then multiply them by some large constant (2^64 for example) and then represent them as `i128`s, then we should be able to perform addition and subtraction without precision loss and not run into this issue. Also integer math is very fast so we might not even lose much performance in doing so.

> Generally I think the suggestion of using rational is nice

I still think this is a good idea to implement in the long run as it us more accurate estimates that better model the exchange. It also solves some other floating point ugliness such as the issue @e00E found here #1297.

fleupold:
superseeded by #1374

PR #1372, opened 2020-08-27
Fix sleep not being awaited

e00E:
And added marked function as must_use. Usually rust warns about not awaiting the result of async functions but it looks like this doesn't trigger when returning BoxFuture. Technically with this reasoning we should put it everywhere where we are returning BoxFuture but not sure if that is really needed since usually we notice because we want to use the return value. sleep is special because we call it for the side effect.

### Test Plan
Tests don't notice this because mockall still registers the call.

### DIFF
--- a/core/src/driver/scheduler/system.rs
+++ b/core/src/driver/scheduler/system.rs
@@ -145,7 +145,7 @@ async fn wait_for_batch_id(
     //   one accepting orders and does not yet accept solutions.
     while batch_id.0 as u32 >= contract.get_current_auction_index().await? {
         log::info!("Solved batch is not yet accepting solutions, waiting for next batch.");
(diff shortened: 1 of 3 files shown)


nlordell (approved):

fleupold (approved):
Nice catch!

https://github.com/rust-lang/futures-rs/pull/1691 has a proposal on how this could be worked around (by wrapping the actual BoxFuture in a struct and marking it must_use).

I'd be in favour of using such a type in our codebase wherever we return BoxFuture as these bugs can be really nasty to find (not caught by unit tests) and I really like the rust compiler warnings in these cases.

nlordell:
> by wrapping the actual BoxFuture in a struct and marking it must_use.

This is a very good idea! We can then have our own `boxed_must_use()` extension method that we use instead of `boxed()`.

PR #1373, opened 2020-08-27
Report healthiness at the start of the next batch

e00E:
Previously we would never set the status to healthy. This is in preparation for using the health check on kubernetes.

### Test Plan
Unit tests and manually tested running the driver with system scheduler.

### DIFF
--- a/core/src/driver/scheduler/system.rs
+++ b/core/src/driver/scheduler/system.rs
@@ -2,6 +2,7 @@ use super::{AuctionTimingConfiguration, Scheduler};
 use crate::{
     contracts::stablex_contract::StableXContract,
     driver::stablex_driver::{DriverError, StableXDriver},
+    health::HealthReporting,
     models::{BatchId, Solution},
     util::{self, AsyncSleep, AsyncSleeping, Now},
 };
@@ -18,6 +19,7 @@ const CONTRACT_BATCH_ID_POLL_INTERVAL: Duration = Duration::from_secs(10);
 pub struct SystemScheduler {
     contract: Arc<dyn StableXContract>,
     driver: Arc<dyn StableXDriver>,
+    health: Arc<dyn HealthReporting>,
     auction_timing_configuration: AuctionTimingConfiguration,
     last_solved_batch: Option<BatchId>,
 }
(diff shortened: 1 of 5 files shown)


nlordell (approved):

core/src/driver/scheduler/system.rs
```diff
@@ -192,8 +196,17 @@ fn log_submit_result(batch_id: BatchId, result: &Result<()>) {
     }
 }
 
+fn duration_until_healthy(now: SystemTime) -> Duration {
+    let next_start_solve_time = BatchId::current(now).unwrap().solve_start_time();
```
fleupold, line 200: I think it would be nice here if we return immediately in case we are before `target_start_solve_time` of `BatchId::currently_being_solved`. 
Or do you think the added complexity is not worth the benefit of saving 5 minutes of ready time in 1/6 of the cases (assuming uniform distribution of restarts)?

core/src/health.rs
```diff
@@ -8,7 +8,9 @@ use std::sync::atomic::{AtomicBool, Ordering};
 /// Trait for asyncronously notifying health information.
 #[cfg_attr(test, mockall::automock)]
 pub trait HealthReporting: Send + Sync {
-    /// Notify that the service is ready.
+    /// Notify that the service is ready. Can be called multiple times.
+    /// We use this to signal healthiness only at the start of a batch in order to not interrupt the
```
fleupold, line 12: nit: we signal readiness not healthiness. Not sure we need to document what this is used for (as this is is a detail that might change but the comment likely won't).

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F3ohzdPgwt1E2qUcGFa%2Fgiphy.gif&token=born_ready" data-gitmeme-token="born_ready"></a>

core/src/driver/scheduler/system.rs
e00E: Like you said I didn't think this was worth it and it's inline with the evm scheduler which doesn't use the target solve start time. It's also a little safer in the case we are very close to target_start_solve_time but the time between the two pods is out of sync. Then it could happen that the other pod started solving already even though we think it is safe for us to start.

core/src/health.rs
e00E: To me it's useful when I imagine someone later looking at this and wondering what the point of the health reporting trait is.

core/src/driver/scheduler/system.rs
nlordell: Could using `target_start_solve_time` become an issue if the configuration changes?

fleupold: It would have to use the value from the configuration. I'm not to worried about system time between pods being dramatically out of sync ( we can cancel an ongoing processing unless we are already sending the tx), but I can see how the added logic is not worth the benefit (we can see if long wait times are bothering us).

PR #1374, opened 2020-08-27
Use integer weights

nlordell:
I got super curious and decided to implement @fedgiac 's awesome idea of mapping the logarithmic weights into integer space. This is done by use fixed-point numbers to represent the logarithms instead of `f64`.

### Test Plan

Added new unit tests.

Benchmarks show a performance regression of about +80% more time to reduce an orderbook, and around 30% more time to estimate prices:

<details><summary>Benchmarks</summary>

```
     Running /var/home/nlordell/Developer/dex-service-alt/target/release/deps/pricegraph-656df3332885b719
Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 8.2s, enable flat sampling, or reduce sample count to 20.
Pricegraph::read        time:   [6.4003 ms 6.4073 ms 6.4162 ms]                             
                        change: [+75.985% +76.767% +77.634%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 50 measurements (12.00%)
  2 (4.00%) low mild
  4 (8.00%) high severe

Warning: Unable to complete 50 samples in 5.0s. You may wish to increase target time to 8.0s, enable flat sampling, or reduce sample count to 20.
                        time:   [6.2604 ms 6.2714 ms 6.2886 ms]
                        change: [+82.277% +83.122% +84.217%] (p = 0.00 < 0.05)
                        Performance has regressed.
  4 (8.00%) high severe

                        time:   [48.818 us 48.951 us 49.162 us]
                        change: [-8.5934% -8.3923% -8.1698%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (2.00%) high mild
  6 (12.00%) high severe
                        time:   [53.632 us 53.792 us 54.014 us]
                        change: [+5.2314% +5.7326% +6.1350%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 50 measurements (18.00%)
  5 (10.00%) high mild
  4 (8.00%) high severe
                        time:   [114.65 us 114.81 us 115.03 us]
                        change: [+30.747% +31.297% +31.912%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (2.00%) high mild
                        time:   [203.75 us 204.57 us 205.61 us]
                        change: [+20.676% +21.287% +21.933%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (6.00%) high mild
  4 (8.00%) high severe
                        time:   [481.42 us 482.69 us 484.53 us]
                        change: [+15.011% +15.391% +15.780%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 50 measurements (18.00%)
  1 (2.00%) low mild
  2 (4.00%) high mild
  6 (12.00%) high severe

                        time:   [42.773 us 42.984 us 43.297 us]
                        change: [-0.0354% +0.9136% +1.6789%] (p = 0.03 < 0.05)
Found 11 outliers among 50 measurements (22.00%)
  1 (2.00%) high mild
  10 (20.00%) high severe
                        time:   [136.14 us 141.69 us 147.87 us]
                        change: [+10.307% +13.194% +16.512%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (2.00%) high mild
  3 (6.00%) high severe
                        time:   [245.88 us 260.30 us 273.60 us]
                        change: [+25.446% +29.754% +35.010%] (p = 0.00 < 0.05)
                        Performance has regressed.
                        time:   [518.59 us 534.24 us 552.01 us]
                        change: [+40.942% +43.017% +45.047%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 50 measurements (10.00%)
  1 (2.00%) high mild
  4 (8.00%) high severe
                        time:   [519.24 us 521.69 us 524.49 us]
                        change: [+27.763% +28.833% +29.834%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 50 measurements (18.00%)
  4 (8.00%) high mild
  5 (10.00%) high severe
```

</details>

Also it looks like the results from the historic prices did not change (beyond randomness across runs because of `HashMap`). Also, the historic e2e script ran at 286.39 batches/s instead of 329.23 batches/s which is 15% slower on average, so not as dramatic as suggested by the benchmarks.

<details><summary>Historic prices</summary>

```
$ cargo run --release -p e2e --bin historic_prices -- --orderbook-file target/orderbook-file-mainnet
62717 / 62717 [========================================================] 100.00 % 286.39/s
Processed 17280 token prices: average error 22706.85%, bad 2.85%, missed 9.04%.
$ git checkout master
$ cargo run --release -p e2e --bin historic_prices -- --orderbook-file target/orderbook-file-mainnet
62717 / 62717 [========================================================] 100.00 % 329.23/s
Processed 17280 token prices: average error 22823.76%, bad 2.84%, missed 9.03%.
```

</details>


pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +170,52 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
     debug_assert!(num::is_strictly_positive_and_finite(value));
     value
 }
+
+/// The exchange rate weight used by the pathfinding algorithm.
+#[derive(Clone, Copy, Debug, Default, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Weight(i128);
```
e00E, line 176: A better high level description of this change is that we use fixed point arithmetic instead of floating point. Effectively we use a i128 to store 103 bits of the fractional part and 16 bits of the whole part.

It might be a good idea to use a real fixed point crate for this. I found https://crates.io/crates/fixed which looks good based on the documentation.

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +170,52 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
...
+        // NOTE: We need to reserve some bits in order for the addition not to
+        // overflow, specifically we need to reserve
+        // - 16 bits for the maximum number of nodes (maximum number of tokens)
+        // - 8 bits for the [-128, 128) range
+        // - 1 bit for infinity
+        // Thats a total of 25 bits, which leaves us with 103 bits of precision
+        // for the weight.
+        const FACTOR: f64 = (1u128 << 103) as _;
+
+        let weight = value.log2() * FACTOR;
```
e00E, line 190: At first I thought this would introduce a lot error because of the multiplication with a large number. This isn't the case because the number is a power of 2 so in the float representation it only changes the exponent. This should be commented.

Since f64 uses 53 bits for the fractional part we don't get any benefit by shifting up to 103. Let's document this too and use a lower shift amount.

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +170,52 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
...
+        debug_assert!(-128.0 * FACTOR <= weight && weight < 128.0 * FACTOR);
+
+        Weight(weight as _)
```
e00E, line 193: Thinking more about precision, it might be possible to have less error by creating our log2 fixed point value directly from the u128 numerator and denominator instead of calculating the ratio as f64 (in LimitPrice and ExchangeRate) first and then moving into fixed point.

e00E (approved):
Moving to fixed point arithmetic makes a lot of sense because it makes the algorithms more stable as we only introduce rounding error once in the initial conversion but in the arithmetic operations which also fixes the problems we have with bellman ford being inconsistent.

pricegraph/src/orderbook/scalar.rs
e00E, line 190: Since f64 uses 53 bits for the fractional part we don't get any benefit by shifting up to 103. Let's document this too and use a lower shift amount. (Actually I'm not sure whether this is true, need to think about it more)

pricegraph/src/orderbook/scalar.rs
nlordell: The issue is we use specific "infinity" semantics required by `petgraph` but it's worth trying out.

pricegraph/src/orderbook/scalar.rs
nlordell: > We should shift by as much as possible.

Correct!

pricegraph/src/orderbook/scalar.rs
e00E: If you use the crate https://docs.rs/fixed/1.1.0/fixed/struct.FixedI128.html#method.from_num is probably relevant (there are a *lot* of traits on this type).

pricegraph/src/orderbook/scalar.rs
nlordell: This is a really good idea, but I'd have to research how. Maybe this is provided by the fixed point crate?

e00E: Agree, can be done later.

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +170,52 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
...
+    fn add(self, rhs: Self) -> Self {
+        // NOTE: The Bellman-Ford implementation relies on `∞ * x == ∞`.
+        // TODO(nlordell): Since this branch is in a tight loop, we would
+        // probably benifit quite a bit in optimizing this so that there are no
+        // branches.
+        if self == Weight::infinite() || rhs == Weight::infinite() {
+            Weight::infinite()
+        } else {
+            Weight(self.0 + rhs.0)
+        }
```
e00E, line 219: If we need to optimize this I'd do it in a follow up PR too unless it turns out that performance is terrible without it.
I don't think it will be super important because this is not going to compile to a branch (with jump) anyway. Instead the compiler should be able to use the cmove instruction: https://godbolt.org/z/G4sj3n
Still a lot more instructions for `i128` that can maybe be improved but it's more on the micro optimization level.

nlordell: > If we need to optimize this I'd do it in a follow up PR

Agree, that's why I left a todo.

> Instead the compiler should be able to use the cmove

Nice! I also compared the performance of this vs `saturated_add` and the if statement seems to perform slightly better and also panic on overflow in `debug` mode :metal:.

I'll reword the TODO.

pricegraph/src/orderbook/scalar.rs
nlordell: The crate didn't help much since we only need to use `+`. I ended up just creating a type alias in order to make it very clear what is going on.

nlordell:
@e00E Do you mind re-reviewing with the latest updates to the documentation and the new test to make sure that we got the fixed point precision right?

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +167,122 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
     debug_assert!(num::is_strictly_positive_and_finite(value));
     value
 }
+
+/// A signed fixed point number with 24 magnitude bits and 104 fractional bits.
+///
+/// Note the that size of the magniture and fractional compontents was carefully
```
e00E, line 173:
```suggestion
/// Note the that size of the magnitude and fractional components was carefully
```

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +167,122 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
...
+
+impl FloatMeasure for Weight {
+    fn zero() -> Self {
+        Weight(0)
+    }
+
+    fn infinite() -> Self {
+        // NOTE: Use a special marker value to represent +∞ which is needed by
+        // the `petgraph` Bellman-Ford implementation. `i128::MIN` is chosen so
+        // that the the range of non-infinite values are semetric around `0`.
```
e00E, line 244:
```suggestion
        // that the the range of non-infinite values are symmetric around `0`.
```

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +167,122 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
     debug_assert!(num::is_strictly_positive_and_finite(value));
     value
 }
+
+/// A signed fixed point number with 24 magnitude bits and 104 fractional bits.
```
e00E, line 171:
```suggestion
/// A signed fixed point number with 1 sign bit, 23 magnitude bits and 104 fractional bits.
```
Maybe? Not sure if this is usually counted towards the magnitude or not.

pricegraph/src/orderbook/scalar.rs
nlordell: :see_no_evil:

pricegraph/src/orderbook/scalar.rs
nlordell: :see_no_evil:

pricegraph/src/orderbook/scalar.rs
nlordell: AFAIK, fixed point numbers typically use 2's complement notation so that they reuse integer math.

pricegraph/src/orderbook/scalar.rs
nlordell: What do you think of the `Fixed24x104 = i128` type alias to make it clear that its a fixed point number?

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +167,158 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
...
+/// In the logarithmic scale, this range is:
+/// ```
+/// [-114.71, 114.71]
+/// ```
+///
+/// Furthermore, these weights can be added at most 2^16 times (this is the
+/// maximum number of tokens in the exchange), so the total range that must be
+/// representable by the magnitude bits is:
+/// ```
+/// [-7517784, 7517784]
```
e00E, line 191: Not important in practice because we aren't close to a power of 2, but the `114.71` is rounded which makes a difference here.
Now that I'm writing this maybe it is easier to understand if we ignore MIN_AMOUNT.
Then the maximum log2 is `log2(2**128-1) ~ 128` (a tiny bit smaller than 128) and `128 * 2**16 == 2**23` which is exactly the 23 bits needed.

pricegraph/src/orderbook/scalar.rs
e00E: That's fine. A comment would also be fine. It just took me some time when I first read the PR to realize it was doing fixed point arithmetic because that word wasn't mentioned anywhere.

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +167,158 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
...
+impl FloatMeasure for Weight {
+    fn zero() -> Self {
+        Weight(0)
+    }
+
+    fn infinite() -> Self {
+        // NOTE: Use a special marker value to represent +∞ which is needed by
+        // the `petgraph` Bellman-Ford implementation. `i128::MIN` is chosen so
+        // that the the range of non-infinite values are symmetric around `0`.
+        Weight(i128::MIN)
```
e00E, line 245: Technically we can pick any value that has the left most bit set to 1, right? Just making sure I understand it correctly.

e00E (approved):

pricegraph/src/orderbook/scalar.rs
nlordell: 2**23 needs 24 bits (1 with 23 0's), so the fact that the magnitude is not quite 128 is important. I also added a unit test to make sure we have enough space.

pricegraph/src/orderbook/scalar.rs
nlordell: No, leftmost bit is always 1 for negative numbers, this has to be a specific value. The reason I mentioned we "need 1 bit for ∞` is actually a little inaccurate and a bit of lie. We need it in order to have enough space for the two's complement of `7517784` while still reserving a special value for ∞.

pricegraph/src/orderbook/scalar.rs
nlordell: > but the 114.71 is rounded which makes a difference here.

Yes, but only in the comment. `7517784` was computed with the original values to avoid rounding error. Note that `7517784` is the truncated value, since this is the maximum magnitude for the floating point number (it does not mean the fractional part is 0).

nlordell: Updated the comment to make more sense.

pricegraph/src/orderbook/scalar.rs
nlordell: Updated the comment to make more sense.

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +167,188 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
...
+/// maximum number of tokens in the exchange), so the total range that must be
+/// representable by the magnitude bits is:
+/// ```text
+/// [-7517784, 7517784]
+/// ```
+///
+/// This number fits in 23 bits. However, an additional bit is needed in order
+/// to be able to represent the two's complement of 7517784 (for the --7517784)
+/// while still reserving a special value for ∞.to have a special "infinite"
+/// value which is required by the Bellman-Ford implementation.
```
nlordell, line 197:
```suggestion
/// to be able to represent the two's complement of 7517784 (for the -7517784)
/// while still reserving a special value for ∞ which is required by the
/// Bellman-Ford implementation.
```

pricegraph/src/orderbook/scalar.rs
```diff
@@ -169,3 +167,188 @@ fn assert_strictly_positive_and_finite(value: f64) -> f64 {
     debug_assert!(num::is_strictly_positive_and_finite(value));
     value
 }
+
```
fleupold, line 170: This is some very good documentation. Would it make sense to pull the weight into its own module?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FdyouDOfN1Rin2f2uSR%2Fgiphy.gif&token=everywhere" data-gitmeme-token="everywhere"></a>

PR #1376, opened 2020-08-28
Panic intead of endlessly allocating.

nlordell:
This PR adds a new failing test similar to #1370 except triggered with an actual orderbook instead of one created directly on a `petgraph::Graph` as well as a `panic` to prevent the service from running out of memory in these cases. This will be addressed in #1370 and #1374 at which point the `#[should_panic]` can be removed.

This is just a band-aid to make it so the services don't restart in case we hit these rounding issues as a bandaid fix while the true fix is being worked on.

### Test Plan

Added assert and new unit test covering it.

### DIFF
--- a/pricegraph/src/graph/shortest_paths.rs
+++ b/pricegraph/src/graph/shortest_paths.rs
@@ -68,7 +68,7 @@ where
         let mut path = Vec::with_capacity(max_path_len);
         let mut current = dest;
         while current != self.source {
-            debug_assert!(path.len() <= max_path_len, "undetected negative cycle");
+            assert!(path.len() <= max_path_len, "undetected negative cycle");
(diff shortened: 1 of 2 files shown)


e00E (approved):
> This is just a band-aid to make it so the services don't restart in case we hit these rounding issues as a bandaid fix while the true fix is being worked on.

It will still restart though, right? The panic causes the price estimator to exit which causes the pod to restart I assume.

nlordell:
> The panic causes the price estimator to exit which causes the pod to restart I assume.

:facepalm: I thought that somewhere we had a `catch_unwind` in `warp` or something.

fleupold (approved):
I like it. Let's keep in the assert for release builds as well.

PR #1383, opened 2020-08-31
Convert TokenInfoFetching to async_trait

e00E:
A welcome side effect of the conversion is that the mockall workaround
for lifetimes is no longer needed.
In this commit I also had to change one test slightly and move the the
default implementation of find_token_by_address into its own function.
Otherwise we cannot test it as mockall overrides the default
implementation.

### Test Plan
No logic change. CI.

### DIFF
--- a/services-core/src/token_info.rs
+++ b/services-core/src/token_info.rs
@@ -1,7 +1,6 @@
 use crate::models::TokenId;
 use anyhow::Result;
 use ethcontract::Address;
-use futures::future::{BoxFuture, FutureExt as _};
 use lazy_static::lazy_static;
 use std::{borrow::Borrow, collections::HashMap, num::NonZeroU128};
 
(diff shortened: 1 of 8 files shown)


services-core/src/token_info.rs
```diff
@@ -289,23 +241,14 @@ mod tests {
 
     #[test]
     fn find_token_info_by_address_finds_result() {
-        let mut info = MockTokenInfoFetching::new();
```
e00E, line 292: This is the test that changed.

services-core/src/token_info.rs
```diff
@@ -82,47 +67,12 @@ where
...
-            TokenInfoFetching_::all_ids(self)
-        }
-        fn find_token_by_symbol<'a>(
-            &'a self,
-            symbol: &'a str,
-        ) -> BoxFuture<'a, Result<Option<(TokenId, TokenBaseInfo)>>> {
-            TokenInfoFetching_::find_token_by_symbol(self, symbol)
-        }
-    }
+fn find_token_by_address(
```
e00E, line 70: This is the function I moved out of the trait.

nlordell (approved):

fleupold (approved):
Words can hardly describe how happy I am to see these wrapper traits go. But maybe a gif can:

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2Fotnqsqqzmsw7K%2Fgiphy.gif&token=penguins" data-gitmeme-token="penguins"></a>

PR #1385, opened 2020-08-31
Move StableXContract to async_trait

e00E:
Here it did not work perfectly. Mockall only allows setting immediately
ready values on expectations. However for `submit_solution` and
`send_noop_transaction` we have tests that explicitly make use of
expectations that block for some time. This is not possible with
mockall. I filed a bug report.
Until that is fixed I've kept these methods as BoxFuture.

### Test Plan
No logic change. CI.


fleupold (approved):
👍 this is the issue: https://github.com/asomers/mockall/issues/189

PR #1391, opened 2020-09-01
Add query parameter to disable rounding buffer

### LINKED ISSUE #1346: Add query parameter to disable roundin buffer
It could be useful to use the true raw data ignoring modifications we apply related to how the solver works internally. Would be a simple query parameter like "baseunit" that defaults to use the rounding buffer.

e00E:
It could be useful to use the true raw data ignoring modifications we apply related to how the solver works internally. Would be a simple query parameter like "baseunit" that defaults to use the rounding buffer.

Fixes #1346 

### Test Plan
CI and manually tested openapi comparing prices with and without the parameter.

### DIFF
--- a/price-estimator/openapi.yml
+++ b/price-estimator/openapi.yml
@@ -33,6 +33,7 @@ paths:
         - $ref: "#/components/parameters/BatchId"
         - $ref: "#/components/parameters/IgnoreAddresses"
         - $ref: "#/components/parameters/BlockNumber"
+        - $ref: "#/components/parameters/RoundingBuffer"
   /api/v1/markets/{market}/estimated-amounts-at-price/{price}:
     get:
       summary: Estimated Amounts At Price
@@ -56,6 +57,7 @@ paths:
         - $ref: "#/components/parameters/BatchId"
         - $ref: "#/components/parameters/IgnoreAddresses"
         - $ref: "#/components/parameters/BlockNumber"
+        - $ref: "#/components/parameters/RoundingBuffer"
   /api/v1/markets/{market}/estimated-best-ask-price:
     get:
       summary: Estimated Best Ask Price
@@ -75,6 +77,7 @@ paths:
         - $ref: "#/components/parameters/BatchId"
         - $ref: "#/components/parameters/IgnoreAddresses"
         - $ref: "#/components/parameters/BlockNumber"
+        - $ref: "#/components/parameters/RoundingBuffer"
   /api/v1/markets/{market}:
     get:
       summary: Market
@@ -112,6 +115,11 @@ components:
       enum: [baseunits, atoms]
       default: baseunits
       example: baseunits
+    RoundingBuffer:
+      type: string
+      enum: [enabled, disabled]
+      default: enabled
+      example: enabled
     IgnoreAddressesParameter:
       type: string
       default: ""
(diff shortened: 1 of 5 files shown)


price-estimator/openapi.yml
```diff
@@ -101,6 +104,11 @@ components:
       enum: [baseunits, atoms]
       default: baseunits
       example: baseunits
+    OrderbookKind:
+      type: string
+      enum: [withroundingbuffer, raw]
+      default: withroundingbuffer
+      example: withroundingbuffer
```
fleupold, line 111: maybe it's just me but I find `orderbookKind=withroundingbuffer` a bit weird to read. Would maybe addRoundingBuffer=true/false be more intuitive?

fleupold (approved):

price-estimator/openapi.yml
e00E: Comes down to the enum vs bool debate maybe? I don't really have a preference.

fleupold: Yes, it's almost equivalent (in the code I don't mind the verbosity and use of an enum). Just from the perspective of API user friendly-ness I find the above mentioned param a bit ugly. But no strong feelings.

nlordell: I'm OK either way, both `disableRoundingBuffer=true` and `orderbookKind=raw` sound nice to me. However, I think the former would be a little easier to understand when looking at a URL in isolation since the value of `raw` doesn't really capture that the rounding buffer is being disabled, and is the one that you would usually see since `orderbookKind=withroundingbuffer` is the default value.

nlordell (approved):

price-estimator/src/amounts_at_price.rs
```diff
@@ -6,9 +6,15 @@ pub fn order_at_price_with_rounding_buffer(
...
     pricegraph: &Pricegraph,
-    rounding_buffer: f64,
+    rounding_buffer: Option<f64>,
 ) -> Option<TransitiveOrder> {
     let order = pricegraph.order_for_limit_price(token_pair, limit_price)?;
+
+    let rounding_buffer = match rounding_buffer {
+        None => return Some(order),
+        Some(rounding_buffer) => rounding_buffer,
+    };
```
bh2smith, line 16: This looks like some kind of weird type manipulation. I read, if there is a rounding buffer, then rounding buffer is itself unwrapped, otherwise return the order.

Perhaps it would be more clear if we simply use an if statement

price-estimator/src/models/query.rs
```diff
@@ -52,6 +53,21 @@ pub enum EstimationTime {
...
 }
 
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum OrderbookKind {
+    // pricegraph instance with the original orders from the orderbook
+    Raw,
+    // pricegraph instance with the orders to which the rounding buffer has been applied
+    WithRoundingBuffer,
+}
```
bh2smith, line 63: Wasn't orderbook kind being used for something entirely different in the past? I am guessing that we no longer use the outdated kind of orderbooks and now only use the streamed one.

bh2smith (approved):

price-estimator/openapi.yml
e00E: What about `roundingBuffer=enabled` and `disabled`?

price-estimator/src/amounts_at_price.rs
e00E: If I use an if then the whole rest of the function has to be part of a new block which I'd like to avoid. I wouldn't call this a weird manipulation either, just an early return in case rounding buffer is none.

price-estimator/openapi.yml
nlordell: Love it.

PR #1392, opened 2020-09-01
Update the default log filter after `core` name change

e00E:
I forgot this when changing the name of the core crate.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -44,7 +44,7 @@ struct Options {
     #[structopt(
         long,
         env = "DFUSION_LOG",
-        default_value = "warn,driver=info,core=info"
+        default_value = "warn,driver=info,services_core=info"
     )]
     log_filter: String,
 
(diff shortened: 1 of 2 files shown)


nlordell (commented):
Can we use `module_path!` macro in `std`?

e00E:
You mean instead of `driver` and `price-estimator`?

nlordell:
> You mean instead of driver and price-estimator? It wouldn't help for services_core.

I mean we could define in `services-core/src/logging.rs` we could:
```
    pub const DEFAULT_FILTER: &str = concat!("warn,", module_path!(), "=info");
```

And then in price estimator and driver:
```
default_value = concat!(logging::DEFAULT_FILTER, "driver=info")
```

nlordell (approved):
Now that I think about it a bit more, I don't think my suggestion is worth implementing.

fleupold (approved):

PR #1394, opened 2020-09-01
Buffered Read/Write Orderbook file

### LINKED ISSUE #1352: Large allocation when writing event file to Disk
When profiling memory usage locally I can see a somewhat large allocation (64MB) when we write the event file to disk. Why this is not enough to cause an OOM this might become an issue in the future as the event file grows.

Screen Shot 2020-08-21 at 14 34 45

If possible, we should probably use a buffered write here.

bh2smith:
Fixes #1352 (out of memory error) on read-write of orderbook file. Now using a buffered reader-writer for significant performance enhancement. This PR is a collaborative effort with @nlordell 

### Test Plan
Run this one usually ignored test to check that read-write is still working as intended

```sh
cargo test test_write_read_recover_full_cycle -- --ignored
```

Run the memory profiler that @fleupold was referring to in the issue


nlordell (approved):
Looks good to me!

e00E (approved):

fleupold:
I was using [Mac Instruments](https://help.apple.com/instruments/mac/current/#) in case you still want to repro the memory reduction.

Two questions from reading the [BufWriter docs](https://doc.rust-lang.org/std/io/struct.BufWriter.html)

> BufWriter<W> can improve the speed of programs that make small and repeated write calls to the same file or network socket. It does not help when writing very large amounts at once, or writing just one or a few times.

Is this a concern given that we are writing large amounts at once here? Would be nice to verify that it breaks up allocations when writing a large blob at once.

> It is critical to call flush before BufWriter<W> is dropped. Though dropping will attempt to flush the contents of the buffer, any errors that happen in the process of dropping will be ignored. Calling flush ensures that the buffer is empty and thus dropping will not even attempt file operations.

Should we flush before dropping?

e00E:
We are not writing large amounts at once. The text is referring to `write` calls. This depends on how `bincode::serialize_into` is implemented. I'd expect internally it does do many small writes.

The flushing is a good point which we should fix.

nlordell:
> We are not writing large amounts at once. The text is referring to write calls.

Taking a quick at the code it seams each primitive is `write`-t-en separately, so the BufWriter is pretty important here IMO. We were seeing it take minutes to serialize without this.

https://github.com/servo/bincode/blob/c39317133cc6c5524eea0bb3a4efca7fb2ffd8fb/src/ser/mod.rs#L35

> Should we flush before dropping?

Great catch!

bh2smith:
As per @fleupold suggestion and bincode recommendation, I have included the flush in a separate PR (#1396)

PR #1395, opened 2020-09-01
Change weight special value

nlordell:
Running the fuzzer, I discovered an issue with the new `Weight`, specifically with the chosen special value for infinity. It turns out that the logic behind the range of values a `Weight` could have was a bit off, and the more natural choice of `i128::MAX` works better.

This PR updates the the `Weight::infiinite` special value and fixes the `Arbitrary` implementation to only provide `remaining_sell_amounts` less than, or equal the price denominator.

### Test Plan

Run the fuzzer and see it doesn't find a panic right away. Adjusted unit tests.

### DIFF
--- a/pricegraph/src/encoding.rs
+++ b/pricegraph/src/encoding.rs
@@ -136,10 +136,8 @@ mod abitrary_impl {
     impl Arbitrary for Element {
         fn arbitrary(u: &mut Unstructured<'_>) -> Result<Self> {
             let price = u.arbitrary::<PriceFraction>()?;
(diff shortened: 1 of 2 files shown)


nlordell:
Once again, fuzzer FTW!

e00E (approved):

fleupold (approved):
I noticed that our fuzzing job was no longer running in the background (not sure why). I am reactivating it.

nlordell:
> I noticed that our fuzzing job was no longer running in the background 

#1297

PR #1396, opened 2020-09-02
[Easy] Flush the writer before method finishes

bh2smith:
Related to suggestions made in #1394, we also flush the buffered writer before the method completes.

### Test Plan
Unit tests.

### DIFF
--- a/services-core/src/history/events.rs
+++ b/services-core/src/history/events.rs
@@ -93,9 +93,9 @@ impl EventRegistry {
         // Create temp file to be written completely before rename
         let temp_file = File::create(&temp_path)
             .with_context(|| format!("couldn't create {}", temp_path.display()))?;
-        let buffered_writer = BufWriter::new(temp_file);
-        self.write_to(buffered_writer)?;
-
+        let mut buffered_writer = BufWriter::new(temp_file);
+        self.write_to(&mut buffered_writer)?;
+        buffered_writer.flush()?;
         // Rename the temp file to the originally specified path.
         fs::rename(temp_path, path)?;
         Ok(())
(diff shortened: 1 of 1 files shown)


e00E (approved):

services-core/src/history/events.rs
```diff
@@ -93,9 +93,9 @@ impl EventRegistry {
         // Create temp file to be written completely before rename
         let temp_file = File::create(&temp_path)
             .with_context(|| format!("couldn't create {}", temp_path.display()))?;
-        let buffered_writer = BufWriter::new(temp_file);
-        self.write_to(buffered_writer)?;
-
+        let mut buffered_writer = BufWriter::new(temp_file);
+        self.write_to(&mut buffered_writer)?;
+        buffered_writer.flush()?;
```
nlordell, line 98: There might be an issue here with the fact that the `File` has not been `drop`-ed before the file is renamed? Maybe there should be an explicit scope so that the `buffered_writer` and its inner `File` get cleaned up:
```suggestion
        {
            let mut buffered_writer = BufWriter::new(temp_file);
            self.write_to(&mut buffered_writer)?;
            buffered_writer.flush()?;
        }
```

services-core/src/history/events.rs
```diff
@@ -93,9 +93,11 @@ impl EventRegistry {
         // Create temp file to be written completely before rename
         let temp_file = File::create(&temp_path)
             .with_context(|| format!("couldn't create {}", temp_path.display()))?;
-        let buffered_writer = BufWriter::new(temp_file);
-        self.write_to(buffered_writer)?;
-
+        {
```
fleupold, line 96: Wouldn't the context have to include the line above to make sure the file is dropped before the rename?

fleupold (approved):

services-core/src/history/events.rs
nlordell: The file is moved into the BufWriter so it will be dropped when the buffered writer is, but I like the suggestion as it makes the code less prone to accidental errors during refactoring.

nlordell (approved):

services-core/src/history/events.rs
bh2smith: Easy peasy! Great suggestion. Thanks.

PR #1397, opened 2020-09-03
Price estimator metrics

e00E:
Adds metrics to the price estimator.

We have two metrics that are measured using a global warp log handler. These count status code and response time but cannot differentiate the routes.
Then there is one metric that counts the success response time, separated for each route.

### Test Plan
Ran locally, made some requests and checked out the metrics.

### DIFF
--- a/price-estimator/src/metrics.rs
+++ b/price-estimator/src/metrics.rs
@@ -1,0 +1,64 @@
+use anyhow::Result;
+use prometheus::{Histogram, HistogramOpts, HistogramVec, IntCounterVec, Opts, Registry};
+use std::time::Instant;
+use warp::log::Info;
+
+// There are global metrics (response_status and response_time) that are measured through warp's log
+// functionality. The global metrics work on the final response and do not know about the specific
+// route that handled them.
+// And there are metrics (response_time_success) that are labelled per route. We only measures the
+// success response time here as error response times are short and uninteresting and would skew the
+// result.
+
+pub struct Metrics {
+    response_status: IntCounterVec,
+    response_time: Histogram,
+    response_time_success: HistogramVec,
+}
+
+impl Metrics {
+    pub fn new(registry: &Registry) -> Result<Self> {
+        let opts = Opts::new(
+            "price_estimator_response_status",
+            "The status code of a price estimator response.",
+        );
+        let response_status = IntCounterVec::new(opts, &["status"]).unwrap();
+        registry.register(Box::new(response_status.clone()))?;
+
+        let opts = HistogramOpts::new(
+            "price_estimator_response_time_global",
+            "The duration it takes for the price estimator to respond.",
+        );
+        let response_time = Histogram::with_opts(opts).unwrap();
+        registry.register(Box::new(response_time.clone()))?;
+
+        let opts = HistogramOpts::new(
+            "price_estimator_response_time_success",
+            "The duration it takes for the price estimator to successfully respond for each route.",
+        );
+        let response_time_success = HistogramVec::new(opts, &["route"]).unwrap();
+        registry.register(Box::new(response_time_success.clone()))?;
+
+        Ok(Self {
+            response_status,
+            response_time,
+            response_time_success,
+        })
+    }
+
+    pub fn handle_successful_response(&self, route: &str, start: Instant) {
+        let response_time = start.elapsed().as_secs_f64();
+        self.response_time_success
+            .with_label_values(&[route])
+            .observe(response_time);
+    }
+
+    pub fn handle_response(&self, info: Info<'_>) {
+        let status = info.status();
+        self.response_status
+            .with_label_values(&[status.as_str()])
+            .inc();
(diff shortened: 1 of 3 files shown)


price-estimator/src/metrics.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::Result;
+use prometheus::{Histogram, HistogramOpts, HistogramVec, IntCounterVec, Opts, Registry};
+use std::time::Instant;
+use warp::log::Info;
+
+// There are global metrics (response_status and response_time) that are measured through warp's log
+// functionality. The global metrics work on the final response and do not know about the specific
+// route that handled them.
```
fleupold, line 8: The Info struct contains the route as a field. Could we use that to log the route from the top level filter instead of hooking into each handler individually?

price-estimator/src/metrics.rs
```diff
@@ -0,0 +1,64 @@
...
+        registry.register(Box::new(response_status.clone()))?;
+
+        let opts = HistogramOpts::new(
+            "price_estimator_response_time_global",
+            "The duration it takes for the price estimator to respond.",
+        );
+        let response_time = Histogram::with_opts(opts).unwrap();
+        registry.register(Box::new(response_time.clone()))?;
+
+        let opts = HistogramOpts::new(
```
fleupold, line 35: I'm not a fan of the per route handling, but if we keep it can we rename the metric/methods to make sure it's response_time_per_route? It's well described in the comments but I'd prefer if the variable and metrics were more self explanatory.

fleupold (approved):
Looks good although I would prefer if we can handle all prometheus logging on the top level (rather than hooking into each handler and thus effectively double counting response time).

Also, I find the PR description a bit sparse.

price-estimator/src/metrics.rs
e00E: We definitely want to label per route because the response times are so different and for the same reason also separate success responses from errors.

Using the path from info was my first attempt but it has problems: We would need to parse the url again to figure out which route is being used. There is no nice way to do this except by manually inspecting url. I couldn't find a way to reuse the filters for this. So we end up writing the path handling code again, mirroring our structure from the filters.

Could you explain more why you prefer using only the top level? I don't see how double counting the response times is a problem. The response times are recorded in separate, partially overlapping metrics but that doesn't make the individual metric wrong.

e00E:
> Also, I find the PR description a bit sparse.

I added a bit more info after your comment.

price-estimator/src/metrics.rs
nlordell: I also would like the metrics handling to be at a top level (not necessarily in `main.rs`, but maybe in `filters::all`?), just to reduce the amount of plumbing required. I also really like the labeled response metrics since, as you said, these requests have vastly different response times. I would even go as far as suggesting that other things (such as query parameters) should be labeled as well (for example, there is a significant overhead to querying with `blockNumber` or `batchId` parameters). 

Maybe the sub-filters like `markets` can return a `impl Filter<Extract = (impl Reply, ResponseLabel)>` that so the metrics handling only needs to done in one place, and only the labels need to be returned by the filters? This way we don't need to actually modify the handler code to add metrics, only the filters around them - which I find is a nice and clean solution. I imagine something like `struct ResponseLabel { name: &'static str, time: EstimationTime }` or something, basically including all the important things that need to be labeled for recording response timing.

e00E: > for example, there is a significant overhead to querying with blockNumber or batchId parameters

Yes I thought about this too. We probably want to label "cached" or "now" versus "uncached" or "custom" or something like that. Actually I'm not even sure if the prod price estimator should expose those routes because it's so easy to dos with them and real users do not need them. But that's a bigger discussion.

I am going to try your idea with returning more information from the sub filters.

fleupold: > Could you explain more why you prefer using only the top level?

Mainly for convenience reasons (having to remember to add logging to each new filter) and reasoning about which metric is which (requires understanding why there are two and what the difference is).

Do you have a concern with just logging the entire path as a label? We can aggregate the ones we want using regexs in Graphana later. 
I remember we shouldn't have a completely unbound number of labels for prometheus (since each label is another series), but it feels like for us it should "only" be O(n^2) the number of tokens (due to how markets scale).

e00E: That's already too many labels and it's not taking into account that there are multiple ways to refer to the same token. Also it would include random bogus 404 urls that someone might throw at the service, and we would still have to parse the url to remove the query.

I think Nick's idea works though (still trying to make it a bit nicer).

e00E:
Implemented Nicks' idea of handling the metrics in the `all` filter.

nlordell (approved):
Nice! I like how the actual handlers didn't have to change at all for the metrics to be added.

e00E:
I'm merging this now so that I can start building the grafana dashboard since Felix is on vacation. We can still make changes afterwards obviously.

PR #1399, opened 2020-09-04
Add metric for min_avg_fee

### LINKED ISSUE #1398: report minAvgFee as prometheus metrics
We have seen occasions in the past where it is not entirely clear what the minAvgFee constraints of our system are. At the moment we log this value to kibana and can search for it there (which is cumbersome).

I think this value is important enough to expose it as a prometheus metric and expose it in Grafana. This will make debugging easier and more accessible beyond the core services team.

e00E:
Will make debugging easier by not having to check kibana logs for the
parameter.

I put this into stablex_metrics. Alternatively it could go into solver_metrics but those were intended to capture all the metrics from the solution file which this metric isn't.

Fixes  #1398 

### Test Plan
I tested that the gauge shows. Didn't test that it is actually set because I don't have a solver set up but code is simple so it should be fine.

### DIFF
--- a/services-core/src/metrics/stablex_metrics.rs
+++ b/services-core/src/metrics/stablex_metrics.rs
@@ -3,7 +3,7 @@ use crate::solution_submission::SolutionSubmissionError;
 use anyhow::Result;
 use chrono::Utc;
 use ethcontract::U256;
-use prometheus::{IntCounterVec, IntGaugeVec, Opts, Registry};
+use prometheus::{Gauge, IntCounterVec, IntGaugeVec, Opts, Registry};
 use std::collections::HashSet;
 use std::convert::TryInto;
 use std::sync::Arc;
@@ -15,6 +15,7 @@ pub struct StableXMetrics {
     orders: IntGaugeVec,
     tokens: IntGaugeVec,
     users: IntGaugeVec,
+    min_avg_fee: Gauge,
 }
 
 impl StableXMetrics {
@@ -63,6 +64,13 @@ impl StableXMetrics {
         BookType::initialize_gauges(&users);
         registry.register(Box::new(users.clone())).unwrap();
 
+        let min_avg_fee_opts = Opts::new(
+            "dfusion_service_min_avg_fee",
+            "min avg fee as passed to the solver. This is the *total generated* fee.",
+        );
+        let min_avg_fee = Gauge::with_opts(min_avg_fee_opts).unwrap();
+        registry.register(Box::new(min_avg_fee.clone())).unwrap();
+
         Self {
             processing_times,
             failures,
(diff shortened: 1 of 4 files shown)


services-core/src/metrics/stablex_metrics.rs
```diff
@@ -63,13 +64,22 @@ impl StableXMetrics {
         BookType::initialize_gauges(&users);
         registry.register(Box::new(users.clone())).unwrap();
 
+        let min_avg_fee_opts = Opts::new(
+            "dfusion_service_min_avg_fee",
+            "Computed min avg earned fee per order. This is the *earned* fee. It is doubled \
+              when passed to the solve which expects to receive the total *generated* fee.",
```
e00E, line 70: Not sure if this value, or the doubled value is more natural. What do you think?

services-core/src/metrics/stablex_metrics.rs
```diff
@@ -63,13 +64,22 @@ impl StableXMetrics {
         BookType::initialize_gauges(&users);
         registry.register(Box::new(users.clone())).unwrap();
 
+        let min_avg_fee_opts = Opts::new(
+            "dfusion_service_min_avg_fee",
+            "Computed min avg earned fee per order. This is the *earned* fee. It is doubled \
+             when passed to the solve which expects to receive the total *generated* fee.",
```
e00E, line 70: Not sure if this value, or the doubled value is more natural. What do you think?

nlordell: Personally, I would prefer the value that is passed to the solver. This way it makes it easier for the solver team to diagnose issues with batches not finding trades.

nlordell (approved):

services-core/src/metrics/stablex_metrics.rs
e00E: Changed it to the total generated fee, as passed to the solver.

services-core/src/price_finding/optimization_price_finder.rs
```diff
@@ -292,7 +298,8 @@ impl PriceFinding for OptimisationPriceFinder {
         let solver_type = self.solver_type;
         // The solver expects the fee amount as the total paid fees. Half of the paid fees are
         // burned and half earned.
-        let min_avg_paid_fee_per_order = 2 * self.economic_viability.min_average_fee().await?;
+        let min_avg_fee = 2 * self.economic_viability.min_average_fee().await?;
```
fleupold, line 301: Unrelated to this but maybe we should expose two methods `min_average_trading_fee`, `min_average_solver_reward` (one could have a default implementation calling the other) to make the difference between the two more clear.

fleupold (approved):
⚡ fast. Thanks!

PR #1402, opened 2020-09-07
Allow setting fixed min-avg-fee-per-order

e00E:
We would like to be able to set fixed amounts for the min avg fee and
max gas price. This PR achieves that by making the subsidy factor
optional. If it is not set we only use default_min_avg_fee_per_order
and default_max_gas_price are used.

### Test Plan
I changed the code to print out the min avg fee to make sure it really does return the default. After merging I'd like to test this for one batch or so on staging.

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -4,7 +4,8 @@ use services_core::driver::{
     stablex_driver::StableXDriverImpl,
 };
 use services_core::economic_viability::{
-    EconomicViabilityComputer, FixedEconomicViabilityComputer, PriorityEconomicViabilityComputer,
+    EconomicViabilityComputer, EconomicViabilityComputing, FixedEconomicViabilityComputer,
+    PriorityEconomicViabilityComputer,
 };
 use services_core::gas_price::{self, GasPriceEstimating};
 use services_core::health::{HealthReporting, HttpHealthEndpoint};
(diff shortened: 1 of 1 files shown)


driver/src/main.rs
```diff
@@ -165,14 +166,11 @@ struct Options {
...
-    /// solution as well as the gas cap for economically viable solution.
-    #[structopt(
-        long,
-        env = "ECONOMIC_VIABILITY_SUBSIDY_FACTOR",
-        default_value = "10.0"
-    )]
-    economic_viability_subsidy_factor: f64,
+    /// Subsidy factor used to compute the minimum average fee per order in a solution as well as
+    /// the gas cap for economically viable solution.
+    /// If not set then only default_min_avg_fee_per_order and default_max_gas_price are used.
```
nlordell, line 171: I find the default value for this unintuitive. To me not specifying a subsidy factor is equivalent to not subsidizing, i.e. `1.0`. Maybe this should be a separate flag (although that is also not super nice because we have enough CLI flags as it is).

e00E: @nlordell implemented it by introducing a new enum EconomicViabilityStrategy.

driver/src/main.rs
```diff
@@ -194,6 +196,17 @@ struct Options {
     #[structopt(long, env = "DEFAULT_MAX_GAS_PRICE", default_value = "100000000000")]
     default_max_gas_price: u128,
 
+    /// How to calculate the economic viability constraints. `Dynamic` means that current eth price
+    /// is taken into account while `Static` means that default_min_avg_fee_per_order and
+    /// default_max_gas_price will always be used.
+    #[structopt(
+        long,
+        env = "ECONOMIC_VIABILITY_STRATEGY",
+        possible_values = &EconomicViabilityStrategy::variants(),
```
nlordell, line 205: This is so cool.

nlordell: We should probably fix our other `structopt` properties.

nlordell (approved):
Thanks! looks great.

fleupold:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FxTiTnBMEz7zAKs57LG%2Fgiphy.gif&token=awesome" data-gitmeme-token="awesome"></a>

PR #1403, opened 2020-09-07
[EASY] Runtime argumentss; default to fallback

bh2smith:
The default-XXX runtime arguments were confusing because they have their own default values and their intended use strikes me as a "fallback" value. Thus, these two arguments name changed from default to fallback.

As an added bonus, the readme table of contents had a couple of broken links

### Test Plan
Unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,10 +143,10 @@ OPTIONS:
             Specify the number of blocks to fetch events for at a time for constructing the orderbook for the solver
             [env: AUCTION_DATA_PAGE_SIZE=]  [default: 500]
         --fallback-max-gas-price <fallback-max-gas-price>
-            The default maximum gas price. This is used when computing the maximum gas price based on ether price in owl
+            The fallback maximum gas price. This is used when computing the maximum gas price based on ether price in owl
             fails [env: FALLBACK_MAX_GAS_PRICE=]  [default: 100000000000]
         --fallback-min-avg-fee-per-order <fallback-min-avg-fee-per-order>
-            The default minimum average fee per order. This is passed to the solver in case the computing its value
+            The fallback minimum average fee per order. This is passed to the solver in case the computing its value
             fails. Its unit is [OWL] [env: MIN_AVG_FEE_PER_ORDER=]  [default: 0]
         --earliest-solution-submit-time <earliest-solution-submit-time>
             The earliest offset from the start of a batch in seconds at which point we should submit the solution. This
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -141,10 +142,10 @@ OPTIONS:
         --auction-data-page-size <auction-data-page-size>
             Specify the number of blocks to fetch events for at a time for constructing the orderbook for the solver
             [env: AUCTION_DATA_PAGE_SIZE=]  [default: 500]
-        --default-max-gas-price <default-max-gas-price>
+        --fallback-max-gas-price <fallback-max-gas-price>
             The default maximum gas price. This is used when computing the maximum gas price based on ether price in owl
```
nlordell, line 146:
```suggestion
            The fallback maximum gas price. This is used when computing the maximum gas price based on ether price in owl
```

README.md
```diff
@@ -141,10 +142,10 @@ OPTIONS:
...
-            fails [env: DEFAULT_MAX_GAS_PRICE=]  [default: 100000000000]
-        --default-min-avg-fee-per-order <default-min-avg-fee-per-order>
+            fails [env: FALLBACK_MAX_GAS_PRICE=]  [default: 100000000000]
+        --fallback-min-avg-fee-per-order <fallback-min-avg-fee-per-order>
             The default minimum average fee per order. This is passed to the solver in case the computing its value
```
nlordell, line 149:
```suggestion
            The fallback minimum average fee per order. This is passed to the solver in case the computing its value
```

nlordell (approved):
In general I like this change. This does require extra care when merging since it requires updates to our deployment configuration.

e00E (approved):
Like Nick said. Need to check whether we set this anywhere on staging.

bh2smith:
Confirmed, these values are not used in staging!

e00E:
My PR for the static min avg fee changed some code in main that uses default max gas price. That's probably why the build failed.

e00E:
I've made a change to staging that will use the max gas price env variable whose name has changed but I took this PR into account already by always setting both the old and the new name.

bh2smith:
> I've made a change to staging that will use the max gas price env variable whose name has changed but I took this PR into account already by always setting both the old and the new name.

Is there anything else that needs to be done here? Or should we be waiting for something on the other end?

e00E:
You can merge whenever. The staging change is https://gitlab.gnosisdev.com/devops/gnosis-staging/merge_requests/327 but it's compatible with this PR and without this PR, so no problem.

bh2smith:
> You can merge whenever. The staging change is https://gitlab.gnosisdev.com/devops/gnosis-staging/merge_requests/327 but it's compatible with this PR and without this PR, so no problem.

Should I not at least try and get the build to pass first?

Issue #1404, opened 2020-09-07
Implement `subsidy_factor` (or min_trade) route in price estimator.


nlordell:
The FE needs a new route for retrieving the subsidy factor. It would probably be nice to have a route that returns the minimum trade amount for a specific token directly instead of having two separate implementations.

@e00E had mentioned that this should be its own endpoint/service. I had initially agreed, but given that #1024 will require the price estimator to compute the min average trade for a specific token, and hence know about the subsidy factor. With that in mind, I think it makes sense to include this endpoint as part of the price estimator.

e00E:
The PR above added a basic version of the route that returns the result always in owl. We could extend this with a new route that works per token and returns the amount in that token.

PR #1409, opened 2020-09-08
Add route for current min order size in owl

e00E:
We want to use this for the frontend.

This commit has several todos that can be improved. We will do these
later so that the route is available as quickly as possible so we stop
blocking the frontend.

Related to #1404 .

I introduced a new trait `EthPricing` which has the `get_eth_price` method that used to be in `PriceEstimating`. This is cleaner because EconomicViability only needs this one method and makes it easier to create EconomicViability with the price source we have in PriceEstimator instead of using PriceOracle.

The todos that we can do after this PR:

1. Avoid code duplication for economic viability related arguments in driver and price estimator.
2. Avoid code duplication for creation of economic viability from arguments in driver and price estimator.
3. Introduce a network_id argument to price estimator which is needed for the gas price estimation to not be hardcoded to mainnet.
4. Create a better version of this route that works per token and returns the amount in that token instead of in owl.
5. Add openapi docs for the route.

### Test Plan
Run price estimator with `--economic-viability-subsidy-factor 5`.
`curl -i "http://localhost:8080/api/v1/minimum-order-size-owl"` returns `1838937853604742400000`
which is 1839 owl and fits with the min avg fee we pass to the solver.
To check the solver look at https://dashboard-projects.gnosisdev.com/d/BOeaAdNGk/solver-metrics?orgId=1 . Note that the open solver is currently using subsidy factor 5 while the others are using 10.

### DIFF
--- a/price-estimator/src/filter.rs
+++ b/price-estimator/src/filter.rs
@@ -7,6 +7,7 @@ use crate::{
 };
 use pricegraph::{Market, Pricegraph, TokenPair, TransitiveOrder};
 use services_core::{
+    economic_viability::EconomicViabilityComputing,
     models::TokenId,
     token_info::{TokenBaseInfo, TokenInfoFetching},
 };
@@ -18,12 +19,14 @@ pub fn all(
     orderbook: Arc<Orderbook>,
     token_info: Arc<dyn TokenInfoFetching>,
     metrics: Arc<Metrics>,
+    economic_viability: Arc<dyn EconomicViabilityComputing>,
 ) -> impl Filter<Extract = impl Reply, Error = Infallible> + Clone + Send {
     let markets = markets(orderbook.clone(), token_info.clone());
     let estimated_buy_amount = estimated_buy_amount(orderbook.clone(), token_info.clone());
     let estimated_amounts_at_price =
         estimated_amounts_at_price(orderbook.clone(), token_info.clone());
     let estimated_best_ask_price = estimated_best_ask_price(orderbook, token_info);
+    let minimum_order_size_owl = minimum_order_size_owl(economic_viability);
 
     let label = |label: &'static str| warp::any().map(move || label);
     let routes_with_labels = warp::path!("api" / "v1" / ..).and(
@@ -33,6 +36,8 @@ pub fn all(
             .or(label("estimated-amounts-at-price").and(estimated_amounts_at_price))
             .unify()
             .or(label("estimated-best-ask-price").and(estimated_best_ask_price))
+            .unify()
+            .or(label("minimum-order-size-owl").and(minimum_order_size_owl))
             .unify(),
     );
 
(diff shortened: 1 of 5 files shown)


price-estimator/src/filter.rs
```diff
@@ -68,6 +73,30 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
     Ok(warp::reply::with_status(json, code))
 }
 
+/// Validate a request of the form
+/// `/minimum-order-size-owl` and answer it.
+fn minimum_order_size_owl(
+    economic_viability: Arc<dyn EconomicViabilityComputing>,
+) -> impl Filter<Extract = (Json,), Error = Rejection> + Clone {
+    warp::path!("minimum-order-size-owl")
```
nlordell, line 81: Maybe this route could be `minimum-order-size/:token-id` and then reject if the token ID is not `0`? That way we just have this one route for minimum trades, and we can just add support for other token IDs in the future.

price-estimator/src/main.rs
```diff
@@ -98,6 +103,47 @@ struct Options {
...
+        Dynamic,
+        Static,
+    }
+}
+
+impl Default for EconomicViabilityStrategy {
+    fn default() -> Self {
+        Self::Dynamic
+    }
 }
```
nlordell, line 147: Moving this into `services-core` will be tricky because of the `clap` dependency... :thinking: Maybe this is something we can achieve with another crate that focuses on deriving things for enums?

**edit**: Just to be clear, I'm fine with this for now, we can address it as a follow up as you mentioned in the description.

price-estimator/src/main.rs
```diff
@@ -98,6 +103,47 @@ struct Options {
...
+    )]
+    economic_viability_subsidy_factor: f64,
+    #[structopt(
+        long,
+        env = "ECONOMIC_VIABILITY_MIN_AVG_FEE_FACTOR",
+        default_value = "1.1"
+    )]
+    economic_viability_min_avg_fee_factor: f64,
+    #[structopt(long, env = "MIN_AVG_FEE_PER_ORDER", default_value = "0")]
+    fallback_min_avg_fee_per_order: u128,
```
nlordell, line 123: Do you think these parameters should have doc comments as well so they show up in the `--help`?

services-core/src/economic_viability.rs
```diff
@@ -1,18 +1,23 @@
...
 const GAS_PER_TRADE: f64 = 120_000.0;
 
+#[cfg_attr(test, mockall::automock)]
+#[async_trait::async_trait]
+pub trait EthPricing {
+    /// Retrieves a price estimate for ETH in OWL atoms.
+    /// The amount of OWL in atoms to purchase 1.0 ETH (or 1e18 wei).
+    async fn get_eth_price(&self) -> Option<NonZeroU128>;
+}
+
```
nlordell, line 20: Really like this change!

nlordell (approved):
Looks really good, just some small comments.

nlordell:
I just noticed, we should add this to the OpenAPI spec.

price-estimator/src/filter.rs
e00E: I have a hunch that it could be a good idea to really keep the routes separate to make it easier to move the more advanced route in the future without breaking compatibility. This way we don't need to spend thinking about the best api design for the route with tokens now.

price-estimator/src/main.rs
e00E: In this case I intentionally removed them as a way of saying "the documentation for this is given in the driver" so that I would copy paste less. Hopefully the doc comments can be part of the deduplication effort. I don't mind adding the doc comments back either but this was my initial decision.

e00E:
> I just noticed, we should add this to the OpenAPI spec.

Agree. Will also do this in follow up PR.

price-estimator/src/filter.rs
nlordell: Sounds good to me!

price-estimator/src/main.rs
nlordell: If its going to be addressed in the de-duplication effort, then we don't really need to add it here to.

PR #1412, opened 2020-09-08
Factor out duplicated econonomic viability struct creation

e00E:
It is better to not copy paste the code.

I got rid of the clap macro because we don't want to depend on that in services_core. Alternatively we could introduce a new crate for sharing functionality between the driver and the price estimator but I felt this is a good first step anyway and the new crate needs more discussion.

### Test Plan
No logic change, checked that both binaries still run.

### DIFF
--- a/price-estimator/src/main.rs
+++ b/price-estimator/src/main.rs
@@ -14,10 +14,7 @@ use orderbook::Orderbook;
 use prometheus::Registry;
 use services_core::{
     contracts::{stablex_contract::StableXContractImpl, web3_provider},
-    economic_viability::{
-        EconomicViabilityComputer, EconomicViabilityComputing, FixedEconomicViabilityComputer,
-        PriorityEconomicViabilityComputer,
-    },
+    economic_viability::EconomicViabilityStrategy,
     gas_price,
     health::{HealthReporting, HttpHealthEndpoint},
     http::HttpFactory,
@@ -32,7 +29,7 @@ use std::{
     collections::HashMap, net::SocketAddr, num::ParseIntError, path::PathBuf, sync::Arc,
     time::Duration,
 };
-use structopt::{clap::arg_enum, StructOpt};
+use structopt::StructOpt;
 use tokio::{runtime, time};
 use url::Url;
 use warp::Filter;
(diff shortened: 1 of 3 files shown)


services-core/src/economic_viability.rs
```diff
@@ -10,6 +10,59 @@ use std::{num::NonZeroU128, sync::Arc};
...
+            "Static" => Ok(Self::Static),
+            _ => Err("invalid value"),
+        }
+    }
+}
+
+impl EconomicViabilityStrategy {
+    pub fn variants() -> &'static [&'static str] {
+        &["Dynamic", "Static"]
+    }
```
nlordell, line 34: Its a pity there is no macro for generating this without pulling in `clap`.

nlordell (approved):
Great cleanup!

services-core/src/economic_viability.rs
fleupold: nit: Maybe this method and `from_str` could depend on the same Map so that `variants` is equal to `keys` and `from_str` is just a lookup.

price-estimator/src/main.rs
```diff
@@ -58,6 +55,11 @@ struct Options {
     #[structopt(long, env = "ETHEREUM_NODE_URL")]
     node_url: Url,
 
+    /// The network ID used for gas estimations.
+    /// For example 1 for mainnet, 4 for rinkeby.
+    #[structopt(long, env = "NETWORK_ID", default_value = "1")]
```
fleupold, line 60: Not related to this PR but I think we should be able to get the network ID from the ethereum node URL with `eth_chainId` or `eth_networkId` call.

fleupold (approved):

services-core/src/economic_viability.rs
nlordell: I think we can use #1413 to do this.

services-core/src/economic_viability.rs
```diff
@@ -10,6 +10,45 @@ use std::{num::NonZeroU128, sync::Arc};
 /// much gas is used in the reversion of the previous solution.
 const GAS_PER_TRADE: f64 = 120_000.0;
 
+arg_enum! {
+    #[derive(Debug)]
+    pub enum EconomicViabilityStrategy {
+        Dynamic,
+        Static,
+    }
+}
```
nlordell, line 19: :tada:

PR #1413, opened 2020-09-08
Implement `arg_enum` macro

nlordell:
This PR adds an `arg_enum` style macro for so we don't have to manually implement `FromStr` and `variants` for argument enums. This is done since it is a fairly simple macro and so we don't have to pull in `clap` into `services-core`.

### Test Plan

Check that the new arguments parse correctly:
```
$ cargo run -p driver -- --solver-type BestRingSolver --solver-internal-optimizer Gurobi --scheduler Evm
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/driver --solver-type BestRingSolver --solver-internal-optimizer Gurobi --scheduler Evm`
2020-09-09T04:45:26.574Z INFO [driver] Starting driver with runtime options: Options {
    ...
    solver_type: BestRingSolver,
    solver_internal_optimizer: Gurobi,
    ...
    scheduler: Evm,
    ...
}
```


services-core/src/macros.rs
```diff
@@ -36,6 +36,67 @@ macro_rules! immediate {
...
+            }
+        }
+
+        impl std::str::FromStr for $name {
+            type Err = anyhow::Error;
+
+            fn from_str(value: &str) -> anyhow::Result<Self> {
+                match value {
+                    $(
+                        _ if value.eq_ignore_ascii_case(stringify!($variant)) => {
```
fleupold, line 85: nit: could even strip `_` / `-` here to make it work with the previous values (unless the arg parser already enforces that possible_values matches exactly).

fleupold (approved):
🤩

services-core/src/macros.rs
nlordell: > unless the arg parser already enforces that possible_values matches exactly

It does :disappointed:

e00E (approved):

PR #1416, opened 2020-09-09
Automatically determine chain id through web3

e00E:
instead of having to specify it on the command line.

For the contract using the chain id instead of network id is what was
intended anyway and for the gas station it doesn't matter whether we
match on network id or chain id because they are the same for mainnet
and rinkeby.

Based on Felix' suggestion.

### Test Plan
Added a debug print to the gas station to see which chain id is used and tried it with mainnet and rinkeby node url.

### DIFF
--- a/services-core/src/gas_price.rs
+++ b/services-core/src/gas_price.rs
@@ -3,7 +3,7 @@ mod gas_station;
 
 pub use self::gas_station::GnosisSafeGasStation;
 use crate::{contracts::Web3, http::HttpFactory};
-use anyhow::{Context as _, Result};
+use anyhow::Result;
 use ethcontract::U256;
 use futures::compat::Future01CompatExt as _;
 use std::sync::Arc;
@@ -20,14 +20,8 @@ pub async fn create_estimator(
     http_factory: &HttpFactory,
     web3: &Web3,
 ) -> Result<Arc<dyn GasPriceEstimating + Send + Sync>> {
-    let network_id: u64 = web3
-        .net()
-        .version()
-        .compat()
-        .await?
-        .parse()
-        .context("failed to parse network_id")?;
-    Ok(match gas_station::api_url_from_network_id(network_id) {
+    let network_id = web3.net().version().compat().await?;
+    Ok(match gas_station::api_url_from_network_id(&network_id) {
         Some(url) => Arc::new(GnosisSafeGasStation::new(http_factory, url)?),
         None => Arc::new(web3.clone()),
     })
--- a/services-core/src/gas_price.rs
+++ b/services-core/src/gas_price.rs
@@ -3,7 +3,7 @@ mod gas_station;
 
 pub use self::gas_station::GnosisSafeGasStation;
 use crate::{contracts::Web3, http::HttpFactory};
-use anyhow::Result;
+use anyhow::{Context as _, Result};
 use ethcontract::U256;
 use futures::compat::Future01CompatExt as _;
 use std::sync::Arc;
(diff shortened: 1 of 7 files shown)


services-core/src/gas_price.rs
```diff
@@ -15,12 +16,12 @@ pub trait GasPriceEstimating {
...
 
 /// Creates the default gas price estimator for the given network.
-pub fn create_estimator(
-    network_id: u64,
+pub async fn create_estimator(
     http_factory: &HttpFactory,
     web3: &Web3,
 ) -> Result<Arc<dyn GasPriceEstimating + Send + Sync>> {
-    Ok(match gas_station::api_url_from_network_id(network_id) {
+    let chain_id = web3.eth().chain_id().compat().await?.as_u64();
```
nlordell, line 23: ubernit:
```suggestion
    let network_id = web3.net().version().compat().await?;
```

I think we want to use the network ID here, and not the chain ID (which is for signing). This is really nit-picky, and my main argument here is that the version of Ganache we are using only allows specifying the network ID and not the chain ID (which is hard coded to `1337` if I am not mistaken).

```
$ docker run -d -p 8545:8545 docker.io/trufflesuite/ganache-cli:v6.7.0 -i 5777
$ printf "%d\n" $(curl -s "http://127.0.0.1:8545" -d '{"jsonrpc":"2.0","method":"eth_chainId","params":[]}' | jq -r '.result')
1337
```

services-core/src/gas_price/gas_station.rs
```diff
@@ -12,8 +12,8 @@ use uint::FromDecStrErr;
 const DEFAULT_MAINNET_URI: &str = "https://safe-relay.gnosis.io/api/v1/gas-station/";
 const DEFAULT_RINKEBY_URI: &str = "https://safe-relay.rinkeby.gnosis.io/api/v1/gas-station/";
 
-pub fn api_url_from_network_id(network_id: u64) -> Option<&'static str> {
-    match network_id {
+pub fn api_url_from_chain_id(chain_id: u64) -> Option<&'static str> {
```
nlordell, line 15: Same as above:

```suggestion
pub fn api_url_from_network_id(network_id: &str) -> Option<&'static str> {
```

nlordell (approved):
:tada:

services-core/src/gas_price.rs
e00E: I see, I got confused by the rust web3 documentation. It says "Returns protocol version" when it actually returns the network id. It also returns a string for some reason instead of a number.
I don't think it matters with the current gas station selection anyway because we only need to differentiate mainnet and rinkeby but network id is more appropriate. Maybe use chain id now and switch to network id after making a PR to rust web3 to update the method.

nlordell: > It also returns a string for some reason instead of a number

Yes, this is confusing. Its because the official Ethereum JSON RPC spec says that it returns a string, not a number: https://eth.wiki/json-rpc/API#net_version

fleupold (approved):
If we go with chain_id we make sure it also works for xDAI?

e00E:
xdai's network_id == xdai's chain_id != mainnet or rinkeby

We don't have a gas station url for xdai's id so it will fallback to using web3 to estimate the gas price.

services-core/src/gas_price.rs
```diff
@@ -15,11 +16,17 @@ pub trait GasPriceEstimating {
...
+pub async fn create_estimator(
     http_factory: &HttpFactory,
     web3: &Web3,
 ) -> Result<Arc<dyn GasPriceEstimating + Send + Sync>> {
+    let network_id: u64 = web3
+        .net()
+        .version()
+        .compat()
+        .await?
+        .parse()
```
nlordell, line 28: Ubernit: I think it's slightly more correct to use strings for network IDs, since according to the Ethereum JSON RPC spec, it is a string and not a number. While in practice this makes no difference, it may if some future chain we connect to doesn't return a number.

PR #1417, opened 2020-09-10
Pricegraph: Store order remaining amount and user balances as integers

e00E:
instead of floats.

While working on a using a BigRational type for ExchangeRate in a different branch I noticed
that this change also makes sense and could be refactored it into its own PR.

With the BigRational change Flow::capacity and min_trade would also become u128s.

Benchmark
<details>Pricegraph::read        time:   [5.6543 ms 5.6597 ms 5.6654 ms]
                        change: [-0.6112% -0.4534% -0.2954%] (p = 0.00 < 0.05)

                        time:   [5.5705 ms 5.5748 ms 5.5794 ms]
                        change: [+0.7394% +0.8957% +1.0439%] (p = 0.00 < 0.05)

                        time:   [47.531 us 47.607 us 47.670 us]
                        change: [+8.3493% +8.4750% +8.5964%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [48.837 us 48.874 us 48.922 us]
                        change: [+1.8688% +2.1218% +2.3802%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [94.161 us 94.197 us 94.240 us]
                        change: [+2.0211% +2.2254% +2.4099%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [192.46 us 192.72 us 193.05 us]
                        change: [+1.5835% +1.8608% +2.0939%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [477.25 us 477.63 us 478.07 us]
                        change: [+7.6418% +7.8084% +7.9628%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [42.607 us 42.744 us 42.921 us]
                        change: [+4.3753% +4.8784% +5.4368%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [122.03 us 122.29 us 122.61 us]
                        change: [+1.9331% +2.3399% +2.6983%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [208.11 us 208.35 us 208.58 us]
                        change: [+2.1229% +2.5886% +3.0075%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [452.56 us 452.84 us 453.11 us]
                        change: [+2.9362% +3.0678% +3.1991%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [509.79 us 510.08 us 510.46 us]
                        change: [+6.5567% +6.7537% +6.9483%] (p = 0.00 < 0.05)
                        Performance has regressed.
</details>

Benchmark now with balance as U256
<details>
Pricegraph::read        time:   [5.9581 ms 5.9664 ms 5.9754 ms]
                        change: [+2.0925% +2.6443% +3.1203%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [5.8476 ms 5.8520 ms 5.8571 ms]
                        change: [+2.1179% +2.6124% +3.0897%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [47.819 us 47.846 us 47.873 us]
                        change: [+8.1399% +8.8444% +9.4816%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [49.410 us 49.513 us 49.632 us]
                        change: [+1.9095% +2.8321% +3.6865%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [96.026 us 96.119 us 96.240 us]
                        change: [+1.8636% +2.7113% +3.3794%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [197.57 us 197.73 us 197.95 us]
                        change: [-0.4474% +0.8413% +2.0184%] (p = 0.18 > 0.05)

                        time:   [488.48 us 489.17 us 490.14 us]
                        change: [+7.3693% +8.1000% +8.7454%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [41.826 us 41.905 us 42.000 us]
                        change: [+3.6122% +4.2047% +4.7273%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [121.74 us 121.89 us 122.08 us]
                        change: [-1.0053% -0.2611% +0.3924%] (p = 0.48 > 0.05)

                        time:   [213.23 us 213.31 us 213.41 us]
                        change: [+0.1689% +1.2980% +2.2842%] (p = 0.01 < 0.05)

                        time:   [464.76 us 465.16 us 465.65 us]
                        change: [+5.2298% +5.4621% +5.6739%] (p = 0.00 < 0.05)
                        Performance has regressed.

                        time:   [521.98 us 522.61 us 523.40 us]
                        change: [+8.5424% +8.7334% +8.9668%] (p = 0.00 < 0.05)
                        Performance has regressed.
</details>

### Test Plan
Existing unit tests.

### DIFF
--- a/pricegraph/src/num.rs
+++ b/pricegraph/src/num.rs
@@ -80,13 +80,7 @@ pub fn is_strictly_positive_and_finite(value: f64) -> bool {
 
 /// Returns true if an amount is considered a dust amount. See `MIN_AMOUNT`
 /// documentation for more details.
-pub fn is_dust_amount_f(amount: f64) -> bool {
-    amount < MIN_AMOUNT as f64
-}
-
-/// Returns true if an amount is considered a dust amount. See `MIN_AMOUNT`
-/// documentation for more details.
-pub fn is_dust_amount_i(amount: u128) -> bool {
+pub fn is_dust_amount(amount: u128) -> bool {
     amount < MIN_AMOUNT
 }
 
--- a/pricegraph/src/num.rs
+++ b/pricegraph/src/num.rs
@@ -80,7 +80,13 @@ pub fn is_strictly_positive_and_finite(value: f64) -> bool {
 
 /// Returns true if an amount is considered a dust amount. See `MIN_AMOUNT`
 /// documentation for more details.
-pub fn is_dust_amount(amount: f64) -> bool {
+pub fn is_dust_amount_f(amount: f64) -> bool {
+    amount < MIN_AMOUNT as f64
+}
+
+/// Returns true if an amount is considered a dust amount. See `MIN_AMOUNT`
+/// documentation for more details.
+pub fn is_dust_amount_i(amount: u128) -> bool {
     amount < MIN_AMOUNT
 }
 
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -16,7 +16,7 @@ mod weight;
 
 pub use self::flow::{Flow, Ring};
 pub use self::iter::TransitiveOrders;
-use self::order::{Order, OrderCollector, OrderMap};
+use self::order::{Amount, Order, OrderCollector, OrderMap};
 pub use self::reduced::ReducedOrderbook;
 pub use self::scalar::{ExchangeRate, LimitPrice};
 use self::user::{User, UserMap};
@@ -362,7 +362,7 @@ impl Orderbook {
             if num::is_dust_amount(new_balance) {
                 user.clear_balance(pair.sell);
                 self.update_projection_graph_node(pair.sell);
-            } else if let Some(amount) = &mut order.amount {
+            } else if let Amount::Remaining(amount) = &mut order.amount {
                 *amount = amount.saturating_sub(fill_amount);
                 if num::is_dust_amount(*amount) {
                     self.update_projection_graph_edge(pair);
@@ -433,8 +433,8 @@ fn format_path(path: &[NodeIndex]) -> String {
 /// for trades
 fn is_dust_order(element: &Element) -> bool {
     num::is_dust_amount(element.remaining_sell_amount as _)
-        || (element.balance < U256::from(u128::MAX)
-            && num::is_dust_amount(element.balance.low_u128()))
+        || (num::is_dust_amount(element.balance.low_u128())
+            && element.balance < U256::from(u128::MAX))
 }
 
 /// An error indicating an invalid operation was performed on an overlapping
@@ -711,9 +711,8 @@ mod tests {
                 .orders
                 .best_order_for_pair(TokenPair { buy: 0, sell: 1 })
                 .unwrap()
-                .amount
-                .unwrap(),
-            1_000_000 - (flow.capacity / transitive_xrate_0_1) as u128
+                .amount,
+            Amount::Remaining(1_000_000 - (flow.capacity / transitive_xrate_0_1) as u128)
         );
         assert_eq!(
             orderbook.users[&user_id(1)].balance_of(1),
@@ -725,9 +724,8 @@ mod tests {
                 .orders
                 .best_order_for_pair(TokenPair { buy: 1, sell: 2 })
                 .unwrap()
-                .amount
-                .unwrap(),
-            1_000_000,
+                .amount,
+            Amount::Remaining(1_000_000),
         );
         assert_eq!(orderbook.users[&user_id(5)].balance_of(2), 1_000_000);
 
@@ -736,9 +734,8 @@ mod tests {
                 .orders
                 .best_order_for_pair(TokenPair { buy: 3, sell: 4 })
                 .unwrap()
-                .amount
-                .unwrap(),
-            2_000_000 - (flow.capacity / flow.exchange_rate.value()) as u128
+                .amount,
+            Amount::Remaining(2_000_000 - (flow.capacity / flow.exchange_rate.value()) as u128)
         );
         assert_eq!(
             orderbook.users[&user_id(4)].balance_of(4),
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -363,6 +363,8 @@ impl Orderbook {
                 user.clear_balance(pair.sell);
                 self.update_projection_graph_node(pair.sell);
             } else if let Amount::Remaining(amount) = &mut order.amount {
+                // TODO: add a debug assert to see that we are not over filling orders.
+                // Will do this when we use BigRational.
                 *amount = amount.saturating_sub(fill_amount);
                 if num::is_dust_amount(*amount) {
                     self.update_projection_graph_edge(pair);
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -29,6 +29,7 @@ use crate::graph::subgraph::{ControlFlow, Subgraphs};
 use crate::num;
 use petgraph::graph::{DiGraph, EdgeIndex, NodeIndex};
 use petgraph::visit::NodeIndexable;
+use primitive_types::U256;
 use std::cmp;
 use std::f64;
 use thiserror::Error;
@@ -249,7 +250,7 @@ impl Orderbook {
         while let Some(true) = self
             .orders
             .best_order_for_pair(pair)
-            .map(|order| num::is_dust_amount(order.get_effective_amount(&self.users)))
+            .map(|order| num::is_dust_amount_i(order.get_effective_amount(&self.users)))
         {
             self.orders.remove_pair_order(pair);
         }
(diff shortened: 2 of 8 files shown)


pricegraph/src/num.rs
```diff
@@ -80,7 +80,13 @@ pub fn is_strictly_positive_and_finite(value: f64) -> bool {
...
 /// Returns true if an amount is considered a dust amount. See `MIN_AMOUNT`
 /// documentation for more details.
-pub fn is_dust_amount(amount: f64) -> bool {
+pub fn is_dust_amount_f(amount: f64) -> bool {
+    amount < MIN_AMOUNT as f64
+}
+
+/// Returns true if an amount is considered a dust amount. See `MIN_AMOUNT`
+/// documentation for more details.
+pub fn is_dust_amount_i(amount: u128) -> bool {
```
e00E, line 89: Let me know if you prefer a different name for these functions. I *think* after the BigRational change we will only have 1 version again.

pricegraph/src/orderbook/order.rs
```diff
@@ -127,7 +125,8 @@ pub struct Order {
     /// The maximum capacity for this order, this is equivalent to the order's
     /// remaining sell amount. Note that orders are also limited by their user's
     /// balance.
-    pub amount: f64,
+    /// None means unlimited.
+    pub amount: Option<u128>,
```
e00E, line 129: We could also use u128::MAX to signal unbounded. That saves space in the struct because there is no extra byte needed for the Option.
I originally used Option but now that I write this I think u128::MAX is better. Could also make amount private and use a pub function to do the subtraction which would enforce this invariant.

pricegraph/src/orderbook/user.rs
```diff
@@ -11,49 +11,47 @@ pub type UserMap = Map<UserId, User>;
 #[derive(Clone, Debug, Default)]
 pub struct User {
     /// User balances per token.
-    balances: Map<TokenId, f64>,
+    balances: Map<TokenId, u128>,
```
e00E, line 14: Technically user balances are U256. We round too large balances to u128::MAX. In practice I don't think this is a problem because individual trades are limited to u128::MAX.

nlordell: This should probably be a `U256` or even `BigInt` since the balance can be used for multiple orders.

pricegraph/src/orderbook/order.rs
nlordell: `u128::MAX` is a reserved contract value for an unbounded amount. Maybe we should create a new `Amount(u128)` wrapper type to represent this?

e00E: It has special meaning in `Element`. It had no special meaning in `Order` up until this point. When we create an `Order` we check for this.

pricegraph/src/num.rs
nlordell: Now that I think of it, maybe we should just make this
```
pub fn is_dust_amount(amount: u128) -> bool { ... }
```

And then replace calls with `is_dust_amount(some_f64_value as _)`.

pricegraph/src/orderbook/order.rs
nlordell: > It had no special meaning in Order up until this point. 

Its because we converted `u128::MAX` to `f64::INFINITY` when we created it. We need to keep this behaviour since unlimited orders shouldn't decrease their `amount` when they get partially filled (only user balance).

e00E: The same way I'm checking for `Some(amount)` now I would change this to check for `u128::MAX`. Will push a new commit that has this soon so we can look at it.

nlordell: My suggestion for the wrapper type was just to have a nice abstraction over this type and avoid creating invalid values. For example, `Some(u128::MAX)` is an invalid value (just like `2.0.powi(129)` would be with how it is currently written). But yes, its not strictly necessary, just thought it would be a nice time to make the code better :smile:

e00E: Looks like that isn't actually faster in benchmarks and Option<u128> is more clear so let's keep the option. Or could create a new enum Amount Limited Unlimited.

pricegraph/src/orderbook/user.rs
e00E: Changed it to U256.

pricegraph/src/orderbook/order.rs
nlordell: My only issue is that `Some(u128::MAX)` is an invalid value, so we have to make `amount` private to make sure this invariant is upheld. That's why I suggested a wrapper type in order to make it clear that it is a unsigned integer with "infinity" semantics. Although the same can be achieved with making `Order::amount` private and having an accessor method to it, and upholding the `!= u128::MAX` invariant in the update method.

pricegraph/src/orderbook.rs
```diff
@@ -438,7 +433,8 @@ fn format_path(path: &[NodeIndex]) -> String {
 /// for trades
 fn is_dust_order(element: &Element) -> bool {
     num::is_dust_amount(element.remaining_sell_amount as _)
-        || num::is_dust_amount(num::u256_to_f64(element.balance))
+        || (element.balance < U256::from(u128::MAX)
+            && num::is_dust_amount(element.balance.low_u128()))
```
nlordell, line 437: ubernit: Since it is way more likely for balances to be less than `u128::MAX` I would do the fast path first:
```suggestion
        || (num::is_dust_amount(element.balance.low_u128())
             && element.balance < U256::from(u128::MAX))
```

Not sure if this actually has a performance impact though.

pricegraph/src/orderbook/user.rs
```diff
@@ -62,3 +59,30 @@ impl User {
         self.balances.remove(&token);
     }
 }
+
+fn u256_to_u128_saturating(u256: &U256) -> u128 {
+    if u256.0[2] == 0 && u256.0[3] == 0 {
```
nlordell, line 64: ubernit: It might make this easier to understand if we used some higher-level abstractions over `U256`.
```suggestion
    cmp::min(u256, u128::MAX.into()).low_u128()
```

I would make sure that this doesn't incur a performance impact though.

nlordell (approved):
Very nice! Loving this change. My only complaint is about upholding `Order::amount` invariants closer to the data, and be in `Order` methods.

pricegraph/src/orderbook/order.rs
nlordell: So, reading the code, I agree and think that the `Option` looks really nice. My only complaint is that it isn't clear from outside of `Order` what `amount == None` means. I suggest that the `if let Some(amount) = &mut order.amount { *amount = amount.saturating_sub(fill_amount); }` logic be moved a method on the `Order` type, and that `amount` be private.

pricegraph/src/orderbook.rs
```diff
@@ -355,23 +356,17 @@ impl Orderbook {
...
+            let fill_amount = (flow.capacity / transitive_xrate.value()) as u128;
             let new_balance = user.deduct_from_balance(pair.sell, fill_amount);
 
             if num::is_dust_amount(new_balance) {
                 user.clear_balance(pair.sell);
                 self.update_projection_graph_node(pair.sell);
-            } else if num::is_dust_amount(order.amount) {
-                self.update_projection_graph_edge(pair);
+            } else if let Some(amount) = &mut order.amount {
+                *amount = amount.saturating_sub(fill_amount);
```
nlordell, line 366: nit: Can we re-introduce the `debug_assert` check on `amount -= fill_amount` to make sure that we aren't over-filling orders? I guess that right now this is hard because some of this is in floats and rounding errors can cause this to fail. Maybe we can leave a `TODO` to put the assertion back in once we switch to rational numbers?

pricegraph/src/orderbook/order.rs
fleupold: Agree that using option here is practical but the None case is an implementation detail that we don't need to leak to users. Otherwise I'd advertise for our own Option like type that is either Unbounded or Some(u128).

fleupold (approved):
Thanks for attaching the benchmarks. Could you summarize their outcome (I assume no significant regression)? When reading through the logs it says: `Performance has regressed.` in both runs, so I guess they were not executed directly after one another.

pricegraph/src/orderbook/order.rs
e00E: I've made an Amount enum. Still about as practical as the Option but obvious what it means.

e00E:
> Thanks for attaching the benchmarks. Could you summarize their outcome (I assume no significant regression)? When reading through the logs it says: `Performance has regressed.` in both runs, so I guess they were not executed directly after one another.

Both benchmarks are comparing to master. Summary is only small performance regression <10% which is fine because the individual tests are still fast.

pricegraph/src/orderbook/order.rs
nlordell: While this is easier to understand, I still don't like how it's possible to create invalid values like `Amount::Remaining(u128::MAX)`

PR #1421, opened 2020-09-11
Filter orders with buy token == sell token

e00E:
In the pricegraph high level operations we already check that the two tokens in the market are different. I think the same makes sense for the orders in the orderbook as buying and selling the same token does not make sense.

Edit: This is already enforced in the smart contract so it probably makes more sense to enforce this when parsing the Element but we 'd have to add another possible error type to that function.

### Test Plan
Existing tests still pass and new test.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -61,6 +61,7 @@ impl Orderbook {
         for (order, element) in elements
             .into_iter()
             .filter(|element| !is_dust_order(element))
+            .filter(|element| element.pair.buy != element.pair.sell)
             .filter_map(|element| Order::new(&element).map(move |order| (order, element)))
         {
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Great catch

e00E:
@nlordell do you think this is better handled in the filter as it is now or as another error in the element parsing function?

nlordell:
Hmm, good question. The exchange currently rejects `buyToken == sellToken` so it would also make sense for it to be an error in the `Element` parsing. That being said, it would require `buy` and `sell` properties on `TokenPair` to be private (so an invalid token pair can't be created), and for it to have a constructor that returns an `Option` which is significantly more work. I'm OK with the current implementation for now, but a checked `TokenPair` type that only accepts valid token pairs is more "correct" IMO.

PR #1449, opened 2020-09-29
Make CI enforce that help output is up to date in README

fleupold:
Inspired by #1429 this PR adds a step to the travis config that makes sure we have the exact output of `cargo run --bin driver -- --help\` in the README.

This is done by diffing the output of the command and content of README.md and making sure that README only contains additions.

### Test Plan
```sh
git checkout contracts_4.3
! diff README.md <(cargo run --bin driver -- --help) | grep -E "^>" || (echo "[ERROR] README doesn't contain output of \`cargo run --bin driver -- --help\`"; false)
```

outputs 

```sh
  Finished dev [unoptimized + debuginfo] target(s) in 0.22s
  Running `target/debug/driver --help`
>     driver [OPTIONS] --node-url <node-url> --private-key <private-key>
[ERROR] README doesn't contain output of `cargo run --bin driver -- --help`
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
         # Unit Tests and Linting
         - cargo test
         # Make sure README is up to date
-        - (! diff README.md <(cargo run --bin driver -- --help) | grep -E "^>" || (echo "[ERROR] README.md doesn't contain output of \`cargo run --bin driver -- --help\`"; false))
+        - diff <(sed -n '/^driver /,/^```/p' README.md | ghead -n -1) <(cargo run --bin driver -- --help)
         # Build image with compiled binary
         - docker build --tag stablex-binary-public --build-arg SOLVER_BASE=gnosispm/dex-open-solver:$OPEN_SOLVER_VERSION -f driver/docker/rust/Dockerfile .
         # StableX e2e Tests (Ganache) - open solver
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
         # Unit Tests and Linting
         - cargo test
         # Make sure README is up to date
-        - diff <(sed -n '/^driver /,/^```/p' README.md | ghead -n -1) <(cargo run --bin driver -- --help)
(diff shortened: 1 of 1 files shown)


nlordell (approved):
This is awesome!

nlordell:
It looks like CI is failing because `diff` is returning a non-0 exit code. You could use `sed` to just print the lines starting from the driver help code block until the next <code>```</code>:
```bash
diff <(sed -n '/^driver 0.1.0/,/^```/p' README.md | head -n -1) <(cargo run --bin driver -- --help)
```

bh2smith (approved):
🆒 😎

fleupold:
> It looks like CI is failing because diff is returning a non-0 exit code.

The reason seems to be that travis is ignoring the ! which should turn the non-0 exit code into a zero one. I will try again with brackets around it. I also like your command because it's doing the diff just on the code that is actually important. My only minor concern is that it would break if we ever updated to driver 0.1.1 or the like.

nlordell:
> concern is that it would break if we ever updated to driver 0.1.1 or the like.

You could always `sed -n '/^driver \d+.\d+.\d+/,/^```/p'`

nlordell:
One issue with removing all new lines is that you get false negatives if we remove options:
```
$ git diff
diff --git a/README.md b/README.md
index 7b6fee9..c55f0c4 100644
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ FLAGS:
 
 
 OPTIONS:
+        --some-no-longer-supported-option
         --auction-data-page-size <auction-data-page-size>
             Specify the number of blocks to fetch events for at a time for constructing the orderbook for the solver
             [env: AUCTION_DATA_PAGE_SIZE=]  [default: 500]
$ (! diff README.md <(cargo run --bin driver -- --help) | grep -E "^>" || (echo "[ERROR] README.md doesn't contain output of \`cargo run --bin driver -- --help\`"; false))
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/driver --help`
$ echo $?
0
```

PR #1460, opened 2020-10-06
Use services_core http transport instead of web3 default http transport

e00E:
In preparation for the ethcontract upgrade. In the new ethcontract
version we get a new web3 version whose default http transport uses
hyper (tokio) which breaks the code here that uses `futures::block_on`
because these futures can only be executed on the tokio runtime.
We avoid this by using the custom http transport we already have for the
driver. Its futures can be run on any executor.
Another benefit of this change that is that hyper and tokio don't
need to be compiled for the e2e test because in the next web3 version this is a feature that can be disabled.

### Test Plan
CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,9 +12,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.13"
+version = "0.7.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2Fw48WeMIN73QsM%2Fgiphy.gif&token=snape" data-gitmeme-token="snape"></a>

bh2smith (approved):

PR #1461, opened 2020-10-06
Use tokio executor in deploy binary

e00E:
Like in #1460 the new web3 version we will get through the ethcontract
upgrade will have futures that must be run on tokio if the default http
transport is used.
In this case we cannot use services_core's http transport because we
would end up with circular dependencies. We could resolve that by
creating a new crate just for our custom http transport. Or we could
move the deploy binary into it's own subcrate so that it can depend on
contracts and services_core.
I don't mind doing that later but using tokio and keeping the the
default web3 http transport is easier initially.

### Test Plan
CI runs the deploy script.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.32"
+version = "1.0.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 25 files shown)


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F3o7abB06u9bNzA8lu8%2Fgiphy.gif&token=lgtm" data-gitmeme-token="lgtm"></a>

josojo (approved):

bh2smith (approved):

PR #1465, opened 2020-10-07
Upgrade ethcontract

e00E:
The new version uses the real future type which allows us to get rid of
`compat`. It also forwards web3's default transport implementations
which use hyper and tokio. We usually do not need this feature which
saves compilation time because we have our own transport implementation.

### Test Plan
CI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,3 @@ default-members = [
     "price-estimator",
     "pricegraph",
 ]
-
-[patch.crates-io]
-# Needed to make coverage work on MacOS (https://github.com/servo/core-foundation-rs/pull/357)
-core-foundation = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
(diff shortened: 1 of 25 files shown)


Cargo.toml
```diff
@@ -19,8 +19,3 @@ default-members = [
     "price-estimator",
     "pricegraph",
 ]
-
-[patch.crates-io]
-# Needed to make coverage work on MacOS (https://github.com/servo/core-foundation-rs/pull/357)
-core-foundation = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
-core-foundation-sys = { git = "https://github.com/servo/core-foundation-rs", tag="v0.20.1"}
```
e00E, line 26: Removed this because I got a warning when building
> ... was not used in the crate graph.

This probably means the dependency is no longer used so overriding it doesn't make sense.

fleupold (approved):
Nice, thanks for digging through all the roadblocks and separating them in different PRs.

bh2smith (approved):

PR #1468, opened 2020-10-09
Fix warnings for new rust version

e00E:
Yesterday there was a no rust version release which introduced two
warnings our code.
In scalar.rs we got https://rust-lang.github.io/rust-clippy/master/index.html#derive_ord_xor_partial_ord
and in search_batches we use `matches!` instead of `match`.

### Test Plan
CI


fleupold (approved):
🎉

PR #1477, opened 2020-10-19
Add flag to ignore external price sources

fleupold:
On rinkeby we occasionally see negative utility solutions which lead to a subtraction overflow (cf. https://github.com/gnosis/dex-solver/issues/325)

The assumption is that external price estimates are too far off the weird prices we might see on an economically not incentivised test net like rinkeby. Therefore the plan is to ignore these external source on rinkeby going forward.

### Test Plan

```bash
cargo run --bin driver -- --node-url https://xdai.poanetwork.dev  --rpc-timeout 60000 --orderbook-file target/orderbook_xdai.bin --private-key c87509a1c067bbde78beb793e6fa76530b6382a4c0241e5e4a9ec0a0f44dc0d3 --solver-type opensolver --native-token-id 0 --log-filter "warn,driver=info,services_core=info,services_core::price_estimation=debg" --ignore-external-price-source true
```

See no log entries coming from `price_estimation::clients::generic_client`

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -242,7 +242,7 @@ struct Options {
 
     /// Whether to ignore external price sources (e.g. 1Inch, Kraken etc)
     /// when estimating token prices
-    #[structopt(long, env="IGNORE_EXTERNAL_PRICE_SOURCE")]
+    #[structopt(long, env="IGNORE_EXTERNAL_PRICE_SOURCE", parse(try_from_str), default_value="false")]
     ignore_external_price_source: bool,
 }
 
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -239,6 +239,11 @@ struct Options {
     /// target chain (e.g. WETH on mainnet, DAI on xDAI).
     #[structopt(long, env = "NATIVE_TOKEN_ID", default_value = "1")]
     native_token_id: u16,
+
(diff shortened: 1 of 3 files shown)


driver/src/main.rs
```diff
@@ -239,6 +239,11 @@ struct Options {
     /// target chain (e.g. WETH on mainnet, DAI on xDAI).
     #[structopt(long, env = "NATIVE_TOKEN_ID", default_value = "1")]
     native_token_id: u16,
+
+    /// Whether to ignore external price sources (e.g. 1Inch, Kraken etc)
+    /// when estimating token prices
+    #[structopt(long, env="IGNORE_EXTERNAL_PRICE_SOURCE", parse(try_from_str), default_value="false")]
+    ignore_external_price_source: bool,
```
e00E, line 246: instead of ignore_... with default false we could avoid the negative and call it use_external_price_sources with default true. Not sure which I like it more, so just throwing it out there.

e00E (approved):

driver/src/main.rs
fleupold: 👍 In the beginning I thought I could use the negative so that not specifying the flag means the defaults. But it turned out that this didn't work with ENV vars which we need in our kubernetes config. Changed it...

PR #1479, opened 2020-10-20
Add mix of dynamic and static economic viability strategy

e00E:
called DynamicBoundedByStatic: Use Dynamic first. If it fails or the
result is worse (larger min-avg-fee, lower max-gas-price) than Static,
use Static instead.

This is useful because for ease of use we want to guarantee a minimum
viable order size but still lower it when this is economically viable
because of low gas price.

Note that this doesn't affect our running solvers until we switch them to the new strategy.

### Test Plan
New unit tests

### DIFF
--- a/services-core/src/driver/stablex_driver.rs
+++ b/services-core/src/driver/stablex_driver.rs
@@ -35,7 +35,7 @@ pub struct StableXDriverImpl {
     price_finder: Arc<dyn PriceFinding + Send + Sync>,
     orderbook_reader: Arc<dyn StableXOrderBookReading>,
     solution_submitter: Arc<dyn StableXSolutionSubmitting + Send + Sync>,
-    economic_viability: Arc<dyn EconomicViabilityComputing + Send + Sync>,
+    economic_viability: Arc<dyn EconomicViabilityComputing>,
     metrics: Arc<StableXMetrics>,
 }
 
@@ -44,7 +44,7 @@ impl StableXDriverImpl {
         price_finder: Arc<dyn PriceFinding + Send + Sync>,
         orderbook_reader: Arc<dyn StableXOrderBookReading>,
         solution_submitter: Arc<dyn StableXSolutionSubmitting + Send + Sync>,
-        economic_viability: Arc<dyn EconomicViabilityComputing + Send + Sync>,
+        economic_viability: Arc<dyn EconomicViabilityComputing>,
         metrics: Arc<StableXMetrics>,
     ) -> Self {
         Self {
--- a/services-core/src/economic_viability.rs
+++ b/services-core/src/economic_viability.rs
@@ -4,6 +4,7 @@
 use crate::{gas_price::GasPriceEstimating, models::solution::EconomicViabilityInfo};
 use anyhow::{anyhow, Context as _, Result};
 use ethcontract::U256;
+use futures::future;
 use std::{num::NonZeroU128, sync::Arc};
 
 /// The approximate amount of gas used in a solution per trade. In practice the value depends on how
@@ -13,8 +14,9 @@ const GAS_PER_TRADE: f64 = 120_000.0;
 arg_enum! {
     #[derive(Debug)]
     pub enum EconomicViabilityStrategy {
-        Dynamic,
         Static,
+        Dynamic,
+        DynamicBoundedByStatic,
     }
 }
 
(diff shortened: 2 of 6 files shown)


services-core/src/driver/stablex_driver.rs
```diff
@@ -35,7 +35,7 @@ pub struct StableXDriverImpl {
     price_finder: Arc<dyn PriceFinding + Send + Sync>,
     orderbook_reader: Arc<dyn StableXOrderBookReading>,
     solution_submitter: Arc<dyn StableXSolutionSubmitting + Send + Sync>,
-    economic_viability: Arc<dyn EconomicViabilityComputing + Send + Sync>,
+    economic_viability: Arc<dyn EconomicViabilityComputing>,
```
bh2smith, line 38: Noob question: What is the difference now without `Send` and `Sync` and why did we need them before, but not now?

services-core/src/economic_viability.rs
```diff
@@ -13,8 +14,9 @@ const GAS_PER_TRADE: f64 = 120_000.0;
 arg_enum! {
     #[derive(Debug)]
     pub enum EconomicViabilityStrategy {
-        Dynamic,
         Static,
+        Dynamic,
+        DynamicBoundedByStatic,
```
bh2smith, line 19: Could we keep this alphabetical? Or is there a reason for this reordering?

services-core/src/economic_viability.rs
```diff
@@ -30,22 +32,29 @@ impl EconomicViabilityStrategy {
...
-        economic_viabilities.push(Box::new(FixedEconomicViabilityComputer::new(
+    ) -> Arc<dyn EconomicViabilityComputing> {
+        let fixed = FixedEconomicViabilityComputer::new(
             Some(fallback_min_avg_fee_per_order),
             Some(fallback_max_gas_price.into()),
-        )));
-        PriorityEconomicViabilityComputer::new(economic_viabilities)
+        );
+        let dynamic = EconomicViabilityComputer::new(
+            native_token_price,
```
bh2smith, line 41: I assume `native_token_price` is specified at runtime? Although I suppose it would be possible to fetch this price based on the existing orderbook for token 0, perhaps this would be too involved and not guaranteed to be deterministic.

services-core/src/driver/stablex_driver.rs
e00E: Good question. On the trait `EconomicViabilityComputing` we already had the requirement `Send + Sync` (which means that it is safe to use from multiple threads). So even before this PR, specifying this requirement here again wasn't necessary. I noticed that while making this change so included it.

services-core/src/economic_viability.rs
e00E: I don't have a strong preference either way. Semantically I thought this made sense because it is the same order as we explain the strategies in the readme. There I use this order because the explanation of dynamic depends on static.

services-core/src/economic_viability.rs
e00E: The logic of how we get native_token_price hasn't changed. From the perspective of this function it's just a trait so this function doesn't know how it's implemented. Iirc we use the usual token price prediction here which is third party price estimators and also the onchain orderbook.

services-core/src/economic_viability.rs
```diff
@@ -229,11 +238,67 @@ impl EconomicViabilityComputing for PriorityEconomicViabilityComputer {
...
+        Future: std::future::Future<Output = Result<T>>,
+    {
+        let (first, second) = future::join(
+            operation(self.first.as_ref()),
+            operation(self.second.as_ref()),
+        )
+        .await;
+        let second = second?;
+        Ok(match first {
+            Ok(first) => comparison(first, second),
```
bh2smith, line 271: I see here that comparison is passed in as a function which is a "binary operator"! This is cute.

bh2smith (approved):
Looks nice!

services-core/src/economic_viability.rs
```diff
@@ -298,4 +363,61 @@ mod tests {
...
+            .expect_min_average_fee()
+            .return_once(|| Err(anyhow!("")));
+        let mut second = MockEconomicViabilityComputing::new();
+        second.expect_min_average_fee().return_once(|| Ok(5));
+        let bounded = BoundedEconomicViabilityComputer::new(Box::new(first), Box::new(second));
+        assert_eq!(
+            bounded.min_average_fee().now_or_never().unwrap().unwrap(),
+            5
+        );
+    }
```
fleupold, line 422: This file is getting quite large. Maybe it makes sense to split up the three implementations into separate submodules?

services-core/src/economic_viability.rs
```diff
@@ -229,11 +238,67 @@ impl EconomicViabilityComputing for PriorityEconomicViabilityComputer {
...
+    ) -> Result<T>
+    where
+        Future: std::future::Future<Output = Result<T>>,
+    {
+        let (first, second) = future::join(
+            operation(self.first.as_ref()),
+            operation(self.second.as_ref()),
+        )
+        .await;
+        let second = second?;
```
fleupold, line 269: I find it quite arbitrary that we fall back to the second one if the first fails but fail if the second passes. Maybe making it clear that one is the `bound` (which we fall back to) and the other one is the `optimistic` (I'm lacking a good name here) would make that seem less arbitrary?

fleupold (approved):
One concern here is that we could now have to worst out of both worlds. If e.g. if our gas estimator is off (showing a lower than actual value) we would derive a small `min_average_fee` but would then continue to generously double our gas price up to the `fallback_max_gas_price`. 

It would be nice if we could use one strategy per batch but that seems a bit hard to do in a clean way.

e00E:
> One concern here is that we could now have to worst out of both worlds.

This is a good point and I don't want to merge until this is fixed. An idea is to change how we specify the fallback gas price. Currently this is one absolute value but it makes more sense as something in relation to fees earned and orders in solution. If it was specified that way then a lower min avg fee wouldn't cause problems because it would just scale down the fallback gas price. I am currently thinking more about this and if there is a good solution will make a PR for that first.

fleupold:
We could store the "current dominant" strategy on the instance and only decide to switch that whenever we compute the min avg fee. It does introduce some implicit timing assumptions (minAvgFee called before maxGasPrice) but maybe that's an ok tradeoff to make.

e00E:
I made another idea here https://github.com/gnosis/dex-services/pull/1481

PR #1480, opened 2020-10-21
Give more control over economic viability computation

e00E:
Instead of storing an EconomicViabilityComputing in
OptimizationPriceFinder we pass the min_avg_fee directly in the
find_prices method.
This makes the price finding code simpler because it has to access one
les external interface and gives us more control over the economic
viability strategy in StableXDriver.
I plan on making use of to implement a new economic viability strategy
which is a mix of dynamic and static.

Related to #1479 

### Test Plan
Existing tests, no logic change

### DIFF
--- a/driver/src/main.rs
+++ b/driver/src/main.rs
@@ -310,7 +310,6 @@ fn main() {
         Some(Fee::default()),
         options.solver_type,
         price_oracle,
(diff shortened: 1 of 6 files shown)


fleupold (approved):
Looks good

PR #1481, opened 2020-10-21
Add mix of dynamic and static economic viability strategy

e00E:
which uses the better of the dynamic and the static strategy.

This is useful because for ease of use we want to guarantee a minimum
viable order size but still lower it when this is economically viable
because of low gas price.

successor of https://github.com/gnosis/dex-services/pull/1479 (sorry for making you review again, but the implementation is different enough that I think it's a good idea)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,9 +153,10 @@ OPTIONS:
             slightly between solution computation and submission [env: ECONOMIC_VIABILITY_MIN_AVG_FEE_FACTOR=]
             [default: 1.1]
         --economic-viability-strategy <economic-viability-strategy>
-            How to calculate the economic viability constraints. `Dynamic` means that current native token price is
-            taken into account while `Static` means that static_min_avg_fee_per_order and static_max_gas_price will
-            always be used [env: ECONOMIC_VIABILITY_STRATEGY=]  [default: Dynamic]  [possible values: Dynamic, Static]
+            How to calculate the economic viability constraints. `Static`: Use fallback_min_avg_fee_per_order and
+            fallback_max_gas_price. `Dynamic`: Use current native token price, gas price and subsidy factor. `Combined`:
+            Use the better (lower min-avg-fee) of the above [env: ECONOMIC_VIABILITY_STRATEGY=]  [default: Dynamic]
+            [possible values: Static, Dynamic, Combined]
         --economic-viability-subsidy-factor <economic-viability-subsidy-factor>
             Subsidy factor used to compute the minimum average fee per order in a solution as well as the gas cap for
             economically viable solution [env: ECONOMIC_VIABILITY_SUBSIDY_FACTOR=]  [default: 10.0]
--- a/services-core/src/economic_viability.rs
+++ b/services-core/src/economic_viability.rs
@@ -13,8 +13,9 @@ const GAS_PER_TRADE: f64 = 120_000.0;
 arg_enum! {
     #[derive(Debug)]
     pub enum EconomicViabilityStrategy {
-        Dynamic,
         Static,
+        Dynamic,
+        Combined,
     }
 }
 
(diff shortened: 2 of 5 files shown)


fleupold:
👍 Looks like a reasonable option too me (I don't really see the advantage of renaming the fixedComputer to a staticComputer but 🤷)

services-core/src/economic_viability.rs
```diff
@@ -31,29 +32,31 @@ impl EconomicViabilityStrategy {
         native_token_price: Arc<dyn NativeTokenPricing + Send + Sync>,
         gas_station: Arc<dyn GasPriceEstimating + Send + Sync>,
     ) -> Result<Arc<dyn EconomicViabilityComputing>> {
+        let make_dynamic = || {
```
e00E, line 35: The reason for making this a helper closure is that in the match statement below it avoids code duplication.

services-core/src/economic_viability.rs
```diff
@@ -171,12 +174,12 @@ fn gas_price_cap(native_token_price: f64, earned_fee: f64, num_trades: usize) ->
 
 /// Fixed values.
 pub struct FixedEconomicViabilityComputer {
-    min_average_fee: Option<u128>,
-    max_gas_price: Option<U256>,
+    min_average_fee: u128,
+    max_gas_price: U256,
```
e00E, line 178: I don't see a reason to ever store None and it makes the code simpler.

bh2smith: Seems legit and fine by me!

services-core/src/economic_viability.rs
```diff
@@ -187,13 +190,41 @@ impl FixedEconomicViabilityComputer {
...
+pub struct CombinedEconomicViabilityComputer {
+    fixed: FixedEconomicViabilityComputer,
+    // This is always DynamicViabilityComputing but for easier mock testing we use box.
+    dynamic: Box<dyn EconomicViabilityComputing>,
+}
+
+#[async_trait::async_trait]
+impl EconomicViabilityComputing for CombinedEconomicViabilityComputer {
+    async fn min_average_fee(&self) -> Result<u128> {
+        let dynamic = self.dynamic.min_average_fee().await?;
```
fleupold, line 211: Do we really want to fail if the dynamic fails? Why not just fallback to fixed?

services-core/src/economic_viability.rs
```diff
@@ -250,4 +282,52 @@ mod tests {
...
+
+    #[test]
+    fn combined_strategy_picks_correct_max_gas_price() {
+        let fixed_gas = U256::from(1);
+        let fixed_fee = 15u128;
+        let dynamic_gas = U256::from(2);
+        // In the first run the average fee of the solution is >= the fixed_fee so the fixed gas
+        // price should be chosen and in the second run < so the dynamic should be chosen.
+        for (earned_fee, expected_gas) in &[(15, 1), (14, 2)] {
+            let fixed = FixedEconomicViabilityComputer::new(fixed_fee, fixed_gas);
```
fleupold, line 312: ubernit: Everything until info could go outside the loop to make the distinction of the two test cases shorter. 

I'd even prefer two separate tests `combined_strategy_picks_fixed_gas_price_if_fee_greater_fixed_average_fee` and `combined_strategy_picks_dynamic_gas_price_if_fee_smaller_fixed_average_fee` (which would avoid the vague word "correct" in the test description). but I remember you don't like duplicated code and it's not a big deal.

fleupold (approved):

services-core/src/economic_viability.rs
e00E: That makes sense. Changed it to warn log and fallback.

services-core/src/economic_viability.rs
e00E: Moved the code outside of the loop.

README.md
```diff
@@ -153,9 +153,10 @@ OPTIONS:
             slightly between solution computation and submission [env: ECONOMIC_VIABILITY_MIN_AVG_FEE_FACTOR=]
             [default: 1.1]
         --economic-viability-strategy <economic-viability-strategy>
-            How to calculate the economic viability constraints. `Dynamic` means that current native token price is
-            taken into account while `Static` means that static_min_avg_fee_per_order and static_max_gas_price will
-            always be used [env: ECONOMIC_VIABILITY_STRATEGY=]  [default: Dynamic]  [possible values: Dynamic, Static]
+            How to calculate the economic viability constraints. `Static`: Use fallback_min_avg_fee_per_order and
+            fallback_max_gas_price. `Dynamic`: Use current native token price, gas price and subsidy factor. `Combined`:
+            Use the better (lower min-avg-fee) of the above [env: ECONOMIC_VIABILITY_STRATEGY=]  [default: Dynamic]
+            [possible values: Static, Dynamic, Combined]
```
bh2smith, line 159: Here, we replaced static with fallback. Investigation complete.

PR #1485, opened 2020-10-22
Fix build on master

e00E:
From travis I can see that the error message is that the lock file is
out of date. Somehow the dependabot must have not updated the lock file
correctly.
To fix this I ran `cargo check` on master.

### Test Plan
CI


fleupold (approved):
Thanks for the quick fix. Would be curious to learn how the PR branch passed in the first place (as it should also check the lock file)

e00E:
I agree. This is the PR branch: https://github.com/gnosis/dex-services/pull/1482 . I am going to take a look.

PR #1486, opened 2020-10-22
Fix PR CI being less strict than master CI

e00E:
To ensure that Cargo.lock is up to date we build with `--locked` on CI.
However the setup_contracts script which is run only for PRs but not for
master didn't have this flag which resulted in a PR being accepted into
master where Cargo.lock was not up to date.

https://github.com/gnosis/dex-services/pull/1482
https://github.com/gnosis/dex-services/pull/1485

### Test Plan
CI

### DIFF
--- a/ci/setup_contracts.sh
+++ b/ci/setup_contracts.sh
@@ -3,4 +3,4 @@
 set -e
 
 docker-compose up -d ganache-cli
-(cd contracts; cargo run --bin deploy --features bin)
+(cd contracts; cargo run --locked --bin deploy --features bin)


fleupold (approved):
But aren't we compiling the entire workspace with the locked flag? How could the deploy binary slip through there?

https://github.com/gnosis/dex-services/blob/c7be392090358ec4b187838cb7702277b4906336/.travis.yml#L38

e00E:
What I think happened is that first we run setup_contracts.sh which builds without --locked , so this updated the Cargo.lock file in the PR which made the following locked workspace build succeed.
On master we only run the locked build without without setup_contracts so there Cargo.lock doesn't get updated and the locked build fails.

fleupold:
That makes perfect sense - great 🕵️‍♂️ work!

bh2smith (approved):

PR #1487, opened 2020-10-22
Do not use fallback min_avg_fee in Dynamic strategy

e00E:
Initially we only had the dynamic strategy and used fallback min_avg_fee
and max_gas_price in case getting the gas price or ether price would
fail.
Now these options are also used for the static strategy.
I want to remove them from the dynamic strategy because it is misleading
to use the same config value for different things and we rely on a
working gas station and token price estimation anyway. It would be
unexpected that we fallback to very different fallback values (see our default fallbacks) when the
strategy is Dynamic. It is better to have this be a real error.
It is also better to have a hard error on startup if the Static strategy is selected but the static values haven't been given.

This is a config change because I renamed the arguments. I am going to do the rename in staging before merging this.

### Test Plan
Run driver


services-core/src/economic_viability.rs
```diff
@@ -26,26 +26,29 @@ impl EconomicViabilityStrategy {
...
+        static_max_gas_price: Option<u128>,
         native_token_price: Arc<dyn NativeTokenPricing + Send + Sync>,
         gas_station: Arc<dyn GasPriceEstimating + Send + Sync>,
-    ) -> impl EconomicViabilityComputing {
-        let mut economic_viabilities = Vec::<Box<dyn EconomicViabilityComputing>>::new();
-        if matches!(self, EconomicViabilityStrategy::Dynamic) {
-            economic_viabilities.push(Box::new(EconomicViabilityComputer::new(
+    ) -> Result<Arc<dyn EconomicViabilityComputing>> {
+        Ok(match self {
+            Self::Dynamic => Arc::new(EconomicViabilityComputer::new(
```
fleupold, line 35: Should we also assert here that static_max_gas_price and static_min_avg_fee_per_order are not set to avoid semantically undefined config state?

fleupold (approved):

services-core/src/economic_viability.rs
e00E: That sounds useful, added it.

README.md
```diff
@@ -154,17 +154,11 @@ OPTIONS:
...
             always be used [env: ECONOMIC_VIABILITY_STRATEGY=]  [default: Dynamic]  [possible values: Dynamic, Static]
         --economic-viability-subsidy-factor <economic-viability-subsidy-factor>
             Subsidy factor used to compute the minimum average fee per order in a solution as well as the gas cap for
             economically viable solution [env: ECONOMIC_VIABILITY_SUBSIDY_FACTOR=]  [default: 10.0]
-        --fallback-max-gas-price <fallback-max-gas-price>
-            The fallback maximum gas price. This is used when computing the maximum gas price based on ether price in
-            owl fails [env: FALLBACK_MAX_GAS_PRICE=]  [default: 100000000000]
-        --fallback-min-avg-fee-per-order <fallback-min-avg-fee-per-order>
-            The fallback minimum average fee per order. This is passed to the solver in case the computing its value
-            fails. Its unit is [OWL] [env: FALLBACK_MIN_AVG_FEE_PER_ORDER=]  [default: 0]
```
bh2smith, line 167: Here, I see we renamed these, but removed the default values.

README.md
```diff
@@ -154,17 +154,11 @@ OPTIONS:
             [default: 1.1]
         --economic-viability-strategy <economic-viability-strategy>
             How to calculate the economic viability constraints. `Dynamic` means that current native token price is
-            taken into account while `Static` means that fallback_min_avg_fee_per_order and fallback_max_gas_price will
+            taken into account while `Static` means that static_min_avg_fee_per_order and static_max_gas_price will
```
bh2smith, line 157: This was a correct change, but somehow got reverted back to fallback in a later PR.

PR #1492, opened 2020-10-27
Calculate gas limit based on number of trades

e00E:
instead of setting it to a fixed 6 million. This could mean faster
transaction confirmation as miners might be more likely to include a
transaction with a lower gas limit.

One problem I see is that there could be a large solution (30 trades) that we need to revert but our solution only has 2 trades. In this case the estimate would likely be wrong because we only scale it with the number of new trades but without any information about trades that need to be reverted. This situation seems unlikely so I'm not sure if that means we don't want this change. I would like to discuss that in the PR.

The worst case is 30 trades reverted, 30 trades submitted, which is probably what the 6 million has been calculated at. We could still improve this estimate by using 30 orders reverted and actual number of orders submitted.
We could also check whether we're going to revert another solution and how many trades it has by querying the contract but this will not give us a certain result as the other solution's transaction hasn't been mined yet.

### Test Plan
Existing unit tests.

### DIFF
--- a/services-core/src/contracts/stablex_contract.rs
+++ b/services-core/src/contracts/stablex_contract.rs
@@ -136,6 +136,7 @@ pub trait StableXContract: Send + Sync {
         claimed_objective_value: U256,
         gas_price: U256,
         nonce: U256,
+        gas_limit: U256,
     ) -> BoxFuture<'a, Result<(), MethodError>>;
 
     async fn past_events(
@@ -260,6 +261,7 @@ impl StableXContract for StableXContractImpl {
         claimed_objective_value: U256,
         gas_price: U256,
         nonce: U256,
+        gas_limit: U256,
     ) -> BoxFuture<'a, Result<(), MethodError>> {
         async move {
             let (prices, token_ids_for_price) = encode_prices_for_contract(&solution.prices);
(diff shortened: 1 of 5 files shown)


bh2smith (approved):
Cool idea! Nice clean implementation.

fleupold (commented):
Given our solver landscape, I would say it is somewhat unlikely that a solution with less trades will yield a significantly higher objective value. How about taking the gas estimate * 2 so that for each trade in *this* solution we could revert a trade in a previous solution?

fleupold:
We could also use the max of the RPC gas estimate and our heuristic, this way we would set a high enough limit if the racing tx was already mined.

e00E:
I like the times-two idea. Using the RPC gas estimate also is a bit more complicated than I want to make this PR but can keep that in mind for later.
For times-two a question is whether the economic viability and the gas limit for the transaction should use the same value. Are we saying, solutions are only economically viable if they can revert a solution with the same number of trades first? I would say yes.  I will implement that in this PR.

fleupold:
> Are we saying, solutions are only economically viable if they can revert a solution with the same number of trades first?

I'm not sure. I think it might be ok to be optimistic here as we don't revert that often (and could probably prevent most of it if our solvers were communicating amongst each another). But then you are right in order to guarantee economic viability we would have to be able to pay both.

nlordell:
> We could still improve this estimate by using 30 orders reverted and actual number of orders submitted.

I like this idea.

PR #1493, opened 2020-10-27
Add more fine grained gas price estimation method

e00E:
In preparation for using more advanced gas estimation this commit adds a
new method to the trait which estimates gas price based on the
transaction's gas use and time limit.

After this is merged I want to implement the trait for other "gas stations" than the gnosis one.

### Test Plan
Ci. New function is not used anywhere.

### DIFF
--- a/services-core/src/gas_price.rs
+++ b/services-core/src/gas_price.rs
@@ -5,13 +5,15 @@ pub use self::gas_station::GnosisSafeGasStation;
 use crate::{contracts::Web3, http::HttpFactory};
 use anyhow::Result;
 use ethcontract::U256;
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait GasPriceEstimating {
-    /// Retrieves gas prices from the Gnosis Safe Relay api.
-    async fn estimate_gas_price(&self) -> Result<U256>;
+    /// Estimate the gas price for a transaction to be mined "quickly".
+    async fn estimate(&self) -> Result<U256>;
+    /// Estimate the gas price for a transaction that uses <gas> to be mined within <time_limit>.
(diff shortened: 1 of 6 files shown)


services-core/src/gas_price.rs
```diff
@@ -5,13 +5,15 @@ pub use self::gas_station::GnosisSafeGasStation;
...
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait GasPriceEstimating {
-    /// Retrieves gas prices from the Gnosis Safe Relay api.
-    async fn estimate_gas_price(&self) -> Result<U256>;
+    /// Estimate the gas price for a transaction to be mined "quickly".
+    async fn estimate_simple(&self) -> Result<U256>;
```
fleupold, line 14: I really don't like the name `_simple` here as it doesn't give any information about what it does. Maybe we can just call it estimate and the other one `estimate_with_limit` or call this one `estimate_fast_execution`?

services-core/src/gas_price.rs
```diff
@@ -5,13 +5,15 @@ pub use self::gas_station::GnosisSafeGasStation;
...
+    /// Estimate the gas price for a transaction that uses <gas> to be mined within <time_limit>.
+    async fn estimate(&self, gas: U256, time_limit: Duration) -> Result<U256>;
```
fleupold, line 16: ubernit: maybe rename gas to `gas_limit`

fleupold (approved):
Looks good modulo some name pedantry...

services-core/src/gas_price.rs
e00E: Both functions do the same thing but one does it based on more parameters. That's why I picked "simple" and implicitly "complex". Could also be coarse and fine grained. But i have no strong feelings about the name so I'll go with your suggestion "estimate" and "estimate_with_limits".

PR #1494, opened 2020-10-27
Refactor Unbounded logic out of bellman ford

fleupold:
Part of #971

This PR prepares for the feedback given in #1368 by moving the existing unbounded bellman ford logic into its own file. It abstracts all necessary methods for both bounded/unbounded searches behind the `PredecessorStoring` trait, so that the next PR just needs to implement this trade for the bounded algorithm and add unit tests.

### Test Plan

CI as no logic should have changed.


pricegraph/src/graph/shortest_paths.rs
```diff
@@ -2,51 +2,65 @@
...
+        -> Option<NegativeCycle<G::NodeId>>;
 }
 
 /// Structure that can be used to derive the shorthest path from a source to any
 /// reachable destination in the graph.
-pub struct ShortestPathGraph<G: GraphBase + Data> {
+pub struct ShortestPathGraph<'a, G: Data> {
     graph: G,
-    predecessor_store: PredecessorStore<G>,
+    predecessor_store: Box<dyn PredecessorStoring<G> + 'a>,
```
e00E, line 52: This could slow down the compiled coded because the compiler might be unable to see through the trait object (`dyn`) which prevents inlining. To fix that, it should be a type parameter `<T: PredecessorStoring>`. But then I can see that this would make the code wants to use ShortestPathGraph harder because we would have to turn it into a trait and a return a `Box<dyn ShortestPathGraph>` when creating it.
Maybe performance impact is very small and we can still do this change later so this isn't a blocker.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -177,31 +139,33 @@ where
...
-    G: NodeCount + IntoNodeIdentifiers + IntoEdges + NodeIndexable,
+    G: 'a + NodeCount + IntoNodeIdentifiers + IntoEdges + NodeIndexable,
+    G::NodeId: Ord,
     G::EdgeWeight: FloatMeasure,
 {
     let mut shortest_path_graph = ShortestPathGraph::empty(g, source);
 
     // scan up to |V| - 1 times.
-    for _ in 1..g.node_count() {
+    for _ in 1..=g.node_count() - 1 {
```
e00E, line 154: Why this change?

pricegraph/src/orderbook.rs
```diff
@@ -164,7 +163,11 @@ impl Orderbook {
 
         let (base, quote) = (node_index(market.base), node_index(market.quote));
 
-        while let Err(cycle) = ShortestPathGraph::new(&self.projection, quote) {
+        loop {
+            let cycle = match ShortestPathGraph::new(&self.projection, quote) {
+                Ok(_) => break,
+                Err(cycle) => cycle,
+            };
```
e00E, line 170: I liked the previous version more but no strong opinion.

pricegraph/src/graph/shortest_paths.rs
```diff
@@ -64,80 +78,40 @@ where
...
+    fn prepare_next_relaxation_step(&mut self) {
+        self.predecessor_store.prepare_next_relaxation_step();
+    }
+
     /// Checks for negative weight cycle and, if any is found, creates loop in
     /// the predecessor store
     fn mark_cycle(&mut self) -> Option<G::NodeId> {
-        // This is the last step of the Bellman-Ford algorithm. It tries to relax
-        // each node: if a node can be relaxed then a negative cycle exists and is
-        // created in the predecessor store; otherwise no such cycle exists.
```
e00E, line 127: This comment is gone. Not sure if we need to have it just pointing out in case that was done by accident.

pricegraph/src/orderbook.rs
fleupold: Me too. However, I got another weird compiler error, because a `a temporary with access to the immutable borrow is created` in `ShortestPathGraph::new` and then still alive when I try to `self.fill_path` which also requires an immutable borrow.

Happy for suggestions how to work around that.

pricegraph/src/graph/shortest_paths.rs
fleupold: Good point. It was taken from how we changed the statement in the bounded case.

pricegraph/src/graph/shortest_paths.rs
fleupold: 👍 will put it back

pricegraph/src/graph/shortest_paths.rs
fleupold: I can benchmark it...

pricegraph/src/orderbook.rs
e00E: For me the error is more specific (maybe you can update rustc). It says `the immutable borrow might be used here, when that temporary is dropped and runs the destructor for type `std::result::Result<graph::shortest_paths::ShortestPathGraph ...`.
With that I can see why the compiler doesn't like it and why the match helps. It is something that rustc could improve but I don't think there is a better workaround. Maybe I'll make a short reproduction and create an issue for rustc.

pricegraph/src/orderbook.rs
```diff
@@ -130,17 +130,16 @@ impl Orderbook {
     /// Reduces the orderbook by matching all overlapping ring trades.
     pub fn reduce_overlapping_orders(mut self) -> ReducedOrderbook {
         Subgraphs::new(self.projection.node_indices()).for_each(|token| loop {
-            match ShortestPathGraph::new(&self.projection, token) {
+            let cycle = match ShortestPathGraph::new(&self.projection, token) {
```
fleupold, line 133: This change is related to the same borrow checker error as below.

pricegraph/src/graph/shortest_paths.rs
fedgiac: It should not overflow in the context because the function input `source` is a node of the graph.

pricegraph/src/graph/shortest_paths.rs
fleupold: Performance for this PR regressed much more than what @fedgiac saw in his, so we might want to take a look at improving this:

||This PR|#1368
|---|---|---
|Pricegraph::read | +74.929%|+25.719%
|Pricegraph::transitive_orderbook/5298183|+75.250%|+25.110%
|Pricegraph::estimate_limit_price/100000000000000000000 |+16.503|+14.982%
|Pricegraph::estimate_limit_price/1000000000000000000000|+21.946%|+19.747%
|Pricegraph::order_for_limit_price/100|+25.230%|+10.584%

I took a crack at implementing the flattending but am not able to resolve the type errors I'm seeing (made a commit [here](https://github.com/gnosis/dex-services/commit/af88a0d089a9802c316c99b02fd0a8dde41c7f46) with my current state)

fleupold: I now have a compilable version of the flattened code (needed to make `ShortestPathGraph::empty/new` free functions), but am still seeing the same performance as in this PR.

My changes are on top of the next PR and committed in https://github.com/gnosis/dex-services/commit/4a48742cf71dcc4ebceb26a88a27732d1ef3fd1d

fleupold: @e00E helped me figure out a way how to get rid of the dynamic dispatch inside the `bellman_ford` method and thus reducing the overhead to similar (even slightly better) levels than the original PR (#1368).

This does however affect a bunch of callsites of the ShortestPath Graph so I would like to do it in a separate PR on top of #1495 if that is ok with you...

e00E (approved):

fedgiac (approved):
The changes look good to me.

I'm not familiar enough with Rust to comment on the overall efficiency.

nlordell (approved):
Looks good!

PR #1496, opened 2020-10-29
Add GasNow gas price estimator

e00E:
They predict based on the pending transaction queue and the currently
mining block (in sparkpool).
This estimation should be more accurate than the gnosis gas station and
react more quickly to changes in gas price.

This is part of an effort to have better gas price estimates (also for oba). I will also make a similar implementation based on pending blocks reported by our ethereum node.

Btw this service also offers sending transactions directly to their mining pool. This allows for potentially faster transactions because the transactions and also potentially private transactions where they ensure that only they mine the transaction and do not broadcast.
Just noting that, not saying we should use it.

### Test Plan
New tests.

### DIFF
--- a/services-core/src/gas_price/gas_station.rs
+++ b/services-core/src/gas_price/gas_station.rs
@@ -1,4 +1,5 @@
 //! Gnosis Safe gas station `GasPriceEstimating` implementation.
+//! Api documentation at https://safe-relay.gnosis.io/ .
 
 use super::GasPriceEstimating;
 use crate::http::{HttpClient, HttpFactory, HttpLabel};
@@ -63,12 +64,8 @@ impl GnosisSafeGasStation {
 
 #[async_trait::async_trait]
 impl GasPriceEstimating for GnosisSafeGasStation {
-    async fn estimate(&self) -> Result<U256> {
-        Ok(self.gas_prices().await?.fast)
-    }
-
     async fn estimate_with_limits(&self, _gas_limit: U256, _time_limit: Duration) -> Result<U256> {
-        self.estimate().await
+        Ok(self.gas_prices().await?.fast)
     }
 }
 
--- a/services-core/src/gas_price/linear_interpolation.rs
+++ b/services-core/src/gas_price/linear_interpolation.rs
@@ -1,0 +1,61 @@
+/// Linearly interpolate `value` between `points`.
+///
+/// `points` must not be empty and contain unique x values sorted in ascending order.
+/// If `value` is smaller than the first point or larger than the last it is clamped.
+pub fn interpolate(value: f64, points: &[(f64, f64)]) -> f64 {
+    assert!(!points.is_empty());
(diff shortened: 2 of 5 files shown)


services-core/src/gas_price/linear_interpolation.rs
```diff
@@ -0,0 +1,40 @@
...
+        points.last().unwrap().1
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use assert_approx_eq::assert_approx_eq;
+
+    const LINEAR_DATA: &[(f64, f64)] = &[(1.0, 1.0), (2.0, 2.0), (3.0, 3.0)];
```
fleupold, line 28: I think a test case where the delta(0,1) != delta(1,2) would be interesting (and even what happens if the y values are not ascending).

services-core/src/gas_price/linear_interpolation.rs
```diff
@@ -0,0 +1,40 @@
+/// Linearly interpolate `value` between `points`.
+///
+/// `points` must not be empty and contain unique x values sorted in ascending order.
+/// If `value` is smaller than the first point or larger than the last it is clamped.
+pub fn interpolate(value: f64, points: &[(f64, f64)]) -> f64 {
+    assert!(!points.is_empty());
+    assert!(points.windows(2).all(|window| window[0].0 < window[1].0));
```
fleupold, line 7: Does this work with an array of size 1? Can we add a test for it?

services-core/src/gas_price/gas_station.rs
```diff
@@ -63,12 +64,8 @@ impl GnosisSafeGasStation {
 
 #[async_trait::async_trait]
 impl GasPriceEstimating for GnosisSafeGasStation {
-    async fn estimate(&self) -> Result<U256> {
-        Ok(self.gas_prices().await?.fast)
-    }
-
     async fn estimate_with_limits(&self, _gas_limit: U256, _time_limit: Duration) -> Result<U256> {
-        self.estimate().await
+        Ok(self.gas_prices().await?.fast)
```
fleupold, line 68: Should we (separately) implement something similar here where fast, fastest etc. map to different target times?

fleupold (approved):
Very nice.

services-core/src/gas_price/gas_station.rs
e00E: We probably will. I'm not exactly sure yet what it would be because the gnosis safe gas station is defined over gas price percentiles in previous blocks which isn't a direct indicator of time.

services-core/src/gas_price/linear_interpolation.rs
e00E: added

services-core/src/gas_price/linear_interpolation.rs
e00E: added

PR #1497, opened 2020-10-29
Enforce requirements on interpolation input more explicitly

e00E:
Instead of relying on the assert we use a wrapper struct which
implements TryFrom. This prevents accidental panics if the assertion
fails by making it more clear when the requirements on points are
checked and turning them into an `Error`.

### Test Plan
New unit tests.


fleupold (approved):
Great change!

nlordell (approved):
:tada:

PR #1498, opened 2020-10-29
Add ethgasstation gas price estimator

e00E:
Another gas estimator that will probably be useful to compare to other implementations.

### Test Plan
Not yet used but added test with real request.

### DIFF
--- a/services-core/src/gas_price/ethgasstation.rs
+++ b/services-core/src/gas_price/ethgasstation.rs
@@ -1,0 +1,100 @@
+use super::{linear_interpolation, GasPriceEstimating};
+use crate::http::{HttpClient, HttpFactory, HttpLabel};
+use anyhow::Result;
+use ethcontract::U256;
+use isahc::http::uri::Uri;
+use pricegraph::num;
+use std::{convert::TryInto, time::Duration};
+
+// Gas price estimation with https://ethgasstation.info/ , api https://docs.ethgasstation.info/gas-price .
+
+const API_URI: &str = "https://ethgasstation.info/api/ethgasAPI.json";
+
+pub struct EthGasStation {
+    client: HttpClient,
+    uri: Uri,
+}
+
+// gas prices in gwei*10 (2 gwei is transmitted as `20`)
+// wait times in minutes
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct Response {
+    fastest: f64,
+    fast: f64,
+    average: f64,
+    safe_low: f64,
+    fastest_wait: f64,
+    fast_wait: f64,
+    avg_wait: f64,
+    safe_low_wait: f64,
+}
+
+impl EthGasStation {
+    #[allow(dead_code)]
+    fn new(http_factory: &HttpFactory) -> Result<Self> {
+        let client = http_factory.create()?;
+        let uri = Uri::from_static(API_URI);
+        Ok(Self { client, uri })
+    }
+
+    async fn gas_price(&self) -> Result<Response> {
+        self.client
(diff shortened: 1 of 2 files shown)


services-core/src/gas_price/ethgasstation.rs
```diff
@@ -0,0 +1,79 @@
...
+// Gas price estimation with https://ethgasstation.info/ , api https://docs.ethgasstation.info/gas-price .
+
+const API_URI: &str = "https://ethgasstation.info/api/ethgasAPI.json";
+
+pub struct EthGasStation {
+    client: HttpClient,
+    uri: Uri,
+}
+
+// gas prices in gwei*10 (2 gwei is transmitted as `20`)
```
fleupold, line 18: What a weird API behavior.

fleupold (approved):

services-core/src/gas_price/ethgasstation.rs
```diff
@@ -0,0 +1,79 @@
+use super::{linear_interpolation, GasPriceEstimating};
+use crate::http::{HttpClient, HttpFactory, HttpLabel};
+use anyhow::Result;
+use ethcontract::U256;
+use isahc::http::uri::Uri;
+use pricegraph::num;
+use std::{convert::TryInto, time::Duration};
+
+// Gas price estimation with https://ethgasstation.info/ , api https://docs.ethgasstation.info/gas-price .
```
nlordell, line 9: Do we need, like the Gnosis Safe gas station, have different URLs for mainnet/rinkeby? At least something along the lines of:
```rust
pub fn api_url_from_network_id(network_id: &str) -> Option<&'static str> {
    match network_id {
        "1" => Some(API_URI),
        _ => None,
    }
}
```

services-core/src/gas_price/ethgasstation.rs
nlordell: :rofl:

services-core/src/gas_price/ethgasstation.rs
```diff
@@ -0,0 +1,79 @@
...
+#[async_trait::async_trait]
+impl GasPriceEstimating for EthGasStation {
+    async fn estimate_with_limits(&self, _gas_limit: U256, time_limit: Duration) -> Result<U256> {
+        let response = self.gas_price().await?;
+        let time_limit_in_minutes = time_limit.as_secs_f64() / 60.0;
+        let points: &[(f64, f64)] = &[
+            (response.fastest_wait, response.fastest),
+            (response.fast_wait, response.fast),
+            (response.avg_wait, response.average),
+            (response.safe_low_wait, response.safe_low),
```
nlordell, line 57: Its nice that they have expected wait time for each gas price! It makes interpolation easier.

Do you think there is value in adding extra points (`fastest * 2` and `safe_low / 2`) for the same reasons as the Gnosis Safe gas station?

nlordell (approved):
Nice!

services-core/src/gas_price/ethgasstation.rs
e00E: Yes, this and the GasNow gas station don't support rinkeby. When I integrate them so that they are actually used I'll add a check that the network id is mainnet.

services-core/src/gas_price/ethgasstation.rs
e00E: For ethgasstation the fastest estimation is already meant to be as quickly as possible (they call it "trader" on the website) and it updates its estimates quickly. So in this case I don't think it is  needed. If you are already at a gas price where you're better than most transactions you will not get more advantage by going even higher.

PR #1500, opened 2020-10-29
Implement gas price interpolation based on time for gnosis safe gas station

e00E:
You can see in the code that I had to make assumptions on how the gnosis safe estimates map to transaction confirmed time. I'm happy to hear other ideas on how to do this.

Edit: It looks like the file shows up as deleted and re-created instead of moved on github. If you want to see it as a diff  you can look at the second commit individually instead of going through `Files changed`.

### Test Plan
Tested that real request still works and print out  gas estimations:
```
GasPrices { last_update: "2020-10-29T11:58:47.411059Z", lowest: 1000000001.0, safe_low: 76000000001.0, standard: 85000000001.0, fast: 93000000001.0, fastest: 4710211381864.0 }
gas price estimate for 0 seconds: 186.000000002 gwei
gas price estimate for 10 seconds: 139.5000000015 gwei
gas price estimate for 20 seconds: 93.000000001 gwei
gas price estimate for 30 seconds: 85.000000001 gwei
gas price estimate for 40 seconds: 80.500000001 gwei
gas price estimate for 50 seconds: 76.000000001 gwei
gas price estimate for 60 seconds: 75.30909091008182 gwei
gas price estimate for 70 seconds: 74.61818181916364 gwei
gas price estimate for 80 seconds: 73.92727272824546 gwei
gas price estimate for 90 seconds: 73.23636363732727 gwei
```


services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -24,22 +25,32 @@ pub fn api_url_from_network_id(network_id: &str) -> Option<&'static str> {
...
-    #[serde(deserialize_with = "deserialize_u256_from_string")]
-    pub safe_low: U256,
-    #[serde(deserialize_with = "deserialize_u256_from_string")]
-    pub standard: U256,
-    #[serde(deserialize_with = "deserialize_u256_from_string")]
-    pub fast: U256,
-    #[serde(deserialize_with = "deserialize_u256_from_string")]
-    pub fastest: U256,
+    #[serde(with = "display_fromstr")]
+    pub lowest: f64,
```
e00E, line 33: I changed this to f64 because it is easier to work with and we don't lose any precision because numbers aren't that high and we are converting to float anyway.

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -64,32 +75,47 @@ impl GnosisSafeGasStation {
...
 #[async_trait::async_trait]
 impl GasPriceEstimating for GnosisSafeGasStation {
-    async fn estimate_with_limits(&self, _gas_limit: U256, _time_limit: Duration) -> Result<U256> {
-        Ok(self.gas_prices().await?.fast)
+    // The default implementation calls estimate_with_limits with 30 seconds which would result in
+    // the standard time instead of fast. So to keep that behavior we implement it manually.
+    async fn estimate(&self) -> Result<U256> {
+        let response = self.gas_prices().await?;
+        Ok(num::f64_to_u256(response.fast))
     }
```
e00E: Not sure if this is the best approach...

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -0,0 +1,157 @@
...
+fn estimate_with_limits(
+    response: &GasPrices,
+    _gas_limit: U256,
+    time_limit: Duration,
+) -> Result<f64> {
+    // We insert two extra points for the linear interpolation because this gas estimator reacts
+    // slowly to gas price changes which means that for example in times of rising gas prices
+    // fast might not be fast enough.
+    let points: &[(f64, f64)] = &[
+        (0.0, response.fast * 2.0),
```
fleupold, line 101: Isn't there also `fastest` and `lowest`?

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -0,0 +1,157 @@
...
+
+const FAST_PERCENTILE: f64 = 0.75;
+const STANDARD_PERCENTILE: f64 = 0.5;
+const SAFE_LOW_PERCENTILE: f64 = 0.3;
+
+const SECONDS_PER_BLOCK: f64 = 15.0;
+// Treat percentiles as probabilities for geometric distribution.
+const FAST_TIME: f64 = SECONDS_PER_BLOCK * 1.0 / FAST_PERCENTILE;
+const STANDARD_TIME: f64 = SECONDS_PER_BLOCK * 1.0 / STANDARD_PERCENTILE;
+const SAFE_LOW_TIME: f64 = SECONDS_PER_BLOCK * 1.0 / SAFE_LOW_PERCENTILE;
```
fleupold, line 52: Can you comment with a bit more context why the expected time is seconds per block divided by percentile and why a geometric distribution is reasonable? Also is the `* 1.0` needed.

@Uxio0 might also have some thoughts here on how the safe gas estimate levels translate into expected execution times.

services-core/src/gas_price/gnosis_safe.rs
nlordell: Makes sense. Also, I think for the kinds of calculations you are doing, `U256` is not well suited.

nlordell (approved):
Looks good. Like @fleupold I have questions around:
```rust
const FAST_PERCENTILE: f64 = 0.75;
const STANDARD_PERCENTILE: f64 = 0.5;
const SAFE_LOW_PERCENTILE: f64 = 0.3;

const SECONDS_PER_BLOCK: f64 = 15.0;
// Treat percentiles as probabilities for geometric distribution.
const FAST_TIME: f64 = SECONDS_PER_BLOCK * 1.0 / FAST_PERCENTILE;
const STANDARD_TIME: f64 = SECONDS_PER_BLOCK * 1.0 / STANDARD_PERCENTILE;
const SAFE_LOW_TIME: f64 = SECONDS_PER_BLOCK * 1.0 / SAFE_LOW_PERCENTILE;
```

Also is there is reason why `fastest` and `slowest` are left out and not used for points when interpolating gas prices/time?

services-core/src/gas_price/gnosis_safe.rs
e00E: I added a comment to the code to explain more. Let me know in this thread if you have more questions.

// For this module we need to estimate confirmation times. So we need some way of converting the
// percentiles into time. The standard percentile is 50% which means transactions at this gas price
// were included in 50% of the blocks. Thus for every block a transaction at this price has a 0.5
// chance to be included. We treat this as a geometric distribution in which the expected time for
// the event to happen (transaction is included in block) is 1/p. Blocks happen every 15 seconds so
// this is how we get a time estimate.
// In reality this estimation can be problematic when gas prices are changing quickly. When prices
// rising, the prices calculated based on the last 200 lagg behind the real price. For this reason
// we insert two extra points for the linear interpolation below in estimate_limits.
// We do not make use of the lowest and fastest gas price because they are too strong outliers. The
// lowest gas price is skewed by miners including their own transactions at 1 gwei. The highest gas
// price can be 1000 times the fast gas price which is not reasonable.
// I suspect this happens when people try to ensure that two of their own transactions are included
// together in the same block by picking a low gas price for the first nonce and a very high one for
// the second. Or maybe just someone making in the wei/gwei unit conversion.

services-core/src/gas_price/gnosis_safe.rs
e00E: Explained in the thread below.

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -0,0 +1,173 @@
...
+const SAFE_LOW_PERCENTILE: f64 = 0.3;
+
+// For this module we need to estimate confirmation times. So we need some way of converting the
+// percentiles into time. The standard percentile is 50% which means transactions at this gas price
+// were included in 50% of the blocks. Thus for every block a transaction at this price has a 0.5
+// chance to be included. We treat this as a geometric distribution in which the expected time for
+// the event to happen (transaction is included in block) is 1/p. Blocks happen every 15 seconds so
+// this is how we get a time estimate.
+// In reality this estimation can be problematic when gas prices are changing quickly. When prices
+// rising, the prices calculated based on the last 200 lagg behind the real price. For this reason
```
fleupold, line 57:
```suggestion
// are rising, the prices calculated based on the last 200 blocks lag behind the real price. For this reason
```

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -0,0 +1,173 @@
...
+// we insert two extra points for the linear interpolation below in estimate_limits.
+// We do not make use of the lowest and fastest gas price because they are too strong outliers. The
+// lowest gas price is skewed by miners including their own transactions at 1 gwei. The highest gas
+// price can be 1000 times the fast gas price which is not reasonable.
+// I suspect this happens when people try to ensure that two of their own transactions are included
```
fleupold, line 62: Love the explanation! 
ubernit: the speculative last two sentences are not really needed (in my experience miners aren't smart enough to look ahead and include a low gas tx in order to unblock a high gas tx).

fleupold (approved):

nlordell (approved):
Thanks very well explained!

services-core/src/gas_price/gnosis_safe.rs
Uxio0: Sorry @fleupold , I've just read that. When the gas station was built we tried to estimate the execution times, but at that time even eth gas station was not doing a good estimation, so @denisgranha and me decided to leave it out as it was not important for the Safe use case

denisgranha: hey @fleupold the percentils were taken based on trial and error experience at that time, and trying to make it simple. We took some codebase as template I don't remember which 

If you want to do it properly, you would need to take into account mempool, but mempool compared to the main chain it's tricky because transactions might not go through your mempool, it depends on your peer and how are you connected so you would need a cluster of nodes with different peers/locations in order to ensure proper sample size

PR #1369, opened 2020-08-26
Number of hops in price estimator

fedgiac:
**This PR will be subject to changes and will be rebased after PR #1368 is merged.
It can be used to estimate the performance cost of using the algorithm from that PR.**

Implementation of the `hops` parameter for the price estimation.
Many, many minor changes needed to implement the number of hops in the price estimator and make the various subprojects of this repo work.

Note that the order graph is intentionally reduced with unlimited hops. This is because even if a user wishes to only see orders with at most `n` hops, matching orders that require a larger number of hops will be cleared anyway by the solver.

Most of the line changes are related to the tests.

### Performance

All benchmarks are run multiple times for different values of `hops`: `None`, 2, 30, 2¹⁶-1.

This allows us to compare the performance of the same function when a different number of hops is used with the default orderbook.

Short version:
| Function | `None` | `Some(2)` | `Some(30)` | `Some(65535)` |
| ----------- | --------- | -------------- |  --------------- | -------------------- |
|estimate_limit_price/volume: 100000000000000000|57.893 us|58.977 us|65.675 us|66.083 us|
|estimate_limit_price/volume: 1000000000000000000|61.671 us|60.801 us|72.253 us|72.690 us|
|estimate_limit_price/volume: 10000000000000000000|107.38 us|60.260 us|175.34 us|175.83 us|
|estimate_limit_price/volume: 100000000000000000000|203.65 us|72.988 us|408.71 us|409.86 us|
|estimate_limit_price/volume: 1000000000000000000000|453.61 us|225.50 us|996.60 us|999.52 us|
|order_for_limit_price/price: 200|51.716 us|52.642 us|57.854 us|58.867 us|
|order_for_limit_price/price: 190|134.43 us|55.603 us|237.08 us|239.17 us|
|order_for_limit_price/price: 180|215.54 us|64.614 us|449.85 us|448.78 us|
|order_for_limit_price/price: 150|443.67 us|212.69 us|1.0070 ms|1.0135 ms|
|order_for_limit_price/price: 100|484.56 us|261.49 us|1.0754 ms|1.1076 ms|

For this run, I incremented the number of samples to `500` (from `50`).
The `change` part in the raw output that follows should be ignored, since no meaningful comparison was performed.
<details>
  <summary>$ cargo bench -p pricegraph</summary>

```
     Running target/release/deps/pricegraph-c4a70d9d337eb8b6
Gnuplot not found, using plotters backend
[...]

PR #1501, opened 2020-10-30
Use query hops param for price estimation

fleupold:
This PR implements part of #1369 namely the integration of the hops Query parameter into the price estimation API (I couldn't reuse the other PR because the merge conflicts were too severe). It reuses the idea of creating a TokenPairRange type that can be passed along when both the pair and hops are important.

Unlike the other PR it continues to use the unbound price estimator in the services (instead of setting it to 30 hops). This is mainly for performance considerations and with the assumption that our orderbook in practice wouldn't use more than 30 hops anyways (we might eventually want to pick a realistic small number but this is beyond the scope of this PR).

This PR also leaves the test cases mostly untouched (using unbounded estimation wherever possible) as the actually bounding logic is covered by unit tests inside the bellman ford module.

I also stripped off the benchmarking code as this seems suitable for another PR.

### Test Plan
Unit tests + loading a local web and making sure we can see the impact of changing the hops parameter in the orderbook view.


fedgiac (approved):
The changes look good.

I'm not sure it's a good idea to leave out the tests with a finite number of hops, since there is quite a bit of logic that could be changed without anyone noticing any issue in this case. (I can imagine some function failing only when detecting cycles with a limited number of hops, but follow-up code expecting that the resulting orderbook is reduced for infinite hops.)

e00E (approved):

pricegraph/src/encoding.rs
```diff
@@ -29,6 +29,41 @@ pub struct TokenPair {
...
+    }
+
+    pub fn into_range(self, hops: Option<usize>) -> TokenPairRange {
+        TokenPairRange { pair: self, hops }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub struct TokenPairRange {
```
nlordell, line 51: ubernit: Not 100% happy with the naming of this, but I also can't think of anything better.

nlordell (approved):
Nice!

pricegraph/src/encoding.rs
fleupold: How about `Itinerary` (consists of start/end some general meta data). Or to make it not too far fetched `BoundedTokenPair` or `TokenPairInfo`

nlordell: `Itinerary` sounds cool, but a little abstract. Lets just leave it as is.

fleupold:
> I'm not sure it's a good idea to leave out the tests with a finite number of hops, since there is quite a bit of logic that could be changed without anyone noticing any issue in this case.

@fedgiac  To me these tests seemed more like integration tests (as they are testing that bellman-ford works, which itself has unit tests) and adding this `for hops in [None, Some(1), Some(5)...]` made the tests less single purpose.

Given the work you put into coming up with the cases, I will add the extra test cases for the one in which the expected behavior differs from the unbounded ones...

PR #1502, opened 2020-10-30
Add Benchmarks for bounded bellman ford

fleupold:
This PR adds benchmarks for the added bounded bellman ford implementation. I only added it to the transitive orderbook computation as this seems to be the most expensive operation anyways

### Test Plan
`cd pricegraph/bench && cargo bench`

### DIFF
--- a/pricegraph/bench/pricegraph.rs
+++ b/pricegraph/bench/pricegraph.rs
@@ -1,6 +1,6 @@
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use pricegraph::{Market, Pricegraph, TokenPair};
-use pricegraph_data::{DEFAULT_BATCH_ID, DEFAULT_ORDERBOOK};
+use pricegraph_data::{DEFAULT_ORDERBOOK};
 use std::time::Duration;
 
 fn read_default_pricegraph() -> Pricegraph {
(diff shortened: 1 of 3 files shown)


pricegraph/bench/pricegraph.rs
```diff
@@ -14,14 +14,19 @@ pub fn read(c: &mut Criterion) {
...
+    let hops = &[None, Some(1), Some(2), Some(5), Some(10), Some(30)];
 
-    c.bench_with_input(
-        BenchmarkId::new("Pricegraph::transitive_orderbook", *DEFAULT_BATCH_ID),
-        &(&pricegraph, dai_weth),
-        |b, &(pricegraph, dai_weth)| {
-            b.iter(|| pricegraph.transitive_orderbook(dai_weth, None, None))
-        },
-    );
+    let mut group = c.benchmark_group("Pricegraph::transitive_orderbook_with_hops");
```
nlordell, line 19: nit: I would leave this the name of the function:
```suggestion
    let mut group = c.benchmark_group("Pricegraph::transitive_orderbook");
```

nlordell (commented):
I think it might be worth including hops in the other benchmarks as well. The reason I say this is that they are commonly used operations, and some changes could regress `estimate_price` performance and we would want to catch that as well.

pricegraph/bench/pricegraph.rs
```diff
@@ -14,29 +15,35 @@ pub fn read(c: &mut Criterion) {
...
     let pricegraph = read_default_pricegraph();
-    let dai_weth = TokenPair { buy: 7, sell: 1 }.into_unbounded_range();
+    let dai_weth = TokenPair { buy: 7, sell: 1 };
     let eth = 1e18;
     let volumes = &[0.1 * eth, eth, 10.0 * eth, 100.0 * eth, 1000.0 * eth];
+    let hops = &[None, Some(1), Some(2), Some(5), Some(10), Some(30)];
 
     let mut group = c.benchmark_group("Pricegraph::estimate_limit_price");
-    for volume in volumes {
+    for (volume, hops) in volumes.iter().cartesian_product(hops) {
```
nlordell, line 41: Fancy! I like this as it reduces nesting, but I'm surprised you didn't just go for nested for loops :stuck_out_tongue:

nlordell (approved):
Looks great!

PR #1505, opened 2020-11-03
Bump the orderbook file version to see if things work now

nlordell:
Just bump the orderbook file version to force it to get re-created to make sure that the rolling update works as expected now.

### Test Plan

Solvers eventually startup.


e00E (approved):
Still need to merge the readinessProbe staging PR (or apply it manually). And I guess we will have the problem that we might overload the node when we have 4 pods all recreating the orderbook from scratch, but we'll see how long it takes.

PR #1511, opened 2020-11-05
Reduce full orderbook sync memory usage

e00E:
By making use of a change in ethcontract that creates a stream for past
events instead of one vector with all of them.
We chunk the stream again so that getting the block timestamps stays
fast as we use a batch transport for that.

This switches to  a specific commit of ethcontract so that we get the above mentioned change. Once ethcontract makes a release with this change we can switch back to a cargo version.

### Test Plan
Unfortunately UpdatingOrderbook cannot easily be tested in the current form because we don't have a mockable web3 trait and because it does file io. I would like to change that but not sure what the best approach is yet and this wouldn't be part of this PR.

### DIFF
--- a/contracts/Cargo.toml
+++ b/contracts/Cargo.toml
@@ -27,13 +27,13 @@ bin = [
 ]
 
 [dependencies]
-ethcontract = { version = "0.8", default-features = false }
+ethcontract = { version = "0.9",  default-features = false }
 serde = "1.0.117"
 
 # [bin-dependencies]
 anyhow = { version = "1.0.34", optional = true }
 env_logger = { version = "0.8.1", optional = true }
-ethcontract-generate = { version = "0.8", optional = true }
+ethcontract-generate = { version = "0.9", optional = true }
 filetime = { version = "0.2.13", optional = true }
 futures = { version = "0.3.8", optional = true }
 log = { version = "0.4.11", optional = true }
@@ -41,4 +41,4 @@ serde_json = { version = "1.0.59", optional = true }
 tokio = { version = "0.2", optional = true, features = ["macros"] }
 
 [build-dependencies]
-ethcontract-generate = "0.8"
+ethcontract-generate = "0.9"
--- a/services-core/src/orderbook/streamed/updating_orderbook.rs
+++ b/services-core/src/orderbook/streamed/updating_orderbook.rs
@@ -7,14 +7,16 @@ use crate::{
 };
 use anyhow::{anyhow, bail, ensure, Result};
 use block_timestamp_reading::{BlockTimestampReading, CachedBlockTimestampReader};
-use ethcontract::{contract::Event, BlockNumber, H256};
-use futures::future::{BoxFuture, FutureExt as _};
-use futures::lock::Mutex;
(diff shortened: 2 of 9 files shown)


services-core/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -209,6 +189,48 @@ impl UpdatingOrderbook {
...
+                let metadata = event
+                    .meta
+                    .as_ref()
+                    .ok_or_else(|| anyhow!("event without metadata: {:?}", event))?;
+                Ok(metadata.block_hash)
+            })
+            .collect::<Result<HashSet<H256>>>()?;
+        context
+            .block_timestamp_reader
+            .prepare_cache(block_hashes, self.block_page_size, latest_block)
```
fleupold, line 231: Do we still need block_page_size in prepare cache? If so, should it be `events.len()` (I see it's the same value but this is not clear from just looking at `prepare_timestamp_cache` in isolation).

fleupold (commented):
Looks good. Waiting for it to get out of Draft state before accepting.

e00E:
Memory usage of driver binary after full orderbook sync on
      master: 750 MB
this branch: 170 MB

So it seems to work.

services-core/src/orderbook/streamed/updating_orderbook.rs
e00E: It's not *needed* but I feel it still makes sense because the batch size we use for events on the updating_orderbook layer doesn't have to be the same batch size as for the CachedBlockTimestampReading layer.
You could argue that with the current code it is clear what the batch size whereas if we use `block_hashes.len()` then you have to look at how the function is called to figure out how many block hashes there are.

contracts/Cargo.toml
```diff
@@ -27,7 +27,7 @@ bin = [
 ]
 
 [dependencies]
-ethcontract = { version = "0.8", default-features = false }
+ethcontract = { git = "https://github.com/gnosis/ethcontract-rs/", rev = "58f82895256189de4ea6efc615ed94f970fc7b0c",  default-features = false }
```
fleupold, line 30: Since we maintain ethcontract, should we do a release nowish (@nlordell)?

fleupold (approved):

bh2smith (approved):
Looks good although I see we had to change the dependency to a git revision in many places. I was just learning yesterday that we could create a patch in the main toml file. Is this something that we might want to consider?

contracts/Cargo.toml
nlordell: Yes, I imagine we release a new version first. Last time @e00E did the same procedure where he created the PR in services with a git revision pending the `ethcontract` version bump just to get a head start in the review here.

services-core/src/orderbook/streamed/updating_orderbook.rs
nlordell: I also prefer slightly using `block_page_size`, this also makes it a bit more resilient to refactors (when we may call `prepare_timestamp_cache` with more events).

nlordell (approved):
Nice!

PR #1513, opened 2020-11-05
Update default values for Economic Viability config

### LINKED ISSUE #1512: Dangerous Default Values for Economic Viability Configuration
Not sure if this has been discussed much, but the current default values for economic viability are:
```
    economic_viability_subsidy_factor: 10.0,
    economic_viability_min_avg_fee_factor: 1.1,
    static_min_avg_fee_per_order: Some(
        0,
    ),
    static_max_gas_price: Some(
        100000000000000000,
    ),
    economic_viability_strategy: Static,
```

Meaning: Static Strategy with 0 as the min average fee.

I consider this to be somewhat dangerous for a Naive user following a startup guide, say our tutorial, that chooses not to read into the "Advanced Configuration". For example, I was experimenting/testing running a solver today and started paying quite some gas money to fill a bunch of old dust orders.

Could we consider making the default values a bit less dangerous for the naive user?

bh2smith:
We are facing an issue of overlapping docker-compose file used both in CI testing and also when following the REDME for running a solver in a container (specifically for macOS). This PR is intended to update those default values to something less dangerous and also keep some undesired overlap between CI configurations and those used in practice.

Fixes #1512

### Test Plan
CLI passes without any logical changes.


.travis.yml
```diff
@@ -41,23 +41,24 @@ jobs:
         # Make sure README is up to date
         - diff --ignore-trailing-space <(sed -n '/^driver /,/^```/p' README.md | head -n -1) <(cargo run --bin driver -- --help)
         # Build image with compiled binary
-        - docker build --tag stablex-binary-public --build-arg SOLVER_BASE=gnosispm/dex-open-solver:$OPEN_SOLVER_VERSION -f driver/docker/rust/Dockerfile .
+        - docker build --tag stablex-binary-public --build-arg SOLVER_BASE=gnosispm/dex-open-solver:$OPEN_SOLVER_VERSION --build-arg ECONOMIC_VIABILITY_STRATEGY=static --build-arg STATIC_MIN_AVG_FEE_PER_ORDER=0 --build-arg STATIC_MAX_GAS_PRICE=100000000000 -f driver/docker/rust/Dockerfile .
```
e00E, line 44: I dont' think the extra build args do anything here. Those are args used in Dockerfiles I don't expect them to be passed to the driver.
We pass them through the env file specified in docker-compose.yml, so by adding the new env file there for the service stablex-test you have changed them.

.travis.yml
```diff
@@ -41,23 +41,24 @@ jobs:
...
         # StableX e2e Tests (Ganache) - open solver
-        - docker-compose -f docker-compose.yml -f driver/docker-compose.open-solver.yml up -d stablex
+        - docker-compose -f docker-compose.yml -f driver/docker-compose.open-solver.yml up -d stablex-test
         - cargo test -p e2e ganache -- --nocapture
         - docker-compose logs
         # Build image with compiled binary
         - docker build --tag stablex-binary-private --build-arg SOLVER_BASE=163030813197.dkr.ecr.eu-central-1.amazonaws.com/dex-solver:$PRIVATE_SOLVER_VERSION -f driver/docker/rust/Dockerfile .
         # StableX e2e Tests (Ganache) - private solver
+        - docker build --tag stablex-binary-private --build-arg SOLVER_BASE=163030813197.dkr.ecr.eu-central-1.amazonaws.com/dex-solver:$PRIVATE_SOLVER_VERSION -f driver/docker/rust/Dockerfile .
```
e00E, line 52: This line is duplicated.
```suggestion
```

e00E:
One thing that confuses is me is whether a user running the driver "for real" would go through docker compose. I thought docker-compose is only useful for us for testing, because it's set up to start ganache automatically. But if you want to actually submit solutions then you wouldn't use ganache.
So I feel it would be simpler to tell those users to run the driver docker image directly without docker-compose. Then we could just not provide any default config values at all and rely on the driver binary's defaults (and adjust those to be safe).

bh2smith:
@e00E - Mac users can’t run a solver without docker. Were you not aware of this? Also I don’t believe there is any indication anywhere in the project that these docket compose files are to be used only for testing.

.travis.yml
bh2smith: They certainly do not. I thought these had been removed.

e00E:
I know some people need docker, I'm talking only about docker-compose. It should be possible to run the driver docker image without docker compose and docker compose is only needed (or makes it easier) for testing with ganache.

.travis.yml
bh2smith: They are gone now.

bh2smith:
> I know some people need docker, I'm talking only about docker-compose. It should be possible to run the driver docker image without docker compose and docker compose is only needed (or makes it easier) for testing with ganache.

Ok, yes this is something I can do, but I will have to update the tutorial. Particularly, I could simply remove the docker-compose command from the first paragraph here

https://github.com/gnosis/dex-docs/blob/develop/docs/tutorial-solver.md#run-the-solver

It would be nice if you could take a look at the tutorial and provide some feedback.

driver/common.env
```diff
@@ -5,4 +5,4 @@ SOLVER_TYPE=NaiveSolver
 SCHEDULER=Evm
 ECONOMIC_VIABILITY_STRATEGY=static
 STATIC_MIN_AVG_FEE_PER_ORDER=0
-STATIC_MAX_GAS_PRICE=100000000000000000
+STATIC_MAX_GAS_PRICE=100000000000
```
bh2smith, line 8:
```suggestion
STATIC_MAX_GAS_PRICE=100000000000000000
```

services-core/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -73,6 +73,11 @@ impl UpdatingOrderbook {
                             "Failed to construct orderbook from path (using default): {:?}",
                             error,
                         );
+                    } else {
+                        warn!(
+                            "Orderbook at specified path not found: {:?}",
+                            error,
+                        );
```
e00E, line 80: Looks like this unrelated change snuck in? It's so small that I don't mind including it here, just mentioning.

e00E (approved):

services-core/src/orderbook/streamed/updating_orderbook.rs
```diff
@@ -73,6 +73,11 @@ impl UpdatingOrderbook {
                             "Failed to construct orderbook from path (using default): {:?}",
                             error,
                         );
+                    } else {
+                        warn!(
```
nlordell, line 77: I don't think this should be a warning, but an `info`. The reason I say this is that you don't expect this file to be there the first time you run, and showing a warning might be a little strong.

driver/src/main.rs
```diff
@@ -175,7 +175,7 @@ struct Options {
     #[structopt(
         long,
         env = "ECONOMIC_VIABILITY_SUBSIDY_FACTOR",
-        default_value = "10.0"
+        default_value = "1.0"
```
nlordell, line 178: Before merging, we explicitly set these values in our deployment.

bh2smith: If we explicitly set these values in our deployment, doesn't that mean there is nothing to do?

services-core/src/orderbook/streamed/updating_orderbook.rs
bh2smith: Yes, seemingly unrelated, but it all relates back to be fixing some stuff for the tutorial and I realized that when I don't put the right path, it doesn't tell me anything about what happened to my orderbook file config. Thanks for allowing it!

driver/src/main.rs
e00E: Maybe Nick meant *make sure* we set them? I do think we set them but it wouldn't hurt to double check with the staging deployment config.

nlordell: Yes, just want to make sure that changing these defaults won't have unexpected results for the solvers we are currently running.

nlordell: I just checked the deployment and it looks like we set the `ECONOMIC_VIABILITY_SUBSIDY_FACTOR` for each, so nothing to do!

services-core/src/orderbook/streamed/updating_orderbook.rs
nlordell: I also think its fine including it. Especially since, as I imagine, it was added while debugging for the current PR.

nlordell (approved):

driver/src/main.rs
nlordell: > Maybe Nick meant make sure we set them?

:see_no_evil:

bh2smith:
Thanks fellas for your review! Looks like it turned out to be much easier than originally expected!

Issue #1455, opened 2020-10-01
Upstream U256 <-> f64 conversion


nlordell:
We should upstream U256 <-> f64 conversion to either the `uint` or `primitive_types` crates.

bh2smith:
So the PR needed to get this rolling has been merged into parity-common, but it doesn't look like there will be a new version of this crate published anytime soon (even though the previous version was published 6 months ago). I have attempted to update our repo by depending on the github master branch itself as 

```
primitive-types = { git = "https://github.com/paritytech/parity-common/tree/master/primitive-types", features = ["fp-conversion"] }
```
and was instructed to set the local env var ` CARGO_NET_GIT_FETCH_WITH_CLI=true` but am still confronted with

```
fatal: repository 'https://github.com/paritytech/parity-common/tree/master/primitive-types/' not found
```

In any case, we are almost there... kinda.

PR #1515, opened 2020-11-06
Use f64 instead of u256 for gas prices

e00E:
Gas prices are technically U256 but most of our code needs to some math
with them so we end up converting between u256 and f64 all the time.
With this commit we use f64 directly everywhere.

### Test Plan
Existing tests

### DIFF
--- a/services-core/src/economic_viability.rs
+++ b/services-core/src/economic_viability.rs
@@ -3,7 +3,7 @@
 
 use crate::{gas_price::GasPriceEstimating, models::solution::EconomicViabilityInfo};
 use anyhow::{anyhow, Context as _, Result};
-use ethcontract::U256;
+use pricegraph::num;
 use std::{num::NonZeroU128, sync::Arc};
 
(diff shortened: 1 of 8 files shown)


bh2smith (approved):
This is a really nice change. Might be worth mentioning here that I have been working on this issue #1455 which relates to those conversion functions. I have already gotten the conversions accepted in the appropriate repo, but have recently gotten stuck trying to upstream the dependence to a commit.

nlordell (approved):
Nice change! Maybe it even makes sense to have the gas price be in GWei everywhere?

e00E:
> Nice change! Maybe it even makes sense to have the gas price be in GWei everywhere?

That is more debatable to me. My first reaction is that I prefer the base unit for less ambiguity and it would only improve log messages.

bh2smith:
I definitely prefer GWei because it sucks to count zeros when you are typing out your gas prices manually. Metamask uses GWei.

PR #1518, opened 2020-11-09
Use smarter gas prices for solution submission

e00E:
Our current algorithm starts at fast gas price and increases the gas price
of the transaction by 50% every 30 seconds that the transaction hasn't
been mined up to the max gas economically viable gas price.
The new algorithm always uses the correct gas price according to our
smarter gas estimation which is based on the time it should take to mine
the transaction. This gas price is updated in short intervals and the
transaction is resubmitted if the gas price increases. The gas price is
still capped by economic viability.
This should lead to lower gas prices as we can pick a lower starting gas
price when there is still more time left to solve the batch and because
we never pick a gas price that is higher than what the gas station
predict we need to pay to get into the next block.

### Test Plan
Adjusted unit tests

### DIFF
--- a/services-core/src/solution_submission.rs
+++ b/services-core/src/solution_submission.rs
@@ -201,7 +201,7 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
             + BatchId::from(batch_index)
                 .solve_end_time()
                 .duration_since(SystemTime::now())
-                .unwrap_or_default();
+                .unwrap_or_else(|_| Duration::from_secs(0));
         let nonce = self.contract.get_transaction_count().await?;
         let submit_future = self.retry_with_gas_price_increase.retry(retry::Args {
             batch_index,
--- a/services-core/src/solution_submission.rs
+++ b/services-core/src/solution_submission.rs
@@ -101,7 +101,7 @@ pub struct StableXSolutionSubmitter<'a> {
 impl<'a> StableXSolutionSubmitter<'a> {
     pub fn new(
         contract: Arc<dyn StableXContract>,
-        gas_price_estimating: Arc<dyn GasPriceEstimating + Send + Sync>,
+        gas_price_estimating: Arc<dyn GasPriceEstimating>,
     ) -> Self {
         Self::with_retry_and_sleep(
             contract.clone(),
--- a/services-core/src/solution_submission.rs
+++ b/services-core/src/solution_submission.rs
@@ -18,7 +18,7 @@ use pricegraph::num;
 use retry::SolutionTransactionSending;
 use std::{
     sync::Arc,
-    time::{Duration, SystemTime},
+    time::{Duration, Instant, SystemTime},
 };
 use thiserror::Error;
 
@@ -94,7 +94,6 @@ impl From<Error> for SolutionSubmissionError {
 
 pub struct StableXSolutionSubmitter<'a> {
     contract: Arc<dyn StableXContract>,
-    gas_price_estimating: Arc<dyn GasPriceEstimating + Send + Sync>,
     retry_with_gas_price_increase: Box<dyn SolutionTransactionSending + Send + Sync + 'a>,
     async_sleep: Box<dyn AsyncSleeping + 'a>,
 }
@@ -106,7 +105,6 @@ impl<'a> StableXSolutionSubmitter<'a> {
     ) -> Self {
         Self::with_retry_and_sleep(
             contract.clone(),
-            gas_price_estimating.clone(),
             retry::RetryWithGasPriceIncrease::new(contract, gas_price_estimating),
             crate::util::AsyncSleep {},
         )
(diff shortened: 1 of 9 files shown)


services-core/src/solution_submission.rs
```diff
@@ -203,30 +197,23 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
         claimed_objective_value: U256,
         gas_price_cap: f64,
     ) -> Result<(), SolutionSubmissionError> {
-        // If the gas price cap is not at least the fast gas price then submitting the solution
-        // is not feasible because it would take too long.
-        match self.gas_price_estimating.estimate().await {
```
e00E, line 208: I have removed this because we already check this when invoking the solver with min-avg-fee so it is unlikely that it fails here. Even with the check the fast gas price could still change afterwards before the transaction is confirmed so it does not guarantee anything and removing it makes the code here simpler.

services-core/src/solution_submission.rs
```diff
@@ -161,14 +157,12 @@ impl<'a> StableXSolutionSubmitter<'a> {
 
     async fn cancel_transaction_after_deadline(
         &self,
```
e00E, line 159: A potential future change is to move cancellation into `retry` where it fits better than here.

services-core/src/solution_submission/retry.rs
```diff
@@ -10,170 +12,153 @@ use futures::{
...
+            "With {} seconds remaining estimated gas price {} capped to {}.",
+            time_remaining.as_secs(),
+            estimated_gas_price,
+            capped_gas_price,
         );
+        Some(capped_gas_price)
+    }
 
-        let solution_submission_future = contract
+    fn submit_solution(&self, args: &Args, gas_price: f64) -> BoxFuture<FutureOutput> {
```
bh2smith, line 115: These args seem oddly more specifically like `solution_args` or maybe `submission_args`. Just wondering if it might be more clear than just "args"

services-core/src/solution_submission.rs
```diff
@@ -203,30 +197,23 @@ impl<'a> StableXSolutionSubmitting for StableXSolutionSubmitter<'a> {
...
-                    )));
-                }
-            }
-            Err(err) => log::warn!("failed to estimate gas price: {:?}", err),
-        }
+        let target_confirm_time = Instant::now()
+            + BatchId::from(batch_index)
+                .solve_end_time()
+                .duration_since(SystemTime::now())
+                .unwrap_or_default();
```
bh2smith, line 204: What is the default here on `duration_since`? Is this something we need to be concerned with?

bh2smith (approved):
This is quite a large change, but looks pretty nice overall.

services-core/src/solution_submission.rs
e00E: It's zero but maybe it's clearer to specifically use zero instead of default. Will change to that.

services-core/src/solution_submission/retry.rs
```diff
@@ -10,170 +12,153 @@ use futures::{
...
+            // Like in `StableXSolutionSubmitter::submit_solution` we need to handle the situation where
+            // we observe a nonce error from one future before the completion of another. It is also
+            // possible that a previous submission transaction completes first instead of the most
+            // recent one.
+            // Unwrap because the timeout future always exists.
+            while let FutureOutput::SolutionSubmission(result) = futures.next().await.unwrap() {
+                // Comparison against `1` instead of `empty` because the timeout future always exists.
+                let is_last_transaction_future = futures.len() == 1;
+                if !result.is_transaction_error() || is_last_transaction_future {
+                    return result;
```
fleupold, line 157: Is this guaranteed to return? Why the change in logic from having the last iteration wait only for the solution submission future?

fleupold (approved):
Overall looks good and correct, but I was reminded that I find our retry logic written in a pretty complicated way and hard to reason about with a little bit of time passed.

services-core/src/solution_submission.rs
e00E: Yes the cancellation still happens.

services-core/src/solution_submission/retry.rs
e00E: In the previous version there was a clear last iteration when we reached the max gas price. At that point we wouldn't continue the `loop` anymore.
In this version we keep updating the gas price forever until a transaction completes or implicitly until the owner of this future drops it.
In both versions we keep track of previously sent transactions and have this while loop as the comment explains.

services-core/src/solution_submission/retry.rs
```diff
@@ -10,170 +12,153 @@ use futures::{
...
-            if !result.is_transaction_error() || futures.is_empty() {
-                return result;
+            futures.push(self.wait_interval());
+
+            // Like in `StableXSolutionSubmitter::submit_solution` we need to handle the situation where
+            // we observe a nonce error from one future before the completion of another. It is also
+            // possible that a previous submission transaction completes first instead of the most
+            // recent one.
+            // Unwrap because the timeout future always exists.
+            while let FutureOutput::SolutionSubmission(result) = futures.next().await.unwrap() {
```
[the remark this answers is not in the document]
e00E: Thanks for pointing out the specific confusing points. I am going to add more comments to make this easier to understand (the loop, the different cases).
One thing to realize about the cancellation logic is that it is quite independent of the gas price logic. As long as the cancellation always cancels at higher gas price I don't see anything that can go wrong. But still I agree that the cancellation should be integrated more with the retry code.
For the state machine it's hard to judge without trying. I will see how it looks when integrating cancellation.
The confusion about `futures` being a vector is unfortunate but here we a have a line `let mut futures = FuturesUnordered::new()` that imo makes it clear enough and which should make the reader check out the FuturesUnordered documentation. But I think it will become more obvious anyway when I add more comments here.

PR #1519, opened 2020-11-09
Use multiple gas stations

e00E:
to be more resilient against failure.

We should use either ethgasstation or gasnow as the highest priority because they react to changing gas prices much faster than the gnosis gas station. Between those I am not totally sure which is better but my gut feeling is that like the way gasnow works more.

### Test Plan
New unit tests of the priority estimator

### DIFF
--- a/services-core/src/gas_price.rs
+++ b/services-core/src/gas_price.rs
@@ -32,7 +32,7 @@ pub async fn create_estimator(
     let network_id = web3.net().version().await?;
     let mut estimators = Vec::<Box<dyn GasPriceEstimating>>::new();
 
-    if network_id == "1" {
+    if is_mainnet(&network_id) {
         let gasnow = gasnow::GasNow::new(http_factory)?;
         estimators.push(Box::new(gasnow));
 
@@ -48,4 +48,8 @@ pub async fn create_estimator(
     estimators.push(Box::new(web3.clone()));
 
     Ok(Arc::new(priority::PriorityGasPrice::new(estimators)))
+}
+
+fn is_mainnet(network_id: &str) -> bool {
+    network_id == "1"
 }
--- a/services-core/src/gas_price.rs
+++ b/services-core/src/gas_price.rs
@@ -3,6 +3,7 @@ mod ethgasstation;
 mod gasnow;
 mod gnosis_safe;
 mod linear_interpolation;
+mod priority;
 
 use crate::{contracts::Web3, http::HttpFactory};
 use anyhow::Result;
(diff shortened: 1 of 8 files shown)


services-core/src/gas_price.rs
```diff
@@ -27,10 +28,24 @@ pub trait GasPriceEstimating: Send + Sync {
...
+        let gasnow = gasnow::GasNow::new(http_factory)?;
+        estimators.push(Box::new(gasnow));
+
+        let ethgasstation = ethgasstation::EthGasStation::new(http_factory)?;
+        estimators.push(Box::new(ethgasstation));
+    }
+
+    if let Some(gnosis_url) = gnosis_safe::api_url_from_network_id(&network_id) {
+        let gnosis_estimator = gnosis_safe::GnosisSafeGasStation::new(http_factory, gnosis_url)?;
+        estimators.push(Box::new(gnosis_estimator));
```
bh2smith, line 45: Why is GnosisSafe the only gas station that requires a URL? Is it dynamic? The others use the `http_factory`, but aren't given a URL?

e00E: The gnosis one uses different urls for rinkeby and mainnet while the gasnow and ethgasstation only support mainnet (so there is no configurable url).

bh2smith (approved):

services-core/src/gas_price.rs
```diff
@@ -27,10 +28,24 @@ pub trait GasPriceEstimating: Send + Sync {
...
-) -> Result<Arc<dyn GasPriceEstimating + Send + Sync>> {
+) -> Result<Arc<dyn GasPriceEstimating>> {
     let network_id = web3.net().version().await?;
-    Ok(match gnosis_safe::api_url_from_network_id(&network_id) {
-        Some(url) => Arc::new(gnosis_safe::GnosisSafeGasStation::new(http_factory, url)?),
-        None => Arc::new(web3.clone()),
-    })
+    let mut estimators = Vec::<Box<dyn GasPriceEstimating>>::new();
+
+    if network_id == "1" {
```
fleupold, line 35: nit: maybe use a `is_mainnet()` helper for this logic (might not be needed right now but in general the network could be worth having as an enum available in the main function to more than just this dependency).

services-core/src/gas_price/priority.rs
```diff
@@ -0,0 +1,76 @@
...
+
+        estimator_0.expect_estimate().times(1).returning(|| Ok(1.0));
+
+        let priority = PriorityGasPrice::new(vec![Box::new(estimator_0), Box::new(estimator_1)]);
+        let result = priority.estimate().now_or_never().unwrap().unwrap();
+        assert_approx_eq!(result, 1.0);
+    }
+
+    #[test]
+    fn prioritize_picks_second() {
```
fleupold, line 62: nit: if first fails (for complete documentation you could add a test that it fails, if all fail).

services-core/src/gas_price/priority.rs
```diff
@@ -0,0 +1,76 @@
...
+pub struct PriorityGasPrice {
+    estimators: Vec<Box<dyn GasPriceEstimating>>,
+}
+
+impl PriorityGasPrice {
+    pub fn new(estimators: Vec<Box<dyn GasPriceEstimating>>) -> Self {
+        Self { estimators }
+    }
+
+    async fn prioritize<'a, T, F>(&'a self, operation: T) -> Result<f64>
```
fleupold, line 15: Totally optional, but seeing how much you like to not repeat code, this could probably be made even more generic by returning the same type as operation. In that case we might be able to use for our other "Priority" components (e.g. price source as well).

fleupold (approved):

PR #1522, opened 2020-11-11
[GasEstimate] refactor gas estimators to use abstract transport trait

fleupold:
This will allow extracting the gas estimation logic into its own crate and reuse it e.g. in the gas token miner as well as in oba services without requiring logic that is specific to `services-core` (e.g. HTTPTransport, MetricLabels, etc). 
The `eth_node` implementation will likely not be extractable as it is bound to the web3 struct and doesn't add much custom logic.

### Test Plan
CI + `cargo test gas_price -- --ignored`

### DIFF
--- a/services-core/src/gas_price.rs
+++ b/services-core/src/gas_price.rs
@@ -9,7 +9,7 @@ use crate::{contracts::Web3, http::HttpClient, http::HttpFactory, metrics::HttpL
 use anyhow::Result;
 use isahc::http::uri::Uri;
 use serde::de::DeserializeOwned;
-use std::{sync::Arc, time::Duration};
+use std::{str::FromStr, sync::Arc, time::Duration};
 
 pub const DEFAULT_GAS_LIMIT: f64 = 21000.0;
 pub const DEFAULT_TIME_LIMIT: Duration = Duration::from_secs(30);
@@ -28,14 +28,14 @@ pub trait GasPriceEstimating: Send + Sync {
 
 #[async_trait::async_trait]
 pub trait Transport: Send + Sync {
-    async fn get_json<T: DeserializeOwned>(&self, url: &'static str) -> Result<T>;
+    async fn get_json<'a, T: DeserializeOwned>(&self, url: &'a str) -> Result<T>;
 }
 
 #[async_trait::async_trait]
 impl Transport for HttpClient {
-    async fn get_json<T: DeserializeOwned>(&self, url: &'static str) -> Result<T> {
-        let uri = Uri::from_static(url);
-        self.get_json_async(uri, HttpLabel::GasStation).await
+    async fn get_json<'a, T: DeserializeOwned>(&self, url: &'a str) -> Result<T> {
+        self.get_json_async(Uri::from_str(url)?, HttpLabel::GasStation)
+            .await
     }
 }
 
@@ -57,7 +57,7 @@ pub async fn create_estimator(
 
     if let Some(gnosis_url) = gnosis_safe::api_url_from_network_id(&network_id) {
         let gnosis_estimator =
-            gnosis_safe::GnosisSafeGasStation::new(http_factory.create()?, gnosis_url);
+            gnosis_safe::GnosisSafeGasStation::new(http_factory.create()?, gnosis_url.into());
         estimators.push(Box::new(gnosis_estimator));
     }
 
@@ -80,9 +80,8 @@ mod tests {
 
     #[async_trait::async_trait]
     impl Transport for TestTransport {
-        async fn get_json<T: DeserializeOwned>(&self, url: &'static str) -> Result<T> {
-            let uri = Uri::from_static(url);
-            let json: String = isahc::get_async(uri).await?.text()?;
+        async fn get_json<'a, T: DeserializeOwned>(&self, url: &'a str) -> Result<T> {
+            let json: String = isahc::get_async(Uri::from_str(url)?).await?.text()?;
             Ok(serde_json::from_str(&json)?)
         }
     }
--- a/services-core/src/gas_price.rs
+++ b/services-core/src/gas_price.rs
@@ -5,8 +5,10 @@ mod gnosis_safe;
 mod linear_interpolation;
 mod priority;
 
-use crate::{contracts::Web3, http::HttpFactory};
+use crate::{contracts::Web3, http::HttpClient, http::HttpFactory, metrics::HttpLabel};
 use anyhow::Result;
+use isahc::http::uri::Uri;
+use serde::de::DeserializeOwned;
 use std::{sync::Arc, time::Duration};
 
 pub const DEFAULT_GAS_LIMIT: f64 = 21000.0;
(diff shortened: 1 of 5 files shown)


services-core/src/gas_price.rs
```diff
@@ -23,14 +25,48 @@ pub trait GasPriceEstimating: Send + Sync {
     async fn estimate_with_limits(&self, gas_limit: f64, time_limit: Duration) -> Result<f64>;
 }
 
+#[async_trait::async_trait]
+pub trait Transport: Send + Sync {
+    async fn get_json<T: DeserializeOwned>(&self, url: &'static str) -> Result<T>;
+}
+
+#[async_trait::async_trait]
+impl Transport for HttpClient {
```
fleupold, line 34: This implementation will stay in the core crate.

services-core/src/gas_price.rs
```diff
@@ -23,14 +25,48 @@ pub trait GasPriceEstimating: Send + Sync {
     async fn estimate_with_limits(&self, gas_limit: f64, time_limit: Duration) -> Result<f64>;
 }
 
+#[async_trait::async_trait]
+pub trait Transport: Send + Sync {
+    async fn get_json<T: DeserializeOwned>(&self, url: &'static str) -> Result<T>;
```
fleupold, line 30: Not sure how to best pass the URL. I didn't want to commit to some specific URL library.

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -59,34 +58,30 @@ const SAFE_LOW_PERCENTILE: f64 = 0.3;
 // price can be 1000 times the fast gas price which is not reasonable.
 const SECONDS_PER_BLOCK: f64 = 15.0;
 // Treat percentiles as probabilities for geometric distribution.
-const FAST_TIME: f64 = SECONDS_PER_BLOCK * FAST_PERCENTILE;
-const STANDARD_TIME: f64 = SECONDS_PER_BLOCK * STANDARD_PERCENTILE;
-const SAFE_LOW_TIME: f64 = SECONDS_PER_BLOCK * SAFE_LOW_PERCENTILE;
+const FAST_TIME: f64 = SECONDS_PER_BLOCK / FAST_PERCENTILE;
+const STANDARD_TIME: f64 = SECONDS_PER_BLOCK / STANDARD_PERCENTILE;
+const SAFE_LOW_TIME: f64 = SECONDS_PER_BLOCK / SAFE_LOW_PERCENTILE;
```
fleupold, line 63: This should probably be its own PR, but can you double check that I'm not crazy and this logic was off? The ignored test was failing.

e00E: Yes you're right. I checked the original PR. It was initially correct and then in https://github.com/gnosis/dex-services/pull/1500/commits/90d4fc130269323e13861913016476c7c249db0f#diff-3af02e6a043d788af05800f15cbd96e304d2eb57faa2c56d3d17e6d78aff8937R64 I got rid off the unnecessary `1.0 / x` and accidentally made it a multiplication :facepalm: .

services-core/src/gas_price.rs
nlordell: I think you just need to specify a lifetime parameter:
```suggestion
    async fn get_json<'a, T: DeserializeOwned>(&self, url: &'a str) -> Result<T>;
```

There might be issues with the `async_trait` (as the boxed future result will have its lifetime tied to this string), maybe @e00E knows more as I'm not super familiar with it.

services-core/src/gas_price.rs
```diff
@@ -23,14 +25,48 @@ pub trait GasPriceEstimating: Send + Sync {
...
+}
+
+#[async_trait::async_trait]
+impl Transport for HttpClient {
+    async fn get_json<T: DeserializeOwned>(&self, url: &'static str) -> Result<T> {
+        let uri = Uri::from_static(url);
```
e00E, line 36: This is problematic because `from_static` panics if the url is invalid.Should use whatever the normal contructor is that cn return an error.

services-core/src/gas_price/gnosis_safe.rs
nlordell: :see_no_evil:

services-core/src/gas_price.rs
e00E: Should be `&str` without `'static`. It is true that for our current clients all the urls are static but in general they don't have to be.

e00E: Passing it as str instead of a specific Url type is fine imo. All urls can be converted to/from strings so the only the disadvantage is a small performance loss because we might be constructing an url twice.

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -59,34 +58,30 @@ const SAFE_LOW_PERCENTILE: f64 = 0.3;
...
 
 /// Retrieve gas prices from the Gnosis Safe gas station service.
 #[derive(Debug)]
-pub struct GnosisSafeGasStation {
-    client: HttpClient,
-    uri: Uri,
+pub struct GnosisSafeGasStation<T> {
+    transport: T,
+    uri: &'static str,
```
nlordell, line 69: This should probably just own the URL string.
```suggestion
    uri: String,
```

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -59,34 +58,30 @@ const SAFE_LOW_PERCENTILE: f64 = 0.3;
...
-        self.client
-            .get_json_async(&self.uri, HttpLabel::GasStation)
-            .await
+        self.transport.get_json(&self.uri).await
     }
 }
 
 #[async_trait::async_trait]
-impl GasPriceEstimating for GnosisSafeGasStation {
+impl<T: Transport + Send + Sync> GasPriceEstimating for GnosisSafeGasStation<T> {
```
nlordell, line 84: Maybe, we can just make the Transport `Send + Sync` so we only need `Transport` trait bound?

e00E (commented):
The approach with the trait makes sense to me.

services-core/src/gas_price/gnosis_safe.rs
```diff
@@ -121,9 +116,10 @@ fn estimate_with_limits(
 
 #[cfg(test)]
 pub mod tests {
+    use super::super::tests::TestTransport;
     use super::super::DEFAULT_GAS_LIMIT;
     use super::*;
-    use crate::util::FutureWaitExt as _;
+    use crate::util::FutureWaitExt;
```
nlordell, line 122: I don't understand this change, as I don't see `FutureWaitExt` used anywhere. The `as _` is used in Rust especially for traits to import them into scope so you can use the methods on implementations but when you don't actually need to reference the trait directly (either implementing it or adding a generic trait bound). I slightly prefer it as it was before.

nlordell (approved):
Nice abstraction!

services-core/src/gas_price.rs
nlordell: > I didn't want to commit to some specific URL library.

You could always use [`http`](https://docs.rs/http/0.2.1/http/) which is what many HTTP clients in Rust use (like hyper, isahc, etc.). I'm also 100% fine with using a string here as well.

services-core/src/gas_price/gnosis_safe.rs
fleupold: Isolated the fix in #1526

e00E (approved):

PR #1524, opened 2020-11-12
Make gas estimators configurable

e00E:
This is useful in case an estimator becomes broken and we want to
disable it without recompilation.

I set the default estimator to web3 because this is the only one that works on every network.

This is a config change because we do not want to use the new default.

### Test Plan
Ran the binary locally to make sure arguments are interpreted correctly.


nlordell (approved):
Very nice

fleupold (approved):
Nice

PR #1525, opened 2020-11-12
Reduce gas estimator error logs

e00E:
The gas now estimator from time to time fails single requests for which we get an alert because this is an error log. Since we use multiple estimators if previous one fails this is not a problem and I don't see value in having this be an alert.

The easiest way to solve this would be to turn the error log into a warning, which I find acceptable too.
This PR instead starts off logging warnings but after an estimator failed N times in a row starts logging an error.

This is has the advantage that we would notice an estimator going down permanently or changing their api in a backwards incompatible way. On the other hand it could still be annoying to get an alert when an estimator goes down for 10 minutes when this hasn't actually impacted our service because of the fallback estimators. I am not sure what others prefer so let's discuss that in this PR.

### Test Plan
Temporarily add a test that with always failing estimator and observe it's errors (also change log to println).

### DIFF
--- a/services-core/src/gas_price/priority.rs
+++ b/services-core/src/gas_price/priority.rs
@@ -1,14 +1,36 @@
 use super::GasPriceEstimating;
 use anyhow::{anyhow, Result};
-use std::{future::Future, time::Duration};
+use std::{
+    future::Future,
+    sync::atomic::{AtomicUsize, Ordering},
+    time::Duration,
+};
+
+// Errors of an individual estimator are logged as warnings until it has failed this many times in
+// a row at which point they are logged as errors.
+// This is useful to reduce alerts for estimators that sometimes fail individual requests while
+// still getting them when the estimator really goes down.
+const LOG_ERROR_AFTER_N_ERRORS: usize = 10;
 
 // Uses the first successful estimator.
 pub struct PriorityGasPrice {
-    estimators: Vec<Box<dyn GasPriceEstimating>>,
+    estimators: Vec<Estimator>,
+}
+
+struct Estimator {
+    estimator: Box<dyn GasPriceEstimating>,
+    errors_in_a_row: AtomicUsize,
 }
 
 impl PriorityGasPrice {
     pub fn new(estimators: Vec<Box<dyn GasPriceEstimating>>) -> Self {
+        let estimators = estimators
+            .into_iter()
+            .map(|estimator| Estimator {
+                estimator,
+                errors_in_a_row: AtomicUsize::new(0),
+            })
+            .collect();
         Self { estimators }
     }
 
@@ -17,10 +39,20 @@ impl PriorityGasPrice {
         T: Fn(&'a dyn GasPriceEstimating) -> F,
         F: Future<Output = Result<f64>>,
     {
-        for estimator in &self.estimators {
-            match operation(estimator.as_ref()).await {
-                Ok(result) => return Ok(result),
-                Err(err) => log::error!("gas estimator failed: {:?}", err),
+        for (i, estimator) in self.estimators.iter().enumerate() {
+            match operation(estimator.estimator.as_ref()).await {
+                Ok(result) => {
+                    estimator.errors_in_a_row.store(0, Ordering::SeqCst);
+                    return Ok(result);
+                }
+                Err(err) => {
+                    let num_errors = estimator.errors_in_a_row.fetch_add(1, Ordering::SeqCst) + 1;
+                    if num_errors < LOG_ERROR_AFTER_N_ERRORS {
+                        log::warn!("gas estimator {} failed: {:?}", i, err);
+                    } else {
+                        log::error!("gas estimator {} failed: {:?}", i, err);
+                    }
+                }
             }
         }
         Err(anyhow!("all gas estimators failed"))


services-core/src/gas_price/priority.rs
```diff
@@ -1,14 +1,36 @@
...
+    future::Future,
+    sync::atomic::{AtomicU64, Ordering},
+    time::Duration,
+};
+
+// Errors of an individual estimator are logged as warnings until it has failed this many times in
+// a row at which point they are logged as errors.
+// This is useful to reduce alerts for estimators that sometimes fail individual requests while
+// still getting them when the estimator really goes down.
+const LOG_ERROR_AFTER_N_ERRORS: u64 = 10;
```
nlordell, line 13: Descriptive constant name, love it.

services-core/src/gas_price/priority.rs
```diff
@@ -1,14 +1,36 @@
...
 
 // Uses the first successful estimator.
 pub struct PriorityGasPrice {
-    estimators: Vec<Box<dyn GasPriceEstimating>>,
+    estimators: Vec<Estimator>,
+}
+
+struct Estimator {
+    estimator: Box<dyn GasPriceEstimating>,
+    errors_in_a_row: AtomicU64,
```
nlordell, line 22: Not that it makes any difference at all (in fact, I don't think its worth changing), buy I think it would be every so slightly more correct if this were a `AtomicUsize` so it would work better on 32-bit platforms (which we don't compile for :see_no_evil:). Also, this most probably doesn't need 64-bits for counting errors, and the platform's word size is almost certainly enough to count these errors.

services-core/src/gas_price/priority.rs
```diff
@@ -17,10 +39,21 @@ impl PriorityGasPrice {
         T: Fn(&'a dyn GasPriceEstimating) -> F,
         F: Future<Output = Result<f64>>,
     {
-        for estimator in &self.estimators {
-            match operation(estimator.as_ref()).await {
-                Ok(result) => return Ok(result),
-                Err(err) => log::error!("gas estimator failed: {:?}", err),
+        const ORDERING: Ordering = Ordering::SeqCst;
```
nlordell, line 42: ubernit: I think you ended up having to type more to use the constant :smile:. Also, I would slightly prefer to inline the ordering with the atomic operations in general, as when reasoning about atomic code, its nice to see the ordering with the atomic operation to more easily reason about how memory is synchronized.

nlordell (approved):
Nice!

services-core/src/gas_price/priority.rs
```diff
@@ -17,10 +39,21 @@ impl PriorityGasPrice {
...
+        for (i, estimator) in self.estimators.iter().enumerate() {
+            match operation(estimator.estimator.as_ref()).await {
+                Ok(result) => {
+                    estimator.errors_in_a_row.store(0, ORDERING);
+                    return Ok(result);
+                }
+                Err(err) => {
+                    let num_errors = estimator.errors_in_a_row.fetch_add(1, ORDERING) + 1;
+                    if num_errors < LOG_ERROR_AFTER_N_ERRORS {
+                        log::warn!("gas estimator {} failed: {:?}", i, err);
```
fleupold, line 52: nit: Could we maybe make GasEstimator implement a display method to print which estimator fails? Printing the index is already a nice improvement but still requires lookup of the current config.

services-core/src/gas_price/priority.rs
```diff
@@ -17,10 +39,21 @@ impl PriorityGasPrice {
...
-                Err(err) => log::error!("gas estimator failed: {:?}", err),
+        const ORDERING: Ordering = Ordering::SeqCst;
+        for (i, estimator) in self.estimators.iter().enumerate() {
+            match operation(estimator.estimator.as_ref()).await {
+                Ok(result) => {
+                    estimator.errors_in_a_row.store(0, ORDERING);
+                    return Ok(result);
+                }
+                Err(err) => {
+                    let num_errors = estimator.errors_in_a_row.fetch_add(1, ORDERING) + 1;
```
fleupold, line 50: Are you are adding 1 twice here? It seems like after the first error num_errors would evaluate to 2.

services-core/src/gas_price/priority.rs
```diff
@@ -1,14 +1,36 @@
...
 }
 
 impl PriorityGasPrice {
     pub fn new(estimators: Vec<Box<dyn GasPriceEstimating>>) -> Self {
+        let estimators = estimators
+            .into_iter()
+            .map(|estimator| Estimator {
+                estimator,
+                errors_in_a_row: AtomicU64::new(0),
+            })
```
fleupold, line 32: ubernanonit: implement .into() trait for style?

fleupold (approved):
👍

services-core/src/gas_price/priority.rs
nlordell: `fetch_add` returns the previous value (`fetch` first then `add`). I *think* it is correct like this.

services-core/src/gas_price/priority.rs
nlordell: I think this is a great idea, but I would just use `Debug` for `#[derive(Debug)]` to do the work for us.

services-core/src/gas_price/priority.rs
e00E: I prefer it without if it's local to this file. Less to type, less traits to understand.

services-core/src/gas_price/priority.rs
e00E: Agree it would be nice if it had the name but I don't think looking up the configuration is too much work for a rare occurrence like this either. If we could get the name without any drawbacks then I would definitely do it now but in this case I'm not sure what the best way to do that is. Like Nick said I don't want the trait to require Debug.

nlordell: Could we add a `fn name(&self) -> &str` to the trait? Still feels a bit icky...

PR #1526, opened 2020-11-12
[Easy] Fix Safe time limit computation

fleupold:
Discovered in #1522 but isolated the fix here in case we want to revert/apply it independently of the feature PR.

### Test Plan
Added unit test


e00E (approved):

nlordell (approved):
Nice! Thanks for splitting into its own PR.

PR #1527, opened 2020-11-12
Move gas estimation logic into its own crate

fleupold:
This PR moves the gas_estimation logic into its own crate so that it can be reused in both oba-services, dex-services and other repos that require gas price estimation.

### Test Plan

CI


services-core/src/gas_price.rs
```diff
@@ -62,45 +38,36 @@ pub async fn create_priority_estimator(
...
 #[cfg(test)]
-mod tests {
-    use super::*;
-    use isahc::ResponseExt;
-
-    #[derive(Default)]
-    pub struct TestTransport {}
+use std::time::Duration;
+#[cfg(test)]
+mockall::mock! {
```
fleupold, line 65: Not sure if that is the best way to recreate a mockable version of the GasPriceEstimating trait in `core-services`. Alternatively we could export it from the `gas-estimation` crate gated behind a feature that `services-core` dev-depends on.

e00E: I find it reasonable to not make any mock related choice in the new crate and instead let services_core mock it like you have implemented.

gas-estimation/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+pub use priority::PriorityGasPriceEstimating;
+
+use anyhow::Result;
+use serde::de::DeserializeOwned;
+use std::time::Duration;
+
+pub const DEFAULT_GAS_LIMIT: f64 = 21000.0;
+pub const DEFAULT_TIME_LIMIT: Duration = Duration::from_secs(30);
+
+#[cfg_attr(test, mockall::automock)]
```
e00E, line 20:
```suggestion
```
Can remove this and the dependency to mockall in Cargo.toml .

gas-estimation/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+authors = ["Gnosis Protocol Team <dex@gnosis.io>"]
+edition = "2018"
+
+[dependencies]
+anyhow = "1.0"
+async-trait = "0.1"
+log = "0.4"
+serde = { version = "1.0", features = ["derive"] }
+serde_with = "1.5"
+web3 = "0.13"
```
e00E, line 13:
```suggestion
web3 = { version = "0.13", default-features = false }
```

e00E (approved):

gas-estimation/Cargo.toml
bh2smith: Is there a particular reason for this suggestion? I can only guess that we do not use any of the default features and want to pull in less dependencies.

gas-estimation/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+primitive-types = { version = "0.7.3", features = ["fp-conversion"] }
```
bh2smith, line 14: Have you already made the appropriate changes related to this updated dependency? I feel like this should be a separate PR, but could be left here if it doesn't include all the old conversion call site changes as well.

gas-estimation/src/eth_node.rs
```diff
@@ -1,18 +1,21 @@
...
+    T: Transport + Send + Sync,
+    <T as Transport>::Out: Send,
+{
     async fn estimate_with_limits(&self, _gas_limit: f64, _time_limit: Duration) -> Result<f64> {
         self.eth()
             .gas_price()
             .await
             .map_err(From::from)
-            .map(num::u256_to_f64)
+            .map(|r| r.to_f64_lossy())
```
bh2smith, line 19: Right... so you have already done this.

bh2smith (approved):
Seems legit!

gas-estimation/src/lib.rs
fleupold: I think it's still needed to the Priority Gas Price estimator

gas-estimation/Cargo.toml
fleupold: Yes, only depend on what you need is a good practice.

gas-estimation/Cargo.toml
fleupold: I have not updated the dependency in other parts (or replaced the num_to_f64 implementation). I think this is its own PR, but since I'm introducing new code here I  though I might as well use the new version instead of copy and pasting deprecated code again.

gas-estimation/src/eth_node.rs
fleupold: There are still other callsites in services_core

PR #1530, opened 2020-11-16
Upstream U256 <-> f64 conversion

### LINKED ISSUE #1455: Upstream U256 <-> f64 conversion

bh2smith:
Closes #1455.

Based on the recent changes to `primiive-types` (https://github.com/paritytech/parity-common/pull/436)
we can now use their `U256 <-> f64` conversion methods.

### Test Plan
Passing build and existing unit tests.


services-core/src/gas_price/eth_node.rs
```diff
@@ -13,6 +13,6 @@ impl GasPriceEstimating for Web3 {
             .gas_price()
             .await
             .map_err(From::from)
-            .map(num::u256_to_f64)
+            .map(|g| U256::to_f64_lossy(g))
```
e00E, line 16:
```suggestion
            .map(U256::to_f64_lossy)
```
Not sure if this works, but slight improvement if it does.

e00E (approved):

services-core/src/gas_price/eth_node.rs
bh2smith: Lets give it a try!

PR #1531, opened 2020-11-16
Remove error for no local account

e00E:
We do in fact guarantee that this cannot happen because `new` always
creates an account. By making use of this fact we simplify the api.

### Test Plan
Adjusted unit test.

### DIFF
--- a/services-core/src/solution_submission.rs
+++ b/services-core/src/solution_submission.rs
@@ -1,7 +1,7 @@
 mod retry;
 
 use crate::{
-    contracts::stablex_contract::{NoopTransactionError, StableXContract},
+    contracts::stablex_contract::StableXContract,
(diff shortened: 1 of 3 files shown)


services-core/src/solution_submission.rs
```diff
@@ -282,16 +282,16 @@ impl IsOpenEthereumTransactionError for ExecutionError {
 impl IsOpenEthereumTransactionError for Result<(), MethodError> {
     fn is_transaction_error(&self) -> bool {
         match self {
-            Ok(()) => false,
             Err(MethodError { inner, .. }) => inner.is_transaction_error(),
+            _ => false,
```
e00E, line 286: This is just so that this function looks more like the similar function below (next change).

fleupold (approved):

PR #1533, opened 2020-11-17
Limit max number of hops

e00E:
Fixes a way of crashing the price estimator with a large number of hops.

### Test Plan
Manually tested that hops > 30 is rejected with 400 status code. curl -i "http://0.0.0.0:8080/api/v1/markets/0-1?hops=31"


fleupold (approved):
Looks great! cc @anxolin

PR #1534, opened 2020-11-17
More aggressive gas price increase

e00E:
It can happen that the new gas price is only slightly larger than the
old one, without being large enough to respect the minimum gas price
increase.
The old algorithm would in this case not resubmit the solution.
The new algorithm always increases the gas price to the minimum increase
if the new gas price is larger than the old gas price.

Also refactored the calculation into its own pure function to make it easier to test.

### Test Plan
New unit test that tests the gas price algorithm directly.

### DIFF
--- a/services-core/src/solution_submission/retry.rs
+++ b/services-core/src/solution_submission/retry.rs
@@ -84,31 +84,33 @@ enum FutureOutput {
     Timeout,
 }
 
+fn new_gas_price_estimate(
+    previous_gas_price: f64,
+    new_gas_price: f64,
+    max_gas_price: f64,
+) -> Option<f64> {
+    let min_gas_price = previous_gas_price * MIN_GAS_PRICE_INCREASE_FACTOR;
+    if min_gas_price > max_gas_price {
+        return None;
+    }
+    if new_gas_price <= previous_gas_price {
+        // Gas price has not increased.
+        return None;
+    }
+    // Gas price could have increased but doesn't respect minimum increase so adjust it up.
+    let new_price = min_gas_price.max(new_gas_price);
+    Some(new_price.min(max_gas_price))
+}
+
 impl<'a> RetryWithGasPriceIncrease<'a> {
-    async fn new_gas_price_for_target_time(&self, args: &Args) -> Option<f64> {
+    async fn gas_price(&self, args: &Args) -> Result<f64> {
         let time_remaining = args
             .target_confirm_time
             .saturating_duration_since(self.now.instant_now());
         // TODO: Use a more accurate gas limit once the gas estimators take that into account.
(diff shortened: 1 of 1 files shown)


services-core/src/solution_submission/retry.rs
```diff
@@ -84,31 +84,38 @@ enum FutureOutput {
     Timeout,
 }
 
+fn calculate_gas_price(
+    previous_gas_price: f64,
+    new_gas_price: f64,
```
fleupold, line 89: nit: `new_gas_price_estimate`

services-core/src/solution_submission/retry.rs
```diff
@@ -84,31 +84,38 @@ enum FutureOutput {
...
+        // Gas price has not increased.
+        None
+    } else if new_gas_price <= min_gas_price {
+        // Gas price has increased but doesn't respect minimum increase so adjust it up.
+        Some(min_gas_price)
+    } else if new_gas_price <= max_gas_price {
+        Some(new_gas_price)
+    } else {
+        // Gas price is larger than the maximum.
+        Some(max_gas_price)
```
fleupold, line 106: maybe slightly reduced branch complexity by using something like:

```rust
if new_estimate <= previous_estimate {
  // no increase
  return None
}
if min_gas_price > cap {
  // Cannot increase without going over cap
  return None
}

let new_price = max(min_gas_price, new_estimate)
return min(new_price, cap)

```

services-core/src/solution_submission/retry.rs
```diff
@@ -138,13 +145,21 @@ impl<'a> RetryWithGasPriceIncrease<'a> {
...
-                    futures.push(self.submit_solution(&args, gas_price));
+            match self.gas_price(&args).await {
+                Ok(gas_price) => {
+                    log::debug!("estimated gas price {}", gas_price);
+                    if let Some(mut gas_price) =
+                        calculate_gas_price(last_used_gas_price, gas_price, args.gas_price_cap)
+                    {
+                        gas_price = gas_price.ceil();
+                        last_used_gas_price = gas_price;
+                        log::info!("submitting solution transaction at gas price {}", gas_price);
```
fleupold, line 158: ubernit: I kind of liked that the old log message also gave us a hint if the current cap was hit (since this cannot be easily found from other informations in the logs). Maybe we can just add the cap to the log message

fleupold (approved):
Looks good. I should have caught that we got rid of the minimum increase when reviewing the original PR.

e00E:
> Looks good. I should have caught that we got rid of the minimum increase when reviewing the original PR.

To be clear, we were still respecting the minimum increase but that meant not resubmitting. Whereas now we bump the gas price up to the minimum increase even it only increased by less.

PR #1538, opened 2020-11-18
Cancellation refactor 3: Implement the new api

e00E:
but do not yet update the tests. Compared to the previous implementation this one:
* knows about the gas price that was used for solution submission, allowing us to make the cancellation transaction cheaper
* does not send a cancellation transaction if the solution transaction was never sent (this doesn't really happen in practice but could happen if the gas estimator keeps failing)
* has one loop with select statement which should make it clearer what happens in what order

### DIFF
--- a/services-core/src/solution_submission/retry.rs
+++ b/services-core/src/solution_submission/retry.rs
@@ -1,3 +1,4 @@
+use super::first_match::FirstMatch;
 use crate::util::{self, AsyncSleeping, Now};
 use crate::{
     contracts::stablex_contract::{StableXContract, SOLUTION_SUBMISSION_GAS_LIMIT},
@@ -10,7 +11,10 @@ use ethcontract::{
     web3::error::Error as Web3Error,
     U256,
 };
-use futures::future::{BoxFuture, FutureExt as _};
+use futures::{
+    future::{BoxFuture, FutureExt as _},
+    stream::{self, FusedStream, StreamExt},
+};
 use gas_estimation::GasPriceEstimating;
 use std::{
     future::Future,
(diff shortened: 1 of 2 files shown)


services-core/src/solution_submission/retry.rs
```diff
@@ -116,18 +120,96 @@ fn new_gas_price_estimate(
...
+
+        // This struct keeps track of all the solution and cancellation futures. If we get a
+        // "nonce already used error" we continue running the other futures. We need to handle this
+        // case because we do not know which transactions will complete or fail or in which order we
+        // observe completion.
+        let mut first_match = FirstMatch::new(|result: &RetryResult| !is_transaction_error(result));
+        let mut last_used_gas_price = 0.0;
+        loop {
+            futures::select! {
+                gas_price = gas_price_stream.next() => match gas_price.unwrap() {
```
fleupold, line 182: could you add in-code comment on why the unwrap is ok please?

services-core/src/solution_submission/retry.rs
```diff
@@ -116,18 +120,96 @@ fn new_gas_price_estimate(
...
+        futures::pin_mut!(cancel_after);
+        futures::pin_mut!(gas_price_stream);
+
+        // This struct keeps track of all the solution and cancellation futures. If we get a
+        // "nonce already used error" we continue running the other futures. We need to handle this
+        // case because we do not know which transactions will complete or fail or in which order we
+        // observe completion.
+        let mut first_match = FirstMatch::new(|result: &RetryResult| !is_transaction_error(result));
+        let mut last_used_gas_price = 0.0;
+        loop {
```
fleupold, line 180: I find this loop still a bit intimidating. Maybe the gas_price_stream handling could be extracted into a helper method?

fleupold (approved):

services-core/src/solution_submission/retry.rs
e00E: I agree it doesn't *look* nice but this is hard to factor out because you'd have to pass so many parameters that are local to this function and at that point I think it's easier to understand to have it inline.

services-core/src/solution_submission/retry.rs
```diff
@@ -116,18 +120,97 @@ fn new_gas_price_estimate(
...
+            futures::select! {
+                // Unwrap because the stream never ends.
+                gas_price = gas_price_stream.next() => match gas_price.unwrap() {
```
nlordell, line 183: ubernit: Instead of a comment, maybe use `expect`?
```suggestion
                gas_price = gas_price_stream.next() => match gas_price.expect("stream should never end") {
```

nlordell (approved):
Very nice!

PR #1539, opened 2020-11-18
Cancellation refactor 4: Update tests for the new retry-cancellation api

e00E:
This is the final PR and should pass CI.

### Test Plan
New unit tests, CI.

### DIFF
--- a/services-core/src/solution_submission/retry.rs
+++ b/services-core/src/solution_submission/retry.rs
@@ -244,10 +244,26 @@ mod tests {
         gas_price::MockGasPriceEstimating,
         util::{FutureWaitExt as _, MockAsyncSleeping},
     };
+    use ethcontract::{transaction::TransactionResult, H256};
     use futures::future;
 
+    pub fn nonce_execution_error() -> ExecutionError {
+        ExecutionError::Web3(Web3Error::Rpc(RpcError {
+            code: ethcontract::jsonrpc::types::ErrorCode::ServerError(-32010),
+            message: "Transaction nonce is too low.".to_string(),
+            data: None,
+        }))
+    }
+
+    fn nonce_method_error() -> MethodError {
+        MethodError {
+            signature: String::new(),
+            inner: nonce_execution_error(),
+        }
+    }
+
     #[test]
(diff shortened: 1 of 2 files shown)


services-core/src/solution_submission/retry.rs
```diff
@@ -399,26 +414,61 @@ mod tests {
...
             .return_once(|_, _, _, _, _| {
                 async move {
-                    receiver.await.unwrap();
+                    submit_receiver.await.unwrap();
                     Ok(())
                 }
                 .boxed()
             });
-        sleep.expect_sleep().times(1).returning(|_| immediate!(()));
+        sleep.expect_sleep().return_once(move |_| {
```
fleupold, line 422: I'm not a big fan of using the sleep method here to kickstart the future resolution as it relies on pretty detailed knowledge of the intrinsics (e.g. the fact that sleep is called after the first solution submission).

Would it work to `cancel_sender.send(()).unwrap();` inside `expect_submit_solution` before `submit_receiver.await.unwrap()`? This way the test would say: we cancel as soon as we have issued the first tx and resolve the first tx as soon as we have started the cancellation.

services-core/src/solution_submission/retry.rs
```diff
@@ -399,26 +414,61 @@ mod tests {
...
             .times(1)
-            .return_once(|_, _, _, _, _| {
-                sender.send(()).unwrap();
-                immediate!(Err(nonce_error()))
-            });
-        sleep
-            .expect_sleep()
-            .returning(|_| future::pending().boxed());
+            .return_once(|_, _, _, _, _| future::pending().boxed());
+        sleep.expect_sleep().return_once(move |_| {
```
fleupold, line 464: ditto: if possible I'd move the cancellation signalling into the `expect_submit_solution` mock.

fleupold (approved):
Good new tests!

services-core/src/solution_submission/retry.rs
e00E: Good point, changed.

nlordell (approved):
Nice! The split up PRs made this much easier to review.

PR #1542, opened 2020-11-19
Limit number of hops in fuzzing

e00E:
We saw this from the fuzzer pod running in kubernetes.

### Test Plan
Run fuzzer.

### DIFF
--- a/pricegraph/fuzz/fuzz_targets/pricegraph.rs
+++ b/pricegraph/fuzz/fuzz_targets/pricegraph.rs
@@ -64,7 +64,7 @@ fuzz_target!(|arguments: Arguments| {
                     return;
                 }
             }
-            pricegraph.transitive_orderbook(market, hops, spread);
+            pricegraph.transitive_orderbook(market, 1+ (hops % 30), spread);
         }
     };
 });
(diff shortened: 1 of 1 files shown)


nlordell (approved):

pricegraph/fuzz/fuzz_targets/pricegraph.rs
```diff
@@ -59,6 +59,11 @@ fuzz_target!(|arguments: Arguments| {
                     return;
                 }
             }
+            if let Some(hops) = hops {
+                if hops > 30 {
+                    return;
+                }
+            }
             pricegraph.transitive_orderbook(market, hops, spread);
```
nlordell, line 67: Could you just
```suggestion
            pricegraph.transitive_orderbook(market, hops % 30, spread);
```

Not sure what works better with the fuzzer...

nlordell:
Also, 0 hops might not make sense either.

fleupold (approved):

pricegraph/fuzz/fuzz_targets/pricegraph.rs
e00E: Dunno which works better but yeah let's use that suggestion.

nlordell: Sorry @e00E, I removed the `merge when green` label as I think there is an issue. It looks like the currently:
- We are using the `if` statement and the modulo expression
- We are computing a modulo on an `Option<usize>`, my suggestion was more pseudo-code than real code :see_no_evil:

e00E: Oops, thanks. Fixed it.

PR #1543, opened 2020-11-19
Refactor minimum gas price logic into it's own module

e00E:
This nicely makes the core gas price logic pure and easily testable
while also simplifying the logic in the big select statement in retry.

### Test Plan
Moved the min increase test into the new module and see how much nicer the test is.

### DIFF
--- a/services-core/src/solution_submission/gas_price_increase.rs
+++ b/services-core/src/solution_submission/gas_price_increase.rs
@@ -1,0 +1,78 @@
+use futures::stream::{Stream, StreamExt};
+
+// openethereum requires that the gas price of the resubmitted transaction has increased by at
+// least 12.5%.
+const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
+
+pub fn minimum_increase(gas_price: f64) -> f64 {
+    (gas_price * MIN_GAS_PRICE_INCREASE_FACTOR).ceil()
+}
+
+fn new_gas_price_estimate(
+    previous_gas_price: f64,
+    new_gas_price: f64,
+    max_gas_price: f64,
+) -> Option<f64> {
+    let min_gas_price = minimum_increase(previous_gas_price);
+    if min_gas_price > max_gas_price {
+        return None;
+    }
+    if new_gas_price <= previous_gas_price {
+        // Gas price has not increased.
+        return None;
+    }
+    // Gas price could have increased but doesn't respect minimum increase so adjust it up.
+    let new_price = min_gas_price.max(new_gas_price);
+    Some(new_price.min(max_gas_price))
+}
+
+// Adapt a stream of gas prices to only yield gas prices that respect the minimum gas price increase
+// while filtering out other values, including those over the cap.
+pub fn enforce_minimum_increase_and_cap(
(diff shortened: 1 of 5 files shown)


services-core/src/solution_submission/gas_price_increase.rs
```diff
@@ -0,0 +1,77 @@
+use futures::stream::{Stream, StreamExt};
+
+// openethereum requires that the gas price of the resubmitted transaction has increased by at
+// least 12.5%.
+const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
+
+pub fn minimum_increase(gas_price: f64) -> f64 {
+    (gas_price * MIN_GAS_PRICE_INCREASE_FACTOR).ceil()
```
e00E, line 8: The `ceil` has moved here which also forced me to change the test to use numbers >= 10.

services-core/src/solution_submission/retry.rs
```diff
@@ -149,24 +127,37 @@ impl<'a> RetryWithGasPriceIncrease<'a> {
...
+                }
+            }
+        })
     }
 
     async fn retry_(&self, args: Args, cancel_after: impl Future) -> RetryResult {
         log::debug!("starting retry with gas price cap {}", args.gas_price_cap);
 
-        let gas_price_stream = self.gas_price_stream(args.target_confirm_time);
+        let gas_price_stream = gas_price_increase::enforce_minimum_increase(
```
fleupold, line 155: nit: Syntactically, I would find it awesome if `enforce_minimum_increase` could be an extension on the Stream to allow this line to look like:

```rust
let gas_price_stream = self
  .gas_price_stream(args.target_confirm_time)
  .enforce_minimum_increase(args.gas_price_cap);
```

services-core/src/solution_submission/gas_price_increase.rs
```diff
@@ -0,0 +1,77 @@
...
+    if new_gas_price <= previous_gas_price {
+        // Gas price has not increased.
+        return None;
+    }
+    // Gas price could have increased but doesn't respect minimum increase so adjust it up.
+    let new_price = min_gas_price.max(new_gas_price);
+    Some(new_price.min(max_gas_price))
+}
+
+// Adapt a stream of gas prices to only yield gas prices that respect the minimum gas price increase.
```
fleupold, line 29: nit: Please add to the comment that prices are capped at a maximum (I think it could even go in the method name as `enforce_minimum_increase_up_to` or even be implemented as a separate chained combinator, but not important).

fleupold (approved):
Very nice refactoring! The retry logic is getting clearer and easier to understand...

services-core/src/solution_submission/retry.rs
e00E: Not a fan because then we have to introduce a new trait and implement this trait for Stream<Item=f64>. I feel this makes the code overall harder to understand just for slightly different syntax here.

nlordell: I also think its not worth it in this case just for some syntactic sugar. If the `gas_price_stream` returned a specific `GasPriceStream: Stream` type, then I 100% agree it would make sense to implement this as a method.

PR #1544, opened 2020-11-19
Use SolutionSubmission Wrapper contract

fleupold:
With this PR we will be able to use the GasToken to submit solutions on mainnet. This is achieved by funneling solutions through a WrapperContract, which has the same interface as the original BatchExchange but on top of forwarding it to the original one it also burns a bunch of gas tokens (if the current gas price is above a certain threshold). Currently the threshold is set to 60 GWEI in the smart contract

### Test Plan
Unit tests still pass. Since this requires a solution submission to test, my plan is to deploy it to staging and check execution of the first couple of solutions.


driver/src/main.rs
```diff
@@ -260,6 +260,16 @@ struct Options {
         use_delimiter = true
     )]
     gas_estimators: Vec<GasEstimatorType>,
+
+    /// Whether to rely on external price sources (e.g. 1Inch, Kraken etc)
+    /// when estimating token prices
```
e00E, line 265: Wrong doc text

driver/src/main.rs
```diff
@@ -260,6 +260,16 @@ struct Options {
...
+    #[structopt(
+        long,
+        env = "USE_SOLUTION_SUBMITTER",
+        parse(try_from_str),
```
e00E, line 269: Is try_from_str needed?

e00E (approved):

nlordell (approved):
Short and sweet.

nlordell:
> Since this requires a solution submission to test, my plan is to deploy it to staging and check execution of the first couple of solutions.

Could we maybe deploy a test gas token on Ganache?

driver/src/main.rs
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FFGTVmzksb2j0k%2Fgiphy.gif&token=oops" data-gitmeme-token="oops"></a>

driver/src/main.rs
fleupold: Yes, because otherwise structopt will take the sheer existence of a `USE_SOLUTION_SUBMITTER` env variable as a true. In our kubernetes env we need to declare the variable unconditionally and rely on setting it to different values (e.g. true/false or 1/0) in different environments. For this, we need to parse from string.

fleupold:
> Could we maybe deploy a test gas token on Ganache?

I found that gas refunds weren't working as expected on ganache. I simulated one of last nights solutions on Tenderly and the exact same call invocation seems to work. I can hold off merging to Monday so that we have one week of testing in staging.

PR #1546, opened 2020-11-20
Fix CI for new clippy warnings

e00E:
I accidentally pushed this commit to master directly because I thought I was on a branch. So I reverted it immediately and now we have a proper PR.

### Test Plan
CI

### DIFF
--- a/services-core/src/http_server/routing.rs
+++ b/services-core/src/http_server/routing.rs
@@ -11,6 +11,7 @@ pub struct DefaultRouter {
 }
 
 impl Handler for DefaultRouter {
+    #[allow(clippy::manual_strip)]
     fn handle_request(&self, request: &Request) -> Result<Response> {
         let handler = router!(request,
             (GET) (/metrics) => { self.metrics.as_ref() },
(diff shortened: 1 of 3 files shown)


services-core/src/http_server/routing.rs
```diff
@@ -11,6 +11,7 @@ pub struct DefaultRouter {
 }
 
 impl Handler for DefaultRouter {
+    #[allow(clippy::manual_strip)]
```
nlordell, line 14: Hmm, weird. Clippy is giving style warnings on macro expansions from foreign crates... That feels wrong.

nlordell (approved):
Yay Rust 1.48! More const-generic support here we come!

services-core/src/http_server/routing.rs
e00E: Wasn't it always like this and we had to change ethcontract macros for the same reason?

nlordell: Thats because we use `ethcontract-generate` to generate source and `include!` it, so it was a bit different. Using the proc macro did not, as far as I remember, have this issue.

PR #1547, opened 2020-11-20
Simplify trait bounds for AsyncSleeping, Now, SolutionTransactionSending

e00E:
For AsyncSleeping and and Now we can safely assume that that this trait
is always 'static which gets rid of some generic lifetimes.
For SolutionTransactionSending we can assume that it is always Send and
Sync but here I didn't want to add 'static because it would be
reasonable to create an implementation based on &StablexContract instead
of Arc for example.

### Test Plan
Still compiles.

### DIFF
--- a/services-core/src/solution_submission.rs
+++ b/services-core/src/solution_submission.rs
@@ -87,13 +87,13 @@ impl From<Error> for SolutionSubmissionError {
     }
 }
 
-pub struct StableXSolutionSubmitter<'a> {
+pub struct StableXSolutionSubmitter {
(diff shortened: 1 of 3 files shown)


nlordell (approved):

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F4UGmPc07F5soU%2Fgiphy.gif&token=simplify" data-gitmeme-token="simplify"></a>

PR #1548, opened 2020-11-20
Make the retry api agnostic of how transactions are sent

e00E:
which is is a big step towards making this a standalone crate. We
accomplish this by using traits to encapsulate how the "solution" and
the cancellation is sent. This way there is no dependency on
StablexContract or even web3.
solution_submission has intentionally not yet been fully updated to the
new api to keep the PR smaller.

### Test Plan
Tests for retry pass but tests for solution_submission have been commented out and code contains a `todo!` so CI won't pass.


services-core/src/solution_submission.rs
```diff
@@ -87,9 +87,9 @@ impl From<Error> for SolutionSubmissionError {
     }
 }
 
-pub struct StableXSolutionSubmitter {
+pub struct StableXSolutionSubmitter<T = RetryWithGasPriceIncrease> {
     contract: Arc<dyn StableXContract>,
-    retry_with_gas_price_increase: Box<dyn SolutionTransactionSending>,
+    retry_with_gas_price_increase: T,
```
e00E, line 92: This can't be a Box<dyn> anymore  because the trait is a generic function.

services-core/src/solution_submission.rs
```diff
@@ -228,6 +211,7 @@ fn convert_cancel_result(
     Err(SolutionSubmissionError::Unexpected(error))
 }
 
+/*
 #[cfg(test)]
```
e00E, line 215: Will be fixed next PR.

services-core/src/solution_submission/retry.rs
```diff
@@ -24,76 +15,97 @@ use std::{
...
-    fn retry<'a>(
-        &'a self,
-        args: Args,
-        cancel_after: BoxFuture<'static, ()>,
-    ) -> BoxFuture<'a, RetryResult>;
+    /// Returns None if no transaction has been submitted.
+    async fn retry<TransactionSender, CancellationSender>(
+        &self,
+        transaction_sender: TransactionSender,
+        cancel_after: BoxFuture<'_, CancellationSender>,
```
e00E, line 54: I made the CancellationSender be returned from the future because it allows users of the function who don't care about cancellation to use a future::pending without having to create a CancellationSender.

services-core/src/solution_submission/retry.rs
```diff
@@ -149,13 +140,15 @@ impl RetryWithGasPriceIncrease {
...
 
-    async fn retry_(&self, args: Args, cancel_after: impl Future) -> RetryResult {
-        log::debug!("starting retry with gas price cap {}", args.gas_price_cap);
-
-        let gas_price_stream = gas_price_increase::enforce_minimum_increase_and_cap(
-            args.gas_price_cap,
-            self.gas_price_stream(args.target_confirm_time),
-        );
+    async fn retry<TransactionSender, CancellationSender>(
+        transaction_sender: TransactionSender,
```
e00E, line 144: This function no longer uses anything from Self anymore which makes the tests a lot simpler to understand because there is no GasPriceEstimating or AsycnSleep going on.

fleupold (approved):
Nice!

I was expecting the logic to be build on top of a TransactionBuilder, but making it ethcontracts agnostic is definitely a cool feature. Additional things that I think this module _could_ take care of if it was more tightly coupled are
- nonce handling
- cancellation tx creation (as this is likely always the same)

but I see how these are "simple enough" to reimplement in the dependent projects.

services-core/src/solution_submission.rs
e00E: `dyn Trait` is a trait object. only "object safe" traits can be converted to trait objects but Retry isn't object safe because it has a method with generics: https://doc.rust-lang.org/reference/items/traits.html#object-safety

e00E:
> Additional things that I think this module could take care of if it was more tightly coupled are

These can potentially be done by having an optional `web3` and `ethcontract` feature which then exposes extra methods with simpler signatures.

PR #1556, opened 2020-11-24
Move retry code into its own crate

e00E:
A potential next step is to add optional dependencies to ethcontract, web3, or gas-estimation to provide more commonly needed helper functionality. 

### Test Plan
CI, no logic change


transaction-retry/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "transaction-retry"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+async-trait = "0.1"
+futures = { version = "0.3", default-features = false, features = ["std", "async-await"] }
```
nlordell, line 8: beautiful

transaction-retry/src/gas_price_increase.rs
```diff
@@ -4,8 +4,9 @@ use futures::stream::{Stream, StreamExt as _};
 // least 12.5%.
 const MIN_GAS_PRICE_INCREASE_FACTOR: f64 = 1.125 * (1.0 + f64::EPSILON);
 
-pub fn minimum_increase(gas_price: f64) -> f64 {
-    (gas_price * MIN_GAS_PRICE_INCREASE_FACTOR).ceil()
+/// The minimum gas price that allows a new transaction to replace an older one.
```
nlordell, line 7: Maybe we should add `#![deny(missing_docs)]` to the crate?

nlordell (approved):
Very nice

transaction-retry/src/gas_price_increase.rs
e00E: I'm slightly opposed to it because it forces documenting some things that are obvious enough from the name that the documentation doesn't add anything. But  no strong preference either way.

fleupold: Let's maybe try it? I'm personally very bad at thinking about docs so having a linter yell at me might be a good thing.

fleupold (approved):
✅ Great work! Looking forward to reusing this complex and powerful piece of our infra.

PR #1557, opened 2020-11-26
Fix cancellation future possibly sleeping too long

e00E:
We do not know when it will be polled for the first time so it is
important to only calculate the sleep duration then.

### Test Plan
There was no explicit test for this and I don't think it's worth writing one because it would end up too convoluted.


nlordell (approved):
Nice catch!

fleupold (approved):

PR #1558, opened 2020-11-26
[Easy] Fix inverted logic in was_mined()

fleupold:
Fixes alerts we have seen twice today in dev channel.

If the result of a Cancellation/Solution submission is not an err, it was mined. At the moment it returns false in that case leading to successful txs not being recognized up by the system.

### Test Plan

Will add a unit test.


nlordell (approved):

e00E (approved):
Good find. This happened because in my PRs I went from is_transaction_error to is_mined and forgot to invert the else statement.

PR #1559, opened 2020-11-26
[Easy] fix price estimator dockerfile

fleupold:
Today, when creating a new top-level subcrate we need to also adjust the dockerfile of the price estimator as otherwise we run into build failure like [this](https://travis-ci.com/github/gnosis/dex-services/jobs/448733177#L309)

This PR simplifies the Dockerfile by copying the entirety of the src repository and relying on .dockerignore to not include any unnecessary data.

### Test Plan
See master turn green after merging.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,13 @@ target
 **/node_modules
 .git
 .vscode
+.github
+.mergify.yml
+.travis.yml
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Real nice!

Issue #1566, opened 2020-12-01
[Price Estimator] Un-reduceable orderbook


nlordell:
It is possible to craft orders in such a way that a path in the orderbook graph would end up with an empty flow because of rounding errors. This causes issues when reducing the orderbook as it would find a negative cycle with an empty flow, reduce it (which effectively does nothing since all the deducted amounts would be 0) and then go back to finding the same cycle, essentially looping forever.

The correct long term fix for this is to implement arbitrary precision rational numbers for exchange rate and capacity when finding and filling flows. This is already captured by #1381.

In the mean time to prevent infinite loops in the price estimator we can:
- [ ] Implement order and token filtering to remove rare offenders
- [ ] Panic on empty flows to prevent infinite loops

PR #1567, opened 2020-12-01
Add support for filtered orderbook in the price estimator

nlordell:
One half of #1566 

This PR adds a "filter" command line argument to allow filtering tokens and accounts 

### Test Plan

Run the price estimator with a token filter:
```
$ cargo run -p price-estimator -- --orderbook-file target/stablex_orderbook_mainnet.bin --orderbook-filter '{"tokens": {"Blacklist": [158]}}'
```

Now check that you can successfully retrieve the orderbook
```
$ curl -s "http://localhost:8080/api/v1/markets/157-4?batchId=5355804" | jq '{ asks: (.asks | length), bids: (.bids | length) }'
{
  "asks": 10,
  "bids": 565
}
```

Previously, in debug mode, the server would error with:
```
ERRO [services_core::logging] thread ‘tokio-runtime-worker’ panicked at ‘assertion failed: num::is_strictly_positive_and_finite(value)’, pricegraph/src/orderbook/scalar.rs:172:5
```

In release mode, it would enter an **infinite loop**.

### DIFF
--- a/price-estimator/src/main.rs
+++ b/price-estimator/src/main.rs
@@ -21,7 +21,7 @@ use services_core::{
     http_server::{DefaultRouter, RouilleServer, Serving},
     logging,
     metrics::{HttpMetrics, MetricsHandler},
-    orderbook::EventBasedOrderbook,
+    orderbook::{EventBasedOrderbook, FilteredOrderbookReader, OrderbookFilter},
     token_info::{cached::TokenInfoCache, hardcoded::TokenData},
     util::FutureWaitExt as _,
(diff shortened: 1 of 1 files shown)


e00E (approved):

fleupold (approved):
For oba services I'm wondering if it makes sense to combine the main.rs of the API and the driver (and allow to specify which components we want to launch), given that most of the configuration is copy pasted between two increasingly large main files (and the one that is specific has reasonable defaults).

PR #1568, opened 2020-12-01
Panic when reducing zero flow

nlordell:
One half of #1566 

This PR adds an additional check when reducing a path's flow to make sure its not empty. This can happen in cases where the computed exchange and capacity along a path would become zero because of rounding errors as the smallest possible float is around `2e-308` while the exchange rate for an order can be `8e-73`, so it doesn't take many orders to reach this limit.

For more details, see the issue linked above and the added unit test.

### Test Plan

Added unit test demonstrating problem. Note that this test **still** passes in release mode which used to not be the case and cause infinite loops.
```
$ cargo test -p pricegraph --release -- panics_on_empty_flow
    Finished release [optimized] target(s) in 0.07s
     Running target/release/deps/pricegraph-4287836eda7f370d

running 1 test
test orderbook::tests::panics_on_empty_flow ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 53 filtered out
```

Additionally, running the price estimator in release mode we can see the error when requesting `curl -s "http://localhost:8080/api/v1/markets/157-4?batchId=5355804"` which used to cause an **infinite loop** previously:
```
$ cargo run --release -p price-estimator -- --orderbook-file target/stablex_orderbook_mainnet.bin
# ...
2020-12-01T18:25:20.243Z ERRO [services_core::logging] thread 'tokio-runtime-worker' panicked at 'stuck reducing path with empty flow: 181->4->157->158->159->160->161->162->163->164->165->166->167->168->169->170->171->172->173->174->175->176->177->178->179->180->181', pricegraph/src/orderbook.rs:368:13
```

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -356,7 +356,18 @@ impl Orderbook {
     ///
     /// Note that currently, user buy token balances are not incremented as a
     /// result of filling orders along a path.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if the provided flow is empty and would not further
+    /// reduce the orderbook. This can happen in certain situations where the
+    /// flow becomes empty due to floating point rounding errors and can cause
+    /// infinite loops when reducing orderbook or enumerating transitive orders.
     fn fill_path_with_flow(&mut self, path: &[NodeIndex], flow: &Flow) -> Option<()> {
+        if flow.is_empty() {
+            panic!("stuck reducing path with empty flow: {}", format_path(path));
+        }
+
         let mut transitive_xrate = ExchangeRate::IDENTITY;
         for pair in pairs_on_path(path) {
             let (order, user) = self.best_order_with_user_for_pair_mut(pair)?;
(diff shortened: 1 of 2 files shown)


e00E (approved):
Panic is better than infinite loop but even better would be if it returned an error. With the panic a problem could be that we will lose a tokio thread forever if it panics (not sure).
Not required just pointing it out because it might not need a lot of code changes to add `Result` in a couple of places.

nlordell:
Good point, it was just a quick fix to prevent infinite loops from happening on the service. Ideally we would use real numbers so this wouldn't be an issue.

e00E:
The error would also be nice because we use the price estimation in the driver code too. So if we removed the current black list it might panic. We can still merge this first and I will convert the panic into an error.

nlordell:
Personally, I think the most benefit would come from using real numbers instead of floats for `ExchangeRate` and `Flow` as then the errors simply cannot happen.

That being said, panicking in the driver code would not be good so it might be worth doing in the interim.

e00E:
I agree but the real number change is very big and maybe not worth to do for gpv1 at this point. We should still merge this because panic is better than infinite loop. Let me know if you want to turn the panic into a result or if I should do it.

nlordell:
> if you want to turn the panic into a result or if I should do it.

You can go for it :smil

fleupold (approved):
Nice change and good idea using a result.

PR #1571, opened 2020-12-02
Turn pricegraph panic into error

e00E:
The meat of the change is Orderbook.rs find_path_flow and fill_path_with_flow which return results now. All the other changes follow from bubbling up the error.

Sadly this makes the code uglier because of a lot of extra `Result`s and
being unable to use `?` in some places because we went from `Option<T>`
to `Result<Option<T>>`.
However, we still need this change to prevent panicking because of
untrusted input (the orderbook).

related to https://github.com/gnosis/dex-services/issues/1566

### Test Plan
Tests still pass. Did a couple of manual requests to my price estimator.

### DIFF
--- a/pricegraph/src/orderbook.rs
+++ b/pricegraph/src/orderbook.rs
@@ -126,22 +126,24 @@ impl Orderbook {
     }
 
     /// Reduces the orderbook by matching all overlapping ring trades.
-    pub fn reduce_overlapping_orders(mut self) -> ReducedOrderbook {
-        Subgraphs::new(self.projection.node_indices()).for_each(|token| loop {
+    pub fn reduce_overlapping_orders(mut self) -> Result<ReducedOrderbook, OrderbookError> {
+        let result = Subgraphs::new(self.projection.node_indices()).for_each_until(|token| loop {
             let cycle = match shortest_path(&self.projection, token, None) {
-                Ok(shortest_path_graph) => break shortest_path_graph.connected_nodes(),
+                Ok(shortest_path_graph) => {
+                    break ControlFlow::Continue(shortest_path_graph.connected_nodes())
+                }
                 Err(cycle) => cycle,
             };
-            self.fill_path(&cycle).unwrap_or_else(|| {
-                panic!(
-                    "failed to fill path along detected negative cycle {}",
-                    format_path(&cycle),
-                )
-            });
+            if let Err(err) = self.fill_path(&cycle) {
+                break ControlFlow::Break(err);
+            }
         });
+        if let Some(err) = result {
+            return Err(err);
+        }
 
         debug_assert!(!self.is_overlapping());
-        ReducedOrderbook(self)
+        Ok(ReducedOrderbook(self))
     }
 
     /// Fills a ring trade over the specified market, and returns the flow
@@ -154,9 +156,12 @@ impl Orderbook {
     /// specifically the market `base`'s subgraph in the case where the `quote`
     /// and `base` token are not part of the same subgraph, may still contain
     /// negative cycles.
-    pub fn fill_market_ring_trade(&mut self, market: Market) -> Option<Ring> {
+    pub fn fill_market_ring_trade(
+        &mut self,
+        market: Market,
+    ) -> Result<Option<Ring>, OrderbookError> {
         if !self.is_token_pair_valid(market.bid_pair()) {
-            return None;
+            return Ok(None);
         }
 
         let (base, quote) = (node_index(market.base), node_index(market.quote));
@@ -178,22 +183,17 @@ impl Orderbook {
                         .fill_path(&bid)
                         .expect("bid transitive path not found after being detected");
 
-                    return Some(Ring { ask, bid });
+                    return Ok(Some(Ring { ask, bid }));
                 }
                 Err(cycle) => {
                     // NOTE: Skip negative cycles that are not along the
                     // specified market.
-                    self.fill_path(&cycle).unwrap_or_else(|| {
-                        panic!(
-                            "failed to fill path along detected negative cycle {}",
-                            format_path(&cycle),
-                        )
-                    });
+                    self.fill_path(&cycle)?;
                 }
             };
         }
 
-        None
+        Ok(None)
     }
 
     /// Returns an iterator over all transitive orders from lowest to highest
@@ -206,7 +206,7 @@ impl Orderbook {
     pub fn transitive_orders(
         self,
         pair_range: TokenPairRange,
-    ) -> Result<TransitiveOrders, OverlapError> {
+    ) -> Result<TransitiveOrders, OrderbookError> {
         TransitiveOrders::new(self, pair_range)
     }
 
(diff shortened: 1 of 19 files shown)


pricegraph/src/orderbook.rs
```diff
@@ -283,27 +283,22 @@ impl Orderbook {
         self.projection.find_edge(buy, sell)
     }
 
-    /// Finds and fills a trading path through the orderbook between the
+    /// Finds a trading path through the orderbook between the
```
e00E, line 286: This comment was out of date already.

fleupold (approved):
Looks correct, but you are right it's not pretty. I'm not sure if it would be better if the `None` case would just become one type of error types (like a "no transitive order found" or the like) and then get handled appropriately by the callsites that care about the difference.

I do think this code base will eventually also be important for v2.

pricegraph/src/orderbook.rs
```diff
@@ -359,20 +359,21 @@ impl Orderbook {
...
+        &mut self,
+        path: &[NodeIndex],
+        flow: &Flow,
+    ) -> Result<(), OrderbookError> {
         let mut transitive_xrate = ExchangeRate::IDENTITY;
         for pair in pairs_on_path(path) {
-            let (order, user) = self.best_order_with_user_for_pair_mut(pair)?;
+            let (order, user) = self
+                .best_order_with_user_for_pair_mut(pair)
+                .unwrap_or_else(|| panic!("missing order for pair {:?}", pair));
```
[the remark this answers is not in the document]
e00E: This function used to return Option<()> where the None case was treated as the error and this error would always have been unwrapped by the caller. The change in behavior is only that I moved the unwrap into here. This is cleaner because this function now returns a Result that we don't want to unwrap.

e00E:
> I do think this code base will eventually also be important for v2.

In that case it might make more sense to do the proper fix of using rational numbers instead floating point. That should completely get rid of this error and these type of panics. But are you sure that this is relevant for v2? The order model is more advanced so I'm not sure how much of this pricegraph implementation still applies.

pricegraph/src/orderbook.rs
```diff
@@ -126,22 +126,24 @@ impl Orderbook {
...
+                Ok(shortest_path_graph) => {
+                    break ControlFlow::Continue(shortest_path_graph.connected_nodes())
+                }
                 Err(cycle) => cycle,
             };
-            self.fill_path(&cycle).unwrap_or_else(|| {
-                panic!(
-                    "failed to fill path along detected negative cycle {}",
-                    format_path(&cycle),
-                )
```
nlordell, line 139: This should still be a panic as it is indicates the logic in the code is wrong.

pricegraph/src/orderbook.rs
```diff
@@ -178,22 +183,17 @@ impl Orderbook {
...
+                    return Ok(Some(Ring { ask, bid }));
                 }
                 Err(cycle) => {
                     // NOTE: Skip negative cycles that are not along the
                     // specified market.
-                    self.fill_path(&cycle).unwrap_or_else(|| {
-                        panic!(
-                            "failed to fill path along detected negative cycle {}",
-                            format_path(&cycle),
-                        )
```
nlordell, line 190: This one as well.

pricegraph/src/orderbook.rs
```diff
@@ -359,20 +359,21 @@ impl Orderbook {
     ///
     /// # Panics
     ///
-    /// This method panics if the provided flow is empty and would not further
-    /// reduce the orderbook. This can happen in certain situations where the
-    /// flow becomes empty due to floating point rounding errors and can cause
-    /// infinite loops when reducing orderbook or enumerating transitive orders.
-    fn fill_path_with_flow(&mut self, path: &[NodeIndex], flow: &Flow) -> Option<()> {
-        if flow.is_empty() {
-            panic!("stuck reducing path with empty flow: {}", format_path(path));
```
nlordell, line 368: I think this is the only panic that should be converted into an `Result::Err`. The other ones indicate things that should never happen unless there is:
- a bug
- an incorrect assumption

nlordell (commented):
I think there are some `panics` that were turned into errors that shouldn't have. Most of the existing panics were things that should never happen unless there was a bug or incorrect assumption.

pricegraph/src/orderbook.rs
nlordell: For this panic, it can happen for certain input orderbooks. The other one should not happen for any inputs.

pricegraph/src/orderbook.rs
e00E: The root problem is that find_path_flow can create an invalid flow in which case we return the error. fill_path calls find_path_flow so we cannot unwrap here.

pricegraph/src/orderbook.rs
e00E: I thought that finding a 0 flow (caused by getting an infinite exchange rate) was the root problem.

pricegraph/src/orderbook.rs
nlordell: Hmm, maybe the errors need to be differentiated so that we can panic if its a negative cycle error but break if its a floating point error.

pricegraph/src/orderbook.rs
nlordell: Yes, but this is caused by certain orderbooks when the transitive exchange rate gets too small or too large and becomes either `0` or `∞`. So legitimate and valid orderbooks can cause these FP rounding errors that should propagate a `Result::Err`. However, in the case where a known good path is being filled, but an edge between two tokens can't be found, we should panic as before.

e00E: We do. I changed the functions to unwrap the get_best_order_for_pair call.

pricegraph/src/orderbook.rs
```diff
@@ -313,36 +308,41 @@ impl Orderbook {
...
         let mut capacity = f64::INFINITY;
         let mut transitive_xrate = ExchangeRate::IDENTITY;
         let mut max_xrate = ExchangeRate::IDENTITY;
         for pair in pairs_on_path(path) {
-            let order = self.orders.best_order_for_pair(pair)?;
-            transitive_xrate *= order.exchange_rate;
+            let order = self
+                .orders
+                .best_order_for_pair(pair)
+                .unwrap_or_else(|| panic!("missing order for pair {:?}", pair));
```
nlordell, line 333: This is what I missed!

nlordell (approved):
I missed the unwrap when filling the path! Looks good.

pricegraph/src/orderbook.rs
nlordell: :see_no_evil:

fleupold:
> But are you sure that this is relevant for v2? The order model is more advanced so I'm not sure how much of this pricegraph implementation still applies.

I'm not sure but I could see how a fast graph based approach only on limit orders can still be interesting e.g. for market makers to get an overview of the protocol native supply/demand. I don't think there is anything actionable right now, but we should also not treat this code as "dead" yet.

PR #1605, opened 2021-01-11
Fix new clippy lints

nlordell:
This PR fixes clippy lints that were added with a new Rust release. These were causing lint errors in CI which made, otherwise passing PRs, impossible to merge (such as some dependency updates).

### Test Plan

CI passes.

### DIFF
--- a/services-core/src/contracts/stablex_contract.rs
+++ b/services-core/src/contracts/stablex_contract.rs
@@ -15,11 +15,7 @@ use ethcontract::{
     transaction::{confirm::ConfirmParams, Account, GasPrice, ResolveCondition, TransactionResult},
     Address, BlockId, BlockNumber, PrivateKey, U256,
 };
-use futures::{
-    future::BoxFuture,
-    stream::{BoxStream, StreamExt},
-    FutureExt as _,
-};
+use futures::stream::{BoxStream, StreamExt};
 
 use lazy_static::lazy_static;
 use std::collections::HashMap;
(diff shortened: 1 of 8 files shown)


services-core/src/contracts/stablex_contract.rs
```diff
@@ -147,14 +143,14 @@ pub trait StableXContract: Send + Sync {
         block_number: Option<BlockNumber>,
     ) -> Result<U256>;
 
-    fn submit_solution<'a>(
-        &'a self,
+    async fn submit_solution(
```
nlordell, line 146: These changes were because of the following lints:
```
warning: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
   --> services-core/src/contracts/stablex_contract.rs:150:5
    |
150 | /     fn submit_solution<'a>(
151 | |         &'a self,
152 | |         batch_index: u32,
153 | |         solution: Solution,
...   |
156 | |         nonce: U256,
157 | |     ) -> BoxFuture<'a, Result<(), MethodError>>;
    | |_______________________________________________^
    |
    = note: `#[warn(clippy::needless_lifetimes)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
```

Instead of removing the explicit lifetimes, I decided to make it use `async-trait` as it also would make the problem go away and touch an equivalent amount of code.

vkgnosis: The reason this was BoxFuture instead of async fn is that we couldn't test some interactions with mockall otherwise. For example with an async fn mockall didn't let us return future::pending. If the tests still pass then async fn is better.

vkgnosis (approved):

PR #1617, opened 2021-01-15
Implement custom benign errors

nlordell:
This PR offers a simple work around to help silence benign errors on a per-configuration basis.

### Test Plan

Added unit test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,9 @@ OPTIONS:
         --auction-data-page-size <auction-data-page-size>
             Specify the number of blocks to fetch events for at a time for constructing the orderbook for the solver
             [env: AUCTION_DATA_PAGE_SIZE=]  [default: 500]
+        --custom-benign-errors <custom-benign-errors>
+            Specify additional custom benign errors that can occur during solution submission [env:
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):
I think that in our kibana alerts filter we could also filter per pod name to accomplish a similar thing but having this here probably more ergonomic.

nlordell:
> I think that in our kibana alerts filter we could also filter per pod name to accomplish a similar thing but having this here probably more ergonomic.

I already installed a kibana filter, the issue is that it counts towards the Grafana metrics as a failed batch.

fleupold (approved):
>  it counts towards the Grafana metrics as a failed batch.

I would have said this is actually intended (since we are not finding a solution for those batches), but I see that we are already excluding subtraction underflows, so the same argument can apply for multiplication overflows.

nlordell:
> I would have said this is actually intended (since we are not finding a solution for those batches),

The thing is we want to suppress the alerts for Rinkeby for multiplication overflows. Maybe for v2 we can have a separate category for this kind of failure?

PR #1626, opened 2021-01-25
Use transaction retry and gas estimation dependencies from git

vkgnosis:
because we want to also use them in gp-v2.

https://github.com/gnosis/gp-gas-estimation
https://github.com/gnosis/gp-transaction-retry

I want to use them as git dependencies from gp v1 and v2 first.  We could eventually publish them on crates.io with the aim of making it easy for projects outside of gp to use them too. That is more work because we should polish them more so I do not want to commit to that now.
The current version of the crates is exactly what was in this repository before. Before using them in v2 we need to upgrade some dependencies which I am going to do as PRs to those repositories. After that v2 can depend on them.

### Test Plan
CI


fleupold (approved):

bh2smith (approved):
I really like the idea here, but it makes me think that these crates wouldn't necessarily be useful to other projects unless they were perhaps named more generally. Of course the GP estimation is GP specific, but the retry functionality is more broad and useful elsewhere. Just a thought.

That is, depending on how GP specific `gp-transaction-retry` is, maybe it could be just `transaction-retry`

PR #1649, opened 2021-03-22
Use custom GPv1 deployement block

fleupold:
This PR addresses an alert we are seeing with the latest geth nodes that by default do not contain tx inclusion logs for more than one year. Since our GPv1 Settlement contract was deployed more than a year ago, we run into issues fetching all events (which tries to get the block number from the deployment tx).

We fix this by manually specifying the block numbers for the the three main deployments:
- https://etherscan.io/tx/0xe74e8e965afc67f96a38caee794f008be33a7fe7ea96b9baad7b2963250ac36a
- https://rinkeby.etherscan.io/tx/0xc5bf4f48747093a79b623c2a2392ccbdb73a3d788e7d1f0f53f640ea18496d90
- https://blockscout.com/xdai/mainnet/tx/0xd5e7da686ace1ae7f1917ad54b8610ee936bdc900e56b495739795fba3d5e7c1/internal-transactions

Due to some version incompatibilities stemming from the new ethcontract-rs version, I had to remove the Cargo.lock file and recreate it from scratch.

### Test Plan
CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.37"
+version = "1.0.39"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee67c11feeac938fae061b232e38e0b6d94f97a9df10e6271319325ac4c56a86"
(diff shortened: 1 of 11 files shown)


nlordell (approved):
Looks great! Thanks for fixing this!

vkgnosis (approved):

